PR #6, opened 2015-01-05
Added CMake and lcov

Teemperor:
This patch adds

* CMake as an additional build system for the tests
* support for creating a coverage report with lcov (branch coverage, line coverage, function coverage).

I added that CMake treats each file in the 'test' directory as a test. So if you want to split up the json_unit.cc to speed up compilation, then the current CMake script should support it.

Here is an example run of lcov for this branch: http://teemperor.de/pub/json/index.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,58 @@
+cmake_minimum_required(VERSION 2.8.4)
+project(json)
+
+# Enable C++11 and set flags for coverage testing
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -g -O0 --coverage -fprofile-arcs -ftest-coverage")
+
+# If not specified, use Debug as build type (necessary for coverage testing)
+if( NOT CMAKE_BUILD_TYPE )
+  set( CMAKE_BUILD_TYPE Debug CACHE STRING
+       ""
+       FORCE )
+endif()
+
+# CMake addons for lcov
+# Only possible with g++ at the moment. We run otherwise just the test
+if(CMAKE_COMPILER_IS_GNUCXX)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -O0 --coverage -fprofile-arcs -ftest-coverage")
+  set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/CMakeModules)
+  include(CodeCoverage)
+
+  setup_coverage(coverage)
+endif()
+
+# Normal sources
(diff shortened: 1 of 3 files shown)


nlohmann:
Thanks!

nlohmann:
Hey @Teemperor,

thanks for your pull request. I checked the commands as described in the README file, but could not get `make coverage` to work. I only get

```
make: *** No rule to make target `coverage'.  Stop.
```

Did I forget something? (As I said - I have no experience with CMake...)

Furthermore, is there a way to delete any file generated by CMake or its Makefiles to return to a directory without any files that are not from the repository?

Thanks, and all the best,
Niels

Teemperor:
Can you clone to a new directory and try to create a project from there? Maybe there are some automake leftovers flying around in your source directory.

I'm not aware of a way to clean generated files as CMake is rather focusing on out-of-source-builds (so, you just build in a different directory). 
If you want to have a clean source directory, then make a 'build' subdirectory and start CMake from there like that.

```sh
mkdir build
cd build
# note the two dots behind cmake, as this is the path to the source directory containing the CMakeLists.txt
cmake ..
make coverage
```

If you want to clean all generated files, then just `rm -rf build` in this case.

nlohmann:
Indeed there was a problem with leftovers from Autoconf/Automake. After cleaning up, I tried:

```sh
[~/json] mkdir build
[~/json] cd build
[~/json/build] cmake ..
-- The C compiler identification is AppleClang 6.0.0.6000034
-- The CXX compiler identification is AppleClang 6.0.0.6000034
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/niels/Documents/Repositories/json/build
[~/json/build] make coverage
make: *** No rule to make target `coverage'.  Stop.
```

I am using CMake 3.1.0 (and still have no idea about it...)

Teemperor:
I see. The problem is that I disabled the 'coverage' target for non-GCC compilers as i only tested lcov with GCC. I think clang is also supported but it totally forgot about it when i wrote the CMake script (my bad!). 

You still need two other tools (lcov itself and genhtml) for producing the coverage support. Both are available over homebrew to my knowledge.

I'll add some remarks about that to the README.

PR #14, opened 2015-01-06
Fix double delete on std::bad_alloc exception

elliotgoodrich:
If the new operator throws in the json::find methods then result.oi_ is
deleted again in the destructor of json::iterator/json::const_iterator,
which is a double delete and undefined behaviour.


nlohmann:
Thanks!

PR #19, opened 2015-01-09
add the std iterator typedefs to iterator and const_iterator

kirkshoop:
really enjoying this library.
when composing it with another library, I needed the iterators to have the std typedefs.
This also required that ```json::value_type``` be disambiguated in the iterators.
NOTE: since collections are also supposed to have ```value_type```, the enum name may need to change in the future.


nlohmann:
Thank you!

PR #22, opened 2015-01-10
Unicode escaping

Teemperor:
Implements the \uXXXX and \uXXXX\uXXXX unicode escape mechanisms of JSON.

This branch has a problem with code coverage as there is one method which has a untestable case (line 2215).
This code is only reachable if
1. there is a bug somewhere else that calls the function with the wrong argument
2. we have access to the private method.

So we need a way to directly access the `codePointToUTF8` method from the test header to get the line coverage back to 100%.

Any recommendation how we do that in a nice way?

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -1,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectCodeStyleSettingsManager">
+    <option name="PER_PROJECT_SETTINGS">
+      <value>
+        <Objective-C-extensions>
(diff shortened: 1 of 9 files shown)


nlohmann:
Thanks, I'll have a look.

FYI: In the unit tests, you can access private methods - at least when the code is compiled with the Automake files, because `-Dprivate=public` is used. (I know this is ugly, but it allows better coverage without adjusting the tested code...)

Teemperor:
Sounds good to me, updated the branch!

Teemperor:
Just realized I effectively reverted your commit `0cd2ecf` about the code style patch. I'll apply the changes to this patch.

nlohmann:
Thanks a lot! This Unicode business is really a pain!

PR #24, opened 2015-01-10
Fixed dumping of strings

Teemperor:
Dumping strings containing special characters such as quotes, backslashes, newlines and so on results in invalid results.

This patch adds a escaping mechanism that fixes this problem.

A possible (but not required by the standard) escaping of unicode characters is not included in this patch.


nlohmann:
Thanks!

PR #26, opened 2015-01-12
Fix string conversion under Clang

wancw:
Explicit conversion to `std::string` by `static_cast<>` causes compilation error when using Clang (3.5).
(Casting to `const std::string` doesn't work either.)

This PR fixed it by removing `const` from the casting operator.
And I think it also makes casting operator more consistent with `get<>()` method.

### DIFF
--- a/src/json.cc
+++ b/src/json.cc
@@ -442,7 +442,7 @@ json::object_t json::get() const
     }
 }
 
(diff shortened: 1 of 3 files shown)


nlohmann:
Thanks a lot for fixing that!

PR #29, opened 2015-01-17
support visual studio compilers (2013 & 2014) on windows

kirkshoop:
If you like, login to http://appveyor.com to add windows CI for the json project. the configuration is in appveyor.yml

PR #34, opened 2015-02-04
(partial) msvc support

Zoadian:
some changes to support compilation with msvc120
I used .h and .cc files directly without cmake.

also line 2374 is probably wrong

so this is just the first step to support msvc.

### DIFF
--- a/src/json.cc
+++ b/src/json.cc
@@ -414,7 +414,7 @@ int json::get() const
     switch (type_)
     {
         case (value_t::number):
-            return value_.number;
+            return static_cast<int>(value_.number);
         case (value_t::number_float):
             return static_cast<int>(value_.number_float);
         default:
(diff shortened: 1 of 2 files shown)


nlohmann:
Hi @Zoadian, sorry for not answering earlier. Though I like the idea of having a Windows version, I am currently working on new code that uses even more C++11 features. When I look at your code, I see that you mostly use the preprocessor to exclude those parts that are not yet supported by Visual Studio. I think I would rather wait until Visual Studio has true C++11 support rather than to fragmentize the project like this - in the end, I am using this project to learn the new C++11 features. 

I'm going to close the pull request here, but I would be happy if you would host a Windows-compatible version yourself. I would link to it in the README. You might also want to get in touch with @kirkshoop. Kirk opened a similar pull request #29 earlier.

nlohmann:
Oh, I forgot. The error you mentioned is indeed a mistake. I will fix it soonish.

PR #35, opened 2015-02-05
fixed duplicate symbols errors on header-only version. issue #16

### LINKED ISSUE #16: create a header-only version
As wished in http://www.reddit.com/r/cpp/comments/2rbhn3/json_for_modern_c_what_if_json_was_part_of_modern/cnelnzq:

> I think this would be fare more useful as a header only library. I'm down for sending a PR with the change if you are.

And as I wrote:

> Good idea! But I think it would be more useful if there would be a script that reads the source files json.cc and json.h and creates a header-only version. That way, both approaches can be followed at the same time.

mnpk:
I got duplicate symbols errors when include header-only version in more than one cpp files.
So I just add inline keyword before method defines to fix this. Link errors disappeared. Is this enough to close #16?

### DIFF
--- a/header_only/json.h
+++ b/header_only/json.h
@@ -599,12 +599,12 @@ namespace nlohmann
 // CONSTRUCTORS OF UNION "value" //
 ///////////////////////////////////
 
-json::value::value(array_t* _array): array(_array) {}
-json::value::value(object_t* object_): object(object_) {}
-json::value::value(string_t* _string): string(_string) {}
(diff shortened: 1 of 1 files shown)


nlohmann:
Thanks for the request. This solves the problem for now, but unfortunately the file is generated from the header `src/json.h`. So whenever that file is changed, it will overwrite `header_only/json.h`.

I am currently working on a header-only file myself - this time, not as a byproduct of a .h/.cc pair of files. I hope to have it released in the weekend.

Thanks for your effort anyway!

PR #38, opened 2015-02-28
Feature/small float serialization

### LINKED ISSUE #37: dumping a small number_float just outputs 0.000000
I expected that small doubles would be represented in dumped JSON as a javascript number in scientific notation, but instead it just appears as `0.000000` 

Example:
```c++
#include <iostream>

#include "json.hpp"
using json = nlohmann::json;

main() {
  double d = 1.23456e-89;
  json j = d;

  std::cout << "d=" << d << " j=" << j << std::endl;
}
```

Expected output: `d=1.23456e-89 j=1.23456e-89`

Actual output: `d=1.23456e-89 j=0.000000`

Looking at the json code, it appears this is due to using `std::to_string()` for all numbers, which is equivalent to `sprintf("%f", ...)` whereas `sprintf("%g", ...)` would be more appropriate. To fix this, the dump function would need to use something other than to_string to format floating point numbers. I'll try to put together a PR for this although C++ is far from my best language.

jrandall:
Adds a test for small floats (which previously failed) along with a fix 
for dump() that makes the test pass. 

Now uses std::snprintf() to generate a "%.15g" formatted string
for JSON values of type number_float. 15 decimals digits are
enough to round-trip an IEEE 754 double from string->double->string
and get an identical result.

std::snprintf is called twice. Once to determine the required
buffer size and then again after allocating a buffer of that
size.

Note that the buffer size *could* be hardcoded for better
performance. "%.15g" should result in strings of maximum length
23, plus one character for the terminating null for a buffer size
of 24. Let me know if you'd like me to switch to a fixed buffer 
size to avoid the extra call to snprintf and the overhead of 
dynamic memory allocation. 

Fixes #37

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -1853,7 +1853,8 @@ class basic_json
     recursively. Note that
 
     - strings and object keys are escaped using escape_string()
-    - numbers are converted to a string before output using std::to_string()
+    - integer numbers are converted to a string before output using std::to_string()
+    - floating-point numbers are converted to a string using "%g" format
 
     @param prettyPrint    whether the output shall be pretty-printed
     @param indentStep     the indent level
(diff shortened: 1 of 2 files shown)


jrandall:
I'm not familiar with AppVeyor, but the failure doesn't seem like it could plausibly be my fault:

```
'make' is not recognized as an internal or external command,
operable program or batch file.
```

nlohmann:
Thanks for the pull request! I made some adjustments, because the file `json.hpp` is generated from `json.hpp.re2c`.

PR #63, opened 2015-04-26
Improve the implementations of the comparission-operators

Florianjw:
As the title says, just one thing should be noted:

Until now, `integer<float` used integer-comparission, while `float<integer` worked on floats. I believe that this was at best questionable behavior, at worst a bug. This pull-request changes that to always work on floats, which might break code (that would then however have been broken to begin with).


nlohmann:
Thanks a lot!

PR #71, opened 2015-05-09
support enum

likebeta:
construct enum type to basic_json (proxy by int type)
```
    enum { t = 0 };
    json  j = json::array();
    j.push_back(t);

    j.push_back(json::object({
         {"game_type", t}
    }));
```

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -417,6 +417,11 @@ class basic_json
     inline basic_json(const number_integer_t& value)
         : m_type(value_t::number_integer), m_value(value)
     {}
+    
+    /// create an int number to support enum type (implicit)
+    inline basic_json(int int_enum)
+	    : m_type(value_t::number_integer), m_value((NumberIntegerType)int_enum)
+    {}
 
     /// create an integer number (implicit)
     template<typename T, typename


nlohmann:
Thanks!

nlohmann:
Unfortunately, I have to remove this fix, because the implicit assumption that anonymous enums are `const int` only works as long as `basic_json::number_t` is a different type (`int64_t` by default). If the type is changed to `int32_t` (as reported in issue #89), the conversion is ambiguous. Unless there is a better way, I decided to undo this change. The example code then would need an explicit cast:

```cpp
enum { t = 0 };
json j = json::array();
j.push_back(static_cast<int>(t));

j.push_back(json::object(
{
    {"game_type", static_cast<int>(t)}
}));
```

I hope you don't mind this change - and if you have any ideas how to find a solution that is compatible to issue #89, please let me know!

gregmarr:
You could probably use std::underlying_type<> to find the base type of the enum.

nlohmann:
Good point!

nlohmann:
Ok, I found another way :-)

likebeta:
well done!， I am sorry to make this mistake.

PR #72, opened 2015-05-10
Replace `default_callback` function with `nullptr` and check for null…

aburgh:
… callback function before calling it. This is cleaner and better performing.

This is in response to your question about performance and the temporary object for the callback. This does not eliminate the temporary object, but it replaces the default callback with a check for null on the callback member variable, which is faster and avoids the temp object in the default case. In the case where the caller has supplied a callback function, presumably the overhead of a temp object is negligible compared to the processing that will be performed in the callback.

If you would like to optimize the callback to avoid the temp object, the only solution I can think of at the moment is to change the `basic_json` reference parameter of the callback function to a `void *` and require the callback function to cast it to a `string` pointer or `basic_json` pointer based on the event type. This solution sounds really ugly to me, so my vote is in favor of the temp object.

### DIFF
--- a/src/json.hpp.re2c
+++ b/src/json.hpp.re2c
@@ -291,12 +291,6 @@ class basic_json
     using parser_callback_t = std::function<bool(int depth, parse_event_t event,
                               const basic_json& parsed)>;
 
-    /// default parser callback returns true to keep all elements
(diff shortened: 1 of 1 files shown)


nlohmann:
Thanks a lot! That's a few percent performance improvement!

PR #75, opened 2015-05-26
Serialization to ostream and faster string escaping

EricMCornelius:
Opening this to spur discussion - I'm assuming it should actually be added to the re2c file.

After doing some performance analysis of an xml => json transformer I'm working on, I discovered that the json.dump() call accounted for a disproportionate amount of the total execution time.

These changes enable ostream output instead of needing to first buffer the entire formatted output in a string.  

The alternative string escaping method seems to be quite a bit faster for a broad array of data which I've tested on (~50% on our internal test cases running over approximately 50GB test sets - unfortunately can't share the files as they contain sensitive security data).

Basic concept is to do one pass to find the number of necessary characters to escape, then another pass to actually do the escaping.  This lets it completely avoid another dynamic memory allocation for strings which don't need any escaping, and guarantees only a single allocation in the alternative scenario.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -789,6 +789,244 @@ class basic_json
         }
     }
 
+    /*!
+    @brief serialization
+
+    Calculates the extra space required to escape a JSON string
+    */
+    inline std::size_t extra_space(const string_t& s) const noexcept
+    {
+        std::size_t count = 0;
+
+        for (const auto c : s)
+        {
+            switch (c)
+            {
+                case '"':
+                case '\\':
+                case '\b':
+                case '\f':
+                case '\n':
+                case '\r':
+                case '\t':
+                    count += 1;
+                    break;
+                default:
+                {
+                    if (c >= 0 and c <= 0x1f)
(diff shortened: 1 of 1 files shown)


nlohmann:
Thank you very much! Your performance reports sound impressive. Please give me some time to check and to integrate.

nlohmann:
Hi there again, I finally have some time to check your code. It seems that it currently only works if the output should not be pretty-printed. Was this limitation intentional or just because it was none of your use cases?

EricMCornelius:
Just was not in my initial use cases.

nlohmann:
Hi there! I just made some changes to my code: I used streams wherever possible to avoid creating and resizing strings. In my benchmarks, serialization performance increased by about 20%. Could you try whether this performance is also realized with your large benchmark?

Furthermore, I would like to ask if there is a way to get a anonymized version of your file, because such a large file would be a perfect benchmark for me.

nlohmann:
Thanks a lot, @EricMCornelius!

PR #79, opened 2015-06-02
Make immutable array static and constexpr

ahamez:
This array is only read and is always initialized with the same values, thus it's useless to construct it for each call of operator<() .
I'm not sure it will have any impact on performances, but the immutability is clearly asserted by making this array static and constexpr.


nlohmann:
I already merged it into the re2c file myself. Thanks for the hint!

PR #81, opened 2015-06-04
fix mingw build

Ingener74:
This changes solve problem for MinGW, and compile fine on Ubuntu Linux with g++ 4.8.4


nlohmann:
I already made some changes with regard to `to_string`. I hopefully fixed the rest (see latest commit).

nlohmann:
The issue is fixed.

PR #86, opened 2015-06-09
Remove useless typename

ahamez:
It avoids a warning with GCC 4.9:

```
json.hpp:2361:49: warning: declaration 'struct std::bidirectional_iterator_tag' does not declare anything
         using iterator_category = typename std::bidirectional_iterator_tag;
                                                 ^
```

I don't add the generated `json.hpp` as my re2c version makes a very different version from the one in the repo. Or maybe I should directly modify `json.hpp`?


nlohmann:
Thanks!

ahamez:
Your'e welcome :-)

PR #87, opened 2015-06-12
Keyword 'inline' is useless when member functions are defined in headers

ahamez:
See https://isocpp.org/wiki/faq/inline-functions#inline-member-fns-more. It won't change anything to behavior or linkage, but it makes the code a bit smaller :-)


nlohmann:
Thanks!

PR #105, opened 2015-07-15
Vs14 fixes

kepkin:
Well, that's it. Everything you need to make it compile and fully work under MSVC 2015.
I'm afraid Appveyor won't be able to build it, as it required to much resource to build `unit.cpp` in release mode.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -4286,7 +4286,9 @@ class basic_json
 
         // We only reach this line if we cannot compare values. In that case,
         // we compare types.
-        return lhs_type < rhs_type;
+        // Have to use operator< explicitly in order to
+        // not cause ambiguity on MS compiler
+        return operator<(lhs_type, rhs_type);
     }
 
     /*!
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -2931,8 +2931,8 @@ class basic_json
     */
     template <class InteratorType, typename
               std::enable_if<
-                  std::is_same<InteratorType, typename basic_json::iterator>::value or
-                  std::is_same<InteratorType, typename basic_json::const_iterator>::value
+                  std::is_same<InteratorType, typename __basic_json::iterator>::value or
+                  std::is_same<InteratorType, typename __basic_json::const_iterator>::value
                   , int>::type
               = 0>
     InteratorType erase(InteratorType first, InteratorType last)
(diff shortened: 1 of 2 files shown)


gregmarr:
C++11 allows unions to have members with non-trivial constructors.  VS2015 RC now supports this according to the feature table at http://blogs.msdn.com/b/vcblog/archive/2015/04/29/c-11-14-17-features-in-vs-2015-rc.aspx

When you use types with non-trivial constructors and destructors in the union, you need to do more management of the union to change the active member, manually calling the destructor on the previous active member and placement new on the new active member.

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2544.pdf

gregmarr:
I know this wasn't first introduced in this commit, but this shouldn't use __basic_json as a typedef.  Identifiers that start with two underscores are reserved to the C++ implementation (compiler and standard library).  That should be renamed to something else, like basic_json_t.

nlohmann:
Hi @kepkin, thanks a lot for the PR!

nlohmann:
Hey @gregmarr, you are right with `__basic_json`, I changed the name to `basic_json_t`.

About the union: I applied @kepkin's change to a `struct`, because I do not understand the details. However, you write

> When you use types with non-trivial constructors and destructors in the union, you need to do more management of the union to change the active member, manually calling the destructor on the previous active member and placement new on the new active member.

However, in the code, we never change the active member.

gregmarr:
If you're fine with it being a struct, that's probably the easiest path.  Otherwise you have to be sure to manually construct the objects, and that's probably going to be difficult cross-platform.

nlohmann:
Hi @kepkin again, I have a question with respect to CMake, msbuild, and Appveyor you may help me with: You reported that Appveyor times out when it tries to compile the unit tests. That is very unfortunate but they offer 100% coverage, it would be nice to at least compile them with MSVC to avoid regressions.

In the `CMakeLists.txt` file, I added `-Od` to `CMAKE_CXX_FLAGS` to switch off optimizations. However, these flags seem to be ignored when `cmake --build . --config Release` is started (see https://ci.appveyor.com/project/nlohmann/json/build/515).

Do you have an idea how to influence this? Or whether there are different ways to make Appveyor compile the code?

gregmarr:
http://stackoverflow.com/questions/30416804/cmake-disable-optimization-with-od

nlohmann:
\o/

https://ci.appveyor.com/project/nlohmann/json/build/517

THANKS A LOT!

PR #115, opened 2015-08-25
Use the right variable name in doc string

whoshuu:


nlohmann:
Hi @whoshuu , thanks for the commit! Could you please make the change to `json.hpp.re2c`, because `json.hpp` is created from this file?

nlohmann:
Thanks a lot!

PR #119, opened 2015-09-08
exception-safe object creation, fixes #118

### LINKED ISSUE #118: Memory leak in face of exceptions
The constructor often does a two-step initialisation, an `allocate()` followed by a `construct()`. If the `construct()` throws then the previously allocated memory will leak...

            object = alloc.allocate(1);
            alloc.construct(object, value);

d-frey:
Using `unique_ptr` is preferable to `try { ... } catch(...) { ...; throw; }` as the latter will screw up the backtrace of uncaught exceptions in the debugger or a terminate handler.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -591,7 +591,7 @@ class basic_json
         AllocatorType<T> alloc;
         auto deleter = [&](T* object) { alloc.deallocate(object, 1); };
         std::unique_ptr<T, decltype(deleter)> object(alloc.allocate(1), deleter);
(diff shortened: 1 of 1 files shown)


gregmarr:
Yes, I definitely prefer the unique_ptr approach, and especially this helper function.

nlohmann:
Hi @d-frey , thanks for the commit! Could you please make the change to `json.hpp.re2c`, because `json.hpp` is created from this file?

d-frey:
I've created a new pull request agains `json.hpp.re2c`.

PR #122, opened 2015-09-20
exception-safe object creation, fixes #118

### LINKED ISSUE #118: Memory leak in face of exceptions

d-frey:


nlohmann:
Thanks a lot! This is a very nice fix indeed!

Issue #125, opened 2015-09-22
Visual Studio 14 Debug assertion failed


ruslo:
I've tried to run unit-test with Visual Studio 14 2015 generator and got runtime error:
```
Expression: vector iterator not dereferencable
```

Last backtrace entry comes from: https://github.com/nlohmann/json/blob/0a813539894d8a6d688ca6a404966f17b56f6e01/src/json.hpp#L5119

which came from: https://github.com/nlohmann/json/blob/0a813539894d8a6d688ca6a404966f17b56f6e01/test/unit.cpp#L2879

nlohmann:
Hi @ruslo,

unfortunately, I have no experience with MSVC. There have been numerous attempts to have the code work with MSVC before, but it is breaking again and again (see https://ci.appveyor.com/project/nlohmann/json). It would be great if you could explain to me what is wrong in the code or better propose a fix. 

Thanks,
Niels

ruslo:
The last time I've tried to figure out what is happening I've noticed a hairy chain of `CHECK` macro expanding. It seems that while doing expand there is an access to the iterator which is equal to `end()` and hence is not dereferencable.

dariomt:
I've been taking a look at this issue.
It seems that:
1) CHECK macro is very smart and will detect (lhs op rhs), i.e. the lhs, the rhs and the operator in the expression 
2) to provide error messages if will "stringify" lhs and rhs trying to call the streaming operator
3) the streaming operator of the json iterators stream their value as seen in 
`friend std::ostream& operator<<(std::ostream& o, const iterator_wrapper_internal& w)`
4) in the problematic case, both lhs and rhs are end iterators, so trying to dereference them is **undefined behavior** which is caught by MSVC in Debug mode

To me the problem is 3)
Why would an iterator be Streamable?
I can't see a use case for that! 
Iterators are *not* Streamable, the values they point to could be, but never the iterator.

My proposed resolution is to remove that streaming operator for iterators (which doesn't make sense).

With that, the code compiles and I can see other tests failing for other reasons, which I'll try to investigate.

dariomt:
Doh! I was not running the tests in the right directory.

Now, after removing operator<< for iterators, I see only two tests failing in Debug mode with VS2015:
```
$ time _build/Debug/json_unit.exe "*"
{
    "happy": true,
    "pi": 3.141
}
{
  "happy": true,
  "pi": 3.141
}
"foo"
1
true
"foo"
1
true

```
[...]

PR #131, opened 2015-10-07
disabled "CopyAssignable" test for MSVC in Debug mode, see #125

dariomt:
STL iterators used by json::iterator don't pass this test in Debug mode.
The test does pass in Release mode, so I felt the best thing to do was selectively disable that test.
More details in #125


nlohmann:
Thanks!

Issue #134, opened 2015-10-08
wrong enable_if for const pointer (instead of pointer-to-const)


dariomt:
Hi ,
The following test fails to compile:
```
const json value = {{"one", 1}, {"two", 2}};
const json::object_t* p = value.get_ptr<const json::object_t*>();
```
    
But this compiles (note the extra const):
```
const json value = {{"one", 1}, {"two", 2}};
const json::object_t* const q = value.get_ptr<const json::object_t* const>();
```

I think that the problem is the additional check for the const overload of get_ptr
`and std::is_const<PointerType>::value`

I think what you intended was to check that the template parameter is a "pointer-to-const" but what you are actually checking is if you got a "const pointer".

I believe the right fix is to check for "pointer-to-const" like this:
`and std::is_const< typename std::remove_pointer<PointerType>::type >::value`
It works in VS2015.

dariomt:
PR #137 ready!

nlohmann:
After merging #137, this issue can be closed.

PR #137, opened 2015-10-16
Detect correctly pointer-to-const

dariomt:
The intention of the current code is to detect a pointer-to-const but instead it is detecting a const-pointer. See #134


nlohmann:
Same issue here. Could you please also change `json.hpp.re2c`, because ´json.hpp` is generated out of this file. Thanks and looking forward to merging!

dariomt:
Will do!

dariomt:
Done!

nlohmann:
Thanks!!

PR #140, opened 2015-10-22
kill two warnings

406345:
sprintf(&result[pos + 1], "u%04x", int(c));   ---->    sprintf_s(&result[pos + 1],result.size(),"u%04x", int(c));

 difference_type m_it = std::numeric_limits<std::ptrdiff_t>::min();   ---->    difference_type m_it = std::numeric_limits<std::ptrdiff_t>::denorm_min();

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -4678,7 +4678,7 @@ class basic_json
                     if (c >= 0x00 and c <= 0x1f)
                     {
                         // print character c as \uxxxx
-                        sprintf(&result[pos + 1], "u%04x", int(c));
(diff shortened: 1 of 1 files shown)


benloong:
`sprintf_s` is not cross platform, you may use `snprintf`(since c++11) instead.

nlohmann:
Thanks for the CR! After #153, `sprintf`is not called any more. I shall commit the change to using `denorm_min` separately.

nlohmann:
@406345, just out of curiosity, which tool gave you a warning about `denorm_min()`?

PR #141, opened 2015-10-22
using c++11 snprintf

benloong:
change `sprintf` to safe version `snprintf`


nlohmann:
After #153, `sprintf` is not called any more.

PR #146, opened 2015-11-13
Fix character skipping after a surrogate pair

robertmrk:
In a string the first character following a surrogate pair is skipped by the
lexer, but the rest of the string is parsed as usual.


nlohmann:
Thanks a lot!!

PR #153, opened 2015-12-07
Replace sprintf with hex function, this fixes #149

### LINKED ISSUE #149: Unable to compile on MSVC 2015 with SDL checking enabled: This function or variable may be unsafe.
When this library is compiled under MSVC 2015 with SDL checking on, compilation fails with this error:

C4996 'sprintf': This function or variable may be unsafe. Consider using sprintf_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.

The error occurs here at line 4694:
image

edit: This occurs with the latest release available from the releases list.

whackashoe:


nlohmann:
Thanks!

Issue #164, opened 2015-12-29
std::unorderd_map cannot be used as ObjectType


nlohmann:
The following code does not compile:

``` cpp
using json = nlohmann::basic_json<std::unordered_map, std::vector, std::string, bool, int64_t, double, std::allocator>;
```

The reason for this seems to be that the definition of `object_t` seems to rely on specifics of `std::map`:

``` cpp
using object_t = ObjectType<StringType,
      basic_json,
      std::less<StringType>,
      AllocatorType<std::pair<const StringType,
      basic_json>>>;
```

erichkeane:
I've been looking at this for a bit, and a 1st run solution isn't that bad.  Just doing SFINAE/enable_if to switch based on unordered_map vs map is pretty easy using is_same or is_base_of.  

However, I'm still trying to figure out a way to not break existing code that specializes the ObjectType parameter to a non-std::map container (such as a custom map that has object_t compatible template args).

The difference between map and unordered_map is 2 fold:
1- Different NUMBER of arguments
2- Different types of arguments.  Essentially, the 3rd template param for std::map is a Comparer, whereas in unordered_map this is replaced by a HashFun and KeyEqualityFun.

Optimally, the object_t would be defined entirely based on the ObjectType signature, though I'm currently having a tough time coming up with anything.

nlohmann:
I think one can check via `enable_if` whether the container has a function `hash_function()` to differentiate between `std::map` (`false`) and `std::unordered_map` (`true`). It's not perfect this way, but it could be a first step.

erichkeane:
Alright, sounds acceptable.  I'm not a HUGE fan of doing it by anything but signature, since the result is overly dependent on the std::map vs std::unordered_map types, but I'll put a pull request together for sdoing it that way in a few hours.  Thanks!

erichkeane:
I've made an attempt that is ALMOST working.  I differentiate between the two types via SFINAE based on key_comparer and hasher, which makes object_t do exactly what we need it to.  

See the commit here: https://github.com/erichkeane/json/commit/90ab92ade1df8d32ea52e7ad302b41c47b783dce

One might think that type-test-validity would work the same way I'm doing it without having to dive into the member types, however unordered_map will go through resolution successfully with only the 4 template arguments, so it would otherwise be ambiguous.

The calls to void_converter are done with ObjectType<char,char>::... since the Value type is incomplete at the time.

[...]

Issue #144, opened 2015-11-02
Implicit assignment to std::string fails


Cleroth:
Put simply, when constructing a string, you can pass a basic_json:

    std::string s = o["name"];

However, this does not work:

	std::string s;
	s = o["name"];
	s = o["name"].get<std::string>(); // this works

The following compiler errors pop up:

1>g:\files\smoldy_adventure\src\client\units\unitmgr.cpp(53): error C2593: 'operator =' is ambiguous
1>  g:\program files (x86)\microsoft visual studio 14.0\vc\include\xstring(1032): note: could be 'std::basic_string<char,std::char_traits<char>,std::allocator<char>> &std::basic_string<char,std::char_traits<char>,std::allocator<char>>::operator =(_Elem)'
1>          with
1>          \[
1>              _Elem=char
1>          ] (compiling source file units\unitmgr.cpp)
1>  g:\program files (x86)\microsoft visual studio 14.0\vc\include\xstring(1027): note: or       'std::basic_string<char,std::char_traits<char>,std::allocator<char>> &std::basic_string<char,std::char_traits<char>,std::allocator<char>>::operator =(const _Elem *)'
1>          with
1>          \[
1>              _Elem=char
1>          ] (compiling source file units\unitmgr.cpp)
1>  g:\program files (x86)\microsoft visual studio 14.0\vc\include\xstring(1011): note: or       'std::basic_string<char,std::char_traits<char>,std::allocator<char>> &std::basic_string<char,std::char_traits<char>,std::allocator<char>>::operator =(const std::basic_string<char,std::char_traits<char>,std::allocator<char>> &)' (compiling source file units\unitmgr.cpp)
1>  g:\program files (x86)\microsoft visual studio 14.0\vc\include\xstring(972): note: or       'std::basic_string<char,std::char_traits<char>,std::allocator<char>> &std::basic_string<char,std::char_traits<char>,std::allocator<char>>::operator =(std::initializer_list<_Elem>)'
1>          with
1>          \[
1>              _Elem=char
1>          ] (compiling source file units\unitmgr.cpp)
1>  g:\program files (x86)\microsoft visual studio 14.0\vc\include\xstring(914): note: or       'std::basic_string<char,std::char_traits<char>,std::allocator<char>> &std::basic_string<char,std::char_traits<char>,std::allocator<char>>::operator =(std::basic_string<char,std::char_traits<char>,std::allocator<char>> &&) noexcept' (compiling source file units\unitmgr.cpp)
[...]

PR #165, opened 2015-12-30
Unordered map json

erichkeane:
Work for unordered map in the json object.  See #164 for the bug report and description of the solution.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -116,17 +116,11 @@ struct void_converter
 };
 
 // Type to help determine between a map or unordered-map type for the
-// object_t type alias.  Takes advantage of the fact that a map has a
-// key_comparer member type and unordered_map has a hasher member type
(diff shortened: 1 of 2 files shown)


erichkeane:
Looks like this doesn't build on MSVC... Perhaps I have to see if I can mess with a few other things in order to get it working.  It might take a while for me to spin a VM, but I'll push a fix ASAP.

erichkeane:
I seem to have managed to cause the same issue you see in https://github.com/nlohmann/json/issues/144

erichkeane:
As mentioned in #144, I checked master out directly, and reproduce this same ambigious conversion issue on origin/master!

It seems that there is a problem running one of the tests (CHECK(*p2 == value.get<test_type>());, due to os << test_type(json::object_t); being ambigious.  

I'm going to open a new bug for this one.

nlohmann:
The current PR is #209. I close this one.

Issue #161, opened 2015-12-28
Incorrect C++11 allocator model support


glenfe:
Here is a preliminary list of issues:
- Your allocator template parameter should be just `class AllocatorType = std::allocator<ValueType>` instead of `template<class U> class AllocatorType = std::allocator`.  i.e. Do not require that the allocator type is a class template with one template parameter.
- You only support stateless allocators; C++ allocators can be stateful. Your `basic_json` constructors should accept an optional allocator instance: `basic_json(..., const AllocatorType& allocator = AllocatorType());`
- Your `get_allocator()` member should return the value of that stored allocator instance.
- Don't use `AllocatorType<Type> alloc;`, instead rebind the stored allocator instance member variable with `std::allocator_traits<AllocatorType>::rebind_alloc<Type> alloc(allocator);`.
- Don't use `alloc.construct(...)` directly, instead use `std::allocator_traits<AllocatorType>::rebind_traits<Type>::construct(alloc, ...)`. Allocators are not required to provide `construct` or `destroy` so all construction and destruction should go through `std::allocator_traits`.
- A given allocator is allowed to have complex types for `pointer` so preserve those (instead of `T*`) for the result of `alloc.allocate(n)`. Use `std::addressof` with those pointers to obtain the value for use with `construct(...)`.

You'll notice all the C++ standard library facilities which support allocators implement the above, so look to them (or to allocator aware types in Boost) for further examples.

nlohmann:
@glenfe, thanks for your feedback! I copied your items to check them off once they are fixed:

- [ ] Your allocator template parameter should be just `class AllocatorType = std::allocator<ValueType>` instead of `template<class U> class AllocatorType = std::allocator`.  i.e. Do not require that the allocator type is a class template with one template parameter.
- [ ] You only support stateless allocators; C++ allocators can be stateful. Your `basic_json` constructors should accept an optional allocator instance: `basic_json(..., const AllocatorType& allocator = AllocatorType());`
- [ ] Your `get_allocator()` member should return the value of that stored allocator instance.
- [ ] Don't use `AllocatorType<Type> alloc;`, instead rebind the stored allocator instance member variable with `std::allocator_traits<AllocatorType>::rebind_alloc<Type> alloc(allocator);`.
[...]

PR #169, opened 2016-01-01
Abstract object/array type out so that arbitrary types can be used

erichkeane:
As suggested by @palacaze in #161, this allows using an unordered_map
as well as anything that obeys the template<K,V> type for the object
type or template<V> for the array type.

PR #170, opened 2016-01-03
Make basic_json allocate-aware, as mentioned in #161 + unrelated changes for #164

palacaze:
This is an attempt to make it work with #164.
The thing is that the current code attempts to define containers with incomplete types,
this is not supported by the standard and does not work with unordered_map. std::vector and
std::map work but we might be in the realm of undefined behaviour.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -105,6 +105,13 @@ static bool approx(const T a, const T b)
 {
     return not (a > b or a < b);
 }
+
+// facilities to express rebound allocator with less verbosity
+template <typename A, typename T>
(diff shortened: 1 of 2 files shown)


nlohmann:
@palacaze Sorry for not coming back to this PR earlier. There were quite some fixes on the MSVC side, so I would like to ask you to resubmit the PR to work on the latest version.

palacaze:
I would be glad to but as mentioned in #164 the allocator-aware change and #169 are not compatible.
I don't know what to do at this point.

nlohmann:
Unfortunately, this PR staled - allocators are hard. :-(

Issue #178, opened 2016-01-13
Integer conversion to unsigned


Furcube:
It is possible to assign negative json value to unsigned variable.

This is a possible patch
```c++
diff --git a/src/json.hpp b/src/json.hpp
index edda8ed..a8c85e5 100644
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -2316,7 +2316,12 @@ class basic_json
         {
             case value_t::number_integer:
             {
-                return static_cast<T>(m_value.number_integer);
+                auto res = m_value.number_integer;
+                if(std::is_unsigned<T>::value && res < 0) {
+                    throw std::domain_error("can't assign negative value to unsigned type");
+                } else {
+                    return static_cast<T>(res);
+               }
             }
 
             case value_t::number_float:

```

gregmarr:
It is possible that the value was unsigned in the json, and was converted to a negative value for storage in the json object.  It would have to be a really big unsigned number with the default storage size being int64_t, but it's still possible.  The json object would need to store additional data to know whether it was negative or not originally.

Furcube:
It would be better to check integer size and signedness in parser. Successfuly parsed json should contain valid data.

I forgot that NumberIntegerType can be changed, and even be unsigned, condition for conversion check should be like this:
```c++
if(std::is_unsigned<T>::value && std::is_signed<NumberIntegerType>::value && res < 0)
```

Another possibility is to remove static_cast and let the user handle conversions. But I'm not sure that it will not break API.

Anyway, silent incorrect conversion is not a good thing.

gregmarr:
I think this also has the possibility to break existing programs, though at runtime rather than compile time.  Currently, if someone populates a json object with an unsigned integer value larger than the max size for NumberIntegerType, it will be encoded in the JSON string as signed.  Parsing that string and extracting the value as unsigned will currently produce the same result as the original input.  This change would make it throw an exception instead.

Furcube:
If it really ignores integer overflow and does not differentiate between signed and unsigned. At least this behavior should be noted in documentation.

For now i removed static_cast to get compiler errors and made validating wrapper, but considering above, it is possible that value would be in fact undefined in result of overflow or signedness conversion.

Simple example - some 3rd party app can send my app json with integer which need 128 bits, I expect deserialization of this value to fail, not to get invalid value.

twelsby:
@gregmarr, I might be missing something but why would a large unsigned value in the JSON be stored as a negative number in the object?

[...]

Issue #171, opened 2016-01-04
Cannot index by key of type static constexpr const char*


poskadesign:
This will not compile, error: use of overloaded operator '[]' is ambiguous (with operand types 'json' (aka 'basic_json<>') and 'const char *const'). Code:

static constexpr const char* _VAR1 = "MyKey";
void foo() {
    nlohmann::json j;
    j[_VAR1] = 10;
}

nlohmann:
I can confirm the bug for Clang (Apple LLVM version 7.0.2 (clang-700.1.81)) and Clang 3.7:

Code:

```cpp
#include <json.hpp>

using nlohmann::json;

int main()
{
    static constexpr const char* _VAR1 = "MyKey";

    nlohmann::json j;
    j[_VAR1] = 10;
}
```

yields:

```
issue171.cpp:10:6: error: use of overloaded operator '[]' is ambiguous (with operand types 'nlohmann::json' (aka 'basic_json<>') and 'const char *const')
    j[_VAR1] = 10;
    ~^~~~~~
src/json.hpp:2911:15: note: candidate function
    reference operator[](const typename object_t::key_type& key)
              ^
src/json.hpp:2959:21: note: candidate function
    const_reference operator[](const typename object_t::key_type& key) const
                    ^
issue171.cpp:10:6: note: built-in candidate operator[](long, const char *)
    j[_VAR1] = 10;
     ^
issue171.cpp:10:6: note: built-in candidate operator[](long, const volatile char *)
1 error generated.
make: *** [issue171] Error
```

Interestingly, GCC 5.2.0 works... I shall have a look at this later.

nlohmann:
This seems related: http://stackoverflow.com/questions/8914986/should-this-compile-overload-resolution-and-implicit-conversions

nlohmann:
The ambiguity is not really between the const and the non-const version. A different program yields

```
src/json.hpp:2959:21: note: candidate function
    const_reference operator[](const typename object_t::key_type& key) const
                    ^
issue171_1.cpp:13:15: note: built-in candidate operator[](long, const char *)
    auto x = k[_VAR1];
              ^
issue171_1.cpp:13:15: note: built-in candidate operator[](long, const volatile char *)
```

gregmarr:
Is the entire root of the ambiguity this function:
    operator value_t() const noexcept
    {
        return m_type;
    }

which allows a json to be converted to an enum?  If so, should that just be removed in favor of the explicit type() function?

[...]

Issue #167, opened 2015-12-30
MSVC 2015 build fails when attempting to compare object_t


erichkeane:
See #144 and #165 for examples of the build.  

It seems that the following results in a test to see if a value is streamable in catch.hpp:
unit.cpp:2588 CHECK(*p1 == value.get<test_type>());

This causes a call to operator == in catch.hpp (17771): 

    template<typename RhsT>
    ResultBuilder& operator == ( RhsT const& rhs ) {
    return captureExpression<Internal::IsEqualTo>( rhs );
    }

THAT causes a call to the captureExpression function on catch.hpp:1829

    template<Internal::Operator Op, typename RhsT>
    ResultBuilder& captureExpression( RhsT const& rhs ) {
        return m_rb
            .setResultType( Internal::compare<Op>( m_lhs, rhs ) )
             .setLhs( Catch::toString( m_lhs ) )
            .setRhs( Catch::toString( rhs ) )
            .setOp( Internal::OperatorTraits<Op>::getName() );
    }

The Catch::toString results in a call to catch.hpp: 1731's toString->StringMaker:
    
    template<typename T>
    std::string toString( T const& value ) {
    return StringMaker<T>::convert( value );
    }

StringMaker<T> simply offloads to Detail::StringMakerBase, but first goes through the Detail::IsStreamInsertable<T>::value!  This happens on 1575:

     template<typename T>
     struct IsStreamInsertable {
         static std::ostream &s;
         static T  const&t;
         enum { value = sizeof( testStreamable(s << t) ) == sizeof( TrueType ) };
     };

On Linux, this appears to realize there is NO overload for operator << that works, however MSVC doesn't seem to know that, and decides that it is an ambiguous call between just about every possible overload.

erichkeane:
A few more comments before I call it a day:
1- I find it odd that the == operator is coming from catch, rather than comparing the items as std::map objects.  Do we know what clang/gcc does with these?

2- It seems that the IsStreamInsertable intends for the proper overload to be:

    FalseType operator<<(std::ostream const&, BorgType const&);

BorgType has a CTOR that should just work for that:

    template<typename T> BorgType (T const &);

This should be an implcit conversion to BorgType, which should result in the operator << being called above.  That version DOES come up in the function lookup (see error on line 82):

    c:\projects\json\test\catch.hpp(1569): note: or       'Catch::Detail::FalseType Catch::Detail::operator <<(const std::ostream &,const Catch::Detail::BorgType &)'

[...]

PR #183, opened 2016-01-17
Extending support to full uint64_t/int64_t range - issue #178 (Integer conversion to unsigned)

twelsby:
This pull request is to address issues raised in issue #178 related to the use of `uint64_t` and `int64_t` types and the inaccuracies involved in parsing, or creating objects with the full range of these values.  Specifically it addresses the following problems:

1. Creating an object with a `uint64_t` value greater than (2<sup>63</sup>) -1 will cause it to be represented as an incorrect negative value internally, and subsequently serialized as the negative value.
2. Numbers are parsed as a `double` initially and then converted to integers if possible, so parsing a number greater than (2<sup>53</sup>) - 1 or less than -(2<sup>53</sup>) + 1 will result in loss of precision.  While this behavior is actually endorsed by RFC7159, due to the common use of `uint64_t`/`int64_t` types in modern software it would be preferable if the full range could be supported with full precision.

The pull request contains extensive changes to the code base however the majority of these are merely to introduce a new type, `number_unsigned_t`, which is required to support the upper range of `uint64_t`/`int64_t` values.  The flow on effect of this is the addition of companion constructors, operators and miscellaneous code that is required to implement the new type.  There are also very substantial changes to the unit testing (now just shy of 8000 assertions), again to accommodate the new type.  This new code is in the same form as that for the existing types so there will be no surprises.  A small number of additional unit tests check the performance of the parsing against the full range of the `uint64_t`/`int64_t` types.

The most controversial change is the change to the number parsing function `get_number()`.  This now attempts to parse the number both as an integer using `std::strloull()`/`std::strtoll()` and also using `std::strtold()` and then stores using the best representation.  This method was presented in the discussion for issue [#178](https://github.com/nlohmann/json/issues/178).

This method is far from ideal however the most likely alternatives seem to be either writing some custom (non-standard library based) number parsing function, or attempting to scan ahead in the string before parsing, or modifying the lexer to make the determination, or just accepting the reduced range.  Neither of these alternatives seems ideal.


nlohmann:
Hi @twelsby, thanks for your work! I'll have a look at the diff. In the meantime, could you check why the unit tests fail on Travis?

src/json.hpp.re2c
```diff
@@ -2027,20 +2197,43 @@ class basic_json
...
+    {
+        return m_type == value_t::number_unsigned;
     }
 
     /*!
     @brief return whether value is a floating-point number
 
-    This function returns true iff the JSON value is a floating-point number.
-    This excludes integer values.
+    This function returns true if the JSON value is a floating-point number.
```
nlohmann: In fact, I used "iff" as (common?) abbreviation for "if and only iff" here.

twelsby: Sorry I didn't know that one and thought it was a typo.

src/json.hpp.re2c
```diff
@@ -618,7 +700,8 @@ class basic_json
         boolean,        ///< boolean value
         number_integer, ///< number value (integer)
         number_float,   ///< number value (floating-point)
-        discarded       ///< discarded by the the parser callback function
+        discarded,      ///< discarded by the the parser callback function
+        number_unsigned ///< number value (unsigned integer)
```
nlohmann: Could you please move `number_unsigned` behind `number_integer` (with the required adjustment in the ordering), just to have the numbers grouped together?

src/json.hpp.re2c
```diff
@@ -6906,17 +7178,40 @@ class basic_json
 
         @throw std::range_error if passed value is out of range
         */
-        long double get_number() const
+        void get_number(basic_json& result) const
```
nlohmann: In the documentation, there is still the `@return` part which makes no sense any more as the function is `void`. Could you please adjust the documentation to described the side effect of the function?

src/json.hpp.re2c
```diff
@@ -6906,17 +7178,40 @@ class basic_json
...
 
-            // return float_val if the whole number was translated and NAN
-            // otherwise
-            return (reinterpret_cast<lexer_char_t*>(endptr) == m_cursor) ? float_val : NAN;
+            // Parse it as a double
+            const auto float_val = strtold(reinterpret_cast<typename string_t::const_pointer>(m_start),&endptr);
+            long double int_part;
+            const auto frac_part = std::modf(float_val, &int_part);
+
+            // Test if the double or integer is a better representation
```
nlohmann: Could you please add a comment to briefly describe the three parts of the disjunction?

src/json.hpp.re2c
```diff
@@ -4618,14 +4837,16 @@ class basic_json
     */
     friend bool operator<(const value_t lhs, const value_t rhs)
     {
-        static constexpr std::array<uint8_t, 7> order = {{
+        static constexpr std::array<uint8_t, 9> order = {{
```
nlohmann: See the comment regarding the order inside the enum above. I'm not sure if still 9 values are required here.

src/json.hpp.re2c
```diff
@@ -1225,14 +1320,72 @@ class basic_json
...
+    @ref number_unsigned_t.
+
+    @param[in] val  an unsigned integer to create a JSON number from
+
+    @complexity Constant.
+
+    @sa @ref basic_json(const number_unsigned_t) -- create a number value
+    (unsigned)
+
+    @since version 1.0.0
```
nlohmann: Please make sure to use `@since version 1.0.1` in all functions you added.

src/json.hpp.re2c
twelsby: I put it at the end because I was concerned that renumbering the enum might break existing code that was expecting certain types to have certain numerical values.  I agree that after number_integer would be a much better position aesthetically. If you are happy that the renumbering won't cause any flow on problems I'll change it.  The added benefit is that it removes the necessity to insert a filler value in the vector that is used for the type ordering comparison (is null < bool etc).

The Travis problem is caused by Valgrind - I'm not sure why but it also happens when I run the unit tests locally under Valgrind.  For some reason Valgrind is truncating the parsed value.  Seems rather bizarre but I'm looking into it.

nlohmann:
I looked at the diffs and adde some comments. Wow - I am impressed how much effort you put to extend all the unit tests!

src/json.hpp.re2c
nlohmann: Hm. The values of the enum are actually available via public member functions. As this may break existing code, then the extension should be in the 2.0.0 release. I'd rather have clean code than something ugly just to have this edge case covered by a 1.x.x release.

About the tests: it would be great to convince Valgrind to accept the code. If this is some kind of bug, then you give these tests a label so they can be skipped when run with Valgrind...

twelsby: Ok, I'll change the order and add from 2.0.0 to the changes.

I've solved the Valgrind problem.  It was down to documented Valgrind behaviour, which is unable to deal with floating point calculations involving `long double` and as a result it was truncating them to `double` and causing the `if(int_part != m_value.number_unsigned)` test to return true which caused the result to be stored as a `number_float_t` instead of a `number_unsigned_t`. For some reason this only happened with the unsigned case... but in any event it doesn't matter, the solution is simply to parse with `strtod` instead of `strtold`, which I should have been doing anyway.  Introducing a `long double` just introduces compiler dependant behaviour (because VS treats them as `double` by default).

While debugging that issue I came up with a better way of parsing which I am still exploring.  It  only requires one parse for integers and has a much simpler (and importantly more obvious) decision methodology.  I would have had it up and running already only I made the fateful decision to upgrade my Ubuntu installation...

Furcube:
Awesome amount of changes.

A little addition: you should set errno = 0 before calling strto<something> functions.
From linux man page for a strtoull:
```
NOTES
       Since strtoul() can legitimately return 0 or LONG_MAX (LLONG_MAX for strtoull()) on both success and failure, the calling program should set errno to 0 before the call, and then determine if an  error
       occurred by checking whether errno has a non-zero value after the call.
```

twelsby:
Good pickup @Furcube I completely forgot about that.  I'll fix it.

Furcube:
result.m_value.number_integer = std::strtoll

This will be very bad if NumberIntegerType is not a int64_t, but int32_t for example
Need to change conversion function to appropriate or do range check additionaly to ERANGE.

twelsby:
@Furcube, another good point.   I think the best way would be to store it in a temporary `int64_t`` /`uint64_t`, check the range of `NumberIntegerType` and cast to either `NumberIntegerType` or `NumberFloatType` as appropriate.  One extra `if`/`else`, so not a huge problem.

twelsby:
The latest commit has the following changes:

1. Fixed the errno issue identified by @Furcube 
2. Fixed the issue with small integers identified by @Furcube 
3. Cleaned things up a bit to increase code reuse using a helper function
4. Implemented the fix referred to in issue #167 - AppVeyor now passes
5. Removed the use of `approx` as discussed in issue #185.  Understandably this will cause angst for some but as I stated in the issue it isn't doing what it is designed for and what it is designed for shouldn't be done.  I look forward to comments on the issue and have no issues continuing with the current way of doing things if any concerns cannot be placated.

twelsby:
Latest build adds a fix for #171 by adding two extra template overloads for operator[] for T* arguments.  Seems to work well with no apparent side effects.

src/json.hpp.re2c
```diff
@@ -65,6 +65,12 @@ Class @ref nlohmann::basic_json is a good entry point for the documentation.
     #endif
 #endif
 
+// disable float-equal warnings on GCC/clang
+#if defined(__clang__) || defined(__GNUC__) || defined(__GNUG__)
+    #pragma GCC diagnostic push
+    #pragma GCC diagnostic ignored "-Wfloat-equal"
+#endif
+
```
nlohmann: I think this part with the `#ifdef`s and `#pragma`s is really ugly... I am not 100% convinced yet to have something like this in the library...

nlohmann:
Hi @twelsby, could you please check #186 with your `get_number` function?

src/json.hpp.re2c
twelsby: This is necessary to suppress floating point warnings under gcc/clang.  The only real alternatives are to:

1. Accept the warnings - they are only warnings after all (ugly though)
2. Change the build settings in Travis.yml so that -Wfloat-equal is omitted at the command line - this works for us but if the user decides to compile with the warning, and doesn't understand floating point, then they may come to the conclusion that this is a bug in our code
3. Add the #pragma without the conditional block - this creates its own warning about unknown #pragma under VS2015
4. Reinstate approx()
5. Come up with some other means of avoiding the warning if such a thing exists

I will revert this for now and add it to its own pull request as it is really a separate issue.

twelsby:
@nlohmann, can you please close this pull request.  I accidentally created it from my master which is causing all sorts of problems as I would like to point master elsewhere but can't while this is open.  I will create another pull request with just the changes relevant to this subject and add a reference to this one.  Unfortunately github does not allow changing the pull request source branch (although its an open feature request).

nlohmann:
Closed as requested by @twelsby.

Issue #128, opened 2015-10-02
Question about get_ref()


dariomt:
I see there were some discussions about adding get_ref(), but I can't see it in the code.
What happened to get_ref()?

I believe that having get_ref() is convenient and safe. It could merely call get_ptr() and throw if it returns nullptr. It makes it easier to traverse a json object when I'm pretty sure about the types it holds. Otherwise I have to constantly check against nullptr myself.

I'd like the following code to work seamlessly:
```cpp
void func(const std::string&);
json value = "foobar";
func( value.get_ref<const std::string>() );
```

Currently I'd have to write the following:
```cpp
void func(const std::string&);
json value = "foobar";
auto p = value.get_ptr<const std::string*>();
if (!p) throw some_exception;
func( *p );
```

Would you accept a PR for this functionality?

nlohmann:
After a getter for a value and a pointer I did not follow to implement `get_ref`. But I would be glad to have a PR for that.

dariomt:
I'll try to provide a PR including tests and docs.

One question: what would be the preferred form?
`get_ref<const std::string>()`
or
`get_ref<const std::string &>()`

To me get_**ref** is enough to convey that we get a reference, the `&` in the template parameter is redundant (I feel the same way about the `*` required by get_**ptr** by the way).

dariomt:
I have this ready in my branch get_ref.
But I have created that branch off my master where I have all my other PRs already merged.
Do you know how I can produce a PR with only the actual changes done in my branch?

nlohmann:
Hi @dariomt, sorry for not answering earlier.

As for your first question: I would prefer `get_ref<const std::string &>` - it makes the getting of a reference explicit.

For your second question, I unfortunately have no answer.

nlohmann:
Any news here?

dariomt:
My changes already implemented `get_ref<const std::string &>`, so that's good.

However, I'm still trying to get the branch in proper shape to make the PR easy for you, git newbie here  :)

nlohmann:
Ok - looking forward to it!

nlohmann:
@dariomt - any news on this?

dariomt:
Sorry, I haven't had time to look into this.
I should have time this week.

dariomt:
PR ready!
I think I've merged everything correctly.

nlohmann:
Thanks @dariomt for the good job!

PR #184, opened 2016-01-18
Implementation of get_ref()

dariomt:
As per discussion about issue #128


nlohmann:
Hi @dariomt, thanks for the PR. Could you please have a look at the issues raised by Travis?

dariomt:
The travis failures seem related to missing ```typename```

Hum... the error says:
```
error: need ‘typename’ before ‘std::add_pointer<ReferenceType>::type’ because ‘std::add_pointer<ReferenceType>’ is a dependent scope
   using PointerType = std::add_pointer<ReferenceType>::type;
```
I thought ```typename``` was not needed in a ```using``` context, because the compiler *knows* I'm asking for a type there! Doh!

I can only test my changes locally with MSVC. 
Is there a way to test my changes on clang/gcc remotely before making the PR?

dariomt:
Trying to compile in VS2015 I get the errors shown in appveyor (```test\catch.hpp(1575): error C2056: illegal expression```)
It seems some change in catch.hpp is problematic :(
But this should be failing in master too!

dariomt:
OK, now both clang and gcc are happy!  :)
I'll wait for #167 to be resolved before checking the MSVC status.

nlohmann:
Thanks a lot!

PR #189, opened 2016-01-23
Fixed Issue #171 - added two extra template overloads of operator[] for T* arguments

### LINKED ISSUE #171: Cannot index by key of type static constexpr const char*

twelsby:
Fix #171 by adding two extra template overloads for operator[] for T* arguments. Seems to work well with no apparent side effects.  Note that this is dependent on the small change in pull #188 in order for it to pass AppVeyor.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -2675,7 +2675,9 @@ class basic_json
              std::enable_if<
                  not std::is_pointer<ValueType>::value
                  and not std::is_same<ValueType, typename string_t::value_type>::value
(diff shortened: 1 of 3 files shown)


nlohmann:
Thanks a lot!

Issue #186, opened 2016-01-22
Bugs in miloyip/nativejson-benchmark: floating-point parsing


nlohmann:
I checked the library with the latest https://github.com/miloyip/nativejson-benchmark benchmark suite. There are several errors parsing floating-point numbers:

- Parse Double,Nlohmann (C++11),double02,false
- Parse Double,Nlohmann (C++11),double37,false
- Parse Double,Nlohmann (C++11),double40,false
- Parse Double,Nlohmann (C++11),double44,false
- Parse Double,Nlohmann (C++11),double48,false
- Parse Double,Nlohmann (C++11),double53,false
- Parse Double,Nlohmann (C++11),double58,false
- Parse Double,Nlohmann (C++11),double63,false

The respective code is this:

```cpp
// Parse Double
{
    using rapidjson::internal::Double;
    int i = 1;
    #define TEST_DOUBLE(json, expect)\
    {\
        bool result = false;\
        double actual = 0.0;\
        if (test.ParseDouble(json, &actual)) \
            result = Double(expect).Uint64Value() == Double(actual).Uint64Value();\
        printf("double%02d: %s\n", i, result ? "true" : "false");\
        if (!result)\
            printf("JSON: %s\nExpect: %17g\nActual: %17g\n\n", json, expect, actual);\
        fprintf(fp, "2. Parse Double,%s,double%02d,%s\n", test.GetName(), i, result ? "true" : "false");\
        i++;\
    }
    TEST_DOUBLE("[0.0]", 0.0);
    TEST_DOUBLE("[-0.0]", -0.0);
    TEST_DOUBLE("[1.0]", 1.0);
    TEST_DOUBLE("[-1.0]", -1.0);
    TEST_DOUBLE("[1.5]", 1.5);
    TEST_DOUBLE("[-1.5]", -1.5);
    TEST_DOUBLE("[3.1416]", 3.1416);
    TEST_DOUBLE("[1E10]", 1E10);
    TEST_DOUBLE("[1e10]", 1e10);
    TEST_DOUBLE("[1E+10]", 1E+10);
    TEST_DOUBLE("[1E-10]", 1E-10);
    TEST_DOUBLE("[-1E10]", -1E10);
    TEST_DOUBLE("[-1e10]", -1e10);
    TEST_DOUBLE("[-1E+10]", -1E+10);
    TEST_DOUBLE("[-1E-10]", -1E-10);
    TEST_DOUBLE("[1.234E+10]", 1.234E+10);
[...]

Issue #187, opened 2016-01-22
Bugs in miloyip/nativejson-benchmark: roundtrips


nlohmann:
I checked the library with the latest https://github.com/miloyip/nativejson-benchmark benchmark suite. There are several errors in the round trip tests:

- Roundtrip,Nlohmann (C++11),roundtrip20,false
- Roundtrip,Nlohmann (C++11),roundtrip21,false
- Roundtrip,Nlohmann (C++11),roundtrip24,false
- Roundtrip,Nlohmann (C++11),roundtrip25,false
- Roundtrip,Nlohmann (C++11),roundtrip26,false
- Roundtrip,Nlohmann (C++11),roundtrip27,false

These are the values:

- `[0.0]`
- `[-0.0]`
- `[5e-324]`
- `[2.225073858507201e-308]`
- `[2.2250738585072014e-308]`
- `[1.7976931348623157e308]`

I will have a deeper look at this later. I just wanted to paste the results here before I forget.

nlohmann:
I wrote a little tool for the roundtrips:

```cpp
#include <json.hpp>

using nlohmann::json;

int main()
{
    json j;
    std::cin >> j;
    std::cout << j << std::endl;
}
```

Here are the results:

- [ ] `[0.0]` -> `[0]`
- [ ] `[-0.0]` -> `[0]`
- [ ] `[5e-324]` -> `[4.94065645841247e-324]`
- [ ] `[2.225073858507201e-308]` -> `[2.2250738585072e-308]`
- [ ] `[2.2250738585072014e-308]` -> `[2.2250738585072e-308]`
- [ ] `[1.7976931348623157e308]` -> `[1.79769313486232e+308]`

Indeed, the output differs from the input.

nlohmann:
The first two items (`[0.0]` and `[-0.0]`) are parsed and stored as integer `0`, because no precision is lost. As JSON does not define an internal representation of numbers and the value (zero) is the same, I think it's rather nitpicky to treat the output as an error.

nlohmann:
FYI: The roundtrips are already part of the unit tests (test case ""compliance tests from nativejson-benchmark", section "roundtrip"), where test case 13, 18-21, and 24-27 are skipped.

twelsby:
Most of these are silly tests in my opinion and seem pretty contrived.

[...]

PR #191, opened 2016-01-24
Fixed Issue #186 - add strto(f|d|ld) overload wrappers, "-0.0" special case and FP trailing zero

twelsby:
This pull request contains the following changes that have been discussed in issues #186 and #187:

1. `get_number()` now calls overload wrappers for `strtof()`, `strtod()` and `strtold(), instead of calling `strtold()` directly.  This prevents the double rounding that causes errors in particular cases where `number_float_t` is not a `long double` due to the cast that is then required.

2. `parse_internal()` now tests for the special case of `-0.0` and ensures that such numbers are stored as a float as they cannot be accurately represented with integer types.

3. `dump()` now tests for floating point numbers that do not have a fractional component (e.g. `1.0`, `42.0`, `-36.0`) and in this case uses a `std::fixed` and `std:setprecision(1)` representation.  This is necessary for certain round trip tests to pass successfully.


nlohmann:
I see that the Travis builds still fail and there is some debug output. Could you please drop a note once this is ready to merge?

twelsby:
Note that this PR is currently a work in progress.  It passes all tests locally including using identical settings as used by Travis for both CLANG 3.7 and GCC 4.9 (as well as GCC 5.2.1) and, obviously, AppVeyor.  The only test that fails under Travis is a new test for "1.00000000000000011102230246251565404236316680908203126" which is expected to parse to 1.00000000000000022 but instead parses to exactly 1.

The DEBUG ONLY commit (f79d52b) results shows that it is only the least significant bit that differs.  The interesting thing is that doing a simple local test using `strtod()` under GCC produces the expected result of 1.00000000000000022 however if it is first parsed using `strtold()` and then cast to a `double` then the result is exactly 1.  This hints that an earlier version of libstdc++ may be being used in Travis that uses an implementation of `strtod()` that itself calls `strtold()` and then does the cast.  Unfortunately it is not clear what version of libstdc++ Travis is using.  This matter is still being investigated but one solution would be simply to only run this particular test under AppVeyor as this implementation still has better compliance than the current master (which currently also fails this test).

twelsby:
Debug dumps added to pull #193 reveal that the error is definitely coming from `strtod()` so it cannot be fixed from the library end.  The only solution is to exclude that test under GCC/clang until Travis is updated.

twelsby:
@nlohmann, the issue is now resolved and this is ready for merge.

nlohmann:
Thanks a lot!!

Issue #195, opened 2016-01-25
GCC/clang floating point parsing bug in strtod()


twelsby:
I have managed to identify the source of the failure to parse the following number:

`1.00000000000000011102230246251565404236316680908203126`

which should parse to:

`1.00000000000000022`

However under gcc/clang *on Travis only* it instead parsed as exactly `1`.  Local gcc/clang builds and AppVeyor builds all ran correctly.

This was mentioned in issue #186 where I speculated that it was likely caused by a bug in `strtod()` in what must have been an older version on Travis.  To address that issue and get the rest of the tests to run on Travis, pull #191 (now merged) disabled this test for gcc/clang.

Research identified the following blog post: [How strtod() Works (and Sometimes Doesn’t)](http://www.exploringbinary.com/how-strtod-works-and-sometimes-doesnt/), which indicates that it was caused by the double rounding error that was suspected all along to have caused it.  Although the bug was fixed back in 2013, it appears that the bug is still present in both gcc and clang on the version of Ubuntu on which the Travis builds run (precise).

Travis also has a more recent version of Ubuntu (trusty), which is currently in beta.  I decided see if upgrading `.travis.yml` to build on trusty instead might address the bug and allow the test to also be done under gcc/clang - it works.  The trusty environment is beta but there do not appear to be any issues with the build and it is inevitable that we will need to migrate to trusty eventually so it might as well be now.

nlohmann:
Fixed by merging #196.

PR #196, opened 2016-01-25
Issue #195 - update Travis to Trusty due to gcc/clang strtod() bug

twelsby:
This pull request updates Travis to use the Ubuntu Trusty distribution instead of Precise.  This is required due to the bug in `strtod()` that is uncorrected in Precise builds of the standard library but fixed in Trusty.  This bug causes a single bit rounding error that may occur in as many as 1 in 4096 of the possible double values.

Note that currently this PR still uses the `llvm-toolchain-precise*` repositories because the  `llvm-toolchain-trusty*` ones have not been added to the Travis white-list.  This works because clang is not the problem, rather it is the standard library. 

There is an open issue on Travis [APT source whitelist request for llvm-toolchain-trusty #199](https://github.com/travis-ci/apt-source-whitelist/issues/199
) to add the repositories to the white-list at which point `.travis.yml` can be further updated.  It is suggested that perhaps this pull be merged but issue #195 be left open until the trusty llvm repos can be added as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ matrix:
       compiler: clang
       addons:
         apt:
-          sources: ['ubuntu-toolchain-r-test', 'llvm-toolchain-trusty-3.6']
+          sources: ['ubuntu-toolchain-r-test', 'llvm-toolchain-precise-3.6']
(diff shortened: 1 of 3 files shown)


nlohmann:
Thanks!

Issue #199, opened 2016-01-30
Small bugs in json.hpp (get_number) and unit.cpp (non-standard integer type test)


twelsby:
There is a small bug in the following area of json.cpp:
```c++
if (attempt_cast(std::strtoll(reinterpret_cast<typename string_t::const_pointer>(m_start), 
                             &endptr, 10), result.m_value.number_unsigned))
```
Here `result.m_value.number_unsigned` should be `result.m_value.number_integer`.

Then in unit.hpp there is another bug:
```c++
// integer object creation - expected to wrap and still be stored as an integer
j = -2147483649LL; // -2^31-1
CHECK(static_cast<int>(j.type()) == static_cast<int>(custom_json::value_t::number_integer));
CHECK(j.get<int32_t>() == 2147483647.0);  // Wrap

// integer parsing - expected to overflow and be stored as a float
j = custom_json::parse("-2147483648"); // -2^31
CHECK(static_cast<int>(j.type()) == static_cast<int>(custom_json::value_t::number_float));
CHECK(j.get<float>() == -2147483648.0);
```
The second test for parsing uses an incorrect value.  Instead of `-2147483648` it should read `-2147483649`.

The previous value `-2147483648` should not have wrapped as was intended however the bug in json.hpp had the effect of causing the cast to fail anyway.  The second error effectively hid the first error so the unit test passed.

One other small error is that all of the floating point comparisons in this section should have an 'f' after the literals otherwise they are double literals - so the above comparison would read `CHECK(j.get<float>() == -2147483650.0f`.

nlohmann:
Thanks a lot!

PR #200, opened 2016-01-30
Fixed issue #199 - Small bugs in json.hpp (get_number) and unit.cpp (non-standard integer type test)

twelsby:
Issues raised in #199 are all addressed.


nlohmann:
Thanks a lot!

PR #206, opened 2016-02-10
Fixed stream precision issue.

jdisset:
Double->text->double or float->text->float round trip conversions were giving inexact results. Max_digits10 is the correct limit.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -5986,10 +5986,10 @@ class basic_json
             {
                 // If the number is an integer then output as a fixed with with
                 // precision 1 to output "0.0", "1.0" etc as expected for some
-                // round trip tests otherwise  15 digits of precision allows
+                // round trip tests otherwise 17 digits of precision allows
                 // round-trip IEEE 754 string->double->string; to be safe, we
                 // read this value from
-                // std::numeric_limits<number_float_t>::digits10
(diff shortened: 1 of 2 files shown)


jdisset:
Hi,
Sorry, I forgot to run the tests as I first made this small change for my own purposes and didn't think it would break anything. It seems some of the tests relied on what I think is an incorrect behavior:

`            auto s = json(42.23).dump();
            CHECK(s.find("42.23") != std::string::npos);
`
only passes because of the rounding that was happening when using a lower precision for double->text conversions. Now, when using better precision for the decimal representation of the double we actually get "42.229999999999997".
My suggestion would be to change this test (and the other one which relies on the same behavior) and maybe rather test (with the same example values) the correct double->text->double round trip conversion instead.

Cheers and congrats for an amazing library,
Jean.

nlohmann:
Hi @jdisset, thanks for the PR. I see that also a few roundtrip tests are failing. The input there is from miloyip/nativejson-benchmark, and there already has been a lot of discussion about floating-point numbers, see #187. It would be great if you could comment on this.

Furthermore, it would be great to have @twelsby 's opinion on this.

nlohmann:
Hi @jdisset again. Could you please also have a look at #201? This PR may already have fixed your issue.

nlohmann:
This issue is fixed by PR #201.

PR #207, opened 2016-02-12
Add support for afl-fuzz testing

msm-:
I tested this library under afl last night: total of 400M executions, ~400 code paths found, no crashes or hangs. Stopped when it was >30mins that any of the fuzzers had found a new path. This is the code I used to support the fuzzing, in case you'd like to easily re-run against any future changes.
(I didn't test under ASAN, as cross compiling with -m32 didn't work out of the box)

"make fuzz" creates a simple executable that de-serialises stdin
and re-serialises to stdout.
"make fuzz_testcases" extracts the smaller json test cases into
a testcases directory.

The library can then be fuzzed as follows:
    CC=afl-clang-fast make fuzz
    make fuzz_testcases
    mkdir out
    afl-fuzz -i testcases -o out ./fuzz

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ all: json_unit
 
 # clean up
 clean:
-	rm -f json_unit json_benchmarks
+	rm -f json_unit json_benchmarks fuzz
 
 
 ##########################################################################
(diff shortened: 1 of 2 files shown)


nlohmann:
Thanks a lot!

PR #209, opened 2016-02-16
abstract object/array type out so that arbitrary types can be used

nlohmann:
This pull request fixes some conflicts from #169, but is basically the work of @erichkeane (thanks for that!). I only fixed a conflict and added documentation.

Changes: The container types `object_t` and `array_t` can now be set to any container provided that it is aliased to a type with 2 template arguments (`object_t` - defaulting to `std::map<Key,T>`) and 1 argument (`array_t` - defaulting to `std::vector<T>`).

As the public API changes in a subtle way, the feature can be added in 2.0.0 at the earliest.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -72,6 +72,36 @@ struct has_mapped_type
 }
 
 /*!
+@brief default type for objects
+
+This type is used to store objects if no other type is provided to @ref
+basic_json via the @a ObjectType template parameter. Defaults to `std::map<Key,
+T>`.
+
+@tparam Key  type of the object keys
+@tparam T  type of the object elements
+
+@since version 2.0.0
+*/
+template<typename Key, typename T>
+using default_object_type = std::map<Key, T>;
+
+/*!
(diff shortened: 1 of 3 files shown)


erichkeane:
+1, glad I could help!

nlohmann:
There is a conceptual problem: using an incomplete type in an `unordered_map` (here: using `std::unordered_map<std::string, basic_json>` within the `basic_json` class) is undefined behavior and the code does not compile with GCC 5. The issue already occurred before, but it seems that `std::map` (or at least its implementation in GCC and Clang) is more robust as no hashes are calculated.

Possible solution (http://stackoverflow.com/a/8487164/266378):
> The solution here is to have an unordered map to a pointer:

That is, changing `std::unordered_map<std::string, basic_json>` to `std::unordered_map<std::string, std::shared_ptr<basic_json>>`.

nlohmann:
As noted earlier, there is a conceptual issue that needs to be addressed separately.

Issue #136, opened 2015-10-11
json.hpp:5746:32: error: 'to_string' is not a member of 'std'


Ingener74:
Hello.
On Android i have this messages

* json.hpp:5746:32: error: 'to_string' is not a member of 'std'
* json.hpp:6911:36: error: 'strtold' is not a member of 'std'

android ndk have no to_strign realization

gregmarr:
The first one works fine for us, we use the c++-shared library.

We had to add this function to our codebase to handle the second one.
inline long double strtold(const char *nptr, char **endptr)
{
    return strtod(nptr, endptr);
}

nlohmann:
Hey @Ingener74, could you please tell me which compiler you are using?

Ingener74:
Hello @nlohmann, i am using arm-linux-androideabi-4.9 toolchain for compiling

nlohmann:
Could this maybe solve your problem: http://stackoverflow.com/a/27589053 ?

Ingener74:
Thanks @nlohmann, i known this solution :) but that about strtold? 
I wanted to ask whether it will be added to your library? for android

nlohmann:
Ok, I'll have a look at `strtold`. It would be easier to check on OSX - is there an OSX version of the compiler you use?

Ingener74:
On Mac OS X i use compiler
```
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/usr/include/c++/4.2.1
Apple LLVM version 7.0.0 (clang-700.0.72)
Target: x86_64-apple-darwin14.5.0
Thread model: posix
```
but on mac os everythink compiles ok

nlohmann:
I meant the compiler to create code for Android. Is there a way to cross-compile from OSX?

gregmarr:
You need the android NDK.

gregmarr:
Yes, it is available for OS X.

Ingener74:
Hello @nlohmann i agree with @gregmarr, you need Android NDK for compiling 
https://developer.android.com/ndk/downloads/index.html

ZahlGraf:
Hi,

for a current project I compiled this library with Android NDK r10e, (it includes gcc 4.9) together with gnustl-shared library.
[...]

PR #215, opened 2016-02-28
Update the README to give better instructions for Android.

zewt:
The MingW link doesn't help for Android, and points people down the wrong path.  On Android all you need to do is tell the build system to use a newer compiler and standard library.

The unit tests run successfully with this configuration on Android.

Related to #136.

(I've left Android mentioned in the MingW line, so people searching for "to_string" and finding that line are pointed upwards.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,16 @@ I would be happy to learn about other compilers/versions.
 Please note:
 
 - GCC 4.8 does not work because of two bugs ([55817](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55817) and [57824](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57824)) in the C++11 support.
-- For GCC running on MinGW or Android SDK, the error `'to_string' is not a member of 'std'` (or similarly, for `strtod`) may occur. Note this is not an issue with the code,  but rather with the compiler itself. Please refer to [this site](http://tehsausage.com/mingw-to-string) and [this discussion](https://github.com/nlohmann/json/issues/136) for information on how to fix this bug.
+- Android defaults to using very old compilers and C++ libraries.  To fix this, add the following to your Application.mk.  
+This will switch to the LLVM C++ library, the Clang compiler, and enable C++11 and other features disabled by default.
+
+```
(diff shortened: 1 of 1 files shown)


zewt:
I've hit an issue with LLVM on Android that can't be worked around (waiting for an upstream fix, which will take a long time to trickle into Android), so I also worked around the issues with gnustl on Android.  It's a simple workaround like the older one, but without polluting std.  I won't send a separate PR to avoid spamming, but I'll just link it here for you to choose from: https://github.com/zewt/json/tree/android-gnustl

I think it's better to just work around the issue so the library can be used.  I think that Clang is still needed (this just avoids requiring the LLVM C++ library), but it probably wouldn't be hard to make g++-4.9 work too.  Spent too much time trying to make LLVM work with my project today, so I need to move on for the time being.

nlohmann:
Hi @zewt, thanks for the effort. I'll check the changes this weekend. One question though: Is there an OSX version of the Android compiler to download so I can use it on my local Jenkins?

zewt:
The Android tools run on OSX.  I suppose you could run the tests in the emulator, but that's probably a bit of a project to automate unless there's existing support for it.

kittinunf:
I got this error when compile json.hpp with Android NDK.

`.../vendors/json/src/json.hpp:7635: error: undefined reference to 'strtold'`

My Android configuration file is

```
NDK_TOOLCHAIN_VERSION := clang3.6
APP_PLATFORM := android-16
APP_ABI := armeabi-v7a,x86
APP_STL := c++_shared
```

Do you have another config to make this work?

whackashoe:
@kittinunf there might be a better way, but an easy way is to yank it out of musl or something.

zewt:
strtold might be missing in llvm.  I don't remember, since I'm building with gnustl now (due to a showstopper bug in llvm, https://llvm.org/bugs/show_bug.cgi?id=26012).

I think strtold should only be needed if NumberFloatType is long double, which it probably isn't.  You can probably just comment out the functions calling it.

nlohmann:
Thanks a lot!

PR #218, opened 2016-03-02
Corrected readme

Annihil:
I found an error in the example of usage when casting map STL container to json, leading users to think that it keeps insertion order instead of alphabetical order of the keys.


nlohmann:
Thanks for noting!

PR #226, opened 2016-03-28
Use namespace std for int64_t and uint64_t

lv-zheng:
According to the c++11 standard, the declarations are within namespace
scope of the namespace `std`. Add `std::` to avoid unnecessary
assumptions of implementations.

References:
http://stackoverflow.com/questions/13129945/use-types-in-cstdint-with-ot-without-namespace
http://en.cppreference.com/w/cpp/header/cstdint
https://msdn.microsoft.com/en-us/library/hh874765.aspx


nlohmann:
Thanks a lot!

PR #234, opened 2016-04-13
Make the indirection and dereference operators of iterator const

### LINKED ISSUE #233: Can't use basic_json::iterator as a base iterator for std::move_iterator
While it's possible to move a range of container elements instead of copying them by using the [std::move](http://en.cppreference.com/w/cpp/algorithm/move) algorithm instead of [std::copy](http://en.cppreference.com/w/cpp/algorithm/copy), some copying algorithms like [std::copy_if](http://en.cppreference.com/w/cpp/algorithm/copy) and [std::copy_n](http://en.cppreference.com/w/cpp/algorithm/copy_n) doesn't have a move counterpart. The only way to use move semantics with these algorithms is to wrap the input iterators into [std::move_iterator](http://en.cppreference.com/w/cpp/iterator/move_iterator) objects.

```c++
json source = {"a", "b", "c"};  
json dest;
std::copy_n(std::make_move_iterator(source.begin()), 2, std::back_inserter(dest));
```

Unfortunately, creating a move_iterator from a basic_json::iterator is not possible, because the indirection operator of move_iterator is a const member function and it can't call the non-const indirection operator of basic_json::iterator.
Compiling the code above with AppleClang 7.0.0 (which is based on clang 3.7 I think), will produce the following error:

```c++
../include/c++/v1/iterator:959:37: error: indirection requires pointer operand ('const nlohmann::basic_json<std::map, std::vector, std::__1::basic_string<char>, bool, long long, unsigned long long, double, std::allocator>::iterator' invalid)
      return static_cast<reference>(*__i);
                                    ^~~~
```
                                    
The issue can be fixed by making the indirection operator of basic_json::iterator const, and while we're at it, we can safely make the dereference operator const too, since the non-const iterators in STL have a const indirection and dereference operator too.

robertmrk:
This is an attempt to fix #233, by making the indirection and dereference operators of `basic_json::iterator` `const`.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -7000,13 +7000,13 @@ class basic_json
         }
 
         /// return a reference to the value pointed to by the iterator
-        reference operator*()
+        reference operator*() const
         {
             return const_cast<reference>(base_iterator::operator*());
         }
 
         /// dereference the iterator
-        pointer operator->()
+        pointer operator->() const
         {
             return const_cast<pointer>(base_iterator::operator->());
         }
--- a/src/json.hpp.re2c
+++ b/src/json.hpp.re2c
@@ -7000,13 +7000,13 @@ class basic_json
         }
 
         /// return a reference to the value pointed to by the iterator
-        reference operator*()
+        reference operator*() const
         {
             return const_cast<reference>(base_iterator::operator*());
         }
 
         /// dereference the iterator
-        pointer operator->()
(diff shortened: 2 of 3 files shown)


nlohmann:
Thanks a lot!

robertmrk:
Don't mention it. I'm glad I could help.

Issue #237, opened 2016-04-25
Additional integration options


robertmrk:
At the moment the only way to integrate the library into a client project is to add the `json.hpp` file to the source tree of the project. In some use cases that might not be the ideal solution.
For instance I'm working on a JSON processing library that uses the nlohmann/json library not only internally but in it's API as well. So client applications which would wan't to use the library would either have to use the `json.hpp` file supplied by the JSON processing library, which is awkward, or if they are already using the `json.hpp` file they can create and pass `json` variables to the library which can cause unpredictable problems if the version of the `json.hpp` file used by the client application and the JSON processing library differs.
Fortunately, CMake has a solution to the above problems. 
With a few lines of CMake code we can create an [interface library](https://cmake.org/cmake/help/v3.0/manual/cmake-buildsystem.7.html#interface-libraries) (a library target that doesn't generates any output, but it can be used to store usage requirements, which in our case would be the include directory of the `json.hpp` file) and export it. This way by including the nlohmann/json library into the client application with `add_subdirectory`, all targets would be able to find and use the library with the `find_package` command.
```CMake
find_package(json)
add_executable(json-client ...) #or add_library(json-client ...)
target_link_libraries(json-client json)
```
With another few lines of CMake we can add an install target to the nlohmann/json project. Users would be able to install it using `make install` and use it, again, with the `find_package` command without including the `json.hpp` file into the project.
So by using `find_package` users can be sure that all the targets that depend on nlohmann/json are using the exact same file, and since we can optionally specify a version number for the `find_package` command too, users can avoid problems that might arise when using different versions.
In the end, we wouldn't lose anything. Users can continue to copy the `json.hpp` file into their projects if they feel that this is the right approach for integrating nlohmann/json into their projects. But we would gain some additional integration options.

mmchen:
Does find_package(json) work already? I am using version 2.0.10, but find_package doesn't quite work. Thanks

nlohmann:
@mmchen I have no idea since I'm no expert in Cmake. I merged PR #238, but I have no idea what this means for `find_package`.

robertmrk:
@mmchen It should work. But the package's name was changed sometimes after this PR got merged to nlohmann_json, so you should try to use `find_package(nlohmann_json)`.

PR #238, opened 2016-04-25
Implement additional integration options

robertmrk:
This PR contains the implementation of the changes suggested in #237.
The listfile is quite different then it was. The required c++ standard is now specified as a target property for the `json_unit` executable which is more portable, and compiler flags and definitions are also specified using target properties using generator expressions instead of variables, but they result in the same compiler flags.
The only longterm change that this PR brings is that from now on the version number of the library must be changed in the listfile as well when a new release is published so CMake can generate the appropriate config version file for the package when it's installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,16 +1,25 @@
 cmake_minimum_required(VERSION 3.0)
 
-project(json CXX)
+# define the project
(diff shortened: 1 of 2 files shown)


nlohmann:
Thanks!

This closes #237.

Issue #230, opened 2016-04-05
Incorrect dumping of parsed numbers with exponents, but without decimal places


nlohmann:
The new floating point representation (#201) has a bug when it comes to negative exponents. Number `2342e-2` is correctly duped as `23.42`, but when it is parsed from a string, the output is `2e01`.

Example code:
```cpp
json j2a = 2342e-2;
json j2b = json::parse("2342e-2");

CHECK(j2a.dump() == "23.42");
CHECK(j2b.dump() == "2342e-2");
```

Output:
```
FAILED:
  CHECK( j2b.dump() == "2342e-2" )
with expansion:
  "2e01" == "2342e-2"
```

@twelsby Any idea about this?

nlohmann:
There is also an issue with other parsed numbers with exponents, but without decimal separator:

`2342e2` and `23420e1` both serialize to `2e05`.

nlohmann:
:exclamation: This bug is currently a blocker for the 2.0.0 release, because the serialization of parsed numbers is not correct.

nlohmann:
@twelsby @gregmarr - Any idea how to fix this?

Kamtron:
Caveat: I'm new to this project (like 1 hour), so please forgive my ignorance.

I drilled down into dump() to try to isolate the problem to either the parser or the formatting.  For this problem case, the underlying data is:
```
m_type.bits.exp_cap    = 0 
m_type.bits.exp_plus   = 0 
m_type.bits.has_exp    = 1 
m_type.bits.parsed     = 1 
m_type.bits.precision  = 0 
m_type.bits.type       = 7 
m_value                = 23.420000
```
It seems to me, that the precision should be 4, not 0.  Am I understanding this correctly?

Kamtron:
Never mind, after looking through the lexer, I realized precision isn't sig figs, but number-of-figs-after-radix in this context.  In which case, 2342e-2 is being lexed and parsed correctly, so I'll look back into the dump formatting again.

Kamtron:
Since we're using snprintf with "%.*e" to generate the string, wouldn't we want 2342e-2 to dump as 2.342e01?  (since it is going to print one leading digit for scientific notation)

If I detect this particular case, and pack the buffer with precision=3, I can get the test to pass.

My spidey sense tells me that we might need a little more logic in the lexer at this point:
```           
            // If no radix point was found then precision would now be set to
            // the number of digits, which is wrong - clear it.
[...]

PR #240, opened 2016-04-29
Patch to fix Issue #230

Kamtron:
## Files to change

   ```sh
   make re2c
   ```

2. [`test/unit.cpp`](https://github.com/nlohmann/json/blob/master/test/unit.cpp) - This contains the [Catch](https://github.com/philsquared/Catch) unit tests which currently cover [100 %](https://coveralls.io/github/nlohmann/json) of the library's code.

   ```sh
   make
   ```

   and can be executed with

   ```sh
   ./json_unit
   ```

## Note

## Please don't


Kamtron:
Oops, I didn't realize how the hpp file was generated.  I will push my changes into re2c instead.

nlohmann:
No worries. If you have a hint how to improve the documentation for committers, please let me know!

Looking forward to your fix!

Kamtron:
Hmm, Travis build is failing because it looks like the build machine can't find cmake.

The builds on my machine and AppVeyor seem ok, though.

Any ideas @nlohmann ? (sorry for being a pain!)

nlohmann:
The reason for the error is that you overwrote the project's Makefile locally and committed this change. Could you please restore the Makefile or just exclude it from your commit?

src/json.hpp
```diff
@@ -8538,9 +8161,8 @@ class basic_json
             result.m_type.bits.parsed = true;
 
             // 'found_radix_point' will be set to 0xFF upon finding a radix
```
nlohmann: The comment needs to be adjusted to the fact that `found_radix_point` is a Boolean now ("set to 0xFF" makes no sense).

src/json.hpp
```diff
@@ -8620,8 +8242,23 @@ class basic_json
...
+            // the number of digits, which is wrong - decrement it to sig figs - 1
+            if(not found_radix_point)
+            {
+                int trailing_zeros = 0;
+                for (int j = precision-1;j>=0;--j)
+                {
+                    auto c = *(m_start+j);
+                    if(c == '0')
+                        ++trailing_zeros;
+                    else if(c > '0' and c <= '9' or c == '-')
```
nlohmann: Clang complains about this the line should be

```cpp
else if ((c > '0' and c <= '9') or c == '-')
```

(added parentheses)

src/json.hpp.re2c
```diff
@@ -7848,9 +7848,8 @@ class basic_json
             result.m_type.bits.parsed = true;
 
             // 'found_radix_point' will be set to 0xFF upon finding a radix
```
nlohmann: The comment needs to be adjusted to the fact that found_radix_point is a Boolean now ("set to 0xFF" makes no sense).

src/json.hpp.re2c
```diff
@@ -7930,8 +7929,23 @@ class basic_json
...
+            // the number of digits, which is wrong - decrement it to sig figs - 1
+            if(not found_radix_point)
+            {
+                int trailing_zeros = 0;
+                for (int j = precision-1;j>=0;--j)
+                {
+                    auto c = *(m_start+j);
+                    if(c == '0')
+                        ++trailing_zeros;
+                    else if(c > '0' and c <= '9' or c == '-')
```
nlohmann: Clang complains about this the line should be

```cpp
else if ((c > '0' and c <= '9') or c == '-')
```

(added parentheses)

test/unit.cpp
```diff
@@ -12740,7 +12742,8 @@ TEST_CASE("regression tests")
 
         CHECK(j2a.dump() == "23.42");
         //issue #230
-        //CHECK(j2b.dump() == "23.42");
+        CHECK(j2b.dump() == "2.342e01");
```
nlohmann: ❗ These two lines do not express the desired behavior: The idea of the adjustment was that dumping parsed numbers yield the same string as before. In the examples, string `""2342e-2""` should be dumped exactly as `"2342e-2"` and not `"2.342e01"`.

Likewise `j3b` should be dumped as `"10E3"`.

nlohmann:
@Kamtron , I added some notes to the changes of `json.hpp.re2c` and `unit.cpp`.

test/unit.cpp
Kamtron: I made  ```CHECK(j2b.dump() == "2.342e01");``` for 2 reasons.

First, to match the existing cases:
```
        CHECK(j3b.dump() == "1E04");
        CHECK(j3c.dump() == "1e04");
```
Second, and more subtly, we fundamentally don't have enough information within ```m_type.bits``` to reconstruct the original string (we would have to know the original exponent).  Because of this, only floating point numbers in scientific notation can make a successful round-trip (hence all of the round-trip regression tests only use scientific notation).

Kamtron: @nlohmann @falemagn 

Opinions on this?

nlohmann:
The roundtrip feature is still buggy and will not make it into 2.0.0. Therefore, I close this PR. Thanks so much @Kamtron for your work. I hope you understand that the this PR does not fix the problem of having "100%" correct roundtrip behavior.

See https://github.com/nlohmann/json/issues/230#issuecomment-227004543 for the discussion.

PR #242, opened 2016-05-11
Out of tree builds and a few other miscellaneous CMake cleanups.

ChrisKitching:
Addresses the stuff I raised in #241. Hopefully this isn't too invasive for you...

I'll attempt to find some time to do more cleanup in the coming days.

Issue #241, opened 2016-05-08
Unit test fails when doing a CMake out-of-tree build


ChrisKitching:
This occurs because it looks for the test data in paths relative to the source tree. Since the test target is not installed, and the test data is not present in the build tree, this fails.

I would suggest making a CMake build option to allow turning off the unit tests, and then making the test data and test binary be installed to the install tree in the conventional way. In-tree builds are, after all, discouraged.

nlohmann:
Hi @ChrisKitching! I only have limited knowledge of CMake and only used it to create a MSVC project for AppVeyor. As my library is header-only, I am not sure why anyone would actually use "my" CMakeLists.txt file, anyway. Before making any changes to it, I would like to understand the benefits for it.

ChrisKitching:
As you know, the cmake script you have builds the tests and then installs the header.

Some plausible uses of this script (my use-case, too, incidentally):
## Pulling in your library as a CMake `ExternalProject` dependency
Even though it's header only, there's still value in pulling it via `ExternalProject` as it is made explicit which version of your library is used. If I copied the header into my own source control system it becomes "detached" from this git repo, which is potentially irritating. It's useful to be able to say in my build script "use the `v1.1.0` tag of nlohmann/json", rather than having to fiddle about making sure I've copied the right version of your header into my source control.
While `ExternalProject` doesn't _require_ CMake, it's very slightly neater if it is, especially because of the next thing...

## Verifying cross-compilation
By building it as an external project, it can be made to automatically pick up the cross-compile flags used in the parent build. This means that if I cross compile my project, I'll automatically end up causing your cmake script to emit a cross-compiled test binary, which will allow me to verify that your library doesn't catch fire when run on whatever silly embedded system I'm using.



Certainly, though, not a big priority. I'm probably going to end up doing this work myself at some point quite soon. If you're interested I'll make a pull request.

nlohmann:
I see. Yes, a PR would be nice. For me, the most important thing is that compiling the unit tests would remain the standard, because otherwise I would have to adjust AppVeyor.

If it is not too much to ask, it would be great if you could double check whether the current Cmakelists.txt is OK and whether it actually has to be that long ;)

PR #247, opened 2016-05-18
Define CMake/CTest tests

robertmrk:
In #241 @ChrisKitching described that tests on out of source builds fail because the test executable can't find the JSON files it needs. It's great that he noticed this and provided a fix for it, I totally missed this issue when I rewrote the CMake lists file.
In #242 the issue is fixed by installing the test executable and test files. There are two problems with this approach:
1. The test executable is used only once by the users of the library. It's executed only after a successful build and never used again, so there is no need to install it on the system level.
2. For executing the tests in out of source builds we don't need to copy any files at all. In CMake we can define tests with the source directory as their working directory, so tests will be able to find the JSON files no matter where the project is built.

In this PR I defined two CMake/CTest test configurations. 
- `json_unit_default` for running the default tests by executing
`json_unit` without any arguments
- `json_unit_all` for running all the tests by executing `json_unit`
with the `"*"` argument

So the tests can be now executed by building the `test` target:
```
make test
```
or by running `ctest` in the build directory.

The tests can be selectively executed as well with `ctest`:
```
ctest -R json_unit_default
```
or
```
ctest -R json_unit_all
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,8 @@ cmake_minimum_required(VERSION 3.0)
 # define the project
 project(nlohmann_json VERSION 2.0.0 LANGUAGES CXX)
 
+enable_testing()
+
 option(BuildTests "Build the unit tests" ON)
 
 # define project variables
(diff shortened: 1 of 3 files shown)


nlohmann:
Thanks a lot! I am lacking Cmake experience, so I am happy about any contributions.

robertmrk:
You're welcome. I'm glad I could help.

Issue #270, opened 2016-06-24
hexify() function emits conversion warning


dtoma:
Hello,

Using gcc `6.1.0`.

Compilation error (`-Wconversion -Werror`):

```
./deps/json.hpp:9995:25:   required from here
./deps/json.hpp:5957:72: error: conversion to 'char' from 'int' may alter its value [-Werror=conversion]
                             : ('a' + static_cast<char>((v - 10) & 0x1f));
                                                                        ^
./deps/json.hpp:5957:72: error: conversion to 'char' from 'int' may alter its value [-Werror=conversion]
```

[hexify](https://github.com/nlohmann/json/blob/master/src/json.hpp.re2c#L5953):

```c++
const auto hexify = [](const int v) -> char
{
  return (v < 10)
  ? ('0' + static_cast<char>(v))
  : ('a' + static_cast<char>((v - 10) & 0x1f));
};
```

Could be written as something like:

```c++
const auto hexify = [](const int v) -> char
{
    static const std::array<char, 16> hex = { '0', '1', '2', '3',
                                              '4', '5', '6', '7',
                                              '8', '9', 'a', 'b',
                                              'c', 'd', 'e', 'f' };
    return hex[v];
};
```

Would you be interested in a PR for such a change?

nlohmann:
Hi @dtoma, thanks for reporting. Yes, a PR would be great!

nlohmann:
Hi @dtoma, if you like, I can add the change myself.

dtoma:
Hi, I'll do it tonight, just got my dev environment setup.

PR #275, opened 2016-06-29
Update hexify to use array lookup instead of ternary (#270)

dtoma:
https://github.com/nlohmann/json/issues/270

Ended up using an array of char rather than std::array, because `std::array::[]` only accepts unsigneds, thus we'd get a conversion warning.


nlohmann:
Thanks!

PR #279, opened 2016-07-09
let the makefile choose the correct sed

murinicanor:
there is no gsed on some systems (i.e. on debian, where i'm building the documentation), so let the makefile choose the correct sed variant

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -1,4 +1,5 @@
 SRCDIR = ../src
+SED:=$(shell command -v sed || gsed)
 
 all: doxygen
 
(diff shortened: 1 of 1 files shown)


nlohmann:
Thanks!

nlohmann:
@murinicanor I realized that on OSX, `sed` and `gsed` are available, but only `gsed` can be used. So the code should be

```
SED:=$(shell command -v gsed || sed)
```

murinicanor:
hi,

i've added a commit to the branch that changes the order of gsed and sed and also uses `which` to have the full path of sed: 0eb9f0cdd533c005ec24770216089d6290f18af2
(sorry, don't know if there's a way to update the pull request after its been merged)

nlohmann:
No worries - I made the change myself.

PR #318, opened 2016-09-26
Include dir relocation

ChristophJud:
Change of CMakeLists.txt and config.cmake.in such that for other projects which include the json package the install-include directory is set.


ChristophJud:
I don't know from where the segfault in unit_test is coming from since I've only changed some minor stuff in the cmake.

nlohmann:
I need to find out why the Xcode build is failing.

nlohmann:
Thanks!

PR #320, opened 2016-09-28
Fix #319

### LINKED ISSUE #319: json_unit runs forever when executed in build directory
It seems to me like a bug that json_unit runs forever when I build the project in a separate build directory with

```text
cmake ../json
make
```
and then try to execute

```text
./test/json_unit
```
in the build directory. 
I would expected it to run the tests and print a result when done which it indeed does if it was build and executed in the source directory.

When I sent an interrupt signal to the running program it terminates with following output:

```test
json_unit is a Catch v1.5.6 host application.
Run with -? for options

-------------------------------------------------------------------------------
constructors
  create a JSON value from an input stream
  std::ifstream
-------------------------------------------------------------------------------
/home/tim/Projekte/github/json/test/src/unit-constructor1.cpp:42
...............................................................................

/home/tim/Projekte/github/json/test/src/unit-constructor1.cpp:1305: FAILED:
due to a fatal error condition:
  SIGINT - Terminal interrupt signal

test cases:   11 |   10 passed | 1 failed
assertions: 3938 | 3937 passed | 1 failed
```

A look at line 1305 in unit-constructor1.cpp
```c++
        SECTION("std::ifstream")
        {
            std::ifstream f("test/data/json_tests/pass1.json");
            json j(f);
        }
```
shows that the input stream test wants to load a file from a subdirectory which doesn't exists in the build directory.

To fix this I added following to 'test/CMakeLists.txt' in the source directory :
```cmake
string(COMPARE NOTEQUAL "${CMAKE_CURRENT_SOURCE_DIR}" "${CMAKE_CURRENT_BINARY_DIR}" JSON_HAS_SEPARATE_BUILDDIR)

if(${JSON_HAS_SEPARATE_BUILDDIR})

file(COPY "${CMAKE_CURRENT_SOURCE_DIR}/data" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")

endif()
```
Now test/data will be copied from the source directory to the build directory when they are not the same.

After rebuilding json_unit finally prints:

```text
All tests passed (8644 assertions in 33 test cases)
```

tboede:
Fixes #319 by adding the changes to CMakeLists.txt in test/ which where suggested in #319.


tboede:
The changes aren't needed anymore, see #319 for details.

PR #324, opened 2016-10-08
make has_mapped_type struct friendly

vpetrigo:
I'd like to propose slightly different implemenation of `struct has_mapped_type` that use Type traits added in C++11. It doesn't change current functionality, but looks more natural.


nlohmann:
Thanks!

Issue #314, opened 2016-09-21
Is there a way to forward declare nlohmann::json?


rachelnertia:
It's preferable not to have to include large header files in other header files, but in my current project I've ended up with json.hpp included in a lot of headers because I can't find a way to forward declare the nlohmann::json type. It's made complicated by all the templating going on. I was wondering if there is actually a way that I just haven't found yet and, if there is, if it could maybe be included as part of the documentation/readme somewhere?

gregmarr:
Looks like this isn't possible for a user to do due to the default template parameters.  The library would need to be updated to provide a forward declaration header with the default template parameters, and remove the defaults from the definition.

SethHamilton:
You could try adding a `#pragma once` to the `json.hpp` right above `#ifndef NLOHMANN_JSON_HPP`. That pragma is supposed to ensure the file isn't even opened on subsequent inclusions. My experience is that compilers are pretty smart about the `#ifndef`.

agauniyal:
Unless you're building your whole project each time, it shouldn't take much time. You could also use ccache to help speed up your build process.

nlohmann:
I think it would be quite cumbersome to realize a forward declaration. All I can think of right now would be a new header with a forward declaration defining a preprocessor symbol and adjusting the current `json.hpp` to use that preprocessor symbol to decide whether the default template parameters should be defined.

ralphtandetzky:
I added a file src/json_fwd.hpp in my [forked repo](https://github.com/ralphtandetzky/json/blob/ralphtandetzky-json_fwd/src/json_fwd.hpp). This header file is minimal and provides a forward declaration of nlohmann::basic_json and adds a typedef nlohmann::json. I've tested it with MinGW and it works. I've also made a pull request to the main repo, so I hope the new header file will be visible to the public there soon.

nlohmann:
@ralphtandetzky Thanks! Just out of curiosity: Does the forward declaration header improve your compilation times? Or maybe @rachelnertia could also have a look at this?

jaredgrubb:
Would it be possible to do something along these lines?
`     struct json : basic_json<...> { using basic_json::basic_json; };
`

I know this is not what std::string does, but it would solve the forward declare problem, and it would solve another annoyance I have. Whenever I try to use a debugger, I see a lot of ugly C++ symbols around json. It would be nice that you only have to "pay for" the symbol noise if you actually do something custom.

For example:

[...]

PR #327, opened 2016-10-12
Fix issue #314: Added json_fwd.hpp.

ralphtandetzky:
The issue asks: "Is there a way to forward declare nlohmann::json?"

This commits helps to do that. The added file helps to reduce compilation dependencies by providing a forward declaration of nlohmann::basic_json and a typedef for nlohmann::json.

### DIFF
--- a/src/json_fwd.hpp
+++ b/src/json_fwd.hpp
@@ -1,0 +1,30 @@
+#ifndef NLOHMANN_JSON_FWD_HPP
+#define NLOHMANN_JSON_FWD_HPP
+
+#include <cstdint>
+#include <map>
+#include <vector>
+#include <string>
+
+namespace nlohmann
+{
+
+template <
+    template<typename U, typename V, typename... Args> class ObjectType,
+    template<typename U, typename... Args> class ArrayType,
+    class StringType,
+    class BooleanType,
+    class NumberIntegerType,
+    class NumberUnsignedType,
+    class NumberFloatType,
+    template<typename U> class AllocatorType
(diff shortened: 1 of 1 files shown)


theodelrieu:
Hi, sorry about the delay since this PR was opened, but due to 2.1.0 changes, could you update the forward declarations to have `adl_serializer` in it?

Also, if you could provide a simple test file, I think @nlohmann will be able to close this PR ;)

nlohmann:
I still need to check for myself whether the compilation times are really that much better.

theodelrieu:
Sure, I guess it became huger since 2.1.0, with all the `from/to_json` prototypes in users' headers. But I didn't make any measurement, this is just an assumption

zetaPRIME:
Here's a version of json_fwd compatible with 2.1.1 if anyone needs one.
https://github.com/zetaPRIME/libstarlight/blob/edf7945acdf567c9b7e65923daf79aee5cc8fc69/libstarlight/source/starlight/_incLib/json_fwd.hpp

AraHaan:
I would really like something like this to get merged for when people need to use things like this.

theodelrieu:
This is something which is likely to be added in 3.0.0

As I proposed in #474 , I'd like to have separate headers for development.
We will continue to release the single header `json.hpp`, but I think we should also release the separate headers version for those who want it.

nlohmann:
This won't be merged in any time soon, so I close the PR. Let's discuss future steps in #474.

PR #334, opened 2016-10-15
Added ability to switch to unix/windows end-of-line format for dump.

prsyahmi:


gregmarr:
Why would you need to do this here instead of when writing to a file?  In general strings in memory use \n and it gets converted to \r\n when writing to a file.

prsyahmi:
Hi, good point. However I use the whole buffer directly to my needs. Feel free to either close or merge this.

nlohmann:
Thanks a lot for your effort, but I think this is out of scope of the library. Sorry.

PR #337, opened 2016-10-16
replace strtold with non locale dependent version

### LINKED ISSUE #302: Parsing of floats locale dependent
With version 2.0.3, the following code: 

```
std::locale l("");
std::string curr_locale = l.name();
float j = nlohmann::json::parse("0.1");
std::setlocale(LC_ALL, "en_US.UTF-8");
float k = nlohmann::json::parse("0.1");
```

yields the following results (read in the debugger): 
`curr_locale = "nb_NO.UTF-8"`
`j = 0`
`k = 0.100000001`

Thus, the decimals are truncated in the Norwegian locale. This is rather unexpected, and should IMO not happen. 

I think it originates from the parser calling strtod which is locale dependent.

whackashoe:
Hi!

I've taken and reworked a strtod implementation from [TI](https://bitbucket.org/C0deMaver1ck/peeklinux/src/9d811be8bcb584caf1da125df80a2b1801f3d98c/peek-build/m23/csw-system/rtslib/v2_70/strtod.c?at=master&fileviewer=file-view-default) in order to provide this. It's rewritten in an attempt to be solid C++11 rather than C. 

Using this fixes #302. It does not support octal or hexadecimal which JSON doesn't recognize in its number type anyway. This seems to be a bit more efficient (possibly due to non support of octal/hexadecimal) though I wouldn't be surprised if these gains disappear once the last few bugs are fixed heh.

Not all tests are passing yet but it's pretty close, perhaps someone can figure out the ones I haven't. 
# BEFORE:

``` text
./json_benchmarks
parse jeopardy.json                           1  1681452236 ns/op
parse canada.json                            50    43518542 ns/op
parse citm_catalog.json                      50    24345478 ns/op
parse twitter.json                          100    12713148 ns/op
parse numbers/floats.json                     5   447398631 ns/op
parse numbers/signed_ints.json                5   280521961 ns/op
parse numbers/unsigned_ints.json              5   278368046 ns/op
dump jeopardy.json                            1  1390675617 ns/op
dump jeopardy.json with indent                1  1578243282 ns/op
./json_benchmarks 21.516s
```
# AFTER:

``` text
./json_benchmarks
parse jeopardy.json                           1  1619098873 ns/op
parse canada.json                            50    36922996 ns/op
parse citm_catalog.json                      50    25375433 ns/op
parse twitter.json                          100    13150869 ns/op
parse numbers/floats.json                     5   376901576 ns/op
parse numbers/signed_ints.json                5   300091862 ns/op
parse numbers/unsigned_ints.json              5   297152348 ns/op
dump jeopardy.json                            1  1390073209 ns/op
dump jeopardy.json with indent                1  1616301685 ns/op
./json_benchmarks 20.828s
```
# UNIT TESTS:

(these are run with `CXX=clang++ make check`)

``` text
src/unit-class_parser.cpp:37

src/unit-class_parser.cpp:258: FAILED:
  CHECK( json::parser("0.999").parse() == json(0.999) )
with expansion:
  0.999 == 0.999

regression tests
  issue #89 - nonstandard integer type
src/unit-regression.cpp:36

src/unit-regression.cpp:150: FAILED:
  CHECK( j.get<float>() == 4294967296.0f )
with expansion:
  4294967040.0f == 4294967296.0f

src/unit-regression.cpp:160: FAILED:
  CHECK( j.get<float>() == -2147483650.0f )
with expansion:
  -2147483520.0f == -2147483648.0f

regression tests
  issue #186 miloyip/nativejson-benchmark: floating-point parsing
src/unit-regression.cpp:36

src/unit-regression.cpp:316: FAILED:
  CHECK( j.get<double>() == 2.2250738585072009e-308 )
with expansion:
  0.0 == 0.0

src/unit-regression.cpp:319: FAILED:
  CHECK( j.get<double>() == 0.99999999999999989 )
with expansion:
  1.0 == 1.0

src/unit-regression.cpp:322: FAILED:
  CHECK( j.get<double>() == 1.00000000000000022 )
with expansion:
  1.0 == 1.0

src/unit-regression.cpp:325: FAILED:
  CHECK( j.get<double>() == 72057594037927928.0 )
with expansion:
  72057594037927920.0 == 72057594037927928.0

src/unit-regression.cpp:328: FAILED:
  CHECK( j.get<double>() == 9223372036854774784.0 )
with expansion:
  9223372036854775808.0
  ==
  9223372036854774784.0

src/unit-regression.cpp:331: FAILED:
  CHECK( j.get<double>() == 10141204801825834086073718800384.0 )
with expansion:
  10141204801825832960173811957760.0
  ==
  10141204801825834086073718800384.0

src/unit-regression.cpp:334: FAILED:
  CHECK( j.get<double>() == 5708990770823838890407843763683279797179383808.0 )
with expansion:
  5708990770823843327184944562486185035640602624.0
  ==
  5708990770823838890407843763683279797179383808.0

compliance tests from nativejson-benchmark
  roundtrip
src/unit-testsuites.cpp:103

src/unit-testsuites.cpp:309: FAILED:
  CHECK( j.dump() == json_string )
with expansion:
  "[0.0]" == "[2.2250738585072e-308]"
with message:
  filename := test/data/json_roundtrip/roundtrip29.json

test cases:      35 |      32 passed |  3 failed
assertions: 8905168 | 8905157 passed | 11 failed
```


nlohmann:
Thanks for the PR! I'll check it as soon as I find the time.

nlohmann:
Just one question: Can the code be licensed under the MIT license?

whackashoe:
Yeah I think it's fine, it's basically all rewritten so I don't think there is much of an issue that way... plus if they attempted to sue for using a similar variable name or if statement we'd probably become famous.

If you want to be extra safe we can change the comments to be unique as they are just copied verbatim, though I think that'd be horrid as their comments are quite nice.

src/json.hpp
```diff
@@ -8800,64 +8800,187 @@ class basic_json
...
-        @brief parse floating point number
+            T result = 0;
+            const char *fst = st;
+            bool successful_parse = false;
 
-        This function (and its overloads) serves to select the most approprate
-        standard floating point number parsing function based on the type
-        supplied via the first parameter.  Set this to @a
-        static_cast<number_float_t*>(nullptr).
+            while (nl_isspace(*fst))
```
nlohmann: Whitespace is already skipped by the lexer, so I think it is safe to skip this loop.

whackashoe:
Since it is relying on other conditions now with these 2 changes, I think we can strip out endptr and errno as well.

whackashoe:
The `unit-regression` failures seem to be just an off by 1 error, if you look at expected and result values as hex:

`0x4370000000000000 == 0x436fffffffffffff`

`0x43e0000000000000 == 0x43dfffffffffffff`

src/json.hpp.re2c
```diff
@@ -8097,64 +8097,143 @@ class basic_json
...
-        */
-        double str_to_float_t(double* /* type */, char** endptr) const
-        {
-            return std::strtod(reinterpret_cast<typename string_t::const_pointer>(m_start), endptr);
-        }
+            // read in explicit exponent and calculate real exponent.
+            if (*str == 'e' or *str == 'E')
+            {
+                cp = *++str;
+                bool negative_exp = cp == '-'; // read in exponent sign (+/-)
```
nlohmann: `negative_exp` could be `const bool`

src/json.hpp.re2c
```diff
@@ -8097,64 +8097,143 @@ class basic_json
...
-        This function (and its overloads) serves to select the most approprate
-        standard floating point number parsing function based on the type
-        supplied via the first parameter.  Set this to @a
-        static_cast<number_float_t*>(nullptr).
+                // read in fractional part of number, until an 'e' is reached.
+                // count digits after decimal point.
+                while (nl_isdigit(cp = *str))
+                {
+                     result = result * 10 + (cp - '0');
+                     ++str;
```
nlohmann: Couldn't `str` be incremented in the `while` condition just like in the loop below?

src/json.hpp.re2c
```diff
@@ -8097,64 +8097,143 @@ class basic_json
...
+skip_loop:
+                exp += negative_exp ? -count : count;
+            }
+
+            // adjust number by powers of ten specified by format and exponent.
+            constexpr std::array<long double, 9> powerof10 = {
+                {1.e1L, 1.e2L, 1.e4L, 1.e8L, 1.e16L, 1.e32L, 1.e64L, 1.e128L, 1.e256L}
+            };
+
+            if (exp > std::numeric_limits<long double>::max_exponent10)
```
nlohmann: Starting here, there a now comments in the code. Please briefly describe each conditional and loop.

src/json.hpp.re2c
whackashoe: No, we may not want to increment if nl_isdigit fails so it must be in body.

whackashoe:
@nlohmann could you add help-wanted tag to this, I am too busy right now to do anything with this right now but it is very close.

nlohmann:
Hey @whackashoe do you need help with this?

whackashoe:
@nlohmann Yes. I haven't had time to figure out the last issues with this, though iirc its basically at a level where it works except for an edge case (maybe more than one).

nlohmann:
I'll have a look. I may open a feature branch myself for the moment to fix the conflicts and fiddle around a bit.

nlohmann:
I opened a feature branch. I changed the parser to reject incomplete numbers. This helped to simplify the number parser. There are still three tests failing - I'll check tomorrow.

src/json.hpp
```diff
@@ -8800,64 +8800,149 @@ class basic_json
...
-        This function (and its overloads) serves to select the most approprate
-        standard floating point number parsing function based on the type
-        supplied via the first parameter.  Set this to @a
-        static_cast<number_float_t*>(nullptr).
+                int count = 0; // exponent calculation
+                if (not nl_isdigit(cp))
+                {
+                     if (plus_or_minus)
+                     {
+                         *--str;
```
jaredgrubb: Why is there a '*' here?

src/json.hpp
```diff
@@ -8800,64 +8800,149 @@ class basic_json
...
+                     }
 
-        @param[in] type  the @ref number_float_t in use
+                     *--str;
```
jaredgrubb: ditto.

src/json.hpp
whackashoe: Thanks for taking a look! @nlohmann want to incorp these into your feature branch?

nlohmann: This makes no sense. Please have a look at https://github.com/nlohmann/json/blob/feature/strtold/src/json.hpp.re2c#L8139 where I started an overworked version of this PR.

nlohmann:
Please have a look at https://github.com/nlohmann/json/blob/feature/strtold/src/json.hpp.re2c#L8139 where I started an overworked version of this PR, and also the discussion in #302.

TurpentineDistillery:
I see the code parses to long double, and then the return value is cast to float_number_t.
This will yield different results than parsing into float_number_t directly (unless it is long double, of course).  See issue #186

TurpentineDistillery:
Apparently implementing strtod correctly is harder than it seems.
Here's [mozilla implementation](https://github.com/tmhorne/celtx/blob/master/nsprpub/pr/src/misc/prdtoa.c)

nlohmann:
@whackashoe Any idea how to proceed here? I agree with @TurpentineDistillery that correct floating-point handling is really hard...

nlohmann:
What about this PR? Is it likely that the remaining issues can be fixed? There is also PR #378 dealing with the same issue.

whackashoe:
It seems like it will need significant changes to work. You can close this if it's dirtying up the queue, I don't have time to work on it for a while.

nlohmann:
Thanks for checking back. I close this for now.

PR #338, opened 2016-10-19
Feature/user defined types

theodelrieu:
WIP: This PR is **NOT** ready to be merged, its purpose is to facilitate review/discussions.
## Files to change

   ``` sh
   make re2c
   ```
   
   ``` sh
   make
   ```
   
   and can be executed with
   
   ``` sh
   ./json_unit
   ```
   
## Note
## Please don't

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -154,10 +154,10 @@ struct to_json_fn
 
 struct from_json_fn
 {
-  template <typename T>
-  constexpr auto operator()(T&& val) const -> decltype(from_json(std::forward<T>(val)))
+  template <typename Json, typename T>
+  constexpr auto operator()(Json const& from, T& to) const -> decltype(from_json(from, to))
   {
-    return from_json(std::forward<T>(val));
+    return from_json(from, to);
   }
 };
 
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -102,6 +102,17 @@ namespace nlohmann
 template <typename T, typename = void>
 struct json_traits;
 
+// alias templates to reduce boilerplate
+template <bool B, typename T = void>
+using enable_if_t = typename std::enable_if<B, T>::type;
+
+template <typename T>
+using remove_cv_t = typename std::remove_cv<T>::type;
+
+template <typename T>
+using remove_reference_t = typename std::remove_reference<T>::type;
+
+// TODO update this doc
 /*!
 @brief unnamed namespace with internal helper functions
 @since version 1.0.0
@@ -1304,15 +1315,11 @@ class basic_json
     // auto j = json{{"a", json(not_equality_comparable{})}};
     // 
     // we can remove this constraint though, since lots of ctor are not explicit already
-    template <
-        typename T,
-        typename =
-            typename std::enable_if<detail::has_json_traits<typename std::remove_cv<
-                typename std::remove_reference<T>::type>::type>::value>::type>
+    template <typename T, typename = enable_if_t<detail::has_json_traits<
+                              remove_cv_t<remove_reference_t<T>>>::value>>
     explicit basic_json(T &&val)
-        : basic_json(json_traits<typename std::remove_cv<
-                         typename std::remove_reference<T>::type>::type>::
-                         to_json(std::forward<T>(val))) {}
+        : basic_json(json_traits<remove_cv_t<remove_reference_t<T>>>::to_json(
+              std::forward<T>(val))) {}
     /*!
     @brief create a string (explicit)
 
@@ -2673,17 +2680,12 @@ class basic_json
 
     // get_impl overload chosen if json_traits struct is specialized for type T
     // simply returns json_traits<T>::from_json(*this);
-    // TODO add alias templates (enable_if_t etc)
-    template <
-        typename T,
-        typename = typename std::enable_if<
-            detail::has_json_traits<typename std::remove_cv<
-                typename std::remove_reference<T>::type>::type>::value>::type>
-    auto get_impl(T *) const -> decltype(
-        json_traits<typename std::remove_cv<typename std::remove_reference<
-            T>::type>::type>::from_json(std::declval<basic_json>())) {
-      return json_traits<typename std::remove_cv<
-          typename std::remove_reference<T>::type>::type>::from_json(*this);
+    template <typename T, typename = enable_if_t<detail::has_json_traits<
+                              remove_cv_t<remove_reference_t<T>>>::value>>
+    auto get_impl(T *) const
+        -> decltype(json_traits<remove_cv_t<remove_reference_t<T>>>::from_json(
+            std::declval<basic_json>())) {
+      return json_traits<remove_cv_t<remove_reference_t<T>>>::from_json(*this);
     }
 
     // this one is quite atrocious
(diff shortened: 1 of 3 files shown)


src/json.hpp
```diff
@@ -91,13 +91,34 @@ SOFTWARE.
...
+// alias templates to reduce boilerplate
+template <bool B, typename T = void>
+using enable_if_t = typename std::enable_if<B, T>::type;
+
+template <typename T>
+using remove_cv_t = typename std::remove_cv<T>::type;
+
+template <typename T>
+using remove_reference_t = typename std::remove_reference<T>::type;
+
```
gnzlbg: you probably want `uncvref_t` here as well.

src/json.hpp
```diff
@@ -144,6 +203,23 @@ struct DecimalSeparator : std::numpunct<char>
...
+// taken from ranges-v3
+// TODO add doc
+template <typename T>
+struct __static_const
+{
+  static constexpr T value{};
+};
+
+template <typename T>
+constexpr T __static_const<T>::value;
```
gnzlbg: I remember that identifiers starting with an underscore followed by adjacent underscores are reserved for the implementation in _any_ scope (not only the global one), that is, one cannot use them. But the rules for reserved identifiers were more complex than that (this is C++ after all), so might want to check if that is actually allowed in a non-global scope (we are inside a namespace here).

src/json.hpp
```diff
@@ -1241,15 +1336,14 @@ class basic_json
...
+    value
     from a compatible string container
 
     @since version 1.0.0
     */
-    basic_json(const string_t& val)
-        : m_type(value_t::string), m_value(val)
-    {
-        assert_invariant();
+    basic_json(const string_t &val) : m_type(value_t::string), m_value(val) {
```
gnzlbg: My intuition says that we would like to use the trait system for constructing `basic_json` objects from all types that are not `basic_json`, since in case somebody has a different kind of representation in a string (e.g. XML), they could still define their own trait class to construct `basic_json` objects from that.

src/json.hpp
```diff
@@ -1241,15 +1336,14 @@ class basic_json
 
     @sa @ref basic_json(const typename string_t::value_type*) -- create a
     string value from a character pointer
-    @sa @ref basic_json(const CompatibleStringType&) -- create a string value
+    @sa @ref basic_json(const CompatibleStringType&) -- create a string
+    value
```
gnzlbg: nitpick: you should probably revert all unnecessary formatting changes, since they make going through the relevant parts of the diffs harder

src/json.hpp
```diff
@@ -122,6 +143,44 @@ struct has_mapped_type
...
+  static std::false_type detect(...);
+
+  static constexpr bool value = decltype(detect<T>(0))::value;
+};
+
+template<typename T>
+struct has_json_traits
+{
+  static constexpr bool value = has_destructor<json_traits<T>>::value;
+};
```
gnzlbg: Do we need all this? 

I think we can have some member function like this in `basic_json`:

```c++
// Sketch

template <typename T>
auto from_impl(T const& t, long) -> decltype(do_something_without_trait(t));
template <typename T>
auto from_impl(T const& t, int) -> decltype(json_trait<T>::from_json(t, *this));
template <typename T>
auto from(T const& t) -> decltype(from_impl(t, 0));
```

Notice how `from` calls `from_impl(t, (int)0)`, which will try to call the overload `from_impl(T, int)` (which is a better match that the overload with long). If that works, then that is the one that is called. But if in that overload `json_trait<T>::from` fails, e.g., because there is no specialization for `T`, that overload will be silently removed, and then the second overload `from_impl(T, long)` will kick in, because there is an implicit conversion from `int` to `long`. 

So what that does is "try to call the int overload, and if that fails, fall back to the long overload". You can probably use this "idiom" to easily detect whether `json_traits` implement something that you need, and fall back to some other solution if it doesn't.

(If you want to check some examples that actually work, a good example is how `begin` and `end` are implemented in range-v3, or in STL2).

src/json.hpp
```diff
@@ -2584,20 +2678,52 @@ class basic_json
...
+                           ::nlohmann::from_json(std::declval<basic_json>(),
+                                                 std::declval<T &>()),
+                           std::declval<T>())>>>
+    {
+      remove_cv_t<T> ret;
+      // I guess this output parameter is the only way to get ADL
+      // Even if users can use the get<T> method to have a more 'functional' behaviour
+      // i.e. having a return type, could there be a way to have the same behaviour with from_json?
+      // e.g. auto t = nlohmann::from_json<T>(json{});
+      // this seems to require variable templates though... (at least it did when I tried to implement it)
```
gnzlbg: I'll think about this, you make good points here.

gnzlbg:
Thanks for working on this. I will try to take a more detailed look tomorrow at it, but the tests look good :)

src/json.hpp
theodelrieu: According to [cppreference](http://en.cppreference.com/w/cpp/language/identifiers) : "the identifiers with a double underscore anywhere are reserved"
 
I'll rename it

src/json.hpp
theodelrieu: I agree, although this would be a huge change to the current implementation, it might be out of the scope of this PR.

src/json.hpp
theodelrieu: Indeed, forgot to check that, my bad

src/json.hpp
theodelrieu: If I understand correctly, the `from_impl` method aims to replace the two `get_impl` overloads that I added? I agree this is a lot more readable than what I have now.

theodelrieu:
I'm also still asking myself whether I prefer the `JSONDeserializer` template argument, or the `json_traits` struct in the `nlohmann` namespace.

It would be nice to have a discussion about the pros and cons of both methods.

src/json.hpp
gnzlbg: While I think we probably don't want to do this for string and int, and other primitive types yet, I recall that there was a constructor for "ranges" (types with `value_type`, `begin` and `end`). I think it would be still a good idea to at least move this constructor out to check that ADL works.

gnzlbg:
I think @d-frey wanted to be able to serialize/deserialize the same type in different ways to json objects. I agree with him in that to do that the best thing to do is to have a template parameter that controls this in `basic_json<..., JSONSerialization = json_traits>`, but obviously we should give it a good default value that does the right thing for the 99% of the time in which one doesn't care. 

We could also, inside `basic_json`, try to use the trait first, and if that fails, try to fall back to ADL. So that users can always use ADL but have an easy way to override it (e.g. they would only need to define a trait for the particular type they want to override, but everything else would still work through ADL in the same way). 

It takes a lot of time and effort (and rewrites) to experiment with these things, but I guess once we get a bit of experience with it the best solution would become more clear.

src/json.hpp
theodelrieu: Are you talking about the constructor taking `CompatibleArrayType&` as a parameter?
Could you post a short pseudo-code about what you have in mind?

theodelrieu:
Good point, I shall experiment with the current way first then. Thanks!

theodelrieu:
Hmpf, it seems VS2015 struggles with ADL... It compiles after specifying `::udt::from_json` in the decltype. I hope there is a workaround

nlohmann:
I'm really amazed about the effort and the discussion. I won't find the time to have a deeper look at this in the next days, but I am looking forward to! Keep up the great work!

theodelrieu:
Like all compiler workarounds, this one is a bit surprising...

theodelrieu:
I tried to provide some overloads for commonly used types, however it needs to rewrite a lot of code so I discarded that.
I think it's best to provide the functionality first, then enhance it later, if needed.

theodelrieu:
Also, before trying to implement the `JSONDeserializer` way, I have a question for you @gnzlbg @nlohmann.
In order to give more control to users, I think the `json_traits` struct needs to be declared *inside* the `basic_json` class.

This allows to specialize on multiple `basic_json` instances, one where `StringType = std::wstring` for example.

One of the cons is readability though:

```cpp
template <>
template <>
struct basic_json<>::json_traits<MyType>
{};
```
Is it worth going so far now? Should we support this after a first release?
I admit I don't really know where to go from here, (from the functionality-side), suggestions are welcome!

gnzlbg:
I don't know if I understood you correctly, but if what you mean is that one user might want to provide different behavior in the `from/to` methods of `json_traits` depending on the type parameters of the `basic_json` object I still think that `json_traits` does not need to be a member of `basic_json`:

```c++
template <typename T> struct json_traits {};

template <>
struct json_traits<MyType> { 
   template <typename JSON>
   static void from(JSON const& j, MyType& m) {
     if constexpr (Same<string_type_t<JSON>, std::wstring>) {
       // do something
     } else {
       // do something else
     }
   }

   template <typename JSON>
   static void to(MyType const& m, JSON& j) { ... }
};
```

I just used a concept (`Same<T, U...>`) to branch on the `string_type_t` of `basic_json` (which I don't know if it is provided) using `if constexpr`, but one could use `std::is_same_v`, tag-dispatching, overloading, ... to do the same in C++14/11.

Commodity template aliases like `json::string_type_t<basic_json<...>>` to obtain the template parameters of a `basic_json` type is something we probably want to offer anyways.

theodelrieu:
Thanks for your reply. Sorry I didn't have time to check the PR last week, I will try to work on it this week.
I don't think there is much left to do on this one, (implementation-wise).

I will start another PR for the other implementation.

TriHard8:
I didn't see in the documentation...Is there functionality to convert libconfig to json?  Any desire to add that functionality?

nlohmann:
Continued in #355.

Issue #328, opened 2016-10-12
Add a customization point for user-defined types


theodelrieu:
Hi, I think it would be a cool addition to the library if there was a way to specify a conversion method, from a type to `basic_json`, which would be called by the lib.

Here a sample code of the behaviour I'm thinking about:

``` cpp
// library code
template <typename UserDefinedType>
struct json_traits;

// user code
template <>
struct json_traits<MyType>
{
  static basic_json convert(MyType const& val)
  {
    return {{"first_row", val.first_row}, {"some_value", 42}};
  }
};
```

There could also be a conversion method when `get` is called

``` cpp
auto obj = json::parse(buffer);
auto val = obj["user_defined_type_value"].get<MyType>();
```

These are just thoughts for now, but that would remove a LOT of boilerplate in my code.
Plus, this would add some compile-time checks (if the struct is not defined for example)
What's your opinion on that?

gregmarr:
That would be cool.  I created to_json and from_json functions for a lot of my types, and use to_json/from_json exclusively even for types it already supports.  Making them work automatically out of the box would be even cooler.

nlohmann:
This issue is related to #298 and #280.

nlohmann:
I have not understood the role of `json_traits` completely. Shouldn't there be a constructor in the `basic_json` class for the `json_traits` template? Would this constructor then be usable for any class that is specialized by `json_traits`?

gregmarr:
It's all static functions, the class is never instantiated.

nlohmann:
I still do not understand what to add to the `basic_json` class.

gregmarr:
Maybe something like this:

    template<typename T> basic_json(T&& val) :
        basic_json(json_traits<T>::convert(std::forward<T>(val)))
    {
        assert_invariant();
    }

There might also need to be some enable_if code to keep the current constructors working, or maybe the current constructors become specializations of the json_traits class instead.

gregmarr:
Yeah, this is basically what @gnzlbg and @d-frey described in #298

theodelrieu:
Indeed, looking at taocpp [value.hh](https://github.com/taocpp/json/blob/master/include/tao/json/value.hh) and [traits.hh](https://github.com/taocpp/json/blob/master/include/tao/json/traits.hh) can help to understand the idea

theodelrieu:
I started a POC on my fork. (I didn't modify the re2c file, but I will !)

theodelrieu:
[...]

Issue #290, opened 2016-08-09
Support for iterator-range parsing


theodelrieu:
Hi, it would be handy to add an overload to the `parse` method.

In my code, I deal a lot with `vector<uint8_t>` when communicating with my server.
However I sometimes have to send JSON, and I must convert the buffer to a `std::string`, which is unfortunate.

I know I could typedef basic_json and use `std::basic_string<uint8_t>` as `StringType`.
However, other parts of the code uses the `std::string` version, and I don't want to patch the whole code because of that.

We could add this overload to `parse`, which will behave like `std::string` iterator constructor:

```cpp
template<typename InputIterator>
static basic_json parse(InputIterator begin, InputIterator end, parser_callback_t cb = nullptr);

// user code

auto buffer = receive_bytes();
auto json = json::parse(buffer.begin(), buffer.end());
```

What are your thoughts on that ?
If you agree with the idea, I can take care of the PR.

nlohmann:
I think the current implementation would not benefit from an iterator-range access: Inside the parser, I employ a lexer which relies on access to the input via a pointer of type `lexer_char_t*` (`unsigned char*`). For `std::string`, this can be done with casting the result of `c_str()`, and `std::istream`s are copied line by line to a buffer of type `std::string`.

To process a `std::vector<uint8_t>`, such a pointer could be casted from `data()`. Maybe one could generalize the `static basic_json parse(const string_t& s, const parser_callback_t cb = nullptr)` function to

```cpp
template<typename T>
static basic_json parse(const T& s, const parser_callback_t cb = nullptr)
```

where `T` has a `data()` member function (`std::basic_string`, `std::vector`, and `std::array`) and whose pointers to its value type can be casted to `lexer_char_t*`.

What do you think?

theodelrieu:
So only contiguous containers of 1 byte integrals are supported I believe? (for now at least)
I don't know if counting on the `data()` method is a good thing, what if an user gives a homemade vector, or a `char[]` ?

Maybe it's better to use methods defined in the [Container Concept](http://en.cppreference.com/w/cpp/concept/Container), and use the `std::begin` method. We could `static_assert` that the iterator type must be RandomAccessIterator to ensure the contiguity of the container.

It could look like this:

```cpp
template<typename ByteContainer>
static basic_json parse(const ByteContainer& buf, const parser_callback_t cb = nullptr)
{
  // ensure std::begin(ByteContainer) is valid + retrieve iterator type
[...]

PR #355, opened 2016-11-07
Feature/user defined types bis

theodelrieu:
WIP: This PR is *NOT* ready to be merged, its purpose is to facilitate review and discussion.


theodelrieu:
This is a first version of the `JSONSerializer` template argument, (alternate implementation).
I didn't make a custom serializer in the tests yet. But I already feel that I like the first implementation more.

I find the need to have another `basic_json` typedef for types we want to serialize differently quite problematic.

A use case in my own code is partial specialization on template types (e.g. boost::optional, boost::containers, ...).

In order to rely on ADL, one would need to write a `from_json` method in the `boost` namespace, which is undesirable.

@gnzlbg I know you mentioned something about providing some "c++11 concepts" (e.g. Optional Concept), however I think we should provide UDT support first, then enhance it if users demand it.

What do you think about this ?

gnzlbg:
> In order to rely on ADL, one would need to write a `from_json` method in the boost namespace, which is undesirable.

Not really, if the customization point is implemented as suggested in the paper, then you have something like this (from the top of my head using my brain's imperfect C++ compiler):

```c++
namespace json {

// here you can define from_json for any type 

template <typename JSON, typename T, std::size_t N>
void from_json(JSON const& j, T (t&)[N]) {
  // implementation for C-Arrays
}

template <typename JSON, typename T>
void from_json(JSON const& j, boost::optional<T>& o) {
  // implementation for boost::optional
}

// if you have a type trait to select "Optional-like objects" you could have a 
// generic implementation for those here as well, but you just want it to
// be picked if a concrete implementation is not found

struct from_json_fn {
  template<typename JSON, typename T>
  constexpr auto operator()(JSON const& j, T& t) -> decltype(from_json(j, t)) {
    return from_json(j, t);
  }
};

namespace { 
  static constexpr auto const& from_json = static_const<from_json_fn>::value;
}  // namespace 

}  // namespace json
```

The only thing to take into account is that you don't want `nlohmann::json` to depend on `boost::optional`, so this means you should put those overloads in the `json` namespace in a separate header, e.g., `<json/extensions/boost/optional.hpp>`, that people can then opt into including.

theodelrieu:
It makes sense to have extensions separately, even if I believe we should add them later, if users show the need.

By the way, I re-checked one of your comments in issue #328 and I realized that I forgot to add template arguments to the `JSONSerializer` class. Which enables the use of partial specialization *and* ADL if both are wanted.

This solves the main issue I had with that solution!

src/json.hpp
```diff
@@ -144,6 +183,50 @@ struct DecimalSeparator : std::numpunct<char>
...
+    uncvref_t<T> ret;
+    ::nlohmann::from_json(std::forward<Json>(j), ret);
+    return ret;
+  }
+
+  template <typename T, typename Json>
+  static auto from_json(Json&& j, T& val) -> decltype(::nlohmann::from_json(std::forward<Json>(j), val))
+  {
+    ::nlohmann::from_json(std::forward<Json>(j), val);
+  }
```
jaredgrubb: This is missing a 'return' and maybe you should have a unit test on a from_json type that *does* return a value.

gnzlbg: IIUC  `from_json` should never return a value, that is, its signature should be `(JSON const& input, T& output) -> void`, so maybe it would be better to omit the return value but put a `void` in the `decltype` like this:

```c++
static auto from_json(Json&& j, T& val) -> 
decltype(::nlohmann::from_json(std::forward<Json>(j), val), void)
{
    ::nlohmann::from_json(std::forward<Json>(j), val);
}
```

theodelrieu: I came to the same conclusion, right now, the code is quite messy, I thought at first it would be a nice idea to return the type directly.

However, we cannot know which `basic_json` type will be used by the user in its own method, hence the need to pass it as an argument.

I will remove the overloads in a next commit

theodelrieu:
@nlohmann  there seems to be a problem in a test.

I had an out of memory error once, in `unit-constructor1.cpp`, when constructing a JSON from a `std::ifstream`. Simply launching the executable from elsewhere solved it for me.

I don't know if the Travis' segfault is related though, do you have this problem on other PR?
It didn't trigger on each of my commit, that's why I'm wondering

nlohmann:
Travis seems to have problems every now and then - I restarted the build and it succeeded.

theodelrieu:
In the last commit I faced an issue when providing a user-defined `JSONSerializer`.

Since I did not change the way base types are handled, such as `int` or `std::string`, the following code breaks because of ambiguity, when doing `enable_if` checks: 

``` cpp
// when not specialized, I want to use ADL as well
template <typename T, typename = void>
struct my_serializer
{
  template <typename Json>
  static void from_json(Json const& j, T& val)
  {
    nlohmann::from_json(j, val);
  }
};
```

There are some ways to solve this:
* specialize the serializer for every type
* do some SFINAE to avoid already handled types
* handle basic types in nlohmann with to_json / from_json functions

The first two are quite annoying for users, and the latter requires a bit of refactoring (which will be needed someday if this feature is merged IMHO).

I would rather choose the latter, which doesn't impact users
@gnzlbg  @nlohmann  Do you see another way to solve this issue?

gnzlbg:
> Do you see another way to solve this issue?

Not really. From the point of view of the new serialization/deserialization mechanism, primitive types and standard library types look like user-defined types :/

nlohmann:
> and the latter requires a bit of refactoring (which will be needed someday if this feature is merged IMHO).

About what kind of refactoring are we talking?

gnzlbg:
The constructors / assignments of `basic_json` for `int`, `float`, `std::string`, ranges(`begin`, `end`, `value_type`) ... would be extracted to `from_json`/`to_json` free functions. The only "special" constructor / assignment that would remain would be templates that are constrained on the type `T` being `JSONSerializable`.

theodelrieu:
I am working on a fourth solution, that does not require refactoring. It should be ready around tonight.

theodelrieu:
After a lot of rewrites, I came to this implementation.

I managed to remove constraints on `nlohmann::adl_serializer` and the custom one in the tests, without breaking anything.

I wanted to have your feedback on it, I'm sorry I didn't write comments, so do not hesitate to comment on the diff!

This fourth solution has the advantages of the former discussed one, without its main drawback: the need to rewrite almost of the lib...

src/json.hpp
```diff
@@ -144,6 +337,41 @@ struct DecimalSeparator : std::numpunct<char>
 
 }
 
+// taken from ranges-v3
+// TODO add doc
+template <typename T>
+struct _static_const
```
gnzlbg: nitpick: the leading underscore _might_ be undefined behavior (I always forget the rules...). Worth checking.

src/json.hpp
```diff
@@ -1225,6 +1445,21 @@ class basic_json
...
+                                                                                        detail::is_compatible_array_type<basic_json_t, uncvref_t<T>>::value or
+                                                                                        detail::is_compatible_float_type<number_float_t, uncvref_t<T>>::value or
+                                                                                        detail::is_compatible_integer_type<number_integer_t, uncvref_t<T>>::value or
+                                                                                        detail::is_compatible_unsigned_integer_type<number_unsigned_t, uncvref_t<T>>::value or
+                                                                                        std::is_constructible<string_t, uncvref_t<T>>::value or
+                                                                                        std::is_base_of<std::istream, uncvref_t<T>>::value or
+                                                                                        std::is_same<boolean_t, uncvref_t<T>>::value), int> = 0>
```
gnzlbg: Might want to abstract all these checks into an `detail::is_basic_json_compatible<uncvref_t<T>>` or something like that. It might make the function definition clearer, and it also might make it easier to explain what is going on.

src/json.hpp
```diff
@@ -1476,11 +1707,7 @@ class basic_json
 
     @since version 2.0.0
     */
-    template<typename CompatibleNumberUnsignedType, typename std::enable_if <
-                 std::is_constructible<number_unsigned_t, CompatibleNumberUnsignedType>::value and
-                 std::numeric_limits<CompatibleNumberUnsignedType>::is_integer and
-                 not std::numeric_limits<CompatibleNumberUnsignedType>::is_signed,
-                 CompatibleNumberUnsignedType>::type = 0>
+    template<typename CompatibleNumberUnsignedType, enable_if_t<detail::is_compatible_unsigned_integer_type<number_unsigned_t, CompatibleNumberUnsignedType>::value, int> = 0>
```
gnzlbg: why do you need to pass `number_unsigned_t` to `detail::is_compatible_unsigned_integer_type` ? Shouldn't passing it `CompatibleNumberUnsignedType` be enough? (as in, can't we simplify this a bit?)

src/json.hpp
```diff
@@ -1556,9 +1783,7 @@ class basic_json
 
     @since version 1.0.0
     */
-    template<typename CompatibleNumberFloatType, typename = typename std::enable_if<
-                 std::is_constructible<number_float_t, CompatibleNumberFloatType>::value and
-                 std::is_floating_point<CompatibleNumberFloatType>::value>::type>
+    template<typename CompatibleNumberFloatType, enable_if_t<detail::is_compatible_float_type<number_float_t, CompatibleNumberFloatType>::value, int> = 0>
```
gnzlbg: same as a bove

src/json.hpp
```diff
@@ -2896,11 +3120,35 @@ class basic_json
...
+    }
+
+    template <typename T,
+              enable_if_t<detail::has_from_json<
+                  JSONSerializer, basic_json, uncvref_t<T>>::value 
+                              and not (detail::is_compatible_object_type<object_t, uncvref_t<T>>::value or
+                                                                                        detail::is_compatible_array_type<basic_json_t, uncvref_t<T>>::value or
+                                                                                        detail::is_compatible_float_type<number_float_t, uncvref_t<T>>::value or
+                                                                                        detail::is_compatible_integer_type<number_integer_t, uncvref_t<T>>::value or
+                                                                                        detail::is_compatible_unsigned_integer_type<number_unsigned_t, uncvref_t<T>>::value or
```
gnzlbg: Same here, define a `detail::is_compatible_basic_json_type` (or similar) that abstracts all this and makes it more readable.

src/json.hpp
```diff
@@ -122,6 +135,186 @@ struct has_mapped_type
...
+
+  static void detect(...);
+
+public:
+  static constexpr bool value = std::is_integral<decltype(
+      detect(std::declval<JSONSerializer<T, void>>()))>::value;
+};
+
+void to_json();
+void from_json();
```
gnzlbg: Why do we need these two here? are they really necessary?

src/json.hpp
```diff
@@ -122,6 +135,186 @@ struct has_mapped_type
         std::is_integral<decltype(detect(std::declval<T>()))>::value;
 };
 
+template <template <typename...> class T>
+struct wrapper{};
```
gnzlbg: Is this used somewhere?

src/json.hpp
theodelrieu: Yes indeed, I pushed as soon as I made it work, I also forgot to format the relevant parts, sorry for that

src/json.hpp
theodelrieu: nope I forgot that

src/json.hpp
theodelrieu: I think we can, I simply translated the existing checks, to avoid spending time on it (before making the whole thing work)

src/json.hpp
gnzlbg: I rechecked and IIUC this usage is fine because the underscore:
- is not followed by another underscore,
- is not followed by a capital letter,

and the identifier is not in the global namespace. 

I am still not sure that I understood the rules correctly though... Avoiding identifiers that start with an underscore was always my pragmatic solution.

src/json.hpp
gnzlbg: While reading all the `has_xxxx` classes I was thinking why one cannot just use expression SFINAE for these things, and actually wondered if it might had something to do with MSVC.

src/json.hpp
theodelrieu: I'm still confused by the "external identifier" thing, but it would surprise me if identifiers beginning with an underscore (except the global namespace) were to be reserved.

I would be an outlaw with all my prefixed private variables...

src/json.hpp
theodelrieu: It was more because of clarity at first. I try to avoid expression SFINAE when it becomes too demanding on MSVC too

theodelrieu:
uuh? Is there a problem with travis? I dont see conflicts

theodelrieu:
Hi, I managed to remove SFINAE checks on `adl_serializer` (on custom serializers too), by tweaking quite a bit of enable_if checks.

I added some comments to help the review, my next goal is to rewrite `unit-udt.cpp` entirely, as it does not make a lot of sense right know.

Once that is done, and the technical review is over, I will add real documentation, examples, etc...

I think the hardest part is done!

qis:
@theodelrieu Hey, the latest version of [feature/user_defined_types_bis](https://github.com/theodelrieu/json/tree/4e0ac5993e80a18008701fb3f3622de9afe898c4) does not work in Visual Studio 2015 unless I patch it. Here is the diff:

```diff
--- json.hpp.orig       2016-11-26 01:46:03.000000000 +0100
+++ json.hpp    2016-11-28 11:41:01.072158500 +0100
@@ -373,8 +373,8 @@ constexpr T static_const<T>::value;

 inline namespace
 {
-  constexpr auto const& to_json = static_const<detail::to_json_fn>::value;
-  constexpr auto const& from_json = static_const<detail::from_json_fn>::value;
+  //constexpr auto const& to_json = static_const<detail::to_json_fn>::value;
+  //constexpr auto const& from_json = static_const<detail::from_json_fn>::value;
 }

 // default JSONSerializer template argument, doesn't care about template argument
@@ -3165,7 +3165,8 @@ class basic_json
     @since version 1.0.0
     */
     template <typename ValueType,
-              enable_if_t<not std::is_pointer<ValueType>::value, int> = 0>
+              enable_if_t<not std::is_pointer<ValueType>::value
+                          and detail::is_compatible_basic_json_type<uncvref_t<ValueType>, basic_json_t>::value, int> = 0>
     auto get() const
         -> decltype(this->get_impl(static_cast<ValueType *>(nullptr)))
     {

```

I have no idea if removing the inline namespace constexpr values breaks something. Care to comment? Need more information?

gnzlbg:
```c++
+  //constexpr auto const& to_json = static_const<detail::to_json_fn>::value;
+  //constexpr auto const& from_json = static_const<detail::from_json_fn>::value;
```

Removing these should remove UDT support. Does it still work afterwards?

qis:
@gnzlbg It works **only** when those two lines are removed. Tested with this code:

```cpp
#include "json.hpp"
#include <iostream>
#include <optional>
#include <string>

namespace nlohmann {

template <typename T>
inline void to_json(json& json, const std::optional<T>& value) {
  if (value) {
    json = value.value();
  } else {
    json = nullptr;
  }
}

template <typename T>
inline void from_json(const json& json, std::optional<T>& value) {
  if (json.is_null()) {
    value = {};
  } else {
    value = json.get<T>();
  }
}

}  // namespace nlohmann

int main() {
  nlohmann::json json(std::optional<std::string>("test"));
  std::cout << json.get<std::optional<std::string>>().value() << std::endl;
}
```

gnzlbg:
Can you provide the error message?

theodelrieu:
interesting, do you have Visual Studio Update 3 ? There might be something breaking on older versions of VS2015

qis:
@theodelrieu Yes, 14.0.25431.01 Update 3. Just tested the code in clang 4.0.0 (trunk 284483) and had to patch it again a bit:

```diff
--- json.hpp    2016-11-26 01:46:03.000000000 +0100
+++ json.hpp    2016-11-28 12:13:39.529514500 +0100
@@ -373,8 +373,10 @@ constexpr T static_const<T>::value;

 inline namespace
 {
+#ifndef _MSC_VER
   constexpr auto const& to_json = static_const<detail::to_json_fn>::value;
   constexpr auto const& from_json = static_const<detail::from_json_fn>::value;
+#endif
 }

 // default JSONSerializer template argument, doesn't care about template argument
@@ -3165,7 +3167,8 @@ class basic_json
     @since version 1.0.0
     */
     template <typename ValueType,
-              enable_if_t<not std::is_pointer<ValueType>::value, int> = 0>
+              enable_if_t<not std::is_pointer<ValueType>::value
+                          and detail::is_compatible_basic_json_type<uncvref_t<ValueType>, basic_json_t>::value, int> = 0>
     auto get() const
         -> decltype(this->get_impl(static_cast<ValueType *>(nullptr)))
     {
@@ -11020,7 +11023,9 @@ namespace std

 @since version 1.0.0
 */
+#ifndef __clang__
 template<>
+#endif
 inline void swap(nlohmann::json& j1,
                  nlohmann::json& j2) noexcept(
                      is_nothrow_move_constructible<nlohmann::json>::value and
```

I also tried an explicit specialization `template<> inline void swap<nlohmann::json>(...` but this resulted in a compile time error: `error: no function template matches function template specialization 'swap'`

qis:
@gnzlbg Here are the error messages.

Original version:

```
2>main.cpp(30): error C2668: 'nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>::get': ambiguous call to overloaded function
2>  json.hpp(3186): note: could be 'std::experimental::optional<std::string> nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>::get<std::experimental::optional<std::string>,0>(void) const'
2>  json.hpp(3173): note: or       'T nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>::get<std::experimental::optional<std::string>,0>(void) const'
2>          with
2>          [
2>              T=std::experimental::optional<std::string>
2>          ]
2>  main.cpp(30): note: while trying to match the argument list '()'
2>main.cpp(30): error C2228: left of '.value' must have class/struct/union
```

After the non-serializer `get()` method also checks if `ValueType` is a compatible basic json type:

```
1>json.hpp(396): error C2872: 'to_json': ambiguous symbol
1>  main.cpp(9): note: could be 'void nlohmann::to_json(nlohmann::json &,const std::experimental::optional<T> &)'
1>  json.hpp(377): note: or       'const nlohmann::detail::to_json_fn &nlohmann::`anonymous-namespace'::to_json'
1>  json.hpp(1492): note: see reference to function template instantiation 'void nlohmann::adl_serializer<T,void>::to_json<nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>,_Ty>(Json &,T &&)' being compiled
1>          with
1>          [
1>              T=std::experimental::optional<std::string>,
1>              _Ty=std::experimental::optional<std::string>,
1>              Json=nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>
1>          ]
1>  json.hpp(1492): note: see reference to function template instantiation 'void nlohmann::adl_serializer<T,void>::to_json<nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>,_Ty>(Json &,T &&)' being compiled
1>          with
1>          [
1>              T=std::experimental::optional<std::string>,
1>              _Ty=std::experimental::optional<std::string>,
1>              Json=nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>
1>          ]
1>  main.cpp(29): note: see reference to function template instantiation 'nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>::basic_json<std::experimental::optional<std::string>,0>(T &&)' being compiled
1>          with
1>          [
1>              T=std::experimental::optional<std::string>
1>          ]
1>  main.cpp(29): note: see reference to function template instantiation 'nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>::basic_json<std::experimental::optional<std::string>,0>(T &&)' being compiled
1>          with
1>          [
1>              T=std::experimental::optional<std::string>
1>          ]
1>json.hpp(390): error C2872: 'from_json': ambiguous symbol
1>  main.cpp(18): note: could be 'void nlohmann::from_json(const nlohmann::json &,std::experimental::optional<T> &)'
1>  json.hpp(378): note: or       'const nlohmann::detail::from_json_fn &nlohmann::`anonymous-namespace'::from_json'
1>  json.hpp(3194): note: see reference to function template instantiation 'void nlohmann::adl_serializer<T,void>::from_json<const nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>&,type>(Json,T &)' being compiled
1>          with
1>          [
1>              T=std::experimental::optional<std::string>,
1>              Json=const nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer> &
1>          ]
1>  json.hpp(3194): note: see reference to function template instantiation 'void nlohmann::adl_serializer<T,void>::from_json<const nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>&,type>(Json,T &)' being compiled
1>          with
1>          [
1>              T=std::experimental::optional<std::string>,
1>              Json=const nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer> &
1>          ]
1>  main.cpp(30): note: see reference to function template instantiation 'std::experimental::optional<std::string> nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>::get<std::experimental::optional<std::string>,0>(void) const' being compiled
1>  main.cpp(30): note: see reference to function template instantiation 'std::experimental::optional<std::string> nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>::get<std::experimental::optional<std::string>,0>(void) const' being compiled
```

When both modifications are applied, everything works.

qis:
Oh, I forgot to mention that I compile with `-std=c++1z` in Clang 4.0.0 and `/std:c++latest` in MSVC14. This may cause some problems since, this flags completely remove some deprecated features.

gnzlbg:
The root of the problem seems to be that the `get` function (and `get_impl`) is not properly constrained, but this part of the code is still in a rough state. Thanks for the test @qis !

theodelrieu:
Thanks, I will try with `/std:c++latest` as soon as I have some time.

I didn't spend as much time constraining `get` as I did with constructors.
Could you please attach your code with `std::optional`?

I'd like to have the same configuration as you when testing

qis:
@theodelrieu It's a slightly modified version of <https://github.com/akrzemi1/Optional>: https://github.com/qis/compat/blob/master/include/msvc/optional

theodelrieu:
@qis Thanks, could you attach the code you used to test this PR too?

qis:
@theodelrieu I already did, but here is a small repo: [json_test](https://github.com/qis/json_test)

My changes are commented out and marked with `// qis`.

theodelrieu:
Thanks, (sorry, I missed your first post)
When I try to add your sample code (with the to_json function) in `unit-udt.cpp`, I get a lot of `ambiguous symbol 'to_json'`. I'm quite sure this is also linked to your compiler errors.

Indeed, in the code I call `nlohmann::to_json` explicitely, and the compiler cannot decide which symbol is the right one:

- Either the `const nlohmann::detail::to_json_fn &nlohmann::'anonymous-namespace'::to_json`
- Or your overload

I don't know if there is a workaround to that particular issue, I think users should not add an overload in `nlohmann` namespace in the first place.


However you could either:

- partially specialize `nlohmann::adl_serializer` (I don't think this is a very good idea though, we might add default specializations in the future)
- create those methods in the `std` namespace (just kidding...)
- or create your own serializer (which is a bit annoying since you have to define your own `basic_json` alias..)

If you have something better in mind, I'd be glad to hear it!

qis:
@theodelrieu Since you already committed yourself to the free function approach - no idea. I typically use the struct solution: <https://github.com/qis/sql/blob/master/include/sql/types.h#L76>

Here is how a user defined type looks like. I find it particularly important to support non-default-constructible types: <https://github.com/qis/sql/blob/master/include/sql/types/optional.h>

gnzlbg:
> I think users should not add an overload in nlohmann namespace in the first place.

Yes, that is not supported, `nlohmann::` behaves like `std::`. 

`adl_serializer` works by adding overloads of `to_json`/`from_json` to the associated namespace of their types. 

For `std::optional` you cannot add an overload to `std::`, and you cannot add an overload to `nlohmann`, so you need to specialize `nlohmann::adl_serializer` _or_ define your own serializer trait class and pass it as a template argument to `basic_json`.

theodelrieu:
@qis I would like to support non default-constructible types too, I already tried to implement it, but I couldn't do it with the free function approach, because of ADL.

I think there is a way though, we could detect if there is a `from_json` overload of this form:

```cpp
// Takes only the json in argument, returns a T
JSONSerializer<T>::from_json(Json const&);

// then constraint the current get method on is_default_constructible and add this one
template <typename T, typename = enable_if_t<is_same<T, 
decltype(JSONSerializer<T>::from_json(std::declval<basic_json_t>()))>::value>>
T get()
{
  return JSONSerializer<T>::from_json(*this);
}
```

Thus, non default-constructible types would only be supported via a custom serializer.
It's not ideal though, I'd like to see a better way of doing this

qis:
@gnzlbg If I move the free functions to the global namespace or a named namespace, it stops working in MSVC with and without my changes.

qis:
@theodelrieu If normal users will have to write code similar to your example, they're screwed. No offense. =)

theodelrieu:
@qis I wasn't clear in the example. The user code would be the custom serializer only

gnzlbg:
@qis The default serializer is `nlohmann::adl_serializer`. ADL means Argument Dependent Lookup. It looks for functions in the current namespace _and_ in the _associated_ namespace of the arguments. 

The global namespace is not an associated namespace of `std::optional`, `std::`/`std::experimental` should be, depending on where `std::optional` is defined. 

However, adding overloads to `std::` is _undefined behavior_. So don't do it.

Adding an overload to `nlohmann::` is also undefined behavior. So don't do it. 

The only one that can add an overload to `nlohmann::` is `nlohmann` itself. That overload would need to be properly constrained, as well as the rest of the machinery, to prevent ambiguouities.

Even if you manage to constrain it properly, your code will break if we add this overload in the future.

The _right_ way to solve this problem right now is to specialize `nlohmann::adl_serializer` for `std::optional<T>`. This is something you can do, and we won't break your code if we decide to add an overload later.

@theodelrieu I think that right now, ADL is only used if the `adl_serializer` is not specialized, that is, as a fall back, but we should check this, in case users like @qis specialize it for std types, and we decide to add overloads later with slightly different behavior. If it works like that, their specialization will still be used, and we won't break their code.

gnzlbg:
> @theodelrieu If normal users will have to write code similar to your example, they're screwed. No offense. =)

@qis The user code is just: 

```c++
template <>
struct adl_serializer<my_type>  // +- ",void" {
  template<typename JSON>
  static my_type from_json(JSON const& j) { /* logic here */ }
};
```

For non-default constructible types, ADL cannot work, so specialization is the only way.

theodelrieu:
@gnzlbg I wonder if a partial specialization of `adl_serializer` could cause issues later.

```cpp
namespace nlohmann
{
template <typename T>
struct adl_serializer<std::experimental::optional<T>>
{
  template <typename Json>
  std::experimental::optional<T> from_json(Json const&j) { /* logic */ }
};
}
```

It wouldn't if `nlohmann` never specialize `adl_serializer` at all.

theodelrieu:
I committed the support for non default-constructible udts, I did that quickly, I'll rewrite/improve it as soon as I can.

Especially I think there are a bit more constraints to put, and tests to rewrite.

gregmarr:
> Adding an overload to nlohmann:: is also undefined behavior. So don't do it.

Why would you say that?  Adding an overload to std:: is only undefined behavior because the standard says it is.  The standard says nothing about nlohmann::.  It may be "not-recommended" behavior, but it's definitely not "undefined" behavior.

> The only one that can add an overload to nlohmann:: is nlohmann itself. That overload would need to be properly constrained, as well as the rest of the machinery, to prevent ambiguouities.

There is no such thing as "nlohmann itself".  Any code can legally add any function to the nlohmann namespace.  That's not to say that it won't result in possible ambiguities or conflicts, now or in the future, but it is perfectly legal.

Maybe what you were trying to say was that it could result in undefined behavior if multiple bits of code did it differently for the same types, resulting in ODR violations.

gnzlbg:
@gregmarr 
> Why would you say that? 

The standard, for better or worse, does not have exclusivity on undefined behavior. A library gets to decide which usage of its API does or does not result in undefined behavior, for example, on the API of its functions:

```c++
/// \precondition If `i` is out of bounds the behavior is undefined.
template <typename T> boost::container::vector<T>::operator[](size_t i);
```

Specifying undefined behavior is actually as easy, just adding comment. You can find this in a lot of APIs like POSIX, boost, ... everywhere where you can find a precondition, typically either the behavior on precondition violation is defined (e.g. throwing an exception) or it is not.

> That's not to say that it won't result in possible ambiguities or conflicts, now or in the future, but it is perfectly legal.

That is optimistic. The design used for `adl_serializer` (which is that of N4381) requires that users don't do this. What will happen if a user does so? We cannot actually know, and we cannot define what will happen, hence, the behavior is undefined.

It might work, it might hopefully produce an error (static assert, ambiguity, assertion, throwing an exception), but if instead another overload (which we might add in the future) is silently picked instead and some preconditions are violated it might silently result in memory corruption, a segfault, or a security vulnerability everything works. 

@theodelrieu I think that we should protect users (like @qis) against this, by making it obvious that this is not intended. A way to do this is to put all the machinery into a `detail` namespace:

```c++
namespace nlohmann
  namespace detail {
      // from_json / to_json overloads
      // from_json_fn / to_json_fn classes which use ADL
      namespace { /*from_json/to_json inline variables*/ }
  }
 // adl_serializer trait, that as a fallback calls detail::to_/from_json inline variables
}
```

We could also name the namespace `dont_add_overloads_here` or the typical `adl_barrier` that boost uses.

gregmarr:
"The standard, for better or worse, does not have exclusivity on saying what is undefined behavior."

Actually, it does.  The term has a very specific meaning, which is not what you're using here.

"A library gets to decide which usage of its API does or does not result in undefined behavior, for example, on the API of its functions:"

That may be true, specific things passed to a function may result in undefined behavior because it results in that function doing something which the standard says is undefined behavior.  Passing nullptr as a function parameter to a function that dereferences it without checking results in undefined behavior because the caller violated a precondition.  However, that's not what you said.  You said that adding things to the nlohmann namespace is undefined behavior, which it's not.  Namespaces are specifically and intentionally allowed to be extended at any time, by any code.

"We cannot actually know, and we cannot define what will happen, hence, "the behavior is undefined"."

That's not `undefined behavior` in the C++ sense.  That's unknown behavior on behalf of the library developer.  Unless the code does something which the standard says is undefined behavior, then it's not undefined behavior.

"It's not a good idea to do this, because the library doesn't support it, so don't do it." is not the same as "It's undefined behavior to do this, so don't do it."  Adding functions to the std namespace is not allowed by the standard.  Adding functions to the nlohmann namespace is specifically allowed by the standard.  Conflating the two diminishes your argument about why it shouldn't be done.  If users think that the only reason they shouldn't do something is based on incorrect reasoning, then they will likely think that it's okay to do it.  So provide them with sound reasons not to do it, such as "future additions to the library may cause your program to fail to compile" or "if different parts of the same program implement the function differently for the same type, that's undefined behavior because it violates the One Definition Rule".

TurpentineDistillery:
This is a language-lawyer issue, I think.
My understanding is that adding things to a 3rd party namespace is a "bad thing" because in the future the owner of the namespace may add conflicting declarations with different preconditions/postconditions that are not met by the existing 3rd-party code. Hence it is preemptively declared as undefined-behavior by the standard, but the reasoning is equally applicable to any namespace.

gregmarr:
I agree that it's generally bad practice, I just want to be sure that people give the right reasons why it's bad practice.

> This is a language-lawyer issue, I think.

Pretty much, but it's good to be precise in an extensibility situation like this, when describing why something can't or shouldn't be done.

> My understanding is that adding things to a 3rd party namespace is a "bad thing" because in the future the owner of the namespace may add conflicting declarations with different preconditions/postconditions that are not met by the existing 3rd-party code.

True, and also because it may violate the one definition rule, though this is less likely on the owner side, but more on the multiple user side.  The ODR violation would probably be caught right away with a single header library, as the two definitions would show up in the same translation unit all the time.

See this very timely blog post on the topic:
https://akrzemi1.wordpress.com/2016/11/28/the-one-definition-rule/

> Hence it is preemptively declared as undefined-behavior by the standard, but the reasoning is equally applicable to any namespace.

It is undefined behavior in only these two cases:

>The behavior of a C++ program is undefined if it adds declarations or definitions to namespace std or to a namespace within namespace std unless otherwise specified.

>The behavior of a C++ program is undefined if it adds declarations or definitions to namespace posix or to a namespace within namespace posix unless otherwise specified. 

For any other namespace, it's simply a bad idea unless the namespace was written to expect it, or there is some other method to be sure that it doesn't cause any other problems.  That's difficult at best.  The standard has no concept of "third party libraries" or "the owner of a namespace", and therefore it can't say that no one other than the owner of a namespace can add to it.

gnzlbg:
> For any other namespace, it's simply a bad idea unless the namespace was written to expect it, or there is some other method to be sure that it doesn't cause any other problems. That's difficult at best.

In this particular case, the customization point functionality of `adl_serializer` is actually designed to properly work _only_ if the user does not do this. Otherwise, it will very likely not work, since this is what allows us to add overloads at will to offer fallbacks that support lots of types out of the box, like:

```c++
// fallback for all ranges:
template <typename Rng, typename J>
void to_json (Rng&& rng, J& j) requires(Range<Rng> && JSON<J>) { ... }

// fallback for all "optional-like" objects: std/boost optional, std/boost any, 
// T*, smart pointers ...
template <typename O, typename J>
void to_json (O&& rng, J& j) requires(OptionalLike<O> && JSON<J>) { ... }
```

We can add and remove such overloads at will, and if a user adds an overload to our namespace, his overload might not be picked up because one of these overloads taking an `&&` reference might be a better match. This _might_ result in undefined behavior in the C++ standard sense (we cannot know the consequences of this), not necessarily because of ODR, out of bounds, null ptr dereference..., but maybe just because this results in us violating the precondition of a std function. Even if it doesn't result in "C++ standard undefined behavior", what will the behavior of the program be? Do we want to define it? If not, the behavior of the program is still undefined in the general sense of the word undefined.

IMO, stating in the docs that adding an overload to our namespace for adl dispatch "is undefined behavior" is the most succinct way to put it. We can add a clarification saying that this might result in the wrong overload being picked up for a particular type, and that the consequences of this cannot be known, but might include compiler errors (ambiguities, static asserts), run time errors (assertions, exceptions), or any form of C++ undefined behavior in the C++ standard sense.

gregmarr:
All I'm saying is that C++ has specific meanings for the terms "undefined behavior", "unspecified behavior", and "implementation defined behavior" and we shouldn't use those terms, or similar terms like "behavior is undefined" when we simply mean "we can't easily predict for you what the behavior may be due to complex overload rules" or "the user-defined implementation of this function may contain undefined behavior because it contains a bug" or "future updates of the library may cause such code to break".  It's not that difficult to simply avoid those terms.

It is definitely possible for someone that is intimately familiar with the standard to add overloads to the nlohmann namespace and know exactly what the behavior is going to be.  We can say that it's not supported, only for experts, use at your own risk, etc., but shouldn't say that it's not allowed *by the C++ standard*.

gnzlbg:
> but shouldn't say that it's not allowed by the C++ standard.

I just said that it is not allowed _by this library_.

> We can say that it's not supported, only for experts, use at your own risk, etc.,

We can just say: "Adding overloads to this namespace is ~undefined behavior~ not _allowed_  (by this library)".

theodelrieu:
@gnzlbg I agree on hiding the inline `to_json/from_json` in a namespace.

I tried to move them in the `detail` namespace, however it caused issues with the MSVC ADL workaround : the `void to_json();` declarations.

I shall try that with a bit more thinking later.


On another topic, I started to rewrite `unit-udt.cpp` as I mentioned before, and there are two things that I tried to write, that didn't compile:

```cpp
struct age{int val;};

void some_func()
{
  age a{42};
  json j;

  // (1)
  j["age"] = a.val;

  // (2)
  json j2{{"age", a}};
}
```

Both are tied to a problem I mentioned at the beginning of the first PR:

> The constructor of `basic_json` for UDT is marked explicit

This was intended, to prevent something like this:

``` cpp
struct not_comparable{};

void to_json(json&, not_comparable){}

// ...

// deep in user's code

not_comparable t1, t2;
// should be a compile-time error, but since not_comparable is convertible to basic_json
// (through the to_json free-function) it instead constructs two basic_json objects and call
// operator==(basic_json, basic_json) on them
if (t1 == t2) { /* do stuff */ }

```

In this code, we consider the mentioned `basic_json` constructor to **NOT** be marked explicit.
This problem exists already, since there are a lot of template constructors for `basic_json`
(though it would be an issue for types like a custom `std::vector` without an `operator==`, which is a bit more special than a plain UDT)

So my question is: What do we want to do about this?

Explicitely constructing a json object from a UDT is quite tedious, so I would be tempted to remove the `explicit` keyword.

If we do, there is work to do to constrain all `basic_json` free and friend functions (I don't know if this is even possible).

What do you all think?

matspetter:
Will this feature make it into the library soon or what is you thinking? Will it then be documented on the front page how to do it? It is a very very nice feature!!

theodelrieu:
@matspetter It will be there soon™. Once the technical part is settled, it will be well documented, and I think it should be on the front page too.

@nlohmann I think we should also show #290 (iterator parsing) on the front page too.

gnzlbg:
@theodelrieu 

>I agree on hiding the inline to_json/from_json in a namespace.
>
>I tried to move them in the detail namespace, however it caused issues with the MSVC ADL workaround : the `void to_json();` declarations.

Hm, did you try to move these into the `detail` namespace as well? Maybe they need to be in multiple namespaces? I have no access to MSVC so I cannot really try here :/

> So my question is: What do we want to do about this?

IIUC the problem is that when a type is implicitly serializable to json, it becomes usable in all operations in which a `basic_json` object is, because of the implicit conversion chain. `operator==` is just a situation in which this happens, but it can potentially happen everywhere. One question is: do we already have this problem? For example, the library already has an implicit constructor from ranges (e.g. things that implement begin and end). Does `my_non_comparable_vector` already run into this issue? I think so. 

OTOH, we _do_ want implicit conversions from UDTs. This is what makes the library really easy to use. @nlohmann has the last say on this, but I would say that this library prioritizes ease of use above everything else (as long as the tradeoffs taken are still reasonable).

So I think that it is worth it to keep implicit conversions/construction from UDTs (something that might already be happening for ranges/containers). 

The question is, do we want to allow the users to opt-out of this? 

We could add machinery for this. I would say that another customization point, `explicit_to_json` should be enough. That is, if a user implements:

- `explicit_to_json`, the explicit constructor is the only one that works. 
- `to_json`, the implicit constructor is the only one that works.
- `both`, we produce an error (ambiguity would be enough for a first implementation, but _maybe_ we could produce a nice static assert in the future)

So I would say that we could have our cake and eat it too, but this has a cost: an extra customization point. Is that worth it? I don't know. Maybe?

theodelrieu:
I tried to reproduce the issue with `basic_json`. I couldn't, so I searched a bit, and I think the problem we are discussing does not apply in our case.

Let me explain with two coliru examples:

[implicit-conversion](http://coliru.stacked-crooked.com/a/78809652f71367e7)

[compile-error](http://coliru.stacked-crooked.com/a/38d1a33edd55611e)

In the second one, the converting struct has **template arguments**, which prevents the sneaky implicit conversion, since the compiler cannot deduce `T` when trying to convert `int` to `basic_json<T>`.

So why did I bring up that issue in the first place?
My bad, I tested with a simple example without template arguments, and assumed it would do the same thing with `basic_json` without testing.

This is good news (unless I missed something big), because we could remove the `explicit` keyword, and with a bit of SFINAE where needed, it should work!

nlohmann:
Hi everyone!

I am impressed by the work you are doing here, and I am afraid I lost track of the discussion here. I am really sorry for that, and I hope you understand that it's not out of disrespect that I do not participate in the discussion.

I am really looking forward to this feature, but before finally merging it, I would need someone to walk me through the code - I really want to understand the final outcome, and also some of the design decisions. If this is OK to you, we may find a time for a chat (Slack? Gitter?).

All the best,
Niels

gnzlbg:
@nlohmann Hi Niels, @theodelrieu has done an awesome work exploring the different design solutions, and I the current implementation (on the last commit) seems to already be a design that can make everybody happy. Some tests are already there, but I guess some more will come to try _hard_ to break the current implementation so that we can be sure that it is what we want. This should allow @theodelrieu to refactor and maybe simplify the implementation a bit. The docs still need to be written, but I am confident that we will come up with a simple and short way to explain the system that is currently in place. Writing it requires using some advanced "pattern/tricks", but using it correctly is easy, and the error messages when one makes a mistake can be made pretty good.

theodelrieu:
@nlohmann @gnzlbg Hi guys! I'll be glad to chat with you, I'm in France, and I have time on Sunday.

Even if the PR is not done, I think there won't be major design changes, thus waiting longer before this chat doesn't seem necessary to me.

theodelrieu:
I have good news! I managed to remove the explicit keyword ! (via some quickfixes that will be changed soon) I'll commit tomorrow when I come home.

@nlohmann Niels could you create a slack room and pm it to @gnzlbg and me? It will be easier to plan a meeting if we already have the room IMO.

Have a good night.

nlohmann:
Great news!

test/src/unit-class_const_iterator.cpp
```diff
@@ -220,20 +220,20 @@ TEST_CASE("const_iterator class")
             {
                 json j(json::value_t::null);
                 json::const_iterator it = j.cbegin();
-                CHECK(it.m_it.primitive_iterator == 1);
+                CHECK(it.m_it.primitive_iterator.m_it == 1);
```
nlohmann: Just out of curiosity: why is the `.m_it` suffix necessary?

theodelrieu: Well this a consequence of removing the `explicit` keyword on the new constructor.

As I mentioned a few days ago:

> In the second [example], the converting struct has template arguments, which prevents the sneaky implicit conversion, since the compiler cannot deduce T when trying to convert int to basic_json<T>.

Although code outside the `basic_json` class is protected from this *sneaky implicit conversion*, the code **inside** is not!

Every implicit conversion inside `basic_json` will try to use the `T &&` constructor, if it goes trough the enable_ifs.

It happened with `primitive_iterator` because there was an implicit conversion operator to `difference_type`, that's why I reimplemented the iterator operators (copy-pasted until I refactor, sorry about that).

That's also the reason why I made the `.m_it` suffix public, it was also needed in the tests, because Catch does fancy capture stuff, and seems to be able to call the doomed constructor (through a call to `operator<<(ostream&, primitive_iterator)` if I remember correctly)

I should have added comments explaining this, sorry about that too!

nlohmann: Thanks for the explanation. Sorry for not having seen this in the discussion above. Great work!

theodelrieu:
Hi everyone, we're almost done!

@nlohmann, @gnzlbg and I will have a Slack meeting shortly, to discuss the design decisions, and decide where we go from here.

If you want to join, feel free to send me an email, I'll invite you to the Slack team!

I'd like to plan it for this weekend, if it fits everyone's agenda. Let me know and I'll post the meeting date on Slack and Github.

Thanks

nlohmann:
Sunday or Friday evening (CET) would fit for me.

gnzlbg:
Sunday fits me, today it depends.

theodelrieu:
Sunday is fine for me, I can't make it today unfortunately

nlohmann:
How about 19:00 (Berlin time) today?

theodelrieu:
No problem for me

theodelrieu:
Hi, I was thinking about what we talked about yesterday, i.e. replacing existing `basic_json` constructors by `to_json`/`from_json` overloads.

This is not only about cleaning code, but this enables a feature:

In the code, we call `JSONSerializer<T>::to_json` in the new constructor, but there is a lot of enable_ifs to avoid chosing the others (when T = `int` for example).

The issue with those constraints is that users cannot override the default behaviour for `vector`, `array` or unscoped enums (and all types already handled by the library)

But if we change constructors to free functions, those functions would be called via the default `adl_serializer<T>::from_json`, thanks to ADL.

Which means, if a user specialized `adl_serializer<std::vector<T>>`, that his specialization will be called.
This allows to support unscoped enums, and to give full control to the users that need it.

Also, they would still be able to call the default behaviour:

```cpp
std::vector<int> v{1, 2, 3};

// this is also called by the default adl_serializer too
auto j = nlohmann::to_json(v);
```

This might have been mentioned by @gnzlbg yesterday, I don't remember, but I thought it would be useful to share that thought.

gnzlbg:
No, I didn't mention it (and hadn't really thought about it), but yes, you are correct, if everything goes through the trait system then the trait system would allow that kind of "override" (note: this kind of override is not possible with just ADL because adding overloads to `std::` is undefined behavior).

theodelrieu:
Indeed, but if we add the overloads in the `detail` namespace, it should work:

```cpp
namespace nlohmann::detail {

// this should be found via unqualified call
template <typename JSON, typename T, typename Allocator>
void to_json(JSON&, std::vector<T, Allocator> const&) {}

struct to_json_fn {
template <typename JSON, typename T>
void operator()(JSON& j, T&& val) {
  // for T = std::vector<U>, this should call the overload ahead
  to_json(j, std::forward<T>(val));
}
};
} // nlohmann::detail

```

I haven't tried to compile this, but I think it's valid

gnzlbg:
Yes that would be the way to move the "constructors" out of `basic_json`. What I meant was only that the only way for users to override this would be to either specialize the `adl_serializer` trait or pass `basic_json` a different serializer trait since they cannot add overloads to the namespaces associated with those types. I guess this should be documented.

nlohmann:
This can be closed, right? After all, we have #423 now.

qis:
Looks like it. Thank you very much for caring about this topic and not rushing things.

theodelrieu:
Yes, this can be closed

Issue #362, opened 2016-11-19
Append ".0" to serialized floating_point values that are digits-only.


TurpentineDistillery:
The code is already smart enough to parse "123.0" as floating-point type.
The problem is that this does not round-trip, as the value serializes as "123", so now it is indistinguishable from pure-integer, and the floating-point-ness information is lost.

Proposed implementation for serializing numeric values below.

```c++
    // Convert floating-point type to string in POSIX locale.
    // If result is composed entirely of digits or minus-sign,
    // append ".0" to preserve the type information.
    template<typename T>
    static std::string as_str(const T x, /*is_integral=*/std::false_type)
    {
        static thread_local std::unique_ptr<std::stringstream> sstr;
        // reusing sstr per-thread as construction of stringstream
        // and ::imbue are both expensive.
        if(!sstr) {
            sstr.reset(new std::stringstream);
            sstr->imbue(std::locale::classic());
        }

        sstr->precision(std::numeric_limits<T>::digits10);

        *sstr << x;
        std::string s = sstr->str();
        sstr->str(std::string()); // clear the data

        if(s.find_first_not_of("0123456789-") == std::string::npos) {
            s += ".0";
        }

        return s;
    };

    // Convert any integral type to base-10 representation,
    // using POSIX locale (no grouping characters are used).
    //
    // This is simple-enough so we can bypass using
    // the posix-locale-imbued stream and do it "manually"
    template<typename T>
    static std::string as_str(T x, /*is_integral=*/std::true_type)
    {
        std::string s;
        const bool is_neg = x < 0;

        while(x) {
            s.push_back('0' + abs(x % 10));
            x /= 10;
        }

        if(s.empty()) {
            s.push_back('0');
        }

        if(is_neg) {
            s.push_back('-');
        }

        reverse(s.begin(), s.end());
        return s;
    }

    // Dispatcher to as_str variants dealing with integral vs floating-point types
    template<typename T, typename = typename std::enable_if<std::is_arithmetic<T>::value>::type >
    static std::string as_str(const T x)
    {
        return as_str(x, std::is_integral<T>());
    }
```

nlohmann:
FYI: I checked PR #378 and added some comments there.

TurpentineDistillery:
The implementation details and relevant discussion is in the PR, so I think this can be closed.

PR #378, opened 2016-12-06
locale-independent num-to-str

TurpentineDistillery:
This implements nlohmann/json#362
In order to determine whether a floating-point string representation needs ".0" appended, we need to be able to examine the string content before dumping it into stream. One way of doing that is to reuse thread-local classic-locale-imbued stringstream, but that has additional overhead, and also clang on my mac does not seem to know about `thread_local`. So I went with a different approach - use `snprintf` and then fix the result to be POSIX-locale compliant (erase decimal grouping characters and replace decimal separator with '.').

Also, this gives 2x speedup to `dump numbers/signed_ints.json` benchmarks. Others are unchanged.


nlohmann:
I shall have a look at the PR this week.

nlohmann:
Hey @TurpentineDistillery, thanks for the PR!

I can confirm some of your numbers - here is the output of the benchmark (mean of 100 runs):

Test | develop | this PR | difference
---- | ---- | ---- | ----
dump jeopardy.json | 508.225 ms | 561.056 ms
dump jeopardy.json with indent | 563.668 ms | 599.749 ms
dump numbers/floats.json | 729.318 ms | 693.462 ms
dump numbers/signed_ints.json | 255.533 ms | 123.513 ms

There is a 2x speedup for integers, a slight speedup for float, but for some reason, the jeopardy-example is slower, which makes no real sense.

nlohmann:
I ran another 100 repetitions. Here are the harmonic means:

Test | develop | this PR
---- | ---- | ----
dump jeopardy.json | 569.3 ms | 573.2 ms

TurpentineDistillery:
Based on the previous runs, this appears to be within the experimental error (maybe add CIs to the timing outputs?) . By the way, jeopardy.json contains no numbers, only strings.

nlohmann:
I know - that's why I am so puzzled that the numbers are so fragile...

whackashoe:
Try running `perf stat -Bddd ./whatever` and compare off a few runs of each?

On https://github.com/nlohmann/json/pull/337 I remember jeopardy-indented was consistently worse too but it was very slight.

TurpentineDistillery:
Another strange observation:
Harmonic mean is biased toward the minimum of sample of values (arithmetic mean of a sample with non-zero variance is always larger). However, in the data above we see the opposite: the arithmetic mean for `develop` was 508.225 ms, and the harmonic mean was 569.3 ms, which is indicative of the latter execution being consistently slower - the computer was was doing something else during the latter benchmarking, causing context-switches or IO-related slowdowns. Either that or a bug ; )

Somewhat related to benchmarking in general: https://stackoverflow.com/questions/9006596/is-the-unix-time-command-accurate-enough-for-benchmarks

nlohmann:
@TurpentineDistillery You are right - this all makes no sense. In the develop branch, I am using benchpress with some additional tweaks to avoid cache-related outliers. Then I had a look at nonius which has a nicer API and output options, but comes to these strange results.

I shall try Google Benchmark...

nlohmann:
I added Google Benchmark to a feature branch and let the serialization benchmarks run (to is the code from the develop branch, bottom the code from this PR):

benchmarks

In this benchmark, the PR is consistently more than twice as fast as the code from develop! 👍

src/json.hpp.re2c
```diff
@@ -7892,6 +7867,159 @@ class basic_json
...
+                                      : d == 16 ? "%.17g"
+                                      : d == 17 ? "%.18g"
+                                      :           "%.19g";
+            // I'm not sure why we need to +1 the precision,
+            // but without it there's a unit-test that fails
+            // that asserts precision of the output
+
+            snprintf(m_buf.data(), m_buf.size(), fmt, x);
+
+#if 0
```
nlohmann: Is this `#if 0` branch still relevant?

TurpentineDistillery:
> In this benchmark, the PR is consistently more than twice as fast as the code from develop!

Huh? How is that possible?? The only thing that's supposed to be faster is the writing of integer types that is done "by hand" - I'd expect others to be unaffected, or minimally affected.

src/json.hpp.re2c
TurpentineDistillery: Yes - this is a here-be-dragons note to a future clever contributor (or code peruser) who'll think "let's use c++-flavor locales here!" Feel free to get rid of it.

nlohmann:
I think the fact that the locale is not reset with every dump call may make a difference.

This code is not executed for every dump:

```cpp
        // fix locale problems
        ss.imbue(std::locale::classic());

        // 6, 15 or 16 digits of precision allows round-trip IEEE 754
        // string->float->string, string->double->string or string->long
        // double->string; to be safe, we read this value from
        // std::numeric_limits<number_float_t>::digits10
        ss.precision(std::numeric_limits<double>::digits10);
```

TurpentineDistillery:
I have not observed similar speedup with running json_benchmarks, however.

nlohmann:
I currently have no real way to produce reproducible benchmark results :(

nlohmann:
Google Benchmark *feels* plausible, but I have no real way to verify the results.

TurpentineDistillery:
I ran my own 'ghetto' benchmarking, which is, as you might have expected by now, is in disagreement with all of the results obtained with various approaches above. It is about as simple as "hello world", measuring the time to dump json record in a loop until it totals 100000000 bytes (I didn't bother factoring out the bootstrap time to initially deserialize the record). It shows that PR is about the same as develop in all cases; maybe just marginally faster. I ran it a few times and got consistent numbers.

```bash
>>g++ --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk/usr/include/c++/4.2.1
Apple LLVM version 6.0 (clang-600.0.57) (based on LLVM 3.5svn)
Target: x86_64-apple-darwin13.4.0
Thread model: posix

>>cat main.cpp
#include "src/json.hpp"
#include <iostream>

int main()
{
    nlohmann::json j;
    std::cin >> j;
    while(true) std::cout << j;
    return 0;
}

>>git checkout develop
>>g++ -std=c++11 -O2 main.cpp -DNDEBUG -o a.out.develop

>>git checkout feature/locale_independent_num_to_str
>>g++ -std=c++11 -O2 main.cpp -DNDEBUG -o a.out.pr


>>for file in `find benchmarks/ -name *.json`; do echo $file; cat $file | time ./a.out.pr | head -c 100000000 >/dev/null;  done
benchmarks//files/jeopardy/jeopardy.json
Command terminated abnormally.
       13.02 real        12.83 user         0.18 sys
benchmarks//files/nativejson-benchmark/canada.json
Command terminated abnormally.
       11.97 real        11.94 user         0.02 sys
benchmarks//files/nativejson-benchmark/citm_catalog.json
Command terminated abnormally.
        8.78 real         8.76 user         0.02 sys
benchmarks//files/nativejson-benchmark/twitter.json
Command terminated abnormally.
        8.10 real         8.08 user         0.01 sys
benchmarks//files/numbers/floats.json
Command terminated abnormally.
       13.63 real        13.59 user         0.03 sys
benchmarks//files/numbers/signed_ints.json
Command terminated abnormally.
        9.44 real         9.41 user         0.02 sys
benchmarks//files/numbers/unsigned_ints.json
Command terminated abnormally.
        9.41 real         9.38 user         0.02 sys


>>for file in `find benchmarks/ -name *.json`; do echo $file; cat $file | time ./a.out.develop | head -c 100000000 >/dev/null;  done
benchmarks//files/jeopardy/jeopardy.json
Command terminated abnormally.
       13.05 real        12.85 user         0.18 sys
benchmarks//files/nativejson-benchmark/canada.json
Command terminated abnormally.
       14.25 real        14.22 user         0.02 sys
benchmarks//files/nativejson-benchmark/citm_catalog.json
Command terminated abnormally.
        9.48 real         9.46 user         0.02 sys
benchmarks//files/nativejson-benchmark/twitter.json
Command terminated abnormally.
        8.20 real         8.18 user         0.02 sys
benchmarks//files/numbers/floats.json
Command terminated abnormally.
       14.75 real        14.72 user         0.03 sys
benchmarks//files/numbers/signed_ints.json
Command terminated abnormally.
       10.59 real        10.56 user         0.03 sys
benchmarks//files/numbers/unsigned_ints.json
Command terminated abnormally.
       10.52 real        10.49 user         0.03 sys

```

TurpentineDistillery:
> I think the fact that the locale is not reset with every dump call may make a difference.

The profiler did not identify these as hot-spots. The locale instantiation (prior to using classic) was super expensive, if you remember, but imbueing and setting precision once per dump was not expensive.

nlohmann:
Sigh... I had another look at the benchmarks. Apart of changing the header file (develop/PR), the numbers stay the same

```
$ ./benchmark-develop --benchmark_filter="dump.*" --benchmark_repetitions=30 --benchmark_report_aggregates_only=true
Run on (8 X 2900 MHz CPU s)
2017-01-11 19:34:34
Benchmark                                          Time           CPU Iterations
--------------------------------------------------------------------------------
dump data/jeopardy/jeopardy.json_mean            271 ns        270 ns    2625706   14.1525MB/s normal
dump data/jeopardy/jeopardy.json_stddev           11 ns         10 ns          0   566.813kB/s normal
dump data/jeopardy/jeopardy.json_mean            267 ns        267 ns    2537979   14.3072MB/s pretty
dump data/jeopardy/jeopardy.json_stddev           10 ns         10 ns          0   566.589kB/s pretty
dump data/numbers/floats.json_mean               270 ns        270 ns    2643914   14.1578MB/s normal
dump data/numbers/floats.json_stddev              10 ns         10 ns          0   563.347kB/s normal
dump data/numbers/floats.json_mean               267 ns        266 ns    2761069   14.3434MB/s pretty
dump data/numbers/floats.json_stddev              12 ns         12 ns          0    667.08kB/s pretty
dump data/numbers/signed_ints.json_mean          264 ns        264 ns    2781078   14.4766MB/s normal
dump data/numbers/signed_ints.json_stddev         13 ns         12 ns          0   708.379kB/s normal
dump data/numbers/signed_ints.json_mean          268 ns        267 ns    2760285   14.3026MB/s pretty
dump data/numbers/signed_ints.json_stddev         12 ns         12 ns          0   665.562kB/s pretty

$ ./benchmark-pr --benchmark_filter="dump.*" --benchmark_repetitions=30 --benchmark_report_aggregates_only=true
Run on (8 X 2900 MHz CPU s)
2017-01-11 19:39:48
Benchmark                                          Time           CPU Iterations
--------------------------------------------------------------------------------
dump data/jeopardy/jeopardy.json_mean            116 ns        116 ns    5679375   33.0805MB/s normal
dump data/jeopardy/jeopardy.json_stddev            6 ns          6 ns          0   1.64533MB/s normal
dump data/jeopardy/jeopardy.json_mean            118 ns        118 ns    5818545   32.3685MB/s pretty
dump data/jeopardy/jeopardy.json_stddev            3 ns          3 ns          0   951.512kB/s pretty
dump data/numbers/floats.json_mean               120 ns        119 ns    5542447   31.9539MB/s normal
dump data/numbers/floats.json_stddev               2 ns          2 ns          0   496.115kB/s normal
dump data/numbers/floats.json_mean               119 ns        119 ns    6097083   32.1939MB/s pretty
dump data/numbers/floats.json_stddev               1 ns          1 ns          0   401.886kB/s pretty
dump data/numbers/signed_ints.json_mean          119 ns        118 ns    6104687   32.2889MB/s normal
dump data/numbers/signed_ints.json_stddev          2 ns          2 ns          0   610.253kB/s normal
dump data/numbers/signed_ints.json_mean          118 ns        118 ns    5893595   32.3966MB/s pretty
dump data/numbers/signed_ints.json_stddev          3 ns          3 ns          0   816.329kB/s pretty
```

TurpentineDistillery:
Well, at least we can conclusively say that PR is not worse performance-wise : )

Issue #380, opened 2016-12-06
Bug in overflow detection when parsing integers


TurpentineDistillery:
```c++
                    // multiply last value by ten and add the new digit
                    auto temp = value * 10 + *curptr - '0';

                    // test for overflow
                    if (temp < value || temp > max)
                    {
                        // overflow
                        type = value_t::number_float;
                    }
```
The `temp < value` above will not detect overflow for some values, as we're not just adding a small value, but also multiplying by 10, so this may overflow couple of times over and end-up satisfying `temp >= value`. As for `temp > max` - it looks to me that it will never happen unless the width of `temp` is greater than that of `max`

Try round-tripping 166020696663385964490

nlohmann:
Merged #390.

PR #390, opened 2016-12-10
Fix issue #380: Signed integer overflow check

qwename:
See issue #380.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -9311,437 +9311,915 @@ class basic_json
                 m_start = m_cursor;
                 assert(m_start != nullptr);
 
-                
-    {
-        lexer_char_t yych;
-        unsigned int yyaccept = 0;
-        static const unsigned char yybm[] = {
-              0,   0,   0,   0,   0,   0,   0,   0, 
-              0,  32,  32,   0,   0,  32,   0,   0, 
-              0,   0,   0,   0,   0,   0,   0,   0, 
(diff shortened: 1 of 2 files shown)


qwename:
Previously, `json::parse("166020696663385964490").dump()` gave 18446744073709551562. Now it gives 1.66020696663386e+20.

qwename:
Off-topic, but I had to compile the latest re2c (0.16) as Fedora 25 only provided 0.14.3, which didn't have the `-W` option. The option was added in 0.15, so this might be useful to add in `.github/CONTRIBUTING.md` http://re2c.org/news/changelog/changelog.html.

It seems like the new version also reduced the generated code length.

TurpentineDistillery:
Or, one could make changes to the .hpp, and then have a pre-commit hook do
`git checkout src/json.hpp.re2c && git diff src/json.hpp | patch src/json.hpp.re2c`

nlohmann:
@qwename Thanks for the hint on the re2c version. I shall add a note to the contributing file. For the whitespace, you may want to execute `make pretty` which uses [astyle](http://astyle.sourceforge.net).

qwename:
I missed the part about `make pretty`, fixed it now.

nlohmann:
There are some conversion warnings. Can you fix them?

nlohmann:
Thanks!

nlohmann:
Thanks! I shall add a regression test for `166020696663385964490`.

PR #391, opened 2016-12-10
Clang: quiet a warning

jaredgrubb:
In my project I have a lot of warnings enabled and -Wdocumentation fires a lot in this header (-Wdocuemntation does a check on Doxygen comments to see how they match what they're attached to; I dont see much value in trying to fix the comments so I'd like to quiet the warning).

Note I tried running re2c but it added a lot of other changes, so I think maybe I have a different version or 'develop' is dirty?


nlohmann:
I am using re2c 0.16. I should add this to the contributing guide.

nlohmann:
Thanks!

nlohmann:
We may pull the pragma, because it is Clang-only, see #393.

PR #395, opened 2016-12-13
Replace class iterator and const_iterator by using a single template class to reduce code.

Bosswestfalen:
Class iterator and const_iterator can be merged by a single template class. 
Currently the classes differ by aliases. This can be handled by the template parameter.
Const_iterator can be created using an iterator. Iterator cannot be created from a const_iterator.
This can be achieved by a conversion function in the template class.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -261,10 +261,12 @@ class basic_json
     /// the type of an element const pointer
     using const_pointer = typename std::allocator_traits<allocator_type>::const_pointer;
 
+    // forward declaration for iterators
+    template <typename U> class iter_impl;
     /// an iterator for a basic_json container
-    class iterator;
+    using iterator = iter_impl<basic_json>;
     /// a const iterator for a basic_json container
-    class const_iterator;
+    using const_iterator = iter_impl<const basic_json>;
     /// a reverse iterator for a basic_json container
     using reverse_iterator = json_reverse_iterator<typename basic_json::iterator>;
     /// a const reverse iterator for a basic_json container
@@ -8204,10 +8206,10 @@ class basic_json
 
   public:
     /*!
-    @brief a const random access iterator for the @ref basic_json class
+    @brief a template for a random access iterator for the @ref basic_json class
 
-    This class implements a const iterator for the @ref basic_json class. From
-    this class, the @ref iterator class is derived.
+    This class implements a both iterators (iterator and const_iterator) 
+    for the @ref basic_json class.
 
     @note An iterator is called *initialized* when a pointer to a JSON value
           has been set (e.g., by a constructor or a copy assignment). If the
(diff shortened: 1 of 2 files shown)


src/json.hpp
```diff
@@ -8222,33 +8224,43 @@ class basic_json
...
+    class iter_impl : public std::iterator<std::random_access_iterator_tag, U>
     {
         /// allow basic_json to access private members
         friend class basic_json;
 
+        // make sure U is basic_json or const basic_json
+        static_assert(std::is_same<U, basic_json>::value
+                      or std::is_same<U, const basic_json>::value,
+                      "iter_impl only accepts (const) basic_json");
+
```
nlohmann: Why not move this into the template specification `std::enable_if<...>`?

src/json.hpp
```diff
@@ -8222,33 +8224,43 @@ class basic_json
...
       public:
         /// the type of the values when the iterator is dereferenced
         using value_type = typename basic_json::value_type;
         /// a type to represent differences between iterators
         using difference_type = typename basic_json::difference_type;
         /// defines a pointer to the type iterated over (value_type)
-        using pointer = typename basic_json::const_pointer;
+        using pointer = typename std::conditional<std::is_const<U>::value,
```
nlohmann: Can't you use `typename U::pointer`?

src/json.hpp
```diff
@@ -8222,33 +8224,43 @@ class basic_json
...
+                                                  typename basic_json::const_pointer,
+                                                  typename basic_json::pointer>::type;
         /// defines a reference to the type iterated over (value_type)
-        using reference = typename basic_json::const_reference;
+        using reference = typename std::conditional<std::is_const<U>::value,
```
nlohmann: Same as above - why not use `typename U::reference`?

src/json.hpp
Bosswestfalen: When U is not (const) basic_json this will give a more informative error message. Unfortunately you do not notice this requirement just by looking at the head of the class. 

In fact it can be moved to the specification.

nlohmann: As the iterators are only defined inside the `basic_json` class, I think it's OK to skip the error message and move it to the specification. The code could maybe even simplified by using `std::is_same<U, std::remove_cv<basic_json>>`.

src/json.hpp
Bosswestfalen: I am not familiar with allocators. My understanding is that `std::allocator_traits<allocator_type>::pointer;` (line 260) using `AllocatorType</* no const */ basic_json>` (line 257), is never const. So I did not test it. If I#m wrong I'd prefer `typename U::pointer`.

src/json.hpp
Bosswestfalen: Same as above.

src/json.hpp
Bosswestfalen: Should be `std::is_same<std::remove_cv<U>, basic_json>`. But yes, this looks much better than my 3 lines above.

src/json.hpp
nlohmann: You were right: const_iterator should have a const pointer, see http://en.cppreference.com/w/cpp/iterator/iterator_traits.

src/json.hpp
nlohmann: Same as above.

nlohmann:
Thanks!

PR #430, opened 2017-01-18
Fix documentation error

vjon:
In the main Readme, in the "STL-like access" subsection of the "Examples", move the comparison before "j.clear()" function call, such that the intended return value "true" is actually correct.


nlohmann:
Thanks!

PR #423, opened 2017-01-09
user-defined types ter: clean, refactor basic_json class

theodelrieu:
Hi!

This PR is the follow-up of #355 

I managed to replace a lot of constructors with `from/to_json` overloads, which helped me to remove some SFINAE boilerplate, and to add a new feature: giving full control on type serialization to whom needs it.

I advise reviewers to review commit per commit, I tried my best to separate everything.

# Things that need to be done
- Put back and adapt the docs (I got a heavy hand on the 'delete' button)
- static assert message improvement
- improve README (talk about specializing types supported by the library, improve clarity)
- TODO fill this row if something comes to mind

PR #435, opened 2017-01-24
conversion from/to user-defined types

nlohmann:
This PR is #423 after a small cleanup. Nothing substantial has changed - mostly documentation was touched.


src/json.hpp.re2c
```diff
@@ -193,51 +192,62 @@ inline bool operator<(const value_t lhs, const value_t rhs) noexcept
...
-template <> struct priority_tag<0> {};
+//////////////////
+// constructors //
+//////////////////
 
 // This is an experiment. I need this to move constructors out of basic_json.
-// I'm sure there is a better way, but this might need a big basic_json refactoring
-template <value_t> struct external_constructor;
+// I'm sure there is a better way, but this might need a big basic_json
+// refactoring
```
theodelrieu: I think we can remove a part of this comment, we could only keep the part about the 'there might be a better way'

README.md
```diff
@@ -461,10 +466,14 @@ j["age"] = p.age;
...
-ns::person p {j["name"].get<std::string>(), j["address"].get<std::string>(), j["age"].get<int>()};
+ns::person p {
+    j["name"].get<std::string>(),
+    j["address"].get<std::string>(),
+    j["age"].get<int>()
+};
 ```
 
-It works, but that's quite a lot of boilerplate.. Hopefully, there's a better way:
+It works, but that's quite a lot of boilerplate... Hopefully, there's a better way:
```
theodelrieu: I think you talked about replacing 'Hopefully' with 'Fortunately'

README.md
```diff
@@ -482,152 +491,143 @@ To make this work with one of your types, you only need to provide two methods:
...
 } // namespace ns
 ```
 
-That's all. When calling the json constructor with your type, your custom `to_json` method will be automatically called.
+That's all! When calling the `json` constructor with your type, your custom `to_json` method will be automatically called.
 Likewise, when calling `get<your_type>()`, the `from_json` method will be called.
 
 Some important things:
 
 * Those methods **MUST** be in your type's namespace, or the library will not be able to locate them (in this example, they are in namespace `ns`, where `person` is defined).
```
theodelrieu: About that part, I don't know if we should add a part mentioning that it can work in the global namespace.
It might not be clear for everyone.

README.md
```diff
@@ -482,152 +491,143 @@ To make this work with one of your types, you only need to provide two methods:
...
-* When using `get<your_type>()`, `your_type` **MUST** be DefaultConstructible and CopyConstructible (There is a way to bypass those requirements described later)
+* When using `get<your_type>()`, `your_type` **MUST** be [DefaultConstructible](http://en.cppreference.com/w/cpp/concept/DefaultConstructible) and [CopyConstructible](http://en.cppreference.com/w/cpp/concept/CopyConstructible). (There is a way to bypass those requirements described later.)
```
theodelrieu: I removed the need for CopyConstructible. This way, some types can be used thanks to copy elision.
If you think this is not a really good idea, you could simply put that check back in the `static_assert`

nlohmann:
Thanks for the comments! I shall have a look later this week.

src/json.hpp.re2c
nlohmann: Ok.

README.md
nlohmann: I forgot about that, thanks!

README.md
nlohmann: I think about how to formulate this briefly.

README.md
```diff
@@ -442,6 +443,226 @@ int vi = jn.get<int>();
...
+// conversion: json -> person
+ns::person p2 = j;
+
+// that's it
+assert(p == p2);
+```
+
+#### Basic usage
+
+To make this work with one of your types, you only need to provide two methods:
```
jaredgrubb: Maybe: "... two functions" instead of "... two methods". Method sounds like a class method, whereas here it really is two free-standing functions.

nlohmann: Good point!

theodelrieu:
We could also add a sentence about enums. Right now, only `enum class` are serializable via `from_json/to_json` without specializing `adl_serializer`

doc/Doxyfile
```diff
@@ -109,7 +109,7 @@ RECURSIVE              = NO
 EXCLUDE                =
 EXCLUDE_SYMLINKS       = NO
 EXCLUDE_PATTERNS       =
-EXCLUDE_SYMBOLS        = nlohmann::anonymous_namespace
+EXCLUDE_SYMBOLS        = nlohmann::detail
```
theodelrieu: Should we bump the version number here too?

nlohmann: I shall change the version number once I created a release branch.

nlohmann:
@theodelrieu I would not know what to write about the enums - don't they work the same?

src/json.hpp.re2c
nlohmann: I removed the comment altogether.

README.md
nlohmann: Changed.

README.md
nlohmann: I added a small comment.

README.md
nlohmann: I removed mention of CopyConstructible.

theodelrieu:
Well, I thought they didn't, since I added this overload in `detail`:

```cpp
template<typename BasicJsonType, typename UnscopedEnumType,
         enable_if_t<is_unscoped_enum<UnscopedEnumType>::value, int> = 0>
void to_json(BasicJsonType& j, UnscopedEnumType e) noexcept { /* ... */ }
```

However I tested it with the following code, and it does work:

```cpp
enum test{};
void to_json(json& j, test t) {} // this method is called
```

I thought this would be an ambiguous call, but since the second one is a perfect match, it isn't!

So, nevermind what I said, no need to add it in the doc!

README.md
```diff
@@ -442,6 +443,225 @@ int vi = jn.get<int>();
...
+        // one argument
+        static move_only_type from_json(const json& j) {
+            return {j.get<int>()};
+        }
+        
+        // Here's the catch! You must provide a to_json method! Otherwise you
+        // will not be able to convert move_only_type to json, since you fully
+        // specialized adl_serializer on that type
+        static void to_json(json& j, move_only_type t) {
+            j = t.i;
```
theodelrieu: `t.i` is private here

README.md
```diff
@@ -442,6 +443,225 @@ int vi = jn.get<int>();
...
+} // namespace ns
+```
+
+That's all! When calling the `json` constructor with your type, your custom `to_json` method will be automatically called.
+Likewise, when calling `get<your_type>()`, the `from_json` method will be called.
+
+Some important things:
+
+* Those methods **MUST** be in your type's namespace (which can be the global namespace), or the library will not be able to locate them (in this example, they are in namespace `ns`, where `person` is defined).
+* When using `get<your_type>()`, `your_type` **MUST** be [DefaultConstructible](http://en.cppreference.com/w/cpp/concept/DefaultConstructible). (There is a way to bypass those requirements described later.)
```
theodelrieu: Maybe rephrase the sentence in parenthesis : (There is a way to bypass *this requirement* described later).

README.md
nlohmann: Right... Did I change this?

theodelrieu: I don't know, but just removing `private:` is fine for this example

nlohmann:
Thanks for the feedback, @jaredgrubb !

And thanks for the awesome implementation @theodelrieu !

PR #439, opened 2017-01-29
README: adjust boost::optional example

jaredgrubb:
The "from_json" example for boost::optional is not complete and should also handle the 'none' case. I'm modifying it to be symmetric with the to_json example.


nlohmann:
Thanks!

PR #441, opened 2017-01-31
Compilation issue with g++-7 (experimental)

ibroheem:
line 6086: return *lhs.m_value.array < *rhs.m_value.array;
Should be changed to return (*lhs.m_value.array) < (*rhs.m_value.array); 
because the compiler reported:
/usr/local/include/nlohmann/json-2.1.0.hpp:6086:63: error: wrong number of template arguments (1, should be 2)
                      return *lhs.m_value.array < *rhs.m_value.array; // gcc-version < 7
                                                               ^~~~~
/usr/include/c++/7/array:94:12: note: provided for ‘template<class _Tp, long unsigned int _Nm> struct std::array’
     struct array
            ^~~~~

## Files to change

   ```sh
   make re2c
   ```

   ```sh
   make
   ```

   and can be executed with

   ```sh
   ./json_unit
   ```

## Note

## Please don't

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -6083,7 +6083,8 @@ class basic_json
             {
                 case value_t::array:
                 {
-                    return *lhs.m_value.array < *rhs.m_value.array;
+                    //return *lhs.m_value.array < *rhs.m_value.array; // g++ version < 7
+                    return (*lhs.m_value.array) < (*rhs.m_value.array); /// g++ 7
                 }
                 case value_t::object:
                 {


jaredgrubb:
Are you sure this isn't a compiler bug?

nlohmann:
This looks odd. The [operator precedence](http://en.cppreference.com/w/cpp/language/operator_precedence) should yield the same behavior as if the parentheses were added.

ibroheem:
It's kinda strange. I was trying out C++17 features with g++-7 (experimental).
Honestly I was confused at first, but looking at the error I noticed it was detecting a template type argument for std::array rather that the relational operator<() on *lhs.m_value.array and *rhs.m_value.array.
@jaredgrubb might be right, it could be a compiler bug. It doesn't occur in versions before g++-7.

nlohmann:
For what it's worth: I tried gcc (GCC) 7.0.1 20170131 (experimental) and could successfully compile and execute the testsuite.

I also tried with `-std=c++17` and got a different error:

```
src/unit-regression.cpp: In function ‘void ____C_A_T_C_H____T_E_S_T____36()’:
src/unit-regression.cpp:258:22: error: ambiguous overload for ‘operator=’ (operand types are ‘std::__cxx11::string {aka std::__cxx11::basic_string<char>}’ and nlohmann::basic_json<>::value_type {aka nlohmann::basic_json<>}’)
         s2 = o["name"];
                      ^
In file included from /Users/niels/Documents/projects/gcc/include/c++/7.0.1/string:52:0,
                 from /Users/niels/Documents/projects/gcc/include/c++/7.0.1/bits/locale_classes.h:40,
                 from /Users/niels/Documents/projects/gcc/include/c++/7.0.1/bits/ios_base.h:41,
                 from /Users/niels/Documents/projects/gcc/include/c++/7.0.1/ios:42,
                 from /Users/niels/Documents/projects/gcc/include/c++/7.0.1/istream:38,
                 from /Users/niels/Documents/projects/gcc/include/c++/7.0.1/sstream:38,
                 from thirdparty/catch/catch.hpp:321,
                 from src/unit-regression.cpp:29:
/Users/niels/Documents/projects/gcc/include/c++/7.0.1/bits/basic_string.h:636:7 note: candidate: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>& std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::operator=(const std::__cxx11::basic_string<_CharT, _Traits, _Alloc>&) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]
       operator=(const basic_string& __str)
       ^~~~~~~~
/Users/niels/Documents/projects/gcc/include/c++/7.0.1/bits/basic_string.h:681:7 note: candidate: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>& std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::operator=(std::__cxx11::basic_string<_CharT, _Traits, _Alloc>&&) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]
       operator=(basic_string&& __str)
       ^~~~~~~~
/Users/niels/Documents/projects/gcc/include/c++/7.0.1/bits/basic_string.h:748:7 note: candidate: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>& std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::operator=(std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::__sv_type) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::__sv_type = std::basic_string_view<char>]
       operator=(__sv_type __sv)
       ^~~~~~~~
```

nlohmann:
@ibroheem I also think it is a compiler bug.

ibroheem:
[I mistakenly closed this issue before]

ibroheem:
My GCC version was 10 days behind.
gcc version 7.0.1 20170121 (experimental) [trunk revision 244748] (Debian 7-**20170121**-1)

nlohmann:
I think it does not make sense to adjust working code to have it compiled by a flawed (?) unreleased compiler.

nlohmann:
I think this is a bug in GCC 7's implementation of C++17 features. GCC 7 and Clang 7 (both most recent repo versions) compile the test suite with C++11 and C++14 features. We need to wait until GCC 7 stabilizes. It makes no sense to adjust the code.

ibroheem:
Sure, but it sure makes sense to prepare for the future though!.

gregmarr:
If we think that this is a bug that is going to fixed, then there's nothing to prepare for.

jaredgrubb:
I would file this against GCC so they have this as a test-case. 

Valid C++14 code will be valid C++17 code; there were no C++17 changes that would make previously-valid C++ no longer work, especially at the grammar level (modulo explicit things like removing auto_ptr or reserving new keywords). The standards body is very careful and deliberate about backward-breaking changes. This is undoubtedly a GCC bug and will not be a JSON library issue.

PR #461, opened 2017-02-19
Speedup CI builds using cotire

tusharpm:
All unit-*.cpp files in test/src include catch.hpp and json.hpp. Both are header-only libraries, and can benefit from using precompiled headers.

- Use [cotire](https://github.com/sakra/cotire) for reducing the compile time using a prefix header (test/src/prefix.hpp).
- Change travis config to use cmake instead of the hand-written Makefiles.
- Enable parallel build in Travis config (-- -j4)

_The results_ - comparing a sample without this change and with:

Metric | Without cotire | With cotire
-------|------------------|-----------
Travis (Running time) | [32 min 21 sec](https://travis-ci.org/nlohmann/json/builds/202895982) | [28 min 32 sec](https://travis-ci.org/tusharpm/json/builds/203105591)
Travis (Total time) | 2 hrs 3 min 43 sec | 1 hr 52 min 36 sec
AppVeyor (Total time) | [3 min 34 sec](https://ci.appveyor.com/project/nlohmann/json/build/1684) | [2 min 55 sec](https://ci.appveyor.com/project/tusharpm/json/build/8)

The travis build time can be further reduced by using the cmake build system for the slowest pieces (valgrind and sanitizer - these recompile the tests independently now), and the container-infrastructure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ fuzz-testing
 *.gcno
 *.gcda
 
+build
 working
 
 doc/xml
--- a/.travis.yml
+++ b/.travis.yml
@@ -228,6 +228,14 @@ install:
   # make sure CXX is correctly set
   - if [[ "${COMPILER}" != "" ]]; then export CXX=${COMPILER}; fi
 
+  # get CMake (only for systems with brew - macOS)
+  - |
+     if [[ !(-x $(which cmake)) && (-x $(which brew)) ]]; then
+       brew update
+       brew install cmake
+       cmake --version
+     fi
+
   # install LLVM/clang when LLVM_VERSION is set
   - |
      if [[ "${LLVM_VERSION}" != "" ]]; then
@@ -266,7 +274,10 @@ script:
   - $CXX --version
 
   # compile and execute unit tests
-  - make check
+  - mkdir -p build && cd build
+  - cmake .. && cmake --build . --config Release -- -j4
+  - ctest -C Release -V
+  - cd ..
 
   # check if homebrew works (only checks develop branch)
   - if [ `which brew` ]; then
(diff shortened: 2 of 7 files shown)


nlohmann:
I have some questions:
- What is the role of file test/src/prefix.hpp?
- To what extend is the speedup related to cotire and where is it just using `-j4`?

tusharpm:
- `test/src/prefix.hpp` is the prefix header (manually maintained, instead of the one auto-generated by cotire) including the headers that need to be precompiled. As per my understanding, this (or a precompiled equivalent anyway) is added to all compilation units by cotire.
- I triggered [a build without `-- -j4`](https://travis-ci.org/tusharpm/json/builds/203383648) just for comparing those results as well. Overall, it is a little slower than the original build (no-cotire).
  - The "special" builds are generally-speaking slower (because of the second compilation not using cmake).
  - The remaining (compile-heavy) builds have mixed results.
    - Some are a tad faster (LLVM - 3.8.0 and 3.8.1 saving 15-20 seconds in about 4 minutes, similar for Xcode 8.2).
    - Others are considerably slower (Xcode 6.4 is roughly 2 minutes slower from the original 6&frac12; minutes).
  - I agree this analysis is not statistically sound (very few data points), but it gives a rough idea of the time taken by each variant.
  - The advantage of (only) using cotire is visible on the AppVeyor build which doesn't have an explicit parallel jobs specification.
  - If the explicit `-- -j4` is the problem, should I try with a different generator/build-system ([ninja](https://github.com/ninja-build/ninja)), which can determine the best configuration at runtime?

nlohmann:
I added a branch from this PR, see https://github.com/nlohmann/json/tree/tusharpm-cotire. I shall merge it when Travis/AppVeyor have finished checking it. Thanks a lot!

tusharpm:
My pleasure! 👍

PR #496, opened 2017-03-11
Fix -Weffc++ warnings (GNU 6.3.1)

TedLyngmo:
Adding constructors and assignment operators where needed to keep -Weffc++ happy.
src/json.hpp regenerated with re2c 0.16 and manually edited to keep the indentation ok.

### DIFF
--- a/src/json.hpp.re2c
+++ b/src/json.hpp.re2c
@@ -6168,6 +6168,21 @@ class basic_json
               thousands_sep(!loc->thousands_sep ? '\0' : loc->thousands_sep[0]),
               decimal_point(!loc->decimal_point ? '\0' : loc->decimal_point[0])
         {}
+        serializer(const serializer& cpy)
+            : o(cpy.o),
+              thousands_sep(cpy.thousands_sep),
+              decimal_point(cpy.decimal_point),
+              loc(cpy.loc),
+              indent_string(cpy.indent_string)
+        {}
+        serializer& operator=(const serializer& rhs) {
+            o = rhs.o;
+            thousands_sep = rhs.thousands_sep;
+            decimal_point = rhs.decimal_point;
+            loc = rhs.loc;
+            indent_string = rhs.indent_string;
+            return *this;
+        }
 
         /*!
         @brief internal implementation of the serialization function
--- a/src/json.hpp.re2c
+++ b/src/json.hpp.re2c
@@ -6159,6 +6159,10 @@ class basic_json
     */
     class serializer
     {
+      private:
+        serializer(const serializer&) = delete;
+        serializer& operator=(const serializer&) = delete;
+
       public:
         /*!
         @param[in] s  output stream to serialize to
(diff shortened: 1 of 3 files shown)


src/json.hpp.re2c
```diff
@@ -6168,6 +6168,21 @@ class basic_json
               thousands_sep(!loc->thousands_sep ? '\0' : loc->thousands_sep[0]),
               decimal_point(!loc->decimal_point ? '\0' : loc->decimal_point[0])
         {}
+        serializer(const serializer& cpy)
```
nlohmann: We don't actually need to copy a serializer, so this should rather be

```cpp
        serializer(const serializer&) = delete;
        serializer& operator=(const serializer&) = delete;
```

to make GCC happy without adding unneeded features.

theodelrieu: I disagree, we should `= default` everything, since this is the actual behaviour.

Deleting them would potentially break code for no good reason.

I've read some comments about -Weff-c++ not being very relevant, (and I believe this warning is a bit too much).

But since this PR is really short, I think we can still merge it

nlohmann: The serializer class is private and only used inside the `dump` function, so no code should rely on it. Doing nothing and relying on default behavior is fine (I could live with the `-Weffc++` warnings), but making explicit that we do not intent to copy serializers feels cleaner.

theodelrieu: I agree that there is no need to copy it in the first place, since this is like a traits struct, with no attributes.

I didn't look at the diff, but if we need to add explicit constructors on stuff like `detail::is_nested_class`, just to support this warning, I really think this isn't worth it, the tests do not build with it, and I believe people who want to use it in their codebase should include `json.hpp` with `-isystem`

nlohmann: I added my proposed two lines to `json.hpp` and the `-Weffc++` warnings disappeared. I think there is no need to change other code

theodelrieu: Ok, but I still think we should `= default ` all operations if we're going down that road.

It's an empty struct after all. Also, nobody ever declares a variable of type `std::is_same<int, void> pointless;`.

However `pointless ` is still copyable, moveable etc.

This might be nit-picking, but I don't see a good reason to delete those constructors.

nlohmann: I'm not talking about changing anything but the serializer class.

theodelrieu: Oh wait a Minute, iI thought it was the adl_serializer!

Sorry I'm on my phone , didn't get that!

TedLyngmo: Hi! I made the proposed change (and made the new serializer methods private too).
Do I need to create a new pull request for that?

nlohmann: No, you can change this PR. Could you check why AppVeyor failed?

nlohmann:
@TedLyngmo, thanks! I shall merge when Travis finished!

Issue #525, opened 2017-03-20
Make exception base class visible in basic_json


krzysztofwos:
Let's make the base class of exceptions visible in`basic_json`, along `parse_error` and others. The reason why I think it is useful is that often one wants to catch all exceptions related to a library in order to guarantee that they cannot escape module boundary. It also makes the code future-proof if new exception types are added. Of course, this can be done right now by simply catching `nlohmann::detail::exception`.

@nlohmann on a personal note, I'm sorry this comes after you've merged user-defined exceptions feature branch. I was traveling for the last couple of months and I did not pay attention. Sorry!

nlohmann:
Fixed by merging #526.

PR #526, opened 2017-03-20
Make exception base class visible in basic_json

krzysztofwos:
Implements change proposed in #525 plus sanity check testing whether `re2c` is installed in order to print a nice error message if it's not.


nlohmann:
Thanks!

PR #532, opened 2017-03-23
Remove three unnecessary bitmasks

TedLyngmo:


TedLyngmo:
Question: My re2c generated json.hpp ends up looking very different from yours. I have to edit it by hand afterwards to remove trailing spaces and to get the proper indenting. I'm using [https://github.com/skvadrik/re2c]. Which re2c are you using?

nlohmann:
@TedLyngmo You can run "make pretty" which uses Astyle http://astyle.sourceforge.net/. I am using re2c 0.16.

TedLyngmo:
Ah, ok, great, that worked :-) Btw, I don't know how to interpret the pending coverage check. It seems to be finished but has no data?

nlohmann:
I also have no idea about coveralls. If you add a commit with "make pretty", it should trigger another check.

TedLyngmo:
Utterly confusing ... It's stuck again although the job is finished. I'll close this and create another branch.

PR #533, opened 2017-03-24
UTF8 bitmask enhancement

TedLyngmo:


TedLyngmo:
I don't get why it's still pending. Retrying...

PR #534, opened 2017-03-24
UTF8 encoding enhancement

TedLyngmo:


TedLyngmo:
So, [coveralls/10757553](https://coveralls.io/builds/10757553) is hanging just like my previous attempts.
The [travis build](https://travis-ci.org/nlohmann/json/jobs/214655066) reported successfully to [coveralls/24178102](https://coveralls.io/jobs/24178102) which looks fine.
Is there perhaps some mismatch in the reporting of the coverage data or have I messed up somehow?

nlohmann:
Thanks!

PR #535, opened 2017-03-25
GCC Requirement >= 4.8.5

perara:
Tested on Centos 7 (4.8.5 is default gcc version)

### DIFF
--- a/README.md
+++ b/README.md
@@ -703,7 +703,7 @@ json j_from_msgpack = json::from_msgpack(v_msgpack);
 
 Though it's 2016 already, the support for C++11 is still a bit sparse. Currently, the following compilers are known to work:
 
-- GCC 4.9 - 6.0 (and possibly later)
+- GCC 4.8.5 - 6.0 (and possibly later)
 - Clang 3.4 - 3.9 (and possibly later)
 - Microsoft Visual C++ 2015 / Build Tools 14.0.25123.0 (and possibly later)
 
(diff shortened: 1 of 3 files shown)


nlohmann:
Could you build and run the test suite? With g++-4.8.5 on macOS, there is an error compiling raw strings.

perara:
So far it seems like i'm not able to compile the tests:
```
json.hpp:5882:36: error: no match for ‘operator=’ (operand types are ‘std::vector<nlohmann::basic_json<>, std::allocator<nlohmann::basic_json<> > >::iterator {aka __gnu_cxx::__normal_iterator<nlohmann::basic_json<>*, std::vector<nlohmann::basic_json<>, std::allocator<nlohmann::basic_json<> > > >}’ and ‘void’)
         result.m_it.array_iterator = m_value.array->insert(
```

But in my opinion, an effort should be put into making this compatible with relatively new compilers?
Or is this a difficulty?

nlohmann:
The error you are quoting is is a [known bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55817): GCC 4.8.x does not fully support C++11. The issue I mentioned is also a [known bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57824). Both issues are not fixed until GCC 4.9.0, so I cannot claim that the library works with GCC 4.8.5.

perara:
This is true, however. Should there be attempts in giving the library support for GCC 4.8.x as most LTS distros use relatively old GCC versions?

nlohmann:
Sorry, I cannot support compilers that do not support C++11.

PR #536, opened 2017-03-25
Add Visual Studio 17 image to appveyor build matrix

vpetrigo:
I would like to propose a small addition to the current `appveyor.yml` file that adds Visual Studio 17 image to the build matrix. So, now it is possible to track a build status with brand new MS's Studio.


nlohmann:
Great! I did not know there was such an image at AppVeyor already!

PR #540, opened 2017-03-30
Properly pop diagnostics

tinloaf:
Currently, with clang, the diagnostics are push()ed twice, but pop()ed only once. Fixing this by adding a "pragma diagnostics pop".


nlohmann:
Thanks!

PR #576, opened 2017-05-08
Adding support for Meson.

Type1J:
Although the use of this library is very convenient by just copying over the header to another project, this makes it easy for those who want to checkout a particular commit or the HEAD at the build time of their project when they use the [Meson Build System](http://mesonbuild.com/).

It adds only 1 file (`meson.build`), and some usage info in the `README.md` file.

Meson is a C++ build system, like CMake, but simpler. It aims to cut down on the time spent working on build files, and the time spent compiling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,29 +66,8 @@ to the files you want to use JSON objects. That's it. Do not forget to set the n
 
 :beer: If you are using OS X and [Homebrew](http://brew.sh), just type `brew tap nlohmann/json` and `brew install nlohmann_json` and you're set. If you want the bleeding edge rather than the latest release, use `brew install nlohmann_json --HEAD`.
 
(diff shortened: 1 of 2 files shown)


Type1J:
I'm not sure what happened with CI. I didn't change the .hpp files, just the `README.md` file, and I added `meson.build`, which shouldn't impact the build that's currently in place.

nlohmann:
No worries - Travis is sometimes a bit buggy. I restarted the failed job. It should be fine.

Type1J:
I think Travis is having a bad day.

Type1J:
The commits that I made per your review only touched the README.md file, but Travis CI is still not working correctly, it seems.

nlohmann:
As I said, Travis is sometimes a bit buggy.

nlohmann:
Thanks!

PR #578, opened 2017-05-10
Adding first and second properties to iteration_proxy_internal

### LINKED ISSUE #350: Request: range-based-for over a json-object to expose .first/.second
I would love to have the ability to iterate over a JSON object type in exactly the same way I would iterate over a std::map:

`for(auto& i : map) {
    i.first; 
    i.second;
}`

Would it be possible to have a free function (or member) that would allow me to do something like:
`for(auto& i : as_map(jsonObj)) {...}`
such that the iterator exposed here works more like map::iterator (.first,.second) than the current iterator type (.key,.value).

Type1J:
This PR fixes #350 by adding property-like members to the iterator_proxy_internal class.

The fix proposed in that issue looks as if you pay for it even if you don't use .first or .second (at least the .first shows a copy of a string each time the iterator is incremented). That cost shouldn't have to be payed, but we don't have to sacrifice the interface that we want.

This solution provides the .first and .second interface as outlined in the issue, but shouldn't cost anything, and is likely inlined out of existence after compilation.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -7877,6 +7877,44 @@ class basic_json
     class iteration_proxy
     {
       private:
+        /// helper class for first "property"
+        template<typename ProxyType>
+        class iterator_key_property
+        {
+          private:
+            /// the reference to the proxy
+            ProxyType& proxy;
+
+          public:
+            explicit iterator_key_property(ProxyType& proxyRef) noexcept
+                : proxy(proxyRef) {}
+
+            /// conversion operator (calls key())
+            operator typename basic_json::string_t() const
+            {
+              return proxy.key();
+            }
+        };
+
+        /// helper class for second "property"
+        template<typename ProxyType>
+        class iterator_value_property
+        {
+          private:
+            /// the reference to the proxy
+            ProxyType& proxy;
+
+          public:
+            explicit iterator_value_property(ProxyType& proxyRef) noexcept
+                : proxy(proxyRef) {}
+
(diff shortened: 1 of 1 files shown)


Type1J:
Here's the test that I used:
```
#include <json.hpp>
#include <iostream>

using namespace std;
using namespace nlohmann;

int main()
{
  json j = {{"A", 1}, {"B", 2}};

  for (auto i : json::iterator_wrapper(j))
  {
    auto x = i.first; // x = "A", x = "B"
    auto y = i.second; // y = 1, y = 2

    cout << x << ", " << y << endl;
  }
}
```
Here's the output:
```
"A", 1
"B", 2
```

nlohmann:
This is a nice solution! Thanks!

nlohmann:
Unfortunately, the code does not work for me. If you look into the test suite in `tests/src/unit-iterator_wrapper.cpp`, replacing calls to `key()` with`first` and `value()` with `second` yields compilation errors.

It would be great if you could have a look at this.

Type1J:
I'm using MSVC 2017. What compiler failed for you?

nlohmann:
clang version 5.0.0 (trunk 292523) (llvm/trunk 292718) - in any case, it would be great to have unit tests - I should not have merged the code without.

Type1J:
I'll look into it, and I'll see about adding the unit tests as well.

nlohmann:
That would be great!

PR #579, opened 2017-05-10
Fixing assignement for iterator wrapper second, and adding unit test

Type1J:
To complete #350 and #578, this PR adds an assignment operator and a new unit test for the iterator wrapper functionality with .first and .second.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -7894,6 +7894,20 @@ class basic_json
             {
               return proxy.key();
             }
+
+            /// equal operator (calls key())
+            template<typename KeyType>
+            bool operator==(const KeyType& key)
+            {
+              return proxy.key() == key;
+            }
+
+            /// not equal operator (calls key())
+            template<typename KeyType>
+            bool operator!=(const KeyType& key)
+            {
+              return proxy.key() != key;
+            }
         };
 
         /// helper class for second "property"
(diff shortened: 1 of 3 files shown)


Type1J:
```
Running the test cases, I get:
$ json_unit iterator_wrapper
All tests passed (111 assertions in 1 test case)


$ json_unit iterator_wrapper_first_second
All tests passed (111 assertions in 1 test case)
```

The new `iterator_wrapper_first_second` test is a duplicate of the regular `iterator_wrapper` test with all `.key()` calls replaced with `.first` and all `.value()` calls replaced with `.second`

nlohmann:
As Travis is sometimes buggy, I restarted the failing tests (see https://travis-ci.org/nlohmann/json/builds/230863073), but I fear that there is a problem with Clang, but also newer GCC versions.

nlohmann:
Travis came to the same results: all compilers but GCC 4.9 fail, see see https://travis-ci.org/nlohmann/json/builds/230863073.

Type1J:
I see. I'll look into it. This it probably a good case for `operator.()`, but it's not supported by most compilers. I think I'll need to add `operator==()`, `operator!=()`, and `operator<()` for the key type. The value type may be fine as it is.

gregmarr:
`operator.()` is not standard.  It has been proposed but hasn't yet been approved.

Type1J:
I'm getting Clang to see if I can fix it. I may have to generate a reverse patch to remove the code (apparently, it's not supported by all compilers).

Type1J:
Clang seems to be happy after making the compares const.

nlohmann:
The unit tests succeed now. Great! I'll merge this tonight.

PR #583, opened 2017-05-13
fix doxygen error of basic_json::get()

zhaohuaxishi:
`@ref` command do not  seems to work in markdown code span. so i change
```
`void from_json(const  @ref basic_json&, ValueType&)`
```
into 
```
`void from_json(const ` @ref basic_json`&, ValueType&)`
```
are there better solution for this problem?

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -3447,9 +3447,9 @@ class basic_json
     This overloads is chosen if:
     - @a ValueType is not @ref basic_json,
     - @ref json_serializer<ValueType> has a `from_json()` method of the form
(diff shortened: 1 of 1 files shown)


nlohmann:
The mentioned functions are not meant to be linked in the first place, because they are most likely in the client's code - it can be any `from_json` function based on `ValueType`. I'm not sure whether fixing this so that `basic_json` can be clicked brings too much value here.

zhaohuaxishi:
is that means the `@ref` command should just be removed then.

Issue #550, opened 2017-04-07
Warning on Intel compiler (icc 17)


mark-99:
The warning below is generated when compiling with icc. I realise this is likely the compiler's fault, but perhaps there is a workaround which would make it work.

*edit* Actually Intel seem to think it's a valid warning: https://software.intel.com/en-us/forums/intel-c-compiler/topic/596393

---------------
nlohmann-json/json/json.hpp(9024): error #597: "nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer>::iter_impl<U>::operator nlohmann::basic_json<std::map, std::vector, std::string, bool, int64_t={long}, uint64_t={unsigned long}, double, std::allocator, nlohmann::adl_serializer>::iter_impl<const nlohmann::basic_json<std::map, std::vector, std::string, bool,
          int64_t={long}, uint64_t={unsigned long}, double, std::allocator, nlohmann::adl_serializer>>() const [with ObjectType=std::map, ArrayType=std::vector, StringType=std::string, BooleanType=bool, NumberIntegerType=int64_t={long}, NumberUnsignedType=uint64_t={unsigned long}, NumberFloatType=double, AllocatorType=std::allocator, JSONSerializer=nlohmann::adl_serializer, U=const nlohmann::basic_json<std::map, std::vector, std::string, bool, int64_t={long}, uint64_t={unsigned long}, double,
          std::allocator, nlohmann::adl_serializer>]" will not be called for implicit or explicit conversions
          operator const_iterator() const
          ^
          detected during instantiation of class "nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer>::iter_impl<U> [with ObjectType=std::map, ArrayType=std::vector, StringType=std::string, BooleanType=bool, NumberIntegerType=int64_t={long}, NumberUnsignedType=uint64_t={unsigned long}, NumberFloatType=double, AllocatorType=std::allocator, JSONSerializer=nlohmann::adl_serializer, U=const
                    nlohmann::basic_json<std::map, std::vector, std::string, bool, int64_t={long}, uint64_t={unsigned long}, double, std::allocator, nlohmann::adl_serializer>]"

nlohmann:
[...]

PR #598, opened 2017-05-30
#550 Fix iterator related compiling issues for Intel icc

HenryRLee:
Currently there are two know issues when compiling with Intel icc:
1. The `operator()` in iterator causes warning
2. The use of `internal_iterator` causes error

The first issue can be resolved by replacing the `operator()` with a converting constructor and assignment. Because the existence of the copy constructor may bring ambiguity, I removed it and the copy constructor will be defined implicitly.

The second issue is fixed by adding a `struct` keyword in front of the type name.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -8054,11 +8054,12 @@ class basic_json
         }
 
         /*!
-        @note The conventional copy constructor is not defined. It is replaced
-		      by either of the following two converting constructors.
-              They support copy from iterator to iterator,
-                           copy from const iterator to const iterator, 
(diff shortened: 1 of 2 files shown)


nlohmann:
@HenryRLee Could you please merge the current `develop` branch so the AppVeyor tests have a chance to succeed?

nlohmann:
@HenryRLee Could you please merge the current `develop` branch so the AppVeyor tests have a chance to succeed?

HenryRLee:
Sure. It's updated now. Testing on the way.

nlohmann:
Thanks!

Issue #614, opened 2017-06-10
to_json for pairs, tuples


hsdk123:
Hi, I was wondering if there might be a simple way to to_json() pairs and tuples, instead of having to explicitly create each element for json.

nlohmann:
I'm not sure what you mean. Could you provide more details or an example please?

hsdk123:
Thanks for the reply! For example, let's say i want to encode an std::pair<int, int> into json.
Ideally:
```
struct Person
{
    std::pair<int, int> position
};
```
```
	void to_json( json& j, const Person& d )
	{
		j = json{
			{ "position", d.position }
		};
	}
	void from_json( const json& j, lvn::NTextbox::NChar::Data& d )
	{
		d.position= j.at( "position" ).get<std::pair<int, int>>();
	}
```
But sadly this gives an error.
Instead, I have to encode each element of the pair (or tuple) explicitly as a different json element, which can get slightly troublesome.

Hopefully I've just misunderstood something and there is a direct method, otherwise it would be absolutely great to be able to do this.

gregmarr:
Have you tried something like this?  (I'm not sure if it will work, just speculating)
```
void to_json( json& j, const Person& d )
	{
		to_json(j["position"], d.position);
	}
	void from_json( const json& j, lvn::NTextbox::NChar::Data& d )
	{
		from_json(j["position"], d.position);
	}
```

nlohmann:
There is no default conversion between `json` and `std::pair`. You need to provide translations from/to arrays. This would work:

```cpp
#include "json.hpp"

using json = nlohmann::json;

struct Person
{
    std::pair<int, int> position;
};

void to_json( json& j, const Person& d )
{
    // encode pair as a list
    j = json{ { "position", {d.position.first, d.position.second} } };
}

void from_json( const json& j, Person& d )
{
    auto& pos = j.at("position");
    // decode list as pair
    d.position = std::make_pair(pos[0], pos[1]);
}

int main()
{
    // create a person
    Person p;
    p.position = {17, 42};
    
    // person -> json
    json j = p;
    std::cout << std::setw(2) << j << std::endl;
    
    // json -> person
    Person p2 = j;
    std::cout << std::boolalpha << (p.position == p2.position) << std::endl;
}
```

theodelrieu:
I added support for `std::pair` in a recent PR, however it requires the `.first` to be a compatible string type , so it can represent the combination of a string field, and a JSON object.

[...]

PR #615, opened 2017-06-13
remove std::pair support

theodelrieu:
As stated in #614, I believe the `std::pair` support I introduced in #609 is bogus. How a pair should be represented is dependent of context, so I think it is wiser to remove it, and leave the question open.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -866,14 +866,6 @@ void to_json(BasicJsonType& j, T (&arr)[N])
     external_constructor<value_t::array>::construct(j, arr);
 }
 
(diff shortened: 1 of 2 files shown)


theodelrieu:
@nlohmann Do you know which line got uncovered? I failed to see which ones could be related to my changes

gregmarr:
@theodelrieu My guess is that because you removed covered lines, the percentage `covered/total` decreased.  Say for example it was 7/10, and you removed two lines, it is now 5/8.

nlohmann:
Thanks! We shall continue discussing #614. I also reopened #600. I am sure we will find a good solution for all this.

Issue #566, opened 2017-04-20
Conan.io


Pancir:
Hello guys I think it would be a good idea to make the library to be available through the conan. https://www.conan.io/
I could do it without your help but it is the better way when the developers is responsible for it themselves.

nlohmann:
I think it's already there: https://www.conan.io/source/nlJson/2.0.2/arnemertz/stable

That version is outdated. Would it make sense to add another one?

Pancir:
Anyone can add to the conan's repository any library,  but official supporting is always better. For example it could be uploaded from the CI at once and the conan's recipe could be adjusted properly by the developers and it could have correct naming and user in the canan's repository.
Anyway it is just a suggestion to adjust delivering to the conan's repository.

nlohmann:
@Pancir What exactly needs to be done for this? Is it sufficient to create a conanfile.py like this: https://github.com/arnemertz/conan-nlJson/blob/master/conanfile.py ?

Pancir:
Yes and usually you have to create the test_package for checking the recipe.  
BTW The conanfile.py with the test_package can be in separated repository if you wish, you can see it with the link you have specified, it is in another repository. 
The test_package is just a simple consumer. 
http://conanio.readthedocs.io/en/latest/reference/commands/test_package.html
http://conanio.readthedocs.io/en/latest/packaging/getting_started.html#the-test-package-folder

nlohmann:
I just learned [conan.io is deprecated](http://docs.conan.io/en/latest/move_to_bintray.html).

Pancir:
Yes, they just moved their public repository to another place. I don't learn the issue about this moving yet but I guess it should not affect the conanfile.py and the test_package. 
The conan is still young and it is under active developing.

nlohmann:
I am currently not planning to invest time into this. A PR would be appreciated.

Pancir:
Sorry, I don't know what "PR" means.

nlohmann:
A pull request. If anyone would create the required files (such as https://github.com/nlohmann/json/blob/develop/meson.build for Meson, see https://github.com/nlohmann/json/pull/576), I would be happy to add them to my repository.

Pancir:
Thx for the explanation.

nlohmann:
@Pancir As I described in #619, I cannot maintain Conan.io myself. If someone (like @arnemertz before) can maintain a repository with the required files, I would be happy.

nlohmann:
[...]

PR #619, opened 2017-06-14
Make the library available through Conan package

vthiery:
This PR addresses the issue #566 by adding the conanfile.py, build.py and test_package necessary to package and test the conan package.

The version is set to 2.1.1 and will have to be bumped when releasing.

In order to test that the package works properly, simply do `conan test_package` in the conan folder.
In order to upload the package, follow the [documentation](http://docs.conan.io/en/latest/move_to_bintray.html).

### DIFF
--- a/conan/.gitignore
+++ b/conan/.gitignore
@@ -1,0 +1,4 @@
+test_package/build
+*.pyc
+conanbuildinfo.cmake
+conaninfo.txt
--- a/conan/build.py
+++ b/conan/build.py
@@ -1,0 +1,7 @@
+from conan.packager import ConanMultiPackager
+
+
+if __name__ == "__main__":
+    builder = ConanMultiPackager(username="nlohmann", channel="stable")
+    builder.add_common_builds(pure_c=False)
+    builder.run()
--- a/conan/conanfile.py
+++ b/conan/conanfile.py
@@ -1,0 +1,21 @@
+from conans import ConanFile
+from conans.tools import download
+
+class JsonForModernCppConan(ConanFile):
+    name = "jsonformoderncpp"
+    version = "2.1.1"
+    license = "MIT"
+    url = "https://github.com/nlohmann/json"
+    author = "Niels Lohmann (mail@nlohmann.me)"
+    settings = None
+    options = {"path": "ANY"}
+    default_options = "path="
+
+    def source(self):
+        download("https://github.com/nlohmann/json/releases/download/v%s/json.hpp" % self.version, "json.hpp")
+
+    def package(self):
+        header_dir = "include"
+        if self.options.path != "":
+            header_dir += "/" + str(self.options.path)
+        self.copy("*.hpp", dst=header_dir)
--- a/conan/test_package/.gitignore
+++ b/conan/test_package/.gitignore
@@ -1,0 +1,2 @@
+build/
+conanfile.pyc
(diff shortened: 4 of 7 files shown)


nlohmann:
In the documentation, we have the following information:

> 🍺 If you are using OS X and Homebrew, just type brew tap nlohmann/json and brew install nlohmann_json and you're set. If you want the bleeding edge rather than the latest release, use brew install nlohmann_json --HEAD.

> If you are using the Meson Build System, then you can wrap this repo as a subproject.

How would such a line look like for Conan?

vthiery:
I guess adding this:

> If you are using [Conan](https://www.conan.io/) to manage your dependencies, merely add `jsonformoderncpp/x.y.z@nlohmann/stable` to your conanfile.py's requires, where `x.y.z` is the release version you want to use.

in the README would indeed help users a lot. We could also add this comment if they are completely new to Conan:

> Then, use the command `conan install /path/to/conanfile.py` to install the requirements and add the following lines to your CMakeLists.txt:

```cmake
include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
conan_basic_setup()
```

although a the documentation on  [conan.io](https://www.conan.io/) already explains it all. Maybe a link to the documentation would be better?

nlohmann:
The shorter the better. Frankly, I think a package manager is if little use for a single-header library...

vthiery:
Indeed, it's overkill for a single header-library, I agree. Nevertheless, if one package all his dependencies with Conan, having a package at hand, even for single-header libraries, allows to have a unified way to reference TPLs.

nlohmann:
One question: Can I really use any version with `jsonformoderncpp/x.y.z@nlohmann/stable`?

vthiery:
Assuming all the versions have been packaged up, yes. Now, do we want to package all the versions, or only a subset of them, is another question. I would go for packaging the latest as a first step, and package new versions going forward. Does it make sense?

nlohmann:
What does "packaging" mean in this sense?

vthiery:
I actually refer to this document: http://docs.conan.io/en/latest/packaging.html

In practical, it means running the packaging procedure for the selected versions and uploading the packages on Bintray's servers. I can do it myself, manually, for the latest version, or a couple of them if that is required. If one wants to be a bit more CI aligned, it would mean adding a 'conan packaging' step in the pipeline, and package upon releases/hotfixes.

nlohmann:
Hm. So this means every time I make a release at Github, I also need to do something to have a Conan release?

vthiery:
If one does not automatize the packaging upload in a CI pipeline, then yes, it means packaging manually every release one wants to have in a Conan package.

nlohmann:
Then I would like not to have this in my repository as I do not plan to do this.

vthiery:
It makes some sense and as Pancir said in #566 the Conan files are usually maintained in another repository. Thanks for your feedback!

PS: I really enjoy working with your library, great work!

nlohmann:
Thanks for your work in this PR and sorry for the inconvenience caused. I need to focus on the development and cannot provide additional service for the different package managers. It would be great if someone (@Pancir ?) could maintain such a repository, though.

nlohmann:
Closing this PR and continue in #566.

vthiery:
Sure! I actually packaged and maintain a couple of Conan packages already. I would be happy to create one for this library and maintain it.

Issue #553, opened 2017-04-13
Cannot convert from json array to std::array


wgtdkp:
Conversion from `std::array` to json array works well, but not in reverse.
```c++
std::string r = R"([1, 2, 3, 4])";
std::array<int, 4> arr = json::parse(r);
```
clang++ gives compiling errors:
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/bits/stl_iterator.h:661:20: error: 
      no member named 'insert' in 'std::array<int, 4>'
        iter = container->insert(iter, std::move(__value));
               ~~~~~~~~~  ^
/usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.0/../../../../include/c++/5.4.0/bits/stl_algo.h:4175:12: note: 
      in instantiation of member function 'std::insert_iterator<std::array<int, 4> >::operator='
      requested here
        *__result = __unary_op(*__first);
                  ^
../../src/json.hpp:720:10: note: in instantiation of function template specialization
      'std::transform<nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char>, bool,
      long, unsigned long, double, std::allocator, adl_serializer>::iter_impl<const
      nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char>, bool, long, unsigned
      long, double, std::allocator, adl_serializer> >, std::insert_iterator<std::array<int, 4> >, (lambda
      at ../../src/json.hpp:721:50)>' requested here
    std::transform(j.begin(), j.end(),
         ^
../../src/json.hpp:804:5: note: in instantiation of function template specialization
      'nlohmann::detail::from_json_array_impl<nlohmann::basic_json<std::map, std::vector,
      std::__cxx11::basic_string<char>, bool, long, unsigned long, double, std::allocator,
      adl_serializer>, std::array<int, 4> >' requested here
    from_json_array_impl(j, arr, priority_tag<1> {});
    ^
../../src/json.hpp:902:16: note: in instantiation of function template specialization
      'nlohmann::detail::from_json<nlohmann::basic_json<std::map, std::vector,
      std::__cxx11::basic_string<char>, bool, long, unsigned long, double, std::allocator,
      adl_serializer>, std::array<int, 4>, 0>' requested here
        return from_json(j, val);
               ^
../../src/json.hpp:917:16: note: in instantiation of function template specialization
[...]

PR #624, opened 2017-06-17
add pair/tuple conversions

theodelrieu:
I left a question in the code: Should we impose that the JSON array size must be equal to the number of tuple/pair elements?

If not, some elements in the JSON could be discarded:

```cpp
json j = std::make_tuple(1, 2, 3, 4);
auto p = j.get<std::pair<int, int>>();
```

Some users might want to discard values on purpose, so I would say we should not check the size at all.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -1063,6 +1063,15 @@ auto from_json_array_impl(const BasicJsonType& j, CompatibleArrayType& arr, prio
     });
 }
 
+template <typename BasicJsonType, typename T, std::size_t N>
+void from_json_array_impl(const BasicJsonType& j, std::array<T, N>& arr, priority_tag<2>)
+{
+    for (std::size_t i = 0; i < N; ++i)
+    {
+        arr[i] = j.at(i).template get<T>();
+    }
+}
+
 template<typename BasicJsonType, typename CompatibleArrayType,
          enable_if_t<is_compatible_array_type<BasicJsonType, CompatibleArrayType>::value and
                      std::is_convertible<BasicJsonType, typename CompatibleArrayType::value_type>::value and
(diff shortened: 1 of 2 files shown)


nlohmann:
What would be the behavior if I try to convert `[1]` to a `std::pair<int,int>`?

theodelrieu:
It would throw, I use `at` in the implementation

nlohmann:
Right, and it makes sense to complain about this (I guess it's the same behavior when you try to read four values from a three-value tuple?).

I think discarding values is OK, too, as long as we mention this in the documentation somewhere.

theodelrieu:
Yes you get a compile-time error if you go out of bounds with `tuple`.

nlohmann:
In #553, we excluded a conversion from `json` to `std::array`. If we have compile-time checks for tuples, maybe adding support for arrays in the same fashion is also possible?

src/json.hpp
```diff
@@ -1094,6 +1145,25 @@ void from_json(const BasicJsonType& j, ArithmeticType& val)
     }
 }
 
+template <typename BasicJsonType, typename... Args>
+void from_json(const BasicJsonType& j, std::pair<Args...>& p)
+{
+    // should we check that j.size() exactly matches the number of arguments for tuple and pairs?
```
nlohmann: No, it's OK not to check this. The comment can be removed.

test/src/unit-constructor1.cpp
```diff
@@ -239,6 +239,31 @@ TEST_CASE("constructors")
...
+
+            CHECK(j.type() == json::value_t::array);
+            REQUIRE(j.size() == 4);
+            CHECK(j[0] == std::get<0>(t));
+            CHECK(j[1] == std::get<1>(t));
+            CHECK(j[2] == std::get<2>(t));
+            CHECK(j[3][0] == 0);
+            CHECK(j[3][1] == 1);
+        }
+
```
nlohmann: Could you also add a test case for the case when we have an array with, say, 3 elements and ask for a `std::pair` and something similar with tuples?

theodelrieu: Here it is, sorry for the delay.

theodelrieu:
I will take a look at `std::array` right now.

nlohmann:
@theodelrieu Yes, a nice exception would definitely help. I guess most people would never expect `at` to be called in the first place.

And yes, the pre/post conditions should be documented. I think the README paragraph is the best place to start.

nlohmann:
Thanks a lot, @theodelrieu !

Issue #474, opened 2017-02-26
Moving forward to version 3.0.0


nlohmann:
I kept postponing the step toward version 3.0.0 for quite some time and quite some [issues](https://github.com/nlohmann/json/milestone/5) have been piled up. These issues have in common that fixing them will break the public API - be it very obvious such as removing deprecated functions or changing the type of exceptions or rather subtle such as begin able to store a NaN value.

Question is: How to proceed? So far, I worked in the `develop` branch. As soon as I, for instance, remove deprecated functions and push to `develop`, this code may break existing applications. I can live with that - we still have `master` as a branch containing the code of the last release, yesterday's 2.1.1. Furthermore, OSS-Fuzz for instance, checks the `develop` branch, so it would be a bit dangerous to only work in a feature branch.

I wonder whether it would be bad style to continue working in `develop`. What do you think?

ibroheem:
"I wonder whether it would be bad style to continue working in develop"
Thats why its called develop/experimental etc...

How about doing it the GCC way:

4.9.x
5.x.x
6.x.x

Everything in parallel:
2.x.x
3.x.x

Just my own opinion.

theodelrieu:
That's a good idea, we should separate branches, since we will surely backports bug fixes from 3.0.0 to previous versions.

GCC still releases 4.9.x

nlohmann:
(We have to keep in mind that the GCC project has far more people, and I could live without 2.x.x releases while preparing for 3.0.0.)

ibroheem:
2.x.x: No new feature additions, Only fixes. Define an LTS, few months or years.
3.x.x: New features, break the breakable if possible.

gregmarr:
If you need a definition for what it means to have 3.0.0 vs 2.x.x, I suggest you go with semver.org's definition:

> Summary
> 
> Given a version number MAJOR.MINOR.PATCH, increment the:
> 
> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards-compatible manner, and
> PATCH version when you make backwards-compatible bug fixes.
> Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format. 

As for where you do it, I wouldn't expect anyone to be basing production code on a `develop` branch, so changing that to 3.0.0-alpha and keeping master as 2.x.x until 3.0.0 is ready for master should be fine.  If any additional fixes are needed in 2.x.x until that time, a temporary branch could be created for that.

[...]

PR #643, opened 2017-07-01
Refacto/split basic json

theodelrieu:
Hi, I managed to move every nested struct/class present in `basic_json` to the `detail` namespace.

Excepted `json_pointer` which is now in `nlohmann`, I don't know if input/output adapters are exposed, if so, I will move them there too.

This is an intermediate refactoring, needed before splitting the library into multiple files, and before changing classes and namespace names.

I'm afraid the review will not be very interesting... I advise reviewers to look at each commit separately, it is not readable otherwise.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -1265,6 +1265,270 @@ struct static_const
 
 template<typename T>
 constexpr T static_const<T>::value;
+
+////////////////////
+// input adapters //
+////////////////////
+
+/// abstract input adapter interface
+struct input_adapter
+{
+    virtual int get_character() = 0;
+    virtual std::string read(size_t offset, size_t length) = 0;
+    virtual ~input_adapter() {}
+};
+
+/// a type to simplify interfaces
+using input_adapter_t = std::shared_ptr<input_adapter>;
+
+/// input adapter for cached stream input
+template<std::size_t N>
+class cached_input_stream_adapter : public input_adapter
+{
+  public:
+    cached_input_stream_adapter(std::istream& i)
+        : is(i), start_position(is.tellg())
+    {
+        fill_buffer();
+
+        // skip byte order mark
+        if (fill_size >= 3 and buffer[0] == '\xEF' and buffer[1] == '\xBB' and buffer[2] == '\xBF')
+        {
+            buffer_pos += 3;
+            processed_chars += 3;
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


theodelrieu:
It's weird that codacy only complains now, I did not change this code.

theodelrieu:
I forgot to use the public API in every class i've split (except `parser` that became a friend, since it needs to modify `m_type/m_value` directly.

The tests passed because of the `#define private public` line present in almost each test file.

I think we should get rid of this line, it's quite dangerous when refactoring

nlohmann:
I added this to allow for 100% coverage without changing the code. I'm not a fan of code refactoring just to simplify testing. I need to have a look at this PR when I'm home.

theodelrieu:
Please note that the purpose of this PR is not to facilitate testing, but to prepare the split of `json.hpp`.

The main reason why this line exists is that we test **private nested classes'** public APIs. Once we split the code into multiple files, we'll be able to unit test each component public API directly, without decreasing today's coverage rate.

nlohmann:
Sorry, i misunderstood. I shall have a look on Saturday.

nlohmann:
Sorry, i misunderstood. I shall have a look on Saturday.

theodelrieu:
It's weird, the `unit-cbor.cpp` file compiles on my laptop. I had to compile it by hand though, it seems it's not included when running `ninja`

theodelrieu:
Finally I made that pass \o/

@nlohmann If you could take a look before pushing other commits, it'd be great (rebasing is horrible with all the code moving)

Also I won't be available for the next 2 weeks, so I can only make changes tonight (sorry for the rush :p)

nlohmann:
Sorry for changing to much lately. I kind of forgot what it means to pending PRs...

theodelrieu:
Thanks a lot! We shall talk about the next steps after I come back of vacations (I'll be available on slack during that time anyway)

The first thing should be quite trivial: splitting into multiple files, and add a cmake/makefile rule to amalgamate a single `json.hpp` file. Looking how Catch does that should be helpful.

The second is more hard, it's about the API breakage (namespace naming, single traits class, etc...) as discussed in #474.

I'll be happy to get on that issue when I come back!

nlohmann:
Have a nice vacation!

Issue #456, opened 2017-02-15
basic_json templated on a "policy" class


jaredgrubb:
I'm enjoying the library, however I have a couple comments:
 - the mangled name is really awful:
`
nlohmann::basic_json<std::__1::map, std::__1::vector, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, bool, long long, unsigned long long, double, std::__1::allocator, nlohmann::adl_serializer>
`
 - it's not possible to extend basic_json beyond the types

The mangled name is cosmetic, but it is annoying to have to try to read stackshots or prints in debuggers.

One idea I had was to refactor basic_json to be based on a policy type, aka `nlohmann::basic_json<nlohmann::default_policy>`, where `default_policy` would be defined something like:

```cpp
template <
    template<typename U, typename V, typename... Args> class ObjectType = std::map,
    template<typename U, typename... Args> class ArrayType = std::vector,
    class StringType = std::string,
    class BooleanType = bool,
    class NumberIntegerType = std::int64_t,
    class NumberUnsignedType = std::uint64_t,
    class NumberFloatType = double,
    template<typename U> class AllocatorType = std::allocator,
    template<typename T, typename SFINAE = void> class JSONSerializer = adl_serializer
    >
struct policy_base {
    using object_t = ...;
    using array_t = ...;
    // ...

    using json_value = ... /* union definition */;
    using parser = ...;

    // ... whatever accessors are needed to complete the abstraction
};

struct default_policy : policy_base<> {};
```

It would take a bit of work to get basic_json refactored in terms of the policy, and to figure out exactly what should go in each part. But I think this could make the design a bit cleaner in that it turns basic_json into the interface-glue that makes the type easy to use, and the policy object becomes the low-level details that provide a nice customization point.

The kinds of customizations that this could enable:
* allow the union to be extended to support user types (for example, `std::shared_ptr<json>`, or `std::unique_ptr<LazyJsonLoader>` that only deserializes a sub-object if it gets used, or to allow `EmployeeRecord` to be placed directly in the union and serialized inline, but also provide conversion-operators like other types)
* maybe the parser could be extended to support non-conforming extensions (like trailing comma); I support that this library doesnt support this out of the box, but it would be nice if it was doable without too much work. (I have not looked at the parser, so maybe this is not even remotely possible right now).

Anyway, I just wanted to throw the idea out there to get feedback.

gregmarr:
[...]

PR #650, opened 2017-07-09
Don't include <iostream>, use std::make_shared

olegendo:
I've been evaluating this library for usage in an embedded software that runs on an MCU with on-chip flash.  Memory on these devices is a premium.  Just including the \<iostream> file, makes the ROM image grow by about 200 KByte, and I've got only about 768 KByte in total.  This is a known problem of \<iostream>.  One workaround is not to include it in (library) header files, but only in source files where it is actually needed.  Header files should include \<iosfwd> instead.

While skimming over the code, I've also spotted some opportunities to use std::make_shared.

Tested with 'make' and 'make test' on GCC 5.4.0 x86_64-linux-gnu.
Also briefly checked that the library can be used without including \<iostream> on GCC 6.4.1 rx-elf.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -6634,17 +6634,17 @@ class basic_json
 
         static std::shared_ptr<output_adapter<CharType>> create(std::vector<CharType>& vec)
         {
-            return std::shared_ptr<output_adapter>(new output_vector_adapter<CharType>(vec));
+            return std::make_shared<output_vector_adapter<CharType>>(vec);
         }
 
         static std::shared_ptr<output_adapter<CharType>> create(std::ostream& s)
         {
-            return std::shared_ptr<output_adapter>(new output_stream_adapter<CharType>(s));
+            return std::make_shared<output_stream_adapter<CharType>>(s);
         }
 
         static std::shared_ptr<output_adapter<CharType>> create(std::string& s)
         {
-            return std::shared_ptr<output_adapter>(new output_string_adapter<CharType>(s));
+            return std::make_shared<output_string_adapter<CharType>>(s);
         }
     };
 
@@ -8767,19 +8767,19 @@ class basic_json
         /// input adapter for input stream
         static std::shared_ptr<input_adapter> create(std::istream& i)
         {
-            return std::shared_ptr<input_adapter>(new cached_input_stream_adapter<16384>(i));
+            return std::make_shared<cached_input_stream_adapter<16384>> (i);
         }
 
         /// input adapter for input stream
         static std::shared_ptr<input_adapter> create(std::istream&& i)
         {
-            return std::shared_ptr<input_adapter>(new cached_input_stream_adapter<16384>(i));
+            return std::make_shared<cached_input_stream_adapter<16384>>(i);
         }
 
         /// input adapter for buffer
         static std::shared_ptr<input_adapter> create(const char* b, size_t l)
         {
-            return std::shared_ptr<input_adapter>(new input_buffer_adapter(b, l));
+            return std::make_shared<input_buffer_adapter>(b, l);
         }
 
         // derived support
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -6675,7 +6675,7 @@ class basic_json
         std::vector<CharType>& v;
     };
 
-    /// putput adapter for output streams
+    /// output adapter for output streams
     template<typename CharType>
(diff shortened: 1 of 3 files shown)


nlohmann:
Thanks a lot!

Can you comment on the memory usage of the library? Is it usable in your usecase?

olegendo:
Ugh, sorry,  I didn't notice the examples.

I can't comment on the memory footprint of the library yet.  The \<iostream> was a no-go for my case.  But now I can try using it for some data exchange between the embedded http server and the browser.

I'm wondering why you haven't used a base class and virtual functions for json_value instead of the open coded type-switch-cases?  At the first glance, it looks like it could chop off quite a bit of code, but I might be missing something here.  What's the reasoning behind the choice of data layout for class basic_json?

Notice that the statement in the "Memory efficiency" section of README.md is not complete.  sizeof (basic_json) is greater than sizeof (void*) + sizeof (uint8_t), because of struct alignment rules.

nlohmann:
Good question. I think such a base class would be handy when it comes to using different types, because instead of assuming string types to have the `basic_string` interface, we could make this explicit (this may be relevant to https://github.com/nlohmann/json/issues/474#issuecomment-312450848). I'm not sure about the memory requirements though.

In any case, the library just "evolved" into this state. Using a union was perfectly reasonable when I started.

jaredgrubb:
Great patch!

olegendo:
What I meant is something like that:
`
struct value_base
{
  virtual ~value_base (void) { }
  virtual bool empty (void) const noexcept = 0;
  virtual size_type size (void) const noexcept  = 0;
  virtual void clear (void) noexcept = 0;
  ...
};

class string_value : public value_base
{
  public:
    virtual bool empty (void) const noexcept override { return m_value.empty (); }
    virtual size_type size (void) const noexcept  override { return 1; }
    virtual clear (void) noexcept override { m_value.clear (); }
    ...
  private:
    std::string m_value;
};

class boolean_value : public value_base
{
public:
    virtual bool empty (void) const noexcept override { return false; }
    virtual size_type size (void) const noexcept  override { return 1; }
    virtual clear (void) noexcept override { m_value = false; }
    ...
private:
   bool m_value;
};
...
`
But I just noticed that you store bool and number types by value in the union.  This brings up the idea of storing all the various value types by-value in basic_json (via std::aligned_storage).  On a 32 bit target and libstdc++, sizeof (json) = 12 bytes, sizeof (std:.vector) = 12 bytes, sizeof (std::string) 24 bytes, sizeof (std::map) = 24 bytes.  So sizeof (json) would become 28 bytes, which means json numbers would take double the space than before.  On the other hand, it would eliminate one indirection when accessing json objects, arrays and strings.  Also, for each json object/array/string the heap management overhead per allocation should be taken into account (~8 bytes), which would be eliminated.

Is it worth going down that rabbit hole?  Hmm ... If you have access to some benchmarks, I could give it a try, just out of curiosity.

nlohmann:
This sounds interesting! Related: #456 #474

nlohmann:
I thought about this idea. It would mean a complete refactoring of the library. For instance, we would need to move the serialization implementation to the individual classes, and also a lot of the internal conversions which currently depend on the mere types (e.g., any kind of string representation can be somehow converted to `string_t` - we would now require these constructors in the `string_value` class).

It is a very interesting approach, and I shall open a feature branch. But I am not sure how long it would take to be able to really use it.

(Or maybe I am just missing something and things are much easier ;-))

nlohmann:
I opened a feature branch: c1f4b249732400d64383824168d33ea953af5b1e. There are a lot of open questions (at least to me), but it's late and I'm not in a rush.

pboettch:
@olegendo Completely off-topic to this thread, is your project, where you are trying to use this library, open-source?

olegendo:
@pboettch Sorry, no.  Unfortunately it's not open source.

olegendo:
@nlohmann I have also created an experimental branch here https://github.com/olegendo/json/tree/virtual_funcs

Unfortunately, I can't get it to fully compile yet.  Somehow the to_json machinery is broken ... but I hope that it shows my intention.
I haven't done a lot of replacements that could be done with the json_value_* classes in place and left the type-switches.  Even if this might not make it into mainline as-is, I think it could be a good idea to wrap the direct member accesses (m_type, m_value.number_float and so on) in functions.  It will make future refactoring work easier.

PR #682, opened 2017-08-08
To fix compilation issue for intel OSX compiler

kbthomp1:
o To prevent the compilation issue on OSX with the intel compiler suite.  The
error was found with icpc version 15.0.3.187 where the "__clang_version__" was
not defined correctly, while "__clang__" was.

## Files to change

   ```sh
   make check
   ```

## Note

## Please don't

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -7526,10 +7526,10 @@ class basic_json
         result["platform"] = "unknown";
 #endif
 
-#if defined(__clang__)
-        result["compiler"] = {{"family", "clang"}, {"version", __clang_version__}};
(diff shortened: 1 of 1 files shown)


nlohmann:
Thanks a lot - I haven't tried ICC so far - does the test suite compile without problems? Are there warnings?

kbthomp1:
Actually, the test suite doesn't compile with OS X icc.  I'm getting both errors and warnings from icpc when running:

```
CC=icc CXX=icpc make check
```

[make_check.txt](https://github.com/nlohmann/json/files/1211250/make_check.txt)

Issue #695, opened 2017-08-11
Generated nlohmann_jsonConfig.cmake does not set JSON_INCLUDE_DIR


matthewOConnell:
Looks like CMake refactoring in f434942 clipped setting `JSON_INCLUDE_DIR`.  This breaks `find_package(nlohmann_json)`  and finding the correct header location.

dan-42:
Hi @matthewOConnell can you explain more by what you mean does not set JSON_INCLUDE_DIR?
if you really  need this information you can access it by 
```cmake
get_target_property(JSON_INCLUDE_DIRS nlohmann_json INTERFACE_INCLUDE_DIRECTORIES)
```
all needed information is exposed to the target ```nlohmann_json``` 
there is no need manually calling ``` include_directories(${JSON_INCLUDE_DIRS})``` or
``` target_include_directories(main ${JSON_INCLUDE_DIRS}) ```

by calling ``` target_link_libraries(main nlohmann_json) ``` all needed include paths are set

```cmake
## note the PATHS is only needed because for testing I did not install it in my system!
find_package(nlohmann_json CONFIG REQUIRED PATHS ../install/lib/cmake) 

add_executable(main main.cpp)
target_link_libraries(main nlohmann_json)
```

But yes, the refactored CMake  does set the include so that the user needs to call ``` #include <nlohmann/json.hpp>  ``` with respect to be consistent with the namespace ``` nlohmann::json ```

I hope that helps.

And If it is needed, we can add a CMake-option to use either 
``` #include <nlohmann/json.hpp>  ``` 
or
``` #include <json.hpp>  ``` 


cheers Daniel

matthewOConnell:
Sounds like we're just getting down to preference, and probably my lack of understanding on "modern" CMake patterns.  This is the first I've run into a `find_package()` configuration that didn't set include and lib variables.   And I admit I'm not sure how  `target_link_libraries(main nlohmann_json)` is superior to  `target_include_directories(main ${JSON_INCLUDE_DIRS})` especially given how typical the `target_include_directories(main ${JSON_INCLUDE_DIRS})` pattern is.  (Especially for a header only library it seems a misnomer to say "link libraries").  Maybe what you've shown is more stylistic to newer cmake?

[...]

PR #696, opened 2017-08-11
To add missing cmake variable for include path

matthewOConnell:
This is to address issue #695 where recent CMake refactoring work stopped setting `JSON_INCLUDE_DIR` breaking CMake `find_package(nlohmann_json)`


matthewOConnell:
This alone does not resolve the `find_package()` issue.  I'm closing this merge request while I continue working on it.

Issue #698, opened 2017-08-13
One build system to rule them all


theodelrieu:
Right now, we have mainly two ways of building the library (i.e. tests):

* Makefile
* CMake

There is also a `meson.build` file, which only provides support for meson users, so they don't have to rely on a wrap.

There are some issues with the `CMakeLists.txt` (as #695 shows. I also personally have to clean my build folder each time to regenerate the precompiled header on Linux)

There are also targets that are only available in the Makefile (e.g. `make pretty`)

I think we should adopt a single build system, I will soon need to add a target to amalgamate the library after we split the monolithic `json.hpp`, and I don't know if I should add it only in the Makefile, the CMakeLists.txt etc...

Rewriting everything would not take very long, since this is not a huge project.

What do people think?

My personal preference would go towards Meson, but we could also simply rewrite the CMake to be more in phase with what people call "modern CMake".

nlohmann:
An explanation of the status quo:

- I added the Makefiles to ease my workflows for the development and release process. They do not necessarily belong in this repository, but they make things easier if people open PRs.
- The Cmake files were originally added to use MSVC on AppVeyor. Meanwhile, I also use them for the Linux and macos builds on Travis. They were not intended to replace the Makefiles, because I only added targets to compile and run the test suite.
- I personally do not care about package managers. For those who use them, I started to add files and a small section to the README file. But frankly I think this is overhead for a single-header library.

I personally would not really favor merging the pure maintainer scripts and the Cmake files. If the Makefile in the root is confusing, moving it into a subfolder could help.

nlohmann:
I think it would be a good idea to move all Makefile targets that are used in Travis (i.e., running cppcheck, sanitiziers, Valgrind, etc.) into the CMake file.

theodelrieu:
That would be great!

nlohmann:
FYI: Started with https://github.com/nlohmann/json/tree/feature/issue698.

Strange issue: Clang sanitizers finds issues that were not found before: https://travis-ci.org/nlohmann/json/jobs/268514575 I tried to move the check from the Makefile to the Cmakelists, but it seems the CXXFLAGS differ. Or there really is an issue in the `std::valarray` code that was not detected before and that I cannot find otherwise.

theodelrieu:
Awesome!

This could be related to `valarray` implementation, according to [cppreference](http://en.cppreference.com/w/cpp/numeric/valarray) some implementations use expression templates, in which case using `auto` can be troublesome.

[...]

Issue #462, opened 2017-02-19
Support startIndex for from_cbor/from_msgpack


prsyahmi:
The idea is to not create another copy of std::vector since my cbor data start at certain offset.

```cpp
    static basic_json from_cbor(const std::vector<uint8_t>& v, const size_t startIndex = 0)
    {
        size_t i = startIndex;
        return from_cbor_internal(v, i);
    }
```

nlohmann:
Done.

prsyahmi:
Thanks!

nlohmann:
With 22b59693f110b730d41c8f616f82b1fef72c161a, the offset/startIndex parameter  was removed as this behavior can be mimicked with an iterator range. For instance, instead of calling

```cpp
json j = json::from_cbor(vec, 5);
```

you can write

```cpp
json j = json::from_cbor({vec.begin()+5, vec.end()});
```

theodelrieu:
Are the braces `{}` intentional?

nlohmann:
Yes, I did not find a better way to implicitly create an input adapter from two parameters. And since there is a parameter with a default value following, I cannot use variadic parameters.

(Though I would love to find a solution!)

theodelrieu:
I didn't go into the details, but what about adding a new overload of  `from_{cbor,msgpack}` that take an iterator range, just like parse?

You could then construct the `input_adapter` as you did in your example code.

nlohmann:
That would be one solution, but then I also have an input adapter to parse from a `uint8_t*` and a length, and then I would need another family of overloads...

nlohmann:
(Of course, the example above could be written as

```cpp
json j = json::from_cbor(json::input_adapter(vec.begin()+5, vec.end()));
```

theodelrieu:
Actually, this is also supported by `parse`:

```cpp
auto j = json::parse(vec.begin(), vec.end());
auto j2 = json::parse(vec.data(), vec.data() + vec.size());
```

Both cases are covered with the same overload. Plus, that would be symmetrical with other parts of the API (e.g. `parse` as mentioned).

If those kind of methods could share the same API, that would become really intuitive to use IMO.

nlohmann:
I know that this is desirable, but I don't really like copy/pasting so many functions...

nlohmann:
On the other hand, there are at most two parameters, so I can add two functions for each parser and forward the arguments to the input_adapter constructors.

nlohmann:
[...]

PR #700, opened 2017-08-14
Refactor/split it

theodelrieu:
Here it is.

refactortime

This PR is the follow-up of #643, and the second phase of my planned refactoring (there will likely be one more, but we shall discuss it in a separate thread, since it requires naming/structural changes).

I've added a [tool](https://github.com/VinnieFalco/Amalgamate) to amalgamate the library into a single header (like Catch does), @nlohmann pointed out this was necessary.

There is a new rule in the Makefile: `single_include`, that builds, and executes the tool.

The output is located in the `single_include` folder, at the root of the repo.

The only requirement is that **every** included file that is part of the library must be with the following style:

```cpp
#include <will_not_be_inlined.hpp>
#include "will_be_inlined.hpp"
```

Each include must be done with a path relative to the source folder:

```cpp
// detail/conversions/from_json.hpp

#include "detail/value_t.hpp" // good
#include "../value_t.hpp" // NOPE, not refactor-proof
```

I've also added one test for each source file (except `json.hpp`), to check that we didn't miss any include.
This will be helpful for future refactoring.


theodelrieu:
Well, looks like codacy wakes up a bit late!

nlohmann:
Wow, this is a big PR... I am not sure yet whether I will be able to find my way around in the code. Let's see.

theodelrieu:
Don't worry, this is just code moving from `json.hpp` to different files ;)

nlohmann:
I know - but I got used to that one large file!

theodelrieu:
Ahah time to let it go!

nlohmann:
Let's see. Maybe I see the benefits soon.

theodelrieu:
Well one very big advantage is that people can work on different areas of the code much more easily, before, rebasing was just impossible (I tried, and it hurt).

Also, after phase 3, it will be possible for users to only include what they want (e.g. the file where `json::parse` resides)

It's not possible yet, I didn't touch at `basic_json`, but once (almost) all its methods will be moved in a namespace, it will be.

theodelrieu:
Do you have any idea on travis failure? I don't get why the `clang_sanitize` fails.

By the way, do you know if I can cancel Travis build? They take soooo long, and if I repush twice in a row, I have to wait 2 full hours to know if my last commit passed...

theodelrieu:
Apparently, `-fsanitize=address` is not recognized by clang...

nlohmann:
I overworked the Travis configuration. With 21726d8d8ffb2a58c0a88f03820e53e7f22f7f07, the sanitizier should work again.

theodelrieu:
Great! I will trigger Travis once I arrive to work. As for Codacy, I'd rather fix its warnings after this PR, as they were introduced before.

I'll also rename commits to reflect the actual paths (omitted `detail`).

theodelrieu:
Hmm, coveralls seems to be hanging, is there some configuration files to modify too?

theodelrieu:
I have no idea why VS2017 fails, I ran the same commands than AppVeyor on my laptop, and all tests pass...

nlohmann:
I have the same problems with AppVeyor in the develop branch. :-S

nlohmann:
This seems related: http://help.appveyor.com/discussions/problems/7450-build-error-on-vs-2017-image-suddenly

I give it a try.

theodelrieu:
Looks promising, thanks!

nlohmann:
AppVeyor works now.

theodelrieu:
Thanks, I don't know about Coveralls though, but I just moved code around, so I don't believe it would be relevant if it was up.

As I mentioned earlier about Codacy, I'd like to fix its warnings after the PR, since they are related to previous commits.

theodelrieu:
Fixing codacy warnings would require to put back all the `parse` overloads, and forwarding arguments to `detail::input_adapter`, a bit similar to what we discussed in #462.

I believe it would be the best, since having a public method taking a parameter from the `detail` namespace seems a bit off, and marking constructors of `input_adapter` et al. `explicit` can not hurt.

src/detail/serializer.hpp
```diff
@@ -1,17 +1,18 @@
...
-#include <cassert> // assert
-#include <clocale> // localeconv, lconv
-#include <cmath> // labs, isfinite, isnan, signbit
-#include <cstddef> // size_t, ptrdiff_t
-#include <cstdint> // uint8_t
-#include <cstdio> // snprintf
-#include <iterator> // next
-#include <limits> // numeric_limits
-#include <string> // string
+#include <algorithm>   // reverse, remove, fill, find, none_of
```
Huemac: What on earth is the purpose of this sudden indent? This kind of stuff is just really annoying to maintain.

theodelrieu: It seems that `astyle` doesn't reindent those comments, I shall update that manually after rebasing on `develop`.

theodelrieu:
Some macOS builds seem flaky, related to homebrew apparently.

About coveralls, the [job](https://travis-ci.org/nlohmann/json/jobs/267946687) passes fine, but no data are shown on coveralls.io. Do you have an idea about that?

theodelrieu:
@nlohmann since you are working on #698, I will add Amalgamate to the `CMakeLists.txt` as an external project once you're done, so we don't add 60K lines of code inside the repo for nothing.

nlohmann:
I don't know what you mean.

theodelrieu:
I added a tool in the folder `third_party` named "Amalgamate", in order to make the single header `json.hpp`.

But there is no need to pollute the repo with this tool' 60K LOC, and I intend to use CMake's [ExternalProject](https://cmake.org/cmake/help/v3.0/module/ExternalProject.html) command instead.

nlohmann:
Coveralls is not yet called from the CMakeLists, but from the standard Makefile, see https://github.com/nlohmann/json/blob/feature/issue698/.travis.yml#L77.

theodelrieu:
I followed the link to the blog post, and it seems quite easy to migrate coveralls to be used with CMake.

nlohmann:
I had problems deriving a coverage report for multiple binaries - therefore I still have a target for a single binary `json_unit`. Maybe I give it another try.

theodelrieu:
With some luck it might work out of the box with CMake. I'm crossing fingers.

nlohmann:
I don't get Coveralls to run with several binaries and CMake right now. Even worse, AppVeyor's MSVC 2017 seems to be broken again :-(

theodelrieu:
Well I had issues with CMake too, I couldn't make `ExternalProject` work as I wanted, so I instead added a git submodule for the `Amalgamate` tool. (submodules are a bit clunky, but since it will only be used when releasing, I think it's a good trade-off vs having 60K third party sources).

Furthermore, I tried to fix the Codacy warnings, but they're really useless, all but one are about internal classes' constructors that are not marked `explicit` (which is what we want though) and one that I believe is bugged (about unsigned comparison).

I would vote for removing the `explicit` warnings, the library relies on implicit conversions.

theodelrieu:
I finally understood why coveralls was stuck, I simply forgot to change the tool invocation in `.travis.yml` ...

Is there anything to change code-wise now? It's been a month since I opened this PR, and it's a bit tedious to maintain (I have to solve conflicts each time there is a new commit on `develop`). Also, I'd like to start the next phase, which requires some discussion, do you have some time to meet on Slack this month? 

When AppVeyor was working, the build succeded on VS 2017, there are some spurious fails on Travis, but related to Homebrew.

Codacy's warnings are (IMO) irrelevant as I described earlier.

nlohmann:
Slack sounds great. What about next Tuesday?

theodelrieu:
Sounds good to me, we can decide on the hour on slack?

theodelrieu:
Okay, I managed to fix coveralls locally, let's try if it works. @nlohmann I've read that the `repo_token` should be private, I don't know if you mind having it publicly exposed though.

I'm not sure if the lines I removed in the `.travis.yml` were important, it seemed to me that Coveralls handled everything itself when I tested it.

theodelrieu:
Looks like they were useful after all :P

theodelrieu:
Finally coveralls is fixed \o/

theodelrieu:
Should be good, I removed the `self-contained` tests. Refactoring the tests will accomplish the same thing anyway.

I fixed the `Amalgamate` compilation on macOS, at least it worked on my box.

trilogy-service-ro:

agauniyal:
is this some kind of bot or what? I remember I've seen it on other PRs(exact same message) as well..

nlohmann:
@agauniyal I already reported @trilogy-service-ro at Github, but nothing happened so far.

theodelrieu:
I've managed to add a CMake target `single_header`, thus removed the Makefile one and the submodule.

theodelrieu:
@nlohmann I've modified the CMakeLists.txt to run the amalgamation on each Ninja invocation, the output is then located in `single_include/json.hpp`.

This file should be committed every time, that way we keep a single `json.hpp` file up-to-date, which one can download directly if they want to keep a single file integration.

I do not see what's left to do on this PR. If you could disable codacy that'd be great (or at least disable the troublesome warnings). IIRC, it uses cppcheck which is quite old, replacing that with `clang-tidy` or some other tool would be better IMO.

nlohmann:
:bookmark: I would move this to the post-3.0.0 milestone.

theodelrieu:
Hi, I just rebased on develop (again T_T). I've added the following `pre-commit` git hook to always add the single header version of `json.hpp`:

```bash
#!/bin/sh

echo "Running pre-commit hook"
git add single_header/json.hpp
```

@nlohmann You might want to add it to the contribution guidelines. Also, I would advise you to fork the [Amalgamate](https://github.com/theodelrieu/Amalgamate) repo, and to replace the URL in the `CMakeLists.txt` with yours.

nlohmann:
Thanks a lot!!!

PR #710, opened 2017-08-25
JSON parser instantiation with customized types.

imalyavskiy:
There are changes that make possible to instantiate and use the nlohmann::json with custom map, vector and string types. In my case it was caused by need to use custom allocator. But currently classes under the hood use std::string instead of StringType(the 3rd template parameter of basic_json class). Thus it makes impossible to instantiate the parser with customized types.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -1598,6 +1598,8 @@ class lexer
     using number_integer_t = typename BasicJsonType::number_integer_t;
     using number_unsigned_t = typename BasicJsonType::number_unsigned_t;
     using number_float_t = typename BasicJsonType::number_float_t;
+	using string_t = typename BasicJsonType::string_t;
+	using value_t = typename string_t::_Alloc::value_type;
 
   public:
     /// token types for the parser
@@ -1623,7 +1625,7 @@ class lexer
     };
 
     /// return name of values of type token_type (only used for errors)
-    static const char* token_type_name(const token_type t) noexcept
+    static const value_t* token_type_name(const token_type t) noexcept
     {
         switch (t)
         {
@@ -2217,17 +2219,17 @@ class lexer
         }
     }
 
-    static void strtof(float& f, const char* str, char** endptr) noexcept
+    static void strtof(float& f, const value_t* str, value_t** endptr) noexcept
     {
         f = std::strtof(str, endptr);
     }
 
-    static void strtof(double& f, const char* str, char** endptr) noexcept
+    static void strtof(double& f, const value_t* str, value_t** endptr) noexcept
     {
         f = std::strtod(str, endptr);
     }
 
-    static void strtof(long double& f, const char* str, char** endptr) noexcept
+    static void strtof(long double& f, const value_t* str, value_t** endptr) noexcept
     {
         f = std::strtold(str, endptr);
     }
@@ -2609,7 +2611,7 @@ scan_number_done:
     @param[in] length        the length of the passed literal text
     @param[in] return_type   the token type to return on success
     */
-    token_type scan_literal(const char* literal_text, const std::size_t length,
+    token_type scan_literal(const value_t* literal_text, const std::size_t length,
                             token_type return_type)
     {
         assert(current == literal_text[0]);
@@ -2680,11 +2682,11 @@ scan_number_done:
     }
 
     /// return string value
-    const std::string get_string()
+    const string_t get_string()
     {
         // yytext cannot be returned as char*, because it may contain a null
         // byte (parsed as "\u0000")
-        return std::string(yytext.data(), yylen);
+        return string_t(yytext.data(), yylen);
     }
 
     /////////////////////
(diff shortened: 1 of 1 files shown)


nlohmann:
Could you add a test case for this?

imalyavskiy:
Yes I will but a bit later. Thank you for immediate response. I even did not finish the typing/editing.

src/json.hpp
```diff
@@ -2217,17 +2219,17 @@ class lexer
         }
     }
 
-    static void strtof(float& f, const char* str, char** endptr) noexcept
+    static void strtof(float& f, const value_t* str, value_t** endptr) noexcept
```
gregmarr: Since `std::strtof` takes `const char *`, this will cause a failure if `value_t` is not `char`.
The same for `std::strtod` and `std::strtold`.

src/json.hpp
```diff
@@ -2818,22 +2820,21 @@ class lexer
...
-
     /// buffer for variable-length tokens (numbers, strings)
-    std::vector<char> yytext = std::vector<char>(1024, '\0');
+    std::vector<value_t> yytext = std::vector<value_t>(1024, '\0');
     /// current index in yytext
     std::size_t yylen = 0;
 
     /// a description of occurred lexer errors
-    const char* error_message = "";
+    const value_t* error_message = "";
```
gregmarr: This can cause failures if the current string literals that are assigned to `error_message` are not compatible with `const value_t *`.

src/json.hpp
```diff
@@ -2818,22 +2820,21 @@ class lexer
     std::size_t chars_read = 0;
     /// the start position of the current token
     std::size_t start_pos = 0;
-
     /// buffer for variable-length tokens (numbers, strings)
-    std::vector<char> yytext = std::vector<char>(1024, '\0');
+    std::vector<value_t> yytext = std::vector<value_t>(1024, '\0');
```
gregmarr: Much of the rest of the code expects `yytext` to be `std::vector<char>`, such as line 2658 where a `char` is added to it, or line 2568 where `yytext.data()` is passed to `std::strtoull`.

src/json.hpp
```diff
@@ -1623,7 +1625,7 @@ class lexer
     };
 
     /// return name of values of type token_type (only used for errors)
-    static const char* token_type_name(const token_type t) noexcept
+    static const value_t* token_type_name(const token_type t) noexcept
```
gregmarr: This will fail if `const value_t *` is not compatible with the simple string literals returned by this function.

imalyavskiy: Hi germar. I'm not insisting for these chanes to be pulled in. I'm not so skilled in templates but I will. Anyways are you agree that once the string type is declared as a class that is possible to change then all the library contents must follow this rule? If not then at least 3rd arg of basic_json makes no sence. And if so 1st, 2rd and the allocator make no sence too. So we have a bug! As for me it was enough to solve my problem. If you are able to fix this - fix it, please.

imalyavskiy: And as for more why do you suppose and limit the users that text have to be only as char compatible types?
Do not understand me wrong I want your library to the best.

gregmarr: FYI, I'm just a user of the library.  As I understand it, the internal parsing is done using plain `char`, while the storage of the data in the `json` object uses the defined string type.  @nlohmann can provide more details.  There may be a smaller set of changes needed to do what you need, not sure.  My concern is simply that while it may work for your case, there may be other users that use a different string type that uses something like `wchar_t` instead, which would cause this to break.  I can't remember if someone was able to do that or not, just pointing out potential issues.

nlohmann: Yes, internally, we use `char` in the lexer class. This is independent of the string class used in the library. I am not sure what this PR tries to achieve. Could we maybe have an issue before where a bug or a desired extension is discussed?

imalyavskiy: If this is the best way to get desired chenges - yes, sure! What should I do? I'm newbie here. If thi requires a sample I can made it but tomorrow(simply drunk ;) - having beer )

nlohmann: In any case, I have yet to understand what changes you desire.

imalyavskiy:
Nicolas, what is unclear in the state that: if to put to your parser some string type different from std::string(exactly I used a std::basic_string with custom allocator) then you parser will not work. And this is because in the template args you're proposing to set a type for string, but really contining to use the std::string which comes into conflict). And suddenly I had no time to provide you with the sample. And seems you're agree that you solution is rough and buggy once you're closed the issue. Cheers.

nlohmann:
@imalyavskiy Did you mean me?

imalyavskiy:
@nlohmann Yes, I did.

nlohmann:
I am aware of problems with allocators. I close the PR, because all tests were failing and it seemed to be stale. If you have an idea how to fix it, please let me know. But I think opening an issue would be more helpful rather than just opening a PR.

imalyavskiy:
Ok Your're right.

PR #749, opened 2017-09-19
Show last breath error with JSON_NOEXCEPTION

di9it:
## Files to change

   ```sh
   make check
   ```

## Note

## Please don't

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -95,7 +95,7 @@ SOFTWARE.
     #define JSON_TRY try
     #define JSON_CATCH(exception) catch(exception)
 #else
-    #define JSON_THROW(exception) std::abort()
+    #define JSON_THROW(exception) std::cerr<<exception.what()<<std::endl; std::abort()
     #define JSON_TRY if(true)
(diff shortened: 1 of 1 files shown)


src/json.hpp
```diff
@@ -95,7 +95,7 @@ SOFTWARE.
     #define JSON_TRY try
     #define JSON_CATCH(exception) catch(exception)
 #else
-    #define JSON_THROW(exception) std::abort()
+    #define JSON_THROW(exception) std::cerr<<exception.what()<<std::endl; std::abort()
```
nlohmann: This does not work for several reasons:

1. This line is executed if exceptions are explicitly switched off, but also if the compiler does not support exceptions. In this case, `exception.what()` will yield a compilation error.
2. `std::cerr` cannot be used, as `iostream` is not included in the library. We removed `iostream` some while ago in favor of `iosfwd` to improve compilation speed.

nlohmann:
Closed due to inactivity. I shall reopen it if requested.

PR #753, opened 2017-09-26
Add info for the vcpkg package.

gregmarr:
This adds the information on the package for this library in the vcpkg package manager, and fixes a typo in the description of the hunter package name.


nlohmann:
Thanks!!

ras0219-msft:
@nlohmann vcpkg doesn't add all installed libraries to your include/link/etc paths by default, so you'll need to use standard cmake commands to find and add the header:

```cmake
find_path(NLOHMANN_JSON_DIR NAMES nlohmann/json.hpp)
include_directories(${NLOHMANN_JSON_DIR})
```

nlohmann:
Thanks a lot! That helped!

Issue #714, opened 2017-08-27
Parse throw std::ios_base::failure exception when failbit set to true


dPavelDev:
In brief:

```cpp
std::ifstream is;
is.exceptions(
      is.exceptions() 
    | std::ios_base::failbit 
    | std::ios_base::badbit
); // handle different exceptions as 'file not found', 'permission denied'

try { 
    is.open("my_valid_file.json");
    const auto &jsonFile = nlohmann::json::parse(is); // (*)
    // ...
catch (const std::ios_base::failure &e) {
    std::cerr << e.code().message() << std::endl;
}
```

The line (*) produce std::ios_base::failure exception when parse function reaches end of stream even if json document is valid. Without std::ios_base::failbit function works OK.

nlohmann:
The problem is that inside the lexer, `std::ifstream::read` is called which may throw in the described setting, because it tries to fill a buffer which may be larger than the input.

Example code:

```cpp
#include <iostream>
#include <fstream>
#include <array>

int main() {
    std::ifstream is;
    is.exceptions(
                  is.exceptions()
                  | std::ios_base::failbit
                  | std::ios_base::badbit
                  ); // handle different exceptions as 'file not found', 'permission denied'
    
    try {
        is.open("file.json");
        
        std::array<char, 10000> buffer;
        is.read(buffer.data(), static_cast<std::streamsize>(buffer.size()));
    }
    catch (const std::ios_base::failure &e) {
        std::cerr << "error: " << e.code().message() << std::endl;
    }
}
```

Any ideas how to implement this without a try/catch?

dPavelDev:
I don't know, is a fast and right this method or not:

```cpp
is.get(buffer.data(), static_cast<std::streamsize>(buffer.size()), '\0');
```

but it works with non-empty input streams.

nlohmann:
The null character does not help here, because the same code is also used for binary formats like CBOR or MsgPack.

Changing this line to

```cpp
is.get(buffer.data(), static_cast<std::streamsize>(buffer.size()), std::char_traits<char>::eof());
```

works, but still has the issue that `get` has the [behavior](http://en.cppreference.com/w/cpp/io/basic_istream/get) that

>  If no characters were extracted, calls `setstate(failbit)`.

This happens if we check that the input actually ended with EOF, so we call `get` once more and get no characters.

dPavelDev:
Good news! If `is.get()` reaches end of stream, `is.eof()` marks as `true`.

Also,
[...]

Issue #367, opened 2016-11-23
Behavior of operator>> should more closely resemble that of built-in overloads.


TurpentineDistillery:
Basically:
* eat up leading blanks or newlines
* leave the stream where the parsing ended, allowing for more data downstream, which may or may not be json. 

```c++
#include "json.hpp"
    
template<typename T>
static void test(const std::string& s)
{
    T val{999};
    std::stringstream sstr(s);
    //sstr.exceptions(std::ifstream::failbit | std::ifstream::badbit);

    try {
        size_t i = 0;
        while(sstr >> val) {
            std::cerr << "    result[" << i++ << "]=" << val 
                      << "; streampos=" << sstr.tellg() << std::endl;
        }   

        std::cerr << "    final values: result=" << val 
                  << " i=" << i
                  << " streampos=" << sstr.tellg()
                  << " streamflags=" << sstr.flags() 
                  << std::endl;
    } catch(std::exception& e) {
        std::cerr << "    std::exception:" << e.what() << std::endl;
    }   

    std::cerr << std::endl;
}

int main()
{
    using json = nlohmann::json;
    size_t i = 0;

    for(std::string s : { 
        "", 
        "   ",
        "111",
        "222 \t\n",
        "\n\t 333", 
        " 111 \n222\n \n  333" })
    {   
        std::cerr << "\nTest-case[" << i++ << "]: '" << s << "'" << std::endl;
    
        std::cerr << "With int:" << std::endl;
        test<int>(s);

        std::cerr << "With json:" << std::endl;
        test<json>(s);

        std::cerr << "----" << std::endl;
    }   

    return 0;
}
```

```
Test-case[0]: ''
With int:
    final values: result=999 i=0 streampos=-1 streamflags=4098

With json:
    std::exception:parse error - unexpected end of input

----

Test-case[1]: '   '
With int:
    final values: result=999 i=0 streampos=-1 streamflags=4098

With json:
    std::exception:parse error - unexpected end of input

----

Test-case[2]: '111'
With int:
    result[0]=111; streampos=-1
    final values: result=111 i=1 streampos=-1 streamflags=4098

With json:
    result[0]=111; streampos=-1
    std::exception:basic_string::append

----

Test-case[3]: '222 	
'
With int:
    result[0]=222; streampos=3
    final values: result=222 i=1 streampos=-1 streamflags=4098

With json:
    final values: result=222 i=0 streampos=-1 streamflags=4098

----

Test-case[4]: '
	 333'
With int:
[...]

PR #764, opened 2017-10-02
Simplified istream handing #367

pjkundert:
Simplified cached_input_stream_adapter<> to input_stream_adapter, to avoid redundant buffering (already buffered in underlying std::streambuf), and thus to not pre-load (and discard) a large buffer of input from the istream when parsing a smaller JSON record.  Handles a leading Byte Order Mark using standard istream putback/unget.

Also, instead of trying to re-read the previous token on error (which involved seeking, which may or may not be available on an istream), simply collects the developing token into a std::string.  The space underlying this string is preserved between tokens, so this collection devolves into simple copying, and is therefore quite efficient.

This also opens the door to further simplifications in the future.

Added a simple unit test to confirm operation.  Did not test effect on json::parse.  I feel that the approach suggested by @nlohmann (std::parse demands that the entire buffer be a single JSON object, but operator>> scans just a single upcoming JSON object) is correct...  Perhaps this pull request may help move in that direction.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -1451,9 +1451,12 @@ class input_stream_adapter : public input_adapter_protocol
     input_stream_adapter(const input_stream_adapter&) = delete;
     input_stream_adapter& operator=(input_stream_adapter&) = delete;
 
+    // std::istream/std::streambuf use std::char_traits<char>::to_int_type, to
+    // ensure that std::char_traits<char>::eof() and the character 0xff do not
+    // end up as the same value, eg. 0xffffffff.
     int get_character() override
     {
-        return reinterpret_cast<int>( sb->sbumpc() );
+        return sb->sbumpc();
     }
 
     void unget_character() override
@@ -1489,10 +1492,10 @@ class input_buffer_adapter : public input_adapter_protocol
     {
         if (JSON_LIKELY(cursor < limit))
         {
-            return reinterpret_cast<int>(std::char_traits<char>::to_int_type(*(cursor++)));
+            return std::char_traits<char>::to_int_type(*(cursor++));
         }
 
-        return reinterpret_cast<int>(std::char_traits<char>::eof());
+        return std::char_traits<char>::eof();
     }
 
     void unget_character() noexcept override
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -2656,9 +2656,9 @@ scan_number_done:
     int get()
     {
         ++chars_read;
-        int c = current = ia->get_character();
-        token_string.push_back(static_cast<char>(c));
-        return c;
+        current = ia->get_character();
+        token_string.push_back(static_cast<char>(current));
+        return current;
     }
 
     /// unget current character (return it again on next get)
@@ -2703,7 +2703,7 @@ scan_number_done:
     }
 
     /// return string value
-    const std::string get_string()
+    std::string get_string()
     {
         return std::move( yytext );
     }
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -1397,7 +1397,7 @@ constexpr T static_const<T>::value;
 /// abstract input adapter interface
 struct input_adapter_protocol
 {
-    virtual int get_character() = 0; // returns characters in range [0,255], or eof() (a -'ve value)
+    virtual int get_character() = 0; // returns characters in range [0,255], or eof()
     virtual void unget_character() = 0; // restore the last non-eof() character to input
     virtual ~input_adapter_protocol() = default;
 };
@@ -1447,10 +1447,13 @@ class input_stream_adapter : public input_adapter_protocol
         }
     }
 
+    // delete because of pointer members
+    input_stream_adapter(const input_stream_adapter&) = delete;
+    input_stream_adapter& operator=(input_stream_adapter&) = delete;
+
     int get_character() override
     {
-        int c = sb->sbumpc(); // Avoided for performance: int c = is.get();
-        return c < 0 ? c : ( c & 0xFF ); // faster than == std::char_traits<char>::eof()
+        return reinterpret_cast<int>( sb->sbumpc() );
     }
 
     void unget_character() override
@@ -1486,10 +1489,10 @@ class input_buffer_adapter : public input_adapter_protocol
     {
         if (JSON_LIKELY(cursor < limit))
         {
-            return *(cursor++) & 0xFF;
+            return reinterpret_cast<int>(std::char_traits<char>::to_int_type(*(cursor++)));
         }
 
-        return std::char_traits<char>::eof();
+        return reinterpret_cast<int>(std::char_traits<char>::eof());
     }
 
     void unget_character() noexcept override
@@ -2668,9 +2671,9 @@ scan_number_done:
     {
         ++chars_read;
         current = ia->get_character();
-        if (JSON_LIKELY(current >= 0)) // faster than: != std::char_traits<char>::eof()))
+        if (JSON_LIKELY( current != std::char_traits<char>::eof()))
         {
-            token_string.push_back(static_cast<char>(current));
+            token_string.push_back(std::char_traits<char>::to_char_type(current));
         }
         return current;
     }
@@ -2679,7 +2682,7 @@ scan_number_done:
     void unget()
     {
         --chars_read;
-        if (JSON_LIKELY(current >= 0)) // faster than: != std::char_traits<char>::eof()))
+        if (JSON_LIKELY(current != std::char_traits<char>::eof()))
         {
             ia->unget_character();
             assert(token_string.size() != 0);
@@ -2690,7 +2693,7 @@ scan_number_done:
     /// add a character to yytext
     void add(int c)
     {
-        yytext.push_back(static_cast<char>(c));
+        yytext.push_back(std::char_traits<char>::to_char_type(c));
     }
 
   public:
@@ -5460,14 +5463,14 @@ class binary_reader
     {
         if (expect_eof)
         {
-            if (JSON_UNLIKELY(current >= 0 )) // faster than: != std::char_traits<char>::eof()))
+            if (JSON_UNLIKELY(current != std::char_traits<char>::eof()))
             {
                 JSON_THROW(parse_error::create(110, chars_read, "expected end of input"));
             }
         }
         else
         {
-            if (JSON_UNLIKELY(current < 0)) // faster than: == std::char_traits<char>::eof()))
+            if (JSON_UNLIKELY(current == std::char_traits<char>::eof()))
             {
                 JSON_THROW(parse_error::create(110, chars_read, "unexpected end of input"));
             }
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -1398,119 +1398,60 @@ constexpr T static_const<T>::value;
 struct input_adapter_protocol
 {
     virtual int get_character() = 0;
-    virtual std::string read(std::size_t offset, std::size_t length) = 0;
     virtual ~input_adapter_protocol() = default;
 };
 
 /// a type to simplify interfaces
 using input_adapter_t = std::shared_ptr<input_adapter_protocol>;
 
-/// input adapter for cached stream input
-template<std::size_t BufferSize>
-class cached_input_stream_adapter : public input_adapter_protocol
+
+/// input adapter for a (caching) istream.  Ignores a UFT Byte Order Mark at beginning of input.
+class input_stream_adapter : public input_adapter_protocol
 {
   public:
-    explicit cached_input_stream_adapter(std::istream& i)
-        : is(i), start_position(is.tellg())
+    explicit input_stream_adapter(std::istream& i)
+        : is(i)
     {
-        fill_buffer();
-
-        // skip byte order mark
-        if (fill_size >= 3 and buffer[0] == '\xEF' and buffer[1] == '\xBB' and buffer[2] == '\xBF')
-        {
-            buffer_pos += 3;
-            processed_chars += 3;
-        }
+	// Ignore Byte Order Mark at start of input
+	int c;
+	if (( c = get_character() ) == 0xEF )
+	{
+	    if (( c = get_character() ) == 0xBB )
+	    {
+		if (( c = get_character() ) == 0xBF )
+		{
+		    return; // Ignore BOM
+		}
+		else if ( c != std::char_traits<char>::eof() )
+		{
+		    is.unget();
+		}
+		is.putback( '\xBB' );
+	    }
+	    else if ( c != std::char_traits<char>::eof() )
+	    {
+		is.unget();
+	    }
+	    is.putback( '\xEF' );
+	}
+	else if ( c != std::char_traits<char>::eof() )
+	{
+	    is.unget(); // Not BOM.  Process as usual.
+	}
     }
 
-    ~cached_input_stream_adapter() override
-    {
-        // clear stream flags
-        is.clear();
-        // We initially read a lot of characters into the buffer, and we may
-        // not have processed all of them. Therefore, we need to "rewind" the
-        // stream after the last processed char.
-        is.seekg(start_position);
-        is.ignore(static_cast<std::streamsize>(processed_chars));
-        // clear stream flags
-        is.clear();
-    }
+    ~input_stream_adapter() override {}
 
     int get_character() override
     {
-        // check if refilling is necessary and possible
-        if (buffer_pos == fill_size and not eof)
-        {
-            fill_buffer();
-
-            // check and remember that filling did not yield new input
-            if (fill_size == 0)
-            {
-                eof = true;
-                return std::char_traits<char>::eof();
-            }
-
-            // the buffer is ready
-            buffer_pos = 0;
-        }
-
-        ++processed_chars;
-        assert(buffer_pos < buffer.size());
-        return buffer[buffer_pos++] & 0xFF;
-    }
-
-    std::string read(std::size_t offset, std::size_t length) override
-    {
-        // create buffer
-        std::string result(length, '\0');
-
-        // save stream position
-        const auto current_pos = is.tellg();
-        // save stream flags
-        const auto flags = is.rdstate();
-
-        // clear stream flags
-        is.clear();
-        // set stream position
-        is.seekg(static_cast<std::streamoff>(offset));
-        // read bytes
-        is.read(&result[0], static_cast<std::streamsize>(length));
-
-        // reset stream position
-        is.seekg(current_pos);
-        // reset stream flags
-        is.setstate(flags);
-
-        return result;
+	int c = is.get();
+	return c  == std::char_traits<char>::eof() ? c : ( c & 0xFF );
     }
 
   private:
-    void fill_buffer()
-    {
-        // fill
-        is.read(buffer.data(), static_cast<std::streamsize>(buffer.size()));
-        // store number of bytes in the buffer
-        fill_size = static_cast<size_t>(is.gcount());
-    }
 
     /// the associated input stream
     std::istream& is;
-
-    /// chars returned via get_character()
-    std::size_t processed_chars = 0;
-    /// chars processed in the current buffer
-    std::size_t buffer_pos = 0;
-
-    /// whether stream reached eof
-    bool eof = false;
-    /// how many chars have been copied to the buffer by last (re)fill
-    std::size_t fill_size = 0;
-
-    /// position of the stream when we started
-    const std::streampos start_position;
-
-    /// internal buffer
-    std::array<char, BufferSize> buffer{{}};
 };
 
 /// input adapter for buffer input
@@ -1541,13 +1482,6 @@ class input_buffer_adapter : public input_adapter_protocol
         return std::char_traits<char>::eof();
     }
 
-    std::string read(std::size_t offset, std::size_t length) override
-    {
-        // avoid reading too many characters
-        const auto max_length = static_cast<size_t>(limit - start);
-        return std::string(start + offset, (std::min)(length, max_length - offset));
-    }
-
   private:
     /// pointer to the current character
     const char* cursor;
@@ -1564,11 +1498,11 @@ class input_adapter
 
     /// input adapter for input stream
     input_adapter(std::istream& i)
-        : ia(std::make_shared<cached_input_stream_adapter<16384>>(i)) {}
+        : ia(std::make_shared<input_stream_adapter>(i)) {}
 
     /// input adapter for input stream
     input_adapter(std::istream&& i)
-        : ia(std::make_shared<cached_input_stream_adapter<16384>>(i)) {}
+        : ia(std::make_shared<input_stream_adapter>(i)) {}
 
     /// input adapter for buffer
     template<typename CharT,
@@ -2624,8 +2558,7 @@ scan_number_any2:
 scan_number_done:
         // unget the character after the number (we only read it to know that
         // we are done scanning a number)
-        --chars_read;
-        next_unget = true;
+	unget();
 
         // terminate token
         add('\0');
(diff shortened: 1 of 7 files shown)


nlohmann:
@pjkundert There is a test case failing that breaks all Travis runs, see https://travis-ci.org/nlohmann/json/builds/282450885?utm_source=github_status&utm_medium=notification.

nlohmann:
I ran a quick benchmark parsing file `benchmarks/files/jeopardy/jeopardy.json`

```cpp
#include "json.hpp"
#include <fstream>

using json = nlohmann::json;

int main()
{
    std::ifstream f("benchmarks/files/jeopardy/jeopardy.json");
    json j;
    f >> j;
}

```

compiled with `clang++ -std=c++11 -O3 -DNDEBUG -flto    bench.cpp   -o bench`.

The `develop` code runs in 1.08 seconds, whereas this PR's code runs in 1.46 seconds. Did you run any benchmarks?

pjkundert:
Performance now within about 10%, while eliminating all over-buffering and seeking.  Since we do not need to maintain the std::istream state flags, I used the underlying streambuf I/O API for performance.  The original code was also clearing the state flags before returning, so that has been maintained.

I noticed that the 'chars_read' (used to indicate file position where an error occurs) are incremented even when an EOF is received; this has been maintained (as it caused too many unit tests to fail when fixed).  It is not a huge deal, as it will only cause error messages to contain an error location position one beyond the last character in the input.

Runs clean under valgrind, and on all Linux platforms, and on Windows under Visual Studio 2015; 2017 fails for some reason I do not have the tooling to investigate.

theodelrieu:
The VS2017 problem is fixed upstream. Try to run `git rebase nlohmann/develop` (assuming you added a git remote nlohmann).

That should suffice!

nlohmann:
@theodelrieu is right about MSVC - it would be great if you could rebase so we can let MSVC 2017 check the code.

I do not understand why the performance is worse (about 7% with my naive benchmark).

src/json.hpp
```diff
@@ -2889,13 +2849,10 @@ class lexer
...
 
-    /// whether get() should return the last character again
-    bool next_unget = false;
-
     /// the number of characters read
     std::size_t chars_read = 0;
-    /// the start position of the current token
-    std::size_t start_pos = 0;
+    /// raw input token string (for error messages)
+    std::vector<char> token_string = std::vector<char>();
```
theodelrieu: redundant initialization, default constructor is fine here.

theodelrieu:
I believe some profiling would be useful for that matter.

I'm not very experienced in them though.

nlohmann:
@pjkundert Do you know whether this PR could also be adjusted to fix #714?

pjkundert:
Performance now exceeds that of the develop branch, in at least some benchmarks. Removed some unnecessary attempts to NUL-terminate the yytext string (since yytext may contain NULs anyway).  Converted yytext to a simple std::string, so we could return it directly via std::move(), and avoid many redundant std::string creations.

Let me know how this affects #714 ; since I use the underlying std::streambuf I/O, and only look for EOF (and do not set the iostream's failbit), it will almost certainly alter the behaviour you are seeing.

src/json.hpp
```diff
@@ -1398,119 +1398,68 @@ constexpr T static_const<T>::value;
...
-        // read bytes
-        is.read(&result[0], static_cast<std::streamsize>(length));
-
-        // reset stream position
-        is.seekg(current_pos);
-        // reset stream flags
-        is.setstate(flags);
-
-        return result;
+        int c = is.rdbuf()->sbumpc(); // Avoided for performance: int c = is.get();
```
theodelrieu: Is there a specific reason to not return `char ` instead of `int`? That's what I assumed from the function name.

theodelrieu: Is there a specific reason to not return `char` here? That's what I'd assume from the function name

src/json.hpp
```diff
@@ -2640,7 +2582,7 @@ class lexer
             const auto x = std::strtoull(yytext.data(), &endptr, 10);
 
             // we checked the number format before
-            assert(endptr == yytext.data() + yylen);
+            assert(endptr == yytext.data() + yytext.size());
```
theodelrieu: Could `yytext.end()` be used here?

src/json.hpp
```diff
@@ -2702,32 +2644,39 @@ class lexer
...
     }
 
     /// get a character from the input
     int get()
     {
         ++chars_read;
-        return next_unget ? (next_unget = false, current)
-               : (current = ia->get_character());
+        int c = current = ia->get_character();
+        token_string.push_back(static_cast<char>(c));
```
theodelrieu: Could you only use `current` here? By the way I think we should have a convention about data members, either explicitly use `this->`, or an underscore prefix etc...

But that's not in the PR's scope.

src/json.hpp
```diff
@@ -2702,32 +2644,39 @@ class lexer
...
+            ia->unget_character();
         }
-        assert(yylen < yytext.size());
-        yytext[yylen++] = static_cast<char>(c);
+        if (! token_string.empty())
+            token_string.pop_back();
+    }
+        
+    /// add a character to yytext
+    void add(int c)
```
theodelrieu: Same question about `int` vs `char` here.

src/json.hpp
```diff
@@ -2756,9 +2705,7 @@ class lexer
     /// return string value
     const std::string get_string()
     {
-        // yytext cannot be returned as char*, because it may contain a null
-        // byte (parsed as "\u0000")
-        return std::string(yytext.data(), yylen);
+        return std::move( yytext );
```
theodelrieu: This could lead to nasty bugsif called from a lvalue, could you provide lvalue and rvalue overloads?

src/json.hpp
```diff
@@ -2885,18 +2829,13 @@ class lexer
...
-    /// the start position of the current token
-    std::size_t start_pos = 0;
+    /// raw input token string (for error messages)
+    std::vector<char> token_string = std::vector<char>();
 
     /// buffer for variable-length tokens (numbers, strings)
-    std::vector<char> yytext = std::vector<char>(1024, '\0');
-    /// current index in yytext
-    std::size_t yylen = 0;
+    std::string yytext = "";
```
theodelrieu: Redundant initialization here

theodelrieu:
Sorry for the double comment...

src/json.hpp
pjkundert: std::streambuf I/O deals in ints, returning either traits_type::eof() (-1, usually) or the buffer item, converted to an int.  I'm not certain why; perhaps to not assume that the underlying data can only be 8-bit?

src/json.hpp
pjkundert: Yes, but we're testing the output of a function that was originally passed yytext.data(), so this is probably clearer.

src/json.hpp
pjkundert: Sure, I'll change that.

src/json.hpp
theodelrieu: Indeed, I was thinking about our interface: do we want to expose int or char for callers of `get_character`?

I believe we should `static _cast` between int/char while dealing with the `std::streambuf`, and only exposing `char`

src/json.hpp
nlohmann: I would have to look up the caller, but maybe we can also return a const reference.

src/json.hpp
nlohmann: I'm not a fan of neither namings...

theodelrieu: I do not really care about which naming to use, but I get confused each time, is it a local variable? A global one? A symbol in the namespace?

nlohmann: (Sorry, I'm ok with only using `current` - I meant I don't like adding an underscore prefix or `this->`)

src/json.hpp
pjkundert: Yes; but we're not doing that with any of the other members; I was just trying to be consistent.

src/json.hpp
theodelrieu: We can return a const ref for lvalues and move for rvalues. I remember someone on reddit discussing the rvalue support in the Library, which could be far better.

But that's not an emergency at all

pjkundert: I don't think so; but, on review -- I noticed that the get_string function was (incorrectly) returning a 'const std::string', preventing any move re-use of the returned std::string anyway!  Removing the const appears to have resulted in (another) pretty dramatic speed-up.  New push forthcoming...

src/json.hpp
theodelrieu: I understand, it's just hard to know what the symbol is without performing a search in the current file.

src/json.hpp
theodelrieu: Well simply calling `get_string` twice would become problematic. It adds more boilerplate to add those overloads, but the code will always be correct. We should use `std::move(lexer).get_string()` to be explicit about this side effect

nlohmann: I think consuming from a lexer (as in: moving data out of the lexer) is a common pattern. We can add documentation though.

pjkundert: This get_string method is being used for keys and values; both of which are ultimately moved into a std::map key, or into a BasicJsonType value; there are no lvalue users.

pjkundert: Yup, appears to improve performance by close to 10% on some benchmarks.  Nice.

pjkundert: Yes, calling get_string twice would (surprisingly) result in an empty string.  Perhaps renaming it to `move_string()` would result in less surprising behavior?  I'm not sure I understand your `std::move(lexer).get_string()` recommendation.

src/json.hpp
pjkundert: Since we need to return values in the range [-1,255], we must continue to use int.

src/json.hpp
theodelrieu: I get your point. Let's hope nobody will get bitten by that then :)

src/json.hpp
theodelrieu: I guess it's up to personal preference, not a big deal anyway ;)

nlohmann:
I re-ran the simple benchmark with the latest commit: it is still slower: 1.24 seconds vs. 1.14 seconds parsing jeopardy.json...

pjkundert:
I guess it's compiler dependent.  On g++ 7.2, its substantially faster on that benchmark.

nlohmann:
Also with GCC 8.0.0 20170919 I measure worse runtimes, but the margin became smaller.

src/json.hpp
gregmarr: You are relying on the optimizer to eliminate the call to strlen() here.  They should eliminate the call and just do default initialization, but better to not make it have to do that.  I would run the benchmarks with and without this and see if there's any difference.

src/json.hpp
pjkundert: This implies that there is another issue -- we (presently) allow EOF to be pushed into the token_string, which is a char vector.  Must prevent this.  Also allows simplifying the get_token_string.  Furthermore, a NUL in the token_string should be escaped (not presently handled; they are ignored for the purposes of error messages, which is not correct).

src/json.hpp
theodelrieu: I haven't seen your last answer sorry. You can overload on lvalues and rvalues:

```cpp
class lexer {
std::string const& get_string() & {
  return _content;
}

std::string get_string() && {
  return std::move(_content);
}
};
```

The first function can only be called on lvalues, while the latter can only be called on rvalues.
Hence the `std::move(lexer).get_string()`, which will call the second overload, while `lexer.get_string()` will call the first one.

I've used this pattern on classes holding huge data, and I would call the rvalue overload at the very end of the processing, while calling the lvalue one during processing.

But a `move_string` is fine with me too.

pjkundert:
Further performance improvements.  Since EOF is the sole -'ve value allowed through the get() interface, we can use less costly comparisons to detect its presence.  Further 1-2% performance improvement.

pjkundert:
Just to refocus the debate -- we're bringing the parser into compliance with expected behavior on std::istream; if this requires a small performance penalty, I believe this is acceptable.  The present behavior on std::istream >> is extremely surprising and non-compliant...

pjkundert:
The performance improvement in detecting EOF is probably due to the code re-using a "-'ve value" CPU flag left over from a previous move, instead of issuing a new comparison against a specific value.  Changing that specific value to const or even constexpr wouldn't help.

pjkundert:
The token_string is used solely in formatting error messages; NUL values can't typically exist in well-formed JSON, so they will typically terminate parsing in an error.  We want them to be output as the last character read -- so they must be escaped in the error message, which is now done.  As for EOF: no, they cannot appear in token_string, because it is a char vector containing values in the range [-128,127] (or [0,255] if interpreted as unsigned char).  We specifically detect -'ve values, and avoid introducing them into token_string.

gregmarr:
There are still more places where it compares against `eof()`.  If it is that big of a savings, then there should probably be a single function that implements this test so that it can be done the same everywhere.  Are we sure that `eof()` is the only possible negative value?

pjkundert:
All remaining usages of std::char_traits<char>::eof() are either A) `switch ... case`entries, or B) only performed rarely.  I'd rather leave those B) cases explicit, I think.  Even in the cases where we widen the comparison to simply -'ve (from specifically == eof()), I left the original comparison there for documentary purpose.  I don't really like to do this kind of optimization at all, but in a well-tested, performance-critical tool such as this, it may be worth it.

pjkundert:
Right now, compiler warnings due to -Weffc++ prevent the default-initialization of some members, as this is not possible using `... = <value>;` initializers.  I would like to convert all of the member initializers from the `... = <value>;` form, to initializer-list form: `... { <value> };`.  However, this may cause difficulty with some not-quite compliant c++11 compilers (eg. g++ < 4.9.1?)  What do you all think of this idea?

gregmarr:
Is this not supported by 4.9.0?  The README currently lists the minimum g++ version as 4.9.

nlohmann:
g++-4.9 (Ubuntu 4.9.4-2ubuntu1~14.04.1) 4.9.4 is successfully used by Travis. I don't know about g++ 4.9.1 though.

gregmarr:
https://github.com/google/googletest/issues/898

GCC's -Weffc++ is IMHO effectively obsolete, and should not be used, unless you are trying to ensure your code meets outdated advice from more than ten years ago. There are numerous bug reports about -Weffc++ in GCC's bugzilla and nobody is interested in fixing them. Among its problems are that the suggestions it makes are based on the first edition of Effective C++ and those items were heavily revised for the second edition. Some advice is simply inappropriate in modern code (C++11 or later) and some warnings are simply poorly implemented in GCC (e.g requiring a mem-initializer for all member variables even if they have default constructors that do the right thing).

Changing modern projects to avoid unhelpful -Weffc++ warnings is a huge mistake.

theodelrieu:
I'm all for removing -Weffc++.

nlohmann:
To my defense: I just added as many compiler warning flags as possible... :-/

theodelrieu:
Cool, we might remove -Weffc++ in another commit. I have no other comments. Thanks :)

nlohmann:
Can anyone confirm the performance improvement?

theodelrieu:
Is there anything specific to

nlohmann:
See https://github.com/nlohmann/json/pull/764#issuecomment-333783833

theodelrieu:
I have no significant difference with GCC 7.2 on Debian 8 64bits, however with clang 4.0.1 it goes from 0.64 sec to 1 sec roughly.

But a real profiling would be needed to determine the cause.

Not related to this topic, for some reason my `git log` tries to show the whole commit message on a single line, don't know why.

src/json.hpp
```diff
@@ -2880,34 +2833,30 @@ class lexer
 
   private:
     /// input adapter
-    detail::input_adapter_t ia = nullptr;
+    detail::input_adapter_t ia { nullptr };
```
nlohmann: I removed the `-Weffc++` warnings. Please use the `=`-style initializations again.

src/json.hpp
```diff
@@ -1397,120 +1397,69 @@ constexpr T static_const<T>::value;
 /// abstract input adapter interface
 struct input_adapter_protocol
 {
-    virtual int get_character() = 0;
-    virtual std::string read(std::size_t offset, std::size_t length) = 0;
+    virtual int get_character() = 0; // returns characters in range [0,255], or eof() (a -'ve value)
```
nlohmann: Just a question: what do you mean with `(a -'ve value)`?

gregmarr: `(a negative value)`

src/json.hpp
```diff
@@ -2702,32 +2644,51 @@ class lexer
...
+
+    @return character read from the input
+    */
     int get()
     {
         ++chars_read;
-        return next_unget ? (next_unget = false, current)
-               : (current = ia->get_character());
+        current = ia->get_character();
+        if (JSON_LIKELY(current >= 0)) // faster than: != std::char_traits<char>::eof()))
```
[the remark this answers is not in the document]
pjkundert: Probably not safe.  See: https://stackoverflow.com/questions/24482728/return-value-of-istreamget

In fact, all of the assumed `int get()` and `int get_character()` is also not strictly correct.  It should be istream::int_type.  And, we should be using istream::traits_type::eq_int_type to detect equality w/ eof(), and istream::traits_type::to_char_type to convert to a char (not ... & 0xFF).

So, I agree -- perhaps we should clean up the char handling a bit to bring it more into strict compliance.  This may cost us the 1-2% performance increase we saw by just testing for negative values instead of == eof() specifically.

src/json.hpp
pjkundert: I think using initializer-lists is really the future path of C++ member initialization; shouldn't we generally be going in that direction, instead?

theodelrieu: I agree that direct-list-initialization has some advantages here. No narrowing, and it considers explicit constructors as well.

I'm not a fan of member initialization when it's redundant (e.g. `std::string s{}`), but I would go for brace-style here too.

pjkundert:
I've refactored the `get` / `get_character` methods to handle std::streambuf input int_type in a strictly correct fashion, I think.  I make no assumptions about the negative value of eof(), as required by the spec.  G++ still substantially faster, Clang only slightly slower than stock develop branch; push forthcoming.

src/json.hpp
nlohmann: I'm ok with letting it as it is. But just because braces work everywhere does not mean that `ia = nullptr;` is wrong in any sense. That's why *I* don't think we should change existing code.

src/json.hpp
```diff
@@ -1535,17 +1489,18 @@ class input_buffer_adapter : public input_adapter_protocol
     {
         if (JSON_LIKELY(cursor < limit))
         {
-            return *(cursor++) & 0xFF;
+            return reinterpret_cast<int>(std::char_traits<char>::to_int_type(*(cursor++)));
```
nlohmann: This seems to fail on MSVC, see https://ci.appveyor.com/project/nlohmann/json/build/2302/job/n1ti8cx5qqafk43i.

nlohmann:
> G++ still substantially faster, Clang only slightly slower than stock develop branch; push forthcoming.

I'm afraid I could still not reproduce this. I shall try to run some benchmarks in the weekend.

src/json.hpp
theodelrieu: Right, it's not a refactoring/style PR anyway, I agree changing this code here is not that relevant.

src/json.hpp
theodelrieu: a `static_cast` should suffice here

theodelrieu:
@pjkundert I think I understand why my `git log` gets wrecked when visualizing your commits. Could you insert a blank line between the header and the commit body? I have the whole commits on a single line.

Good job on the body details BTW :)

src/json.hpp
```diff
@@ -1397,120 +1397,77 @@ constexpr T static_const<T>::value;
 /// abstract input adapter interface
 struct input_adapter_protocol
 {
-    virtual int get_character() = 0;
-    virtual std::string read(std::size_t offset, std::size_t length) = 0;
+    virtual int get_character() = 0; // returns characters in range [0,255], or eof()
```
gregmarr: Should this be `std::char_traits<char>::int_type` instead of `int` so that we know that there will be no type changes later?

src/json.hpp
```diff
@@ -1845,9 +1803,9 @@ class lexer
     @brief scan a string literal
 
     This function scans a string according to Sect. 7 of RFC 7159. While
-    scanning, bytes are escaped and copied into buffer yytext. Then the
-    function returns successfully, yytext is null-terminated and yylen
-    contains the number of bytes in the string.
+    scanning, bytes are escaped and copied into buffer yytext. Then the function
+    returns successfully, yytext is *not* null-terminated (as it may contain \0
+    bytes), and yytext.size() is the number of bytes in the string.
```
gregmarr: Would it be more accurate to say that it's null terminated but may also contain embedded nulls?
This currently sounds like there isn't guaranteed to be at least one null.

gregmarr: Ah, that's a change in behavior, I see now.

src/json.hpp
```diff
@@ -2885,18 +2846,14 @@ class lexer
...
-    std::size_t start_pos = 0;
+
+    /// raw input token string (for error messages)
+    std::vector<char> token_string { };
 
     /// buffer for variable-length tokens (numbers, strings)
-    std::vector<char> yytext = std::vector<char>(1024, '\0');
-    /// current index in yytext
-    std::size_t yylen = 0;
+    std::string yytext { };
```
gregmarr: Neither of these need the `{ }` at the end.

src/json.hpp
```diff
@@ -5205,7 +5162,7 @@ class binary_reader
     @brief get next character from the input
 
     This function provides the interface to the used input adapter. It does
-    not throw in case the input reached EOF, but returns
+    not throw in case the input reached EOF, but returns a -'ve valued
```
gregmarr: We should remove this "negative valued" bit, right?

src/json.hpp
pjkundert: That would be an acceptable change, I think.  It's typically that type anyway, and being explicit about it keeps things obvious; that all implementation of the adapter class need to be consistent in producing `std::char_traits<char>::int_type`, too.

gregmarr: A `using int_type = ...` would make it less verbose then the full thing.

src/json.hpp
pjkundert: Yes, I assume that it used to be NUL-terminated, in the mists of history, and was probably being returned as a `char*`'  Since the token can contain NULs (supplied by escaping), any assumptions about NUL and termination are incorrect.  So it was removed.

src/json.hpp
pjkundert: Correct.  However, since the -Weffc++ flag still exists (and arbitrary client code may still use it), and it is cost and risk free for us to still comply with its constraints, I decided to include the default initializer.

src/json.hpp
pjkundert: Agreed.

src/json.hpp
gregmarr: I'm not sure I'd say that there is no cost.  There is a cost whenever you do something unnecessary to appease a broken tool.  It's not a runtime or build-time cost, more of a cognitive burden.  If we intend to keep it there to support the warning, anyone who maintains the code should know why it's there, which means that there should probably be a comment.
Anyway, it's all philosophical.  I'm not going to lose any sleep over it.

theodelrieu:
Are there anything left required before merging this?

nlohmann:
I'll have another look soon - I want to have another look whether this also fixes #714.

About anything else: are we happy with this approach? I could not measure performance improvement, but I understand that less code is preferable here, and the caching always felt like a hack (though it performed better in my bechmarks, and also with the simple example of parsing jeopardy.json).

trilogy-service-ro:

theodelrieu:
I personally have no more issues with this PR.

theodelrieu:
Can we merge this PR?

nlohmann:
Thanks so much for all the patience!

PR #765, opened 2017-10-03
using more CMake

nlohmann:
Done:

- using CMake for the special targets Valgrind (JSON_Valgrind), Clang sanitizers (JSON_Sanitizer), and no-exceptions (JSON_NoExceptions)
- using Ninja to improve compilation speed
- compiling and executing each test suite individually
- removed test cases that broke the MSVC 2017 build

Missing:

- some magic to get Coveralls to work with individual executables (maybe @theodelrieu can help here with his approach from #700)


nlohmann:
(This should fix #698)

nlohmann:
@theodelrieu Would it be OK if I merged this PR tonight?

theodelrieu:
Sure! Even now if you want

PR #785, opened 2017-10-16
Fix warning C4706 on Visual Studio 2017

jseward:
I'm pretty sure the logic is the same, just moves out the updating of variable `keep`. All tests pass at least.

This also fixes a bug in `unit-udt` which has a missing `u8` causing crashes on windows (due to iterator assert)

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -91,6 +91,16 @@ foreach(file ${files})
         set_target_properties(${testcase} PROPERTIES COMPILE_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated -Wno-float-equal")
     endif()
 
+    # https://stackoverflow.com/questions/2368811/how-to-set-warning-level-in-cmake
+    if(MSVC)
+        # Force to always compile with W4
+        if(CMAKE_CXX_FLAGS MATCHES "/W[0-4]")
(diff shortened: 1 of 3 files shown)


trilogy-service-ro:

nlohmann:
@trilogy-service-ro Could you please stop adding automated comments to every PR of this project? Of course, we have a look at each patch.

test/CMakeLists.txt
```diff
@@ -66,6 +66,19 @@ set_target_properties(catch_main PROPERTIES
 )
 target_include_directories(catch_main PRIVATE "thirdparty/catch")
 
+# https://stackoverflow.com/questions/2368811/how-to-set-warning-level-in-cmake
+if(MSVC)
+    # Force to always compile with W4
```
nlohmann: Why do you want to force the `/W4` flag?

jseward: With /W4 always enabled we can ensure no warnings will be emitted when adding this library to an existing project. It feels like a library such as this one should support /W4 with no issues. There are other warnings in unit tests even with /W3, but that would be for another PR. At least there is visibility now.

nlohmann: Ok, thanks. I realized we're up to some 50 warnings now...

nlohmann: I would really appreciate any help with MSVC regarding this warnings, as I have no experience here.

nlohmann:
Thanks a lot!

PR #788, opened 2017-10-17
Fix Visual Studio 2017 warnings

jseward:
With this PR there should now be 0 warnings in Visual Studio 2017 /W4 when compiling all tests.


trilogy-service-ro:

test/src/unit-readme.cpp
```diff
@@ -298,3 +303,7 @@ TEST_CASE("README", "[hide]")
         std::cout.rdbuf(old_cout_buffer);
     }
 }
+
+#if defined(_MSC_VER)
+#pragma warning (pop)
+#endif
```
nlohmann: Could you please add a newline here?

test/CMakeLists.txt
```diff
@@ -75,8 +75,11 @@ if(MSVC)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4")
     endif()
 
-    # Disable warning C4389: '==': signed/unsigned mismatch
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4389")
+    # warning C4389: '==': signed/unsigned mismatch
```
nlohmann: Could you please add to the comment that these warnings are disabled?

jseward:
Done and done!

nlohmann:
In MSVC 2015, there is still one warning left, which does not seem to occur in MSVC 2017:

> json.hpp(7950): warning C4702: unreachable code [C:\projects\json\test\test-allocator.vcxproj]

See https://ci.appveyor.com/project/nlohmann/json/build/2315/job/f15evu7h4lpcxog8

This is really strange. Do you have any ideas about this?

gregmarr:
The allocator used in this test always throws, so that code is indeed unreachable.  It probably takes some really aggressive inlining to determine that.  It could be that VS2017 isn't doing as much inlining as VS2015 in that particular area.  It could also be that the unreachable code warning has been modified to allow code after an unconditional throw.

nlohmann:
@gregmarr Thanks, this makes sense.

nlohmann:
Thanks @jseward !

Issue #464, opened 2017-02-20
VS2017 implicit to std::string conversion fix.


qis:
Hi! Great work so far. I had to patch the library to allow implicit to-string-conversions in VS2017.

```cpp
    template < typename ValueType, typename std::enable_if <
                   not std::is_pointer<ValueType>::value and
                   not std::is_same<ValueType, typename string_t::value_type>::value
//#ifndef _MSC_VER  // fix for issue #167 operator<< ambiguity under VS2015
                   and not std::is_same<ValueType, std::initializer_list<typename string_t::value_type>>::value
                   and not std::is_same<ValueType, typename std::string_view>::value
//#endif
                   , int >::type = 0 >
    operator ValueType() const
    {
        // delegate the call to get<>() const
        return get<ValueType>();
    }
```

Hope this helps.

nlohmann:
@qis Could you please try the most recent version (f2dfa09aecf672a3b9980613bde196a5cb1ef994)?

nlohmann:
My bad - I misunderstood your comment, so I reverted the last commit.

I cannot add the line with `std::string_view` to the code, because this would require C++17.

Could you please post the compiler error for the original code together with the options you are using with VS2017?

qis:
Here are my tests with 1a6d7f5bca9baec12fe6065c6009903c24ea533f:

```cpp
#include <json.hpp>
#include <string>

int main(int argc, char* argv[]) {
  using json = nlohmann::json;
  json v  = "test";
  std::string test;
  test = v;
  return 0;
}
```

Compiled with `/permissive- /std:c++latest /utf-8` it yields this error:

```
1>main.cpp
1>C:\Workspace\test\src\main.cpp(8): error C2593: 'operator =' is ambiguous
1>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.10.24930\include\xstring(2254): note: could be 'std::basic_string<char,std::char_traits<char>,std::allocator<char>> &std::basic_string<char,std::char_traits<char>,std::allocator<char>>::operator =(const _Elem)'
1>        with
1>        [
1>            _Elem=char
1>        ]
[...]

PR #795, opened 2017-10-21
Add transparent comparator and perfect forwarding support to find() and count()

jseward:
This avoids unnecessary string copies and allocation on often used `find()` if `std::string_view` is supported.

NOTE: currently only implemented for MSVS 2017, should be easily enabled for other compilers as well but don't have them available.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -10884,7 +10884,7 @@ class basic_json
     @note This method always returns @ref end() when executed on a JSON type
           that is not an object.
 
-    @param[in] key key value of the element to search for
+    @param[in] key key value of the element to search for.
 
     @return Iterator to an element with key equivalent to @a key. If no such
     element is found or the JSON value is not an object, past-the-end (see
@@ -10895,39 +10895,6 @@ class basic_json
     @liveexample{The example shows how `find()` is used.,find__key_type}
 
     @since version 1.0.0
-    */
-    iterator find(typename object_t::key_type key)
-    {
-        auto result = end();
-
-        if (is_object())
-        {
-            result.m_it.object_iterator = m_value.object->find(key);
-        }
-
-        return result;
-    }
-
-    /*!
-    @brief find an element in a JSON object
-    @copydoc find(typename object_t::key_type)
-    */
-    const_iterator find(typename object_t::key_type key) const
-    {
-        auto result = cend();
-
-        if (is_object())
-        {
-            result.m_it.object_iterator = m_value.object->find(key);
-        }
-
-        return result;
-    }
-
-#ifdef JSON_HAS_CPP_14
-    /*!
-    @brief find an element in a JSON object
-    @copydoc find(typename object_t::key_type)
     */
     template<typename KeyT>
     iterator find(KeyT&& key)
@@ -10944,7 +10911,7 @@ class basic_json
 
     /*!
     @brief find an element in a JSON object
-    @copydoc find(typename object_t::key_type)
+    @copydoc find(KeyT&&)
     */
     template<typename KeyT>
     const_iterator find(KeyT&& key) const
@@ -10958,7 +10925,6 @@ class basic_json
 
         return result;
     }
-#endif
 
     /*!
     @brief returns the number of occurrences of a key in a JSON object
@@ -10981,24 +10947,13 @@ class basic_json
 
     @since version 1.0.0
     */
-    size_type count(typename object_t::key_type key) const
-    {
-        // return 0 for all nonobject types
-        return is_object() ? m_value.object->count(key) : 0;
-    }
-
-#ifdef JSON_HAS_CPP_14
-    /*!
-    @brief returns the number of occurrences of a key in a JSON object
-    @copydoc count(typename object_t::key_type)
-    */
     template<typename KeyT>
     size_type count(KeyT&& key) const
     {
         // return 0 for all nonobject types
         return is_object() ? m_value.object->count(std::forward<KeyT>(key)) : 0;
     }
-#endif
+
     /// @}
 
 
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -110,7 +110,7 @@ SOFTWARE.
 #endif
 
 // string_view support
-#if defined(_MSC_VER) && defined(_HAS_CXX17)
+#if defined(_MSC_VER) && defined(_HAS_CXX17) && _HAS_CXX17 == 1
 	#define JSON_USE_STRING_VIEW
 #endif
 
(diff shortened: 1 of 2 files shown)


jseward:
Found some chat on https://github.com/nlohmann/json/issues/464 for similar issues. Hijacked the solution there for this new functionality.

trilogy-service-ro:

nlohmann:
@trilogy-service-ro Please stop spamming PRs.

src/json.hpp
```diff
@@ -7589,11 +7599,27 @@ class basic_json
...
+	using object_comparator_key_t = std::string_view;
+#else
+	using object_t = ObjectType<StringType,
+		basic_json,
+		std::less<StringType>,
+		AllocatorType<std::pair<const StringType,
+		basic_json>>>;
+
+	using object_comparator_key_t = typename object_t::key_type;
+#endif
```
gregmarr: Any reason not to do the transparent comparator across the board for C++14 and higher?

src/json.hpp
```diff
@@ -10894,9 +10920,9 @@ class basic_json
 
     /*!
     @brief find an element in a JSON object
-    @copydoc find(typename object_t::key_type)
+    @copydoc find(object_transparent_comparator_key_t)
```
gregmarr: Where is this type (with transparent in it) defined?

src/json.hpp
```diff
@@ -10929,7 +10955,7 @@ class basic_json
 
     @since version 1.0.0
     */
-    size_type count(typename object_t::key_type key) const
+    size_type count(object_comparator_key_t key) const
```
gregmarr: Would the `find` and `count` functions be better as templates with perfect forwarding, so we can find or count directly anything that the underlying object class can find or count?

jseward:
@gregmarr thanks for the great feedback. I've implemented your suggested changes

src/json.hpp
```diff
@@ -110,11 +110,11 @@ SOFTWARE.
 #endif
 
 // string_view support
-#if defined(_MSC_VER) && defined(_HAS_CXX17) && _HAS_CXX17 == 1
-	#define JSON_USE_STRING_VIEW
+#if (defined(__cplusplus) && __cplusplus >= 201703L) || (defined(_MSC_VER) && _MSC_VER > 1900 && defined(_HAS_CXX17) && _HAS_CXX17 == 1) // fix for issue #464
```
theodelrieu: Could we use `__has_include` instead?

src/json.hpp
```diff
@@ -10908,6 +10924,42 @@ class basic_json
         return result;
     }
 
+#ifdef JSON_HAS_CPP_14
+    /*!
+    @brief find an element in a JSON object
+    @copydoc find(typename object_t::key_type)
+    */
+    template<typename KeyT>
+    iterator find(KeyT&& key)
```
theodelrieu: Could we define those perfect-forwarded overloads for C++11 as well?

test/CMakeLists.txt
```diff
@@ -80,6 +80,11 @@ if(MSVC)
 	# Disable warning C4566: character represented by universal-character-name '\uFF01' cannot be represented in the current code page (1252)
 	# Disable warning C4996: 'nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>::operator <<': was declared deprecated
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4389 /wd4309 /wd4566 /wd4996")
+
+    if(MSVC_VERSION GREATER_EQUAL 1910)
+        # Enable c++17 support in Visual Studio 2017 (for testing perfect forwarding and transparent comparator in find() / count())
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++17")
+    endif()
```
theodelrieu: IIRC Appveyor already runs with VS2017, you might want to put that flag in the Appveyor file?

nlohmann:
Sorry - AppVeyor fails because of a recent change of mine. I shall fix this.

nlohmann:
The `develop` version now compiles without problems with AppVeyor.

src/json.hpp
gregmarr: The templated member functions don't exist in 11. They were added in 14.

theodelrieu: Yes, but we could keep those overloads in C++11, they will call the regular comparators which will perform implicit conversions, as for:

```cpp
std::cout << std::multiplies<int>{}(2.0f, 2.0f) << std::endl;
```

I haven't tested my suggestion though, so it might very well be bogus.

gregmarr: My first though was that it might be weird for users that aren't familiar with the difference in capabilities between `std::map` in 11 and 14 to have the templated type that only works with things implicitly convertible to the key type, but maybe that's not an issue, as you'd get the same issue with anything that isn't comparable with the key type with the 14 version.  The STL kept the non-template `find()` and `count()`, but likely only for backwards compat.  If it works, I think it would be better with just the templated versions.  If we can't eliminate the fixed type versions, we should add `const &` to them to match the `std::map` versions.  Right now we make a copy unnecessarily.  (I just noticed this now.)

gregmarr:
@jseward You're welcome.  I'm glad it worked out.

test/CMakeLists.txt
nlohmann: Indeed we have three configurations for AppVeyor:

- MSVC 2015
- MSVC 2017
- MSVC 2017 + `/permissive- /std:c++latest /utf-8`

jseward: Ah nice, didn't realize that thanks. I'll this remove this then.

src/json.hpp
jseward: I like the suggestion, pushed changes to just have perfect-forwarding only

jseward:
travis-ci failing with:

```
==> Cloning https://github.com/nlohmann/json.git
Cloning into '/Users/travis/Library/Caches/Homebrew/nlohmann_json--git'...
remote: Counting objects: 1279, done.
remote: Compressing objects: 100% (605/605), done.
error: RPC failed; curl 56 SSLRead() return error -36
fatal: The remote end hung up unexpectedly
fatal: early EOF
fatal: index-pack failed
Error: Failed to download resource "nlohmann_json"
Failure while executing: git clone --depth 1 --branch develop https://github.com/nlohmann/json.git /Users/travis/Library/Caches/Homebrew/nlohmann_json--git
Error: Testing requires the latest version of nlohmann/json/nlohmann_json
```

don't notice an obvious way to retry

nlohmann:
Travis' support for macOS is a bit flaky. I restarted the job.

nlohmann:
Thanks a lot!

Issue #774, opened 2017-10-08
Another idea about type support


nlohmann:
# The problem

The library currently offers some template arguments to configure the types used for numbers, strings, booleans, objects, and arrays. However, this type support is very limited, and even simple things like using `std::unordered_map` for objects or an `std::vector` with a user-defined allocator for arrays do not work. Even worse, there are tons of `std::string` usages in the code even though we actually have a type `basic_json::string_t` which is a typedef for the passed string type (see #766).

With issue #456 came the idea to replace the long template parameter list with a policy class. I like the idea, but I am afraid that this change does not improve things much with respect to the mentioned problems.

# Another solution

One possible solution for a true type configuration could be the following: we define interfaces for each type that must be implemented by concrete implementations to be used in the `basic_json` class which would be refactored to only rely on this interface.

For instance, we could have a class hierarchy

```cpp
class abstract_value
{
public:
    virtual ~abstract_value() = 0;

    virtual bool empty() const noexcept = 0;
    virtual size_type size() const noexcept = 0;
    virtual size_type max_size() const noexcept = 0;
    virtual void clear() noexcept = 0;
    
    virtual bool is_primitive() const noexcept = 0;
    virtual bool is_structured() const noexcept = 0;
    virtual const char* name() const noexcept = 0;
    virtual value_t type() const noexcept = 0;
    
    ...
};
```

```cpp
class abstract_primitive_value : public abstract_value
{
public:
    bool empty() const noexcept override { return false; }
    size_type size() const noexcept override { return 1; }
    size_type max_size() const noexcept override { return 1; }
    
    bool is_primitive() const noexcept override { return true; };
    bool is_structured() const noexcept override { return false; };
    
    ...
};
```

```cpp
class abstract_boolean_value : public abstract_primitive_value
{
public:
    const char* name() const noexcept override { return "boolean"; }
    value_t type() const noexcept override { return value_t::boolean; }
    
    ...
};
```

and finally

```cpp
class boolean_value_imp : public abstract_boolean_value
{
public:
    boolean_value_imp() = default;
    boolean_value_imp(const bool v) : m_value(v) {}
    
    void clear() noexcept override
    {
        m_value = false;
    }
    
    ...
    
private:
    bool m_value = false;
};
```

to implement the Boolean value.

We could then have a traits class like

```cpp
struct default_value_traits {
    using null_value = null_value_imp;
    using boolean_value = boolean_value_imp;
    using string_value = string_value_imp;
[...]

PR #798, opened 2017-10-24
Add no thrown exception support to parse

jseward:
This PR adds support for parse to not throw exceptions.  Use similar pattern as `std::experimental::filesystem` where overload that takes in `std::error_code&` does not throw an exception.

Did not add new tests as changes to existing tests provide full coverage of new codepath.

NOTE: accidentally branched from https://github.com/nlohmann/json/pull/795, ignore that for now.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -10884,7 +10884,7 @@ class basic_json
     @note This method always returns @ref end() when executed on a JSON type
           that is not an object.
 
-    @param[in] key key value of the element to search for
+    @param[in] key key value of the element to search for.
 
     @return Iterator to an element with key equivalent to @a key. If no such
     element is found or the JSON value is not an object, past-the-end (see
@@ -10895,39 +10895,6 @@ class basic_json
     @liveexample{The example shows how `find()` is used.,find__key_type}
 
     @since version 1.0.0
-    */
-    iterator find(typename object_t::key_type key)
-    {
-        auto result = end();
-
-        if (is_object())
-        {
-            result.m_it.object_iterator = m_value.object->find(key);
-        }
-
-        return result;
-    }
-
-    /*!
-    @brief find an element in a JSON object
-    @copydoc find(typename object_t::key_type)
-    */
-    const_iterator find(typename object_t::key_type key) const
-    {
-        auto result = cend();
-
-        if (is_object())
-        {
-            result.m_it.object_iterator = m_value.object->find(key);
-        }
-
-        return result;
-    }
-
-#ifdef JSON_HAS_CPP_14
-    /*!
-    @brief find an element in a JSON object
-    @copydoc find(typename object_t::key_type)
     */
     template<typename KeyT>
     iterator find(KeyT&& key)
@@ -10944,7 +10911,7 @@ class basic_json
 
     /*!
     @brief find an element in a JSON object
-    @copydoc find(typename object_t::key_type)
+    @copydoc find(KeyT&&)
     */
     template<typename KeyT>
     const_iterator find(KeyT&& key) const
@@ -10958,7 +10925,6 @@ class basic_json
 
         return result;
     }
-#endif
 
     /*!
     @brief returns the number of occurrences of a key in a JSON object
@@ -10981,24 +10947,13 @@ class basic_json
 
     @since version 1.0.0
     */
-    size_type count(typename object_t::key_type key) const
-    {
-        // return 0 for all nonobject types
-        return is_object() ? m_value.object->count(key) : 0;
-    }
-
-#ifdef JSON_HAS_CPP_14
-    /*!
-    @brief returns the number of occurrences of a key in a JSON object
-    @copydoc count(typename object_t::key_type)
-    */
     template<typename KeyT>
     size_type count(KeyT&& key) const
     {
         // return 0 for all nonobject types
         return is_object() ? m_value.object->count(std::forward<KeyT>(key)) : 0;
     }
-#endif
+
     /// @}
 
 
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -110,7 +110,7 @@ SOFTWARE.
 #endif
 
 // string_view support
-#if defined(_MSC_VER) && defined(_HAS_CXX17)
+#if defined(_MSC_VER) && defined(_HAS_CXX17) && _HAS_CXX17 == 1
 	#define JSON_USE_STRING_VIEW
 #endif
 
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -111,10 +111,10 @@ SOFTWARE.
 
 // cpp language standard detection
 #if (defined(__cplusplus) && __cplusplus >= 201703L) || (defined(_HAS_CXX17) && _HAS_CXX17 == 1) // fix for issue #464
-	#define JSON_HAS_CPP_17
-	#define JSON_HAS_CPP_14
+    #define JSON_HAS_CPP_17
+    #define JSON_HAS_CPP_14
 #elif (defined(__cplusplus) && __cplusplus >= 201402L) || (defined(_HAS_CXX14) && _HAS_CXX14 == 1)
-	#define JSON_HAS_CPP_14
+    #define JSON_HAS_CPP_14
 #endif
 
 /*!
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -7606,10 +7606,10 @@ class basic_json
     using object_comparator_t = std::less<StringType>;
 #endif
     using object_t = ObjectType<StringType,
-        basic_json,
-        object_comparator_t,
-        AllocatorType<std::pair<const StringType,
-        basic_json>>>;
+      basic_json,
+      object_comparator_t,
+      AllocatorType<std::pair<const StringType,
+      basic_json>>>;
 
     /*!
     @brief a type for an array
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -109,6 +109,16 @@ SOFTWARE.
     #define JSON_UNLIKELY(x)    x
 #endif
 
+// string_view support
+#if defined(_MSC_VER) && defined(_HAS_CXX17)
+	#define JSON_USE_STRING_VIEW
+#endif
+
+#if defined(JSON_USE_STRING_VIEW)
+	#include <string_view>
+#endif
+
+
 /*!
 @brief namespace for Niels Lohmann
 @see https://github.com/nlohmann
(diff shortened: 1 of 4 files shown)


test/src/unit-class_parser.cpp
```diff
@@ -45,7 +45,8 @@ json parser_helper(const std::string& s)
     // if this line was reached, no exception ocurred
     // -> check if result is the same without exceptions
     json j_nothrow;
-    CHECK_NOTHROW(json::parser(nlohmann::detail::input_adapter(s), nullptr, false).parse(true, j_nothrow));
+    std::unique_ptr<json::exception> ex;
```
theodelrieu: What is the advantage of a `unique_ptr` over `std::exception_ptr`?

src/json.hpp
```diff
@@ -2968,20 +2994,35 @@ class parser
     @throw parse_error.101 in case of an unexpected token
     @throw parse_error.102 if to_unicode fails or surrogate error
     @throw parse_error.103 if to_unicode fails
+    @throw out_of_range 406 if float not finite
     */
     void parse(const bool strict, BasicJsonType& result)
     {
+        std::unique_ptr<exception> ex;
+        parse(strict, result, ex);
+        try_throw_any_exception(ex.get());
```
theodelrieu: `std::rethrow_exception` could be used here (assuming we replace `ex` with a `std::exception_ptr`.

src/json.hpp
```diff
@@ -466,16 +490,51 @@ caught.,other_error}
...
   private:
-    other_error(int id_, const char* what_arg) : exception(id_, what_arg) {}
+    static std::string format_what(int id_, const std::string& what_arg)
+    {
+        return exception::name("other_error", id_) + what_arg;
+    }
 };
 
+/*!
+@brief helper function to throw the a downcasted exception.
```
gregmarr: Typo: "throw the a downcasted" too many articles

src/json.hpp
```diff
@@ -466,16 +490,51 @@ caught.,other_error}
...
+    Exception* down = dynamic_cast<Exception*>(ex);
+    if(down != nullptr)
+    {
+        JSON_THROW(*down);
+    }
+#else
+    ex = ex; // suppress unreferenced formal parameter warnings
+    std::abort();
+#endif
+}
```
gregmarr: I think this looks cleaner

    #ifdef JSON_EXCEPTIONS_ENABLED
    template<class Exception>
    void try_throw_exception(exception* ex)
    {
        // Only use RTTI if necessary (exceptions are enabled). This way
        // can be compiled with no RTTI.
        if(Exception* down = dynamic_cast<Exception*>(ex))
        {
            JSON_THROW(*down);
        }
    }
    #else
    template<class Exception>
    void try_throw_exception(exception* /*unused*/)
    {
        std::abort();
    }
    #endif

On the other hand, why is this even needed?  (It's definitely not if @theodelrieu 's suggested change is made, and I would say it's not even if it isn't, just throw `ex.get()`, no need for the downcasting.

test/src/unit-class_parser.cpp
gregmarr: 1. `std::exception_ptr` isn't limited to `json::exception` types.
2. The only things you can do with an `std::exception_ptr` are convert to bool, compare to another pointer, and rethrow.  You can't actually get the exception object out of it, so all the caller would be able to do is determine whether or not an exception was thrown, not what that exception was.

theodelrieu: Right, there's no way to retrieve the exception, forgot about that.

I do think we should not use `std::unique_ptr<json::exception>` though, but `std::error_code` instead (this [blog post](https://akrzemi1.wordpress.com/2017/07/12/your-own-error-code/) is really useful on this topic).

`std::filesystem` uses `std::error_code` for the `noexcept` overloads, one can still embed a lot of information with this facility. I've seen it mentioned in [Outcome's documentation](https://ned14.github.io/outcome/tutorial/error_code) too.

What do you all think?

gregmarr: I don't work without exceptions, so I don't really care one way or another.  I think the best solution in the end is probably to return an `std::expected` or `std::outcome` or whatever it ends up as, but that's years away.

src/json.hpp
jseward: ah interesting thanks! I actually have never heard of `std::exception_ptr` before. Unfortunately it looks like it isn't very useful in this situation. Specifically I want to be able to get at the message on why the parse failed.

test/src/unit-class_parser.cpp
jseward: I agree that long term using `std::expected` or `std::outcome` would be best. My original plan was to use `std::error_code` (as you can see by my branch name) but it has it's own issues.  Mostly creating a custom formatted error message with context does not seem cleanly possible. You can see in the blog post linked (which I used as a source as well) that aspect is glossed over.

src/json.hpp
jseward: Agreed, that does look cleaner thanks. I'll try it out. As for just throwing without downcasting I couldn't get that to work. `catch (const SpecificError& e) {}` would never get hit.

test/src/unit-class_parser.cpp
theodelrieu: Indeed you cannot have context in the `error_code` (e.g. line of the parsing failure).

However I don't think that's relevant, we are used to this already. Trying to copy an non-existing file with `std::filesystem` gives the same behavior:

exception:
> what: filesystem error: cannot copy: No such file or directory [asdfasd] [asdfasdfa2] 

error_code:
> No such file or directory

We will not have the line number or very precise information like that, but @nlohmann already documented all the exceptions thrown by the library, with a code and example message, which are IMO more than enough!

It should be quite easy to map each exception type to one enum/error category, I believe we should also add the `error_code` as a member of the exception.

src/json.hpp
gregmarr: Yeah, I missed that you were throwing by value instead of by pointer, which unfortunately does require all this extra casting over throwing directly when the exception is created.  This delayed throw method does lose some of the call stack context, so it may be better to make the choice whether to throw or place in the unique_ptr<> at the point of the THROW macro.

test/src/unit-class_parser.cpp
jseward: I guess I would argue that losing context on json parsing errors would be quite a step backwards. Whereas not a loss for the filesystem examples as the extra context can be inferred by the calling parameters.

Another option could be yet another overload taking in another parameter `std::string &error_location` or combine both std::error_code and std::string error_location into one object.

theodelrieu: Right now `std::abort` is called, so I don't think this is a step backward :p. If I understand correctly, one must use `dynamic_cast` to down cast the output parameter `unique_ptr<json::exception>`.

If so, that's really a big issue, usually people that turn off exceptions also turn off RTTI. Not even mentioning how painful this is to use, nor the cost of heap allocation each time the `unique_ptr` gets constructed.

I don't see how we could combine an `error_location` parameter with the `error_code`, the former being relevant only in a few cases (e.g. parse errors). Maybe a `extra_detail` field?
Even that would be sub-optimal, as the error type would bear the size of an extra string, even when it's not used.

On my machine it goes from 16 bytes (error_code) to 48 (error_code + string).

I would suggest trying to implement the `error_code` version to see how it goes, we might find the messages detailed enough!

jseward: The thing is that the new overload does not call `std::abort` at all. It just passes the exception with the context back up to the caller.

No RTTI is required either. The virtual `what()` function dumps the context of the parse error with full context for the caller.

That is what I'm referring to as a step backwards.

nlohmann:
Hey there. I am a bit confused about this PR. Can someone please provide a brief explanation on the benefits?

test/src/unit-class_parser.cpp
theodelrieu: Sorry, I misread the latest commits.

I did not realize that the definition of `json::exception` depends on whether or not the exceptions are disabled globally. This is very misleading, if someone wants to use the nothrow version (`noexcept` should be added by the way) with exceptions enabled, they will not have the same `json::exception` struct than someone without exceptions. One should have the choice to use the throw or nothrow version in the code and not through compiler flags.

You could argue that this is not a big deal, since the data member is not exposed (it does change the size of the struct though). Plus, having a parameter of type `json::exception` which is not an exception feels strange.

What I consider a step backwards is the loss of the ability to switch on the error type.
With exceptions, it can still be done with multiple `catch`, or by downcasting.
With `std::error_code`, a simple `switch`, or an approach as discussed in [this paper](http://open-std.org/JTC1/SC22/WG21/docs/papers/2017/p0824r0.pdf) can be used.

Here, users cannot do anything, they have a `std::string`. They might parse it, but frankly that's unusable. The only advantage that I see is for debugging purposes, but it's not possible to take a decision at runtime based on this approach.

I would recommend the reading of the aforementioned paper, especially the part about interoperability with four libraries using `std::error_code`. Also, this mechanism is used in the latest Couroutine/Networking TS, adopting it seems future proof to me.

Let's not forget this library is likely to be used in/with other libraries:
For example, [Boost.Beast](https://github.com/boostorg/beast) uses `error_code` extensively, if someone wants to wrap it with this library (e.g. to parse the json body of an HTTP request), having both libraries using/providing `std::error_code` is quite helpful.

I'm far from being an expert in Beast, though @vinniefalco stated in [this issue](https://github.com/nlohmann/json/issues/774#issuecomment-339323015) that one could use this library to implement a Beast `Body` container (assuming incremental parsing is supported). I might be wrong but it seems that having `error_code` would simplify the implementation of `BodyReader/BodyWriter` concepts too.

jseward: I definitely see where you are coming from @theodelrieu and pretty much agree with every one of your points. I appreciate the links.

The issue is in regards to a specific, but I would argue potentially popular, scenario though. That is using this library for parsing json and getting nice context aware error messages on failures even if exception handling is completely disabled. (most common in game engines, embedded)

Being able to handle various error codes / categories at runtime for parse errors is not important here, and I can't see the ability to do so being useful other than a theoretical exercise.

@nlohmann does the above description answer your question on why this PR exists?

nlohmann: So basically all you want to have is access to parser diagnostics (error message + location) when exceptions are switched off?

jseward: yes

theodelrieu: @jseward I understand the use-case, I think a compromise can be reached.

The good solution would be [outcome](https://github.com/ned14/outcome), you can indeed attach an extra payload to it. Unfortunately, I do not think we can use it in this library.

However, we could have the following type:

```cpp
// not a good name (std::system_error is used as an exception).
struct json_error {
  std::error_code ec;
  std::string context; // empty if ec == 0, or no context provided.
};
```

This type would replace the `std::unique_ptr<json::exception>` in this PR. Plus, the `#ifdef` inside `json::exception` definition can be removed.

I would also suggest adding an `error_code` to `json::exception` too (for the same reasons mentioned before in this thread, similarly to `std::filesystem_error`).

What do you all think?

nlohmann: We already support parsing without exceptions (and return a `discarded` type on parse errors). I think the desired outcome could be achieved by just adding the required member functions to the parser class to query the required diagnostics in case an error occurs. The same functions could be re-used in the code that creates the actual `parse_error` exception so we would not even need to add too much code.

vinniefalco:
LOL wtf? A stale bot?

PR #819, opened 2017-11-05
Adds get_to

arcanis:
This PR adds a `get_to` method that does exactly the same thing as `get`, except that it doesn't create a new value to store the final result (it takes a reference instead).

It's a convenience function that can be easily implemented in userland (which is also true for `get`), but I felt like it was simple and useful enough to be added into the core.

A few things have to be fixed before this PR can be merged, I plan to work on them if this API is accepted:

  - [ ] Fixes the documentation
  - [ ] Adds tests

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -9580,6 +9580,59 @@ class basic_json
     }
 
     /*!
+    @brief get a value and store it into an existing value
+
+    Explicit type conversion between the JSON value and a compatible value.
+    Because the function takes a reference as parameter, the value type doesn't have to be
+    [CopyConstructible](http://en.cppreference.com/w/cpp/concept/CopyConstructible) nor
+    [DefaultConstructible](http://en.cppreference.com/w/cpp/concept/DefaultConstructible).
+    The value is converted by calling the @ref json_serializer<ValueType>
+    `from_json()` method.
+
+    The function is equivalent to executing
+    @code {.cpp}
+    ValueType my_value;
+    JSONSerializer<ValueType>::from_json(*this, my_value);
+    @endcode
+
+    @tparam ValueTypeCV the provided value type
+    @tparam ValueType the returned value type
+
+    @return same reference as the one passed in parameter
+
+    @throw what @ref json_serializer<ValueType> `from_json()` method throws
+
+    @liveexample{The example below shows several conversions from JSON values
+    to other types. There a few things to note: (1) Floating-point numbers can
+    be converted to integers\, (2) A JSON array can be converted to a standard
+    `std::vector<short>`\, (3) A JSON object can be converted to C++
+    associative containers such as `std::unordered_map<std::string\,
+    json>`.,get__ValueType_const}
+
+    @since version 2.1.0
+    */
+    template <
+        typename ValueTypeCV,
+        typename ValueType = detail::uncvref_t<ValueTypeCV>,
+        detail::enable_if_t <
+            not std::is_same<basic_json_t, ValueType>::value and
+            detail::has_from_json<basic_json_t, ValueType>::value and
+            not detail::has_non_default_from_json<basic_json_t, ValueType>::value,
+            int > = 0 >
+    ValueType & get_to(ValueType & target) const noexcept(noexcept(
+                                       JSONSerializer<ValueType>::from_json(std::declval<const basic_json_t&>(), std::declval<ValueType&>())))
+    {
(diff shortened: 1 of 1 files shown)


arcanis:
If you don't like the idea of adding a new member name, the API can also look like this:

```
ValueType val;
json.get<ValueType&>(val);
```

nlohmann:
What is the difference to the existing `get_ref` method?

arcanis:
Unless I got it wrong, `get_ref` only works for underlying value types. `get_to` would work with anything, going through the whole `from_json` pipeline, just like `get`.

src/json.hpp
```diff
@@ -9580,6 +9580,59 @@ class basic_json
...
+            detail::has_from_json<basic_json_t, ValueType>::value and
+            not detail::has_non_default_from_json<basic_json_t, ValueType>::value,
+            int > = 0 >
+    ValueType & get_to(ValueType & target) const noexcept(noexcept(
+                                       JSONSerializer<ValueType>::from_json(std::declval<const basic_json_t&>(), std::declval<ValueType&>())))
+    {
+        // we cannot static_assert on ValueTypeCV being non-const, because
+        // there is support for get<const basic_json_t>(), which is why we
+        // still need the uncvref
+        static_assert(not std::is_reference<ValueTypeCV>::value,
```
theodelrieu: Might want to check that `ValueType` is not const?

theodelrieu:
@arcanis I would personally go for an overload of `get`, I don't find `get_to` expressive enough.

nlohmann:
I do not like the name `get_to`. But before we start to bikeshed, can you please provide some example usages?

arcanis:
Sure! The following snippet showcases the two main use cases I see, which is deserializing a value without overwriting the existing members if the corresponding json fields are missing.

```c++
#include <iostream>
#include "./json.hh"

struct Foobar
{
    int value;
};

void from_json(nlohmann::json const & json, Foobar & val)
{
    auto valueIt = json.find("value");

    if (valueIt != json.end()) {
        valueIt->get_to<decltype(val.value)>(val.value);
    }
}

int main(void)
{
    Foobar foobar;

    nlohmann::json json1 = nlohmann::json::object_t({
        { "value", 42 },
    });

    json1.get_to<Foobar>(foobar);
    std::cout << foobar.value << std::endl;

    nlohmann::json json2 = nlohmann::json::object_t({
        { "not a value", 66 },
    });

    json2.get_to<Foobar>(foobar);
    std::cout << foobar.value << std::endl;
}
```

arcanis:
How is it different? Frankly, the only thing this function do is to make it easier for users to specify their own target object where to deserialize a value instead of always creating a new one.

It seems like a nice way to solve both the partial deserialization and the default construction issues, with a simple and backward compatible API.

src/json.hpp
arcanis: Yup, I'll revisit the asserts once we agree on the API :)

gregmarr:
It's the "deserializing a value without overwriting the existing members" bit that threw me.  It's not quite what I was thinking at first since it's only if the json values are missing, and I was thinking it was if the object fields was already set, which is not what you were saying it all.

theodelrieu:
On a second thought, it seems users could simply call `from_json(j, existing_object)` to get the desired behavior.

theodelrieu:
@arcanis I'm self-bumping my last post but, what do you think about the following:

```cpp
std::string existing_str;
auto j = nlohmann::json::parse(...);
nlohmann::from_json(j, existing_string);
```

This has the benefit of not adding a new API. I'm not against the principle, but it seems to me that the lower level API alreary covers your use-case.

nlohmann:
I agree with @theodelrieu. Unless there is a compelling reason to add a new function to the API, I would rather (re)use the `from_json` function.

nlohmann:
Any further opinion on this? Otherwise I would close this PR.

arcanis:
I can't check right now; would this syntax also work with custom JSON data types? For example in my app I actually do something like `using Json = nlohmann::basic_json<...>`.

I guess syntax-wise I still like `get_to` (or just `get`) better because they look complementary with the already existing `get` and `get_ref` functions, but your call.

arcanis:
> I can't check right now; would this syntax also work with custom JSON data types? For example in my app I actually do something like `using Json = nlohmann::basic_json<...>`.

Seems to work fine. Going to close this PR.

Issue #821, opened 2017-11-06
Code Checking Warnings from json.hpp on VS2017 Community


AraHaan:
**Bug Report**

- What is the issue you have?
A lot of warnings from code checking in Visual Studio 2017.

- Please describe the steps to reproduce the issue. Can you provide a small but working code example?

- What is the expected behavior?
No annoying warnigns from json.hpp

- And what is the actual behavior instead?
Warnings right after another on json.hpp.

- Which compiler and operating system are you using? Is it a [supported compiler](https://github.com/nlohmann/json#supported-compilers)?
The latest Visual Studio 2017 Community compiler.

- Did you use a released version of the library or the version from the `develop` branch?
Release I think of version 2.0.7

- If you experience a compilation error: can you [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests)?

Here are the warnings that are emmited on json.hpp when including and using it:
```
j:\14.4 gb usb\vbtrpdet\libs\json.hpp(10768): warning C26481: Don't use pointer arithmetic. Use span instead. (bounds.1: http://go.microsoft.com/fwlink/p/?LinkID=620413)
j:\14.4 gb usb\vbtrpdet\libs\json.hpp(2204): warning C26409: Avoid calling new and delete explicitly, use std::make_unique<T> instead. (r.11 http://go.microsoft.com/fwlink/?linkid=845485)
j:\14.4 gb usb\vbtrpdet\libs\json.hpp(2223): warning C26424: Failing to delete or assign ownership of allocation at line 2204.
j:\14.4 gb usb\vbtrpdet\libs\json.hpp(6883): warning C26499: Could not find any lifetime tracking information for '**this'
j:\14.4 gb usb\vbtrpdet\libs\json.hpp(6877): warning C26499: Could not find any lifetime tracking information for '**this'
j:\14.4 gb usb\vbtrpdet\libs\json.hpp(6846): warning C26499: Could not find any lifetime tracking information for '**this'
j:\14.4 gb usb\vbtrpdet\libs\json.hpp(6840): warning C26499: Could not find any lifetime tracking information for '**this'
j:\14.4 gb usb\vbtrpdet\libs\json.hpp(6398): warning C26499: Could not find any lifetime tracking information for '*return of ->'
j:\14.4 gb usb\vbtrpdet\libs\json.hpp(6400): warning C26499: Could not find any lifetime tracking information for '*return of ->'
[...]

PR #822, opened 2017-11-06
Fix Compile warning in Visual Studio

AraHaan:
In issue #821 I found that on the latest (since version 2.0.7) there are now min and max warnings. This PR now removes those warnings on the source and hopefully should work on all platforms without degrading anything. This only disables the min and max macros if defined.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -120,16 +120,14 @@ SOFTWARE.
 In Visual Studio 2017 there are warnings on min
 and max in this header having not enough args.
 This is because in C min and max take args and
-wll are not needed anyway so we undefine them
+well are not needed anyway so we undefine them
 first to silence these annoying warnings!!!
 */
-#if (_MSC_VER >= 1911)
 #ifdef min
 #undef min
 #endif
 #ifdef max
 #undef max
-#endif
 #endif
 
 /*!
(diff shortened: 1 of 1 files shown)


AraHaan:
In issue #821 I found that on the latest (since version 2.0.7) there are now min and max warnings. This PR now removes those warnings on the source and hopefully should work on all platforms without degrading anything.

src/json.hpp
```diff
@@ -116,6 +116,19 @@ SOFTWARE.
...
     #define JSON_HAS_CPP_14
 #endif
+/*
+In Visual Studio 2017 there are warnings on min
+and max in this header having not enough args.
+This is because in C min and max take args and
+well are not needed anyway so we undefine them
+first to silence these annoying warnings!!!
+*/
+#ifdef min
```
nlohmann: I don't see the point of undefining a symbol that was not defined by the library. This feels like cleaning up after MSVC...

AraHaan: It only happens in MSVC (1911) for some reason yes.

AraHaan:
Just realized I could have just did an global define of ``NOMINMAX``.

PR #844, opened 2017-11-26
Add .natvis for MSVC debug view

TinyTinni:
This PR adds a .natvis improving the debug view of `nlohmann::basic_json` in the MSVC GUI.
Also, included the ..natvis file into CMake install process.

The supported container types for the template arguments `object `and `array` are generic and the container's .natvis will be shown (if declared).

Here is an example view from the readme test (MSVC2017): [](https://postimg.org/image/a01qe113p/)

There is also a name bug in the CMakeLists for tests. I did not created a second issue, since it's just the name-prefix which is missing.
 I can create a new pull request if you want or you can fix it on your own. (its the first commit).


nlohmann:
Just out of curiosity: does the .natvis file must be put into the src directory?

TinyTinni:
No, the .natvis can be everywhere. I.e. the outcome library from ned14 puts it into include dir, GSL from Microsoft puts it into root.

TinyTinni:
Checked the build error. Its only on one Mac machine.
I guess the error comes from the cmake version. On the machine, cmake v3.2.2 is running, where sources are not getting exported:

[> Targets with INTERFACE_SOURCES may not be exported with the export() or install(EXPORT) commands. This limitation may be lifted in a future version of CMake.](https://cmake.org/cmake/help/v3.2/command/target_sources.html)

This restriction is gone with [cmake v3.3](https://cmake.org/cmake/help/v3.3/command/target_sources.html).

possible ad-hoc solutions:
- adding cmake version and/or compiler check before adding the sources
- increase minimum required cmake version (regarding to the copyright, cmake 3.2 was current in 2015)

TinyTinni:
added compiler & cmake version check.

For non MSVC-Version, the .natvis file is not needed and I guess it's a good idea no not install it on these system. When I already implemented a condition, I see no problem in adding a version check.

src/json.natvis
```diff
@@ -0,0 +1,32 @@
...
+    <!--    skip the pair first/second members in the treeview while traversing a map. 
+            Only works in VS 2015 Update 2 and beyond using the new visualization -->
+    <Type Name="std::pair&lt;*, nlohmann::basic_json&lt;*&gt;&gt;" IncludeView="MapHelper">
+        <DisplayString>{second}</DisplayString>
+        <Expand>
+            <ExpandedItem>second</ExpandedItem>
+        </Expand>
+    </Type>
+    
+</AutoVisualizer>
```
nlohmann: Could you please move this file to the root? I do not want to confuse people that have no idea what a `.natvis` file is. I may clean up the repo and come up with a `misc` folder or something, but right now, the root would be fine.

TinyTinni: Sure, 
had to edit the install procedure for it.
Its now in the root dir and will be installed to the root dir /install-prefix dir.

nlohmann:
Thanks a lot!

PR #848, opened 2017-11-30
 simple .each/.forEach extension

sfinktah:
This is more of a nascent idea, than a ready-to-pull addition.  Although it did work perfectly for me, I am not at all sure it would behave well while iterating through a container of mixed data types.

The idea is pretty simple... I do some work on an underscore/c++ fork https://github.com/sfinktah/lodash-cpp which implements the `.each` method (badly).

I am going to assume you are familiar with underscore/lodash/JavaScript's .forEach, since JSON derived from JavaScript, but I could be wrong.

I'm not at all sure you wish to extend your library in this direction, I only did so because it was beyond my ability to properly handle JSON containers in an external library.

This is what I wanted to achieve, and it worked perfectly for what I required.

```cpp
        // json j;
        // std::vector<std::string> loadedIpls;

        if (j["unloadScenery"].is_array()) {
            j["unloadScenery"].each([&](std::string ipl, auto, auto) {
                if (IS_IPL_ACTIVE(ipl))
                    REMOVE_IPL(ipl);
                _::pull(loadedIpls, ipl);
                _::each(loadedIpls, [](std::string ipl) { REMOVE_IPL(ipl); });
            });
        }
```

To me (eye of the beholder) this is beautiful code, so very close to the simplicity of manipulating arrays in JavaScript (using underscore/lodash for extra "simplicity"), expressed succinctly in C++.

If you're not familiar with underscore/lodash, then this probably look horrendous. 

The `.each` method I propose also handles `objects` (associative containers) with the same ease, something which JavaScript's internal `.forEach` function never managed, but could be achieved with underscore/lodash.

I'm not sure how it will fare with your CI tests, and obviously it's not worth me pursuing the possibly issues to achieve full functionality, if this simply isn't something you want to add to your library.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -279,3 +279,4 @@ script:
     brew install nlohmann_json --HEAD ;
     brew test nlohmann_json ;
     fi
+
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -13659,6 +13659,59 @@ class basic_json
     }
 
     /*!
+    @brief execute a provided function once for each item in a container
+
+    Iterates over a list of elements, yielding each in turn to an iteratee function. The iteratee is bound to the context object, if one is passed. Each invocation of iteratee is called with three arguments: (element, index, list). If list is a JavaScript object, iteratee's arguments will be (value, key, list). Returns the list for chaining.
+
+    @param iteratee  Callable object accepting 3 arguments
+    @param iteratee.value The value of the current item
+    @param iteratee.key The key/index of the current item
+    @param iteratee.object The JSON container being iterated
+
+    @throw type_error.302 if not is_object() or not is_array()
+
+    @complexity Linear in the size the JSON container.
+
+    @example j["list"].each([](auto key, auto value, auto& object){ object[key] = std::to_string(value); });
+
+    @since never
+    */
+    template <typename Function>
+    void each(Function function)
+    {
+        basic_json& value = *this;
+        switch (value.m_type)
+        {
+            case detail::value_t::array:
+            {
+                // iterate array and use index as reference string
+                for (std::size_t i = 0; i < value.m_value.array->size(); ++i)
+                {
+                    function(value.m_value.array->operator[](i), i, value);
+                }
+            }
+            break;
+
+            case detail::value_t::object:
+            {
+                // iterate object and use keys as reference string
+                for (const auto& element : *value.m_value.object)
+                {
+                    function(element.second, element.first, value);
+                }
+                break;
+            }
+
+            default:
+            {
+                // maybe throw an error
+                JSON_THROW(type_error::create(302, "type must be array or object, but is " + std::string(value.type_name())));
+                break;
+            }
+        }
+    }
+
+    /*!
     @brief return flattened JSON value
 
     The function creates a JSON object whose keys are JSON pointers (see [RFC


trilogy-service-ro:

nlohmann:
How is this different to C++'s http://en.cppreference.com/w/cpp/algorithm/for_each?

sfinktah:
I'm glad you asked.

std::for_each only passes the **value**, it doesn't pass the **key**.

It is possible with STL associative containers to obtain the key via 

```cpp
        for (std::pair<std::string, std::string> i : container) {
            // i.first;  i.second
        }
```

but that will cause a run time error if you attempt it on a json array/object.

There was no generic (read; STL compliant) method of iterating the keys in a json container that I could determine.

sfinktah:
If you're not sold on the entire `.each` idea, I could get by with a way to list all the **keys** in an associative container, then I could iterate by key name.

gregmarr:
Have you looked at `iterator_wrapper`, it should do what you need.

sfinktah:
I have looked at iterator_wrapper, but as far as I can tell, it returns an iterator_wrapper_internal which has ```cpp
        /// return key of the iterator
        std::string key() const
```
defined, and as such does not lend itself to a generic STL solution.

You may have ready my unrelated 'issue' (whine) regarding how zen::xml just isn't up to the nlohmann::json standard.  It does however offer one option that I put to use to solve the equivalent problem.

```cpp
                auto itpair = nodeVehicle->getChildren();
                _::each_key_value(itpair.first, itpair.second, [&](auto, auto, auto) {});
```

nlohmann:
With the `iteration_wrapper`, you can use any STL algorithm which relies on iterator pairs. And inside the passed function, you can call `key()` and `value()` to access the parts you need. I am confident that whatever usecase `each`/`forEach` can handle, an STL algorithm with `iteration_wrapper` can also do.

nlohmann:
I understand the idea behind this PR, but it feels strange in a C++ world. As the iteration wrapper provides everything needed to solve `.forEach` and `.each` scenarios, the PR would not add new functionality.

sfinktah:
> With the iteration_wrapper, you can use any STL algorithm which relies on iterator pairs. And inside the passed function, you can call key() and value() to access the parts you need.

It's the `key()` and `value()` parts which cause the issue.

Lets put my PR aside.  The only reason I wrote it, was because I couldn't iterate json keys in my lodash-cpp library using standard STL methods.  This may simply be a reflection of my ineptitude in template specialization, which is limited to about the level of `std::enable_if_has_type<typename T::value_type>`, and comparing the type to a nlohmann::json container isn't going to end well if your library isn't present.
 
I have already mentioned that the well known method for obtaining keys from STL associative containers `for (std::pair... )` won't work, but here is another method that was suggested to me by a far more capable programmer.

I've rendered it as a self-contained test.  I can put it into your `issues` queue if you like, it's possible that the solution may be simpler than attempting to make the internal_iterator compliant with `.first` and `.second`

```cpp
#include <map>
#include <string>
#include <vector>
#include <algorithm>
#include "json.hpp"

using json = nlohmann::json;

template <typename ResultContainer, typename Container, typename Function>
ResultContainer map(const Container& container, Function&& function) {
    ResultContainer result;

    std::transform(std::begin(container), std::end(container),
                   std::back_inserter(result), std::forward<Function>(function));
    return result;
}

template <typename ResultContainer, std::size_t I, typename Container>
ResultContainer get_item(const Container& container) {
    return map<ResultContainer>(container, [ ](auto & value) {
        return std::get<I>(value);
    });
}

template <typename ResultContainer, typename Container>
ResultContainer tuple_keys(const Container& container) {
    return get_item<ResultContainer, 0>(container);
}

int main(int argc, const char **argv) {
    std::map<std::string, std::string> m = {
        {"a", "x"},
        {"b", "y"},
        {"c", "z"}
    };

    tuple_keys<std::vector<std::string>>(m);

    json j(m);
    // will not compile with this line uncommented
    // tuple_keys<std::vector<std::string>>(j);
}
```

sfinktah:
Same with this, FYI.

```cpp
    template <typename ResultContainer, typename Container>
    ResultContainer tuple_values(const Container& container)
    {
        return get_item<ResultContainer, 1>(container);
    }
```

theodelrieu:
The issue here is that the library doesn't specialize `std::get<N>`. Note that it cannot specialize `std::tuple_size`, since the size is not known at compile time.

Even if we define `std::get<nlohmann::basic_json<...>>`, which type should it return? I would expect `std::get` to only work on JSON arrays, not objects.

I don't really see the benefit of having a new function in `basic_json` which would only work for objects (and return an empty container for every other JSON type).

You could add a small utility function `json_keys` in your library which would do such a thing:

```cpp
#include <map>
#include <string>
#include <vector>
#include <iostream>
#include <algorithm>
#include "json.hpp"

using json = nlohmann::json;

int main(int argc, char const *argv[])
{
  json j{{"Zero", 0}, {"One", 1}, {"Two", 2}};
  auto wrapper = json::iterator_wrapper(j);
  auto const begin = wrapper.begin();
  auto const end = wrapper.end();

  std::vector<std::string> keys;
  std::transform(begin, end, std::back_inserter(keys),
                 [](auto elem) { return elem.key(); });
  for (auto const& elem : keys)
  {
    std::cout << elem << std::endl;
  }
}
```

sfinktah:
I'm going to try a different tack.

There's a section in the README: https://github.com/nlohmann/json#conversion-from-stl-containers

But there's no section for "conversion **to** stl containers".

Here's what that section could look like:

```cpp
// converting to a std::map
std::map<std::string, int> m;
j.each( [&] (const std::string& value, const std::string& key) {
    m.emplace({value, key});
};

// converting to a std::vector (silly way)
std::vector<int> m;
j.each( [&] (int value, auto&) {
    m.push_back(value);
};

// converting to a std::vector (using the key as the index)
std::vector<int> m;
j.each( [&] (int value, int key) {
    m[key] = value; 
};
```

and, here's one from the a project I'm working on:
```cpp
if (j.is_object()) {

    // existing code
    for (auto it = j.begin(); it != j.end(); ++it) {
        uint64_t key = std::stoull(it.key());
        json& values = it.value();
        for (auto it2 = values.begin(); it2 != values.end(); ++it2) {
            uintptr_t value = *it2;
            g_hints.insert(std::make_pair(key, GTA5().base() + value));
        }
    }

    // what it **could** look like
    j.each([&](const json& values, int _key) {
        uint64_t key = std::stoull(_key);
        values.each([&](uintptr_t value, auto) {
            g_hints.insert({ key, GTA5().base() + value));
        });
    };

    // what it **could** look like in a perfect world
    // if the library would obligue by converting the key
    j.each([&](const json& values, int key) {
        values.each([&](uintptr_t value, auto) {
            g_hints.insert({ key, GTA5().base() + value));
        });
    };
}
```

sfinktah:
Full disclosure: I actually tested my real-life example, and it required a few small modifications to the PR, and to the code.  This is the actual functional code.

```cpp
                        j.each([&](json& values, json& _key) {
	   					  uint64_t key = std::stoull(_key.get<std::string>());
                            values.each([&](uintptr_t value, auto) {
                                g_hints.insert({ key, GTA5().base() + value });
                            });
                        });
```

gregmarr:
What about this?  It looks very similar and requires no modification to the library.  I've compiled this on Compiler Explorer but not run it.

    auto jw = json::iterator_wrapper(j);
    std::for_each(jw.begin(), jw.end(), [&](auto &x) {
        uint64_t key = std::stoull(x.key());
        std::for_each(x.value().begin(), x.value().end(), [&](auto &value) {
            g_hints.insert({ key, GTA5().base() + value });
        });
    });

gregmarr:
Here is your "conversion to stl containers" section with current `develop` code, with a fix to the last one to make it work:

    // converting to a std::map
    std::map<std::string, int> m = j;

    // converting to a std::vector
    std::vector<int> m = j;

    // converting to a std::vector (using the key as the index)
    std::vector<int> m;
    auto jw = json::iterator_wrapper(j);
    std::for_each(jw.begin(), jw.end(), [&](auto &x) {
        auto key = std::stoull(x.key());
        if(key >= m.size()) {
            m.resize(key + 1);
        }
        m[key] = x.value(); 
    };

nlohmann:
I shall have a further look at this after the 3.0.0 release.

nlohmann:
I think the PR should not be merged for the reasons stated in https://github.com/nlohmann/json/pull/848#issuecomment-348959389.

Thanks a lot for the effort though!

sfinktah:
I want to revisit this general concept now that we're past the 3.0.0 milestone.

This time I'm looking at this:

```cpp
        for (auto& [x, y] : an_stl_map) {} // happy compiler
        for (auto& [x, y] : j) {} // sad compiler
        // 'nlohmann320::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann320::adl_serializer>': 
        // cannot decompose type with non - public members
```

nlohmann:
@sfinktah I see the issue is still not fixed. Any ideas?

sfinktah:
I've decided to learn all about iterators.

It seems somewhat unfair to request that something be added, if I'm not prepared to put in the time to research it myself.    Especially after seeing how much code (and tickets) were involved in the creation of our current iterator.

I actually still use that custom `.each()` function I wrote, which has proved quite versatile.  The trick now is to replicate that versatility in an iterator.

```cpp
        // Example of json.each
        j.each([&](json& values, json& key) {
            values.each([&](uintptr_t value, auto) {
                multimap.insert({key, value});
            });
        });
```

Note how the values are received as `json` in the outer loop, but are implicitly converted to `uint64_t` in the inner loop?

That's what I'm hoping to get out of a `std::pair` (`std::tuple`?) iterator::value_type... otherwise, I don't really think it's worth the effort.

Issue #62, opened 2015-04-25
Windows/Visual Studio (through 2013) is unsupported


smiley:
I was about to use this for a C++ project in Visual Studio 2013, but the header fails compilation entirely, thanks to a myriad of errors:

```
1>------ Build started: Project: jsondummy, Configuration: Debug Win32 ------
1>  main.cpp
1>d:\github\jsondummy\src\json.hpp(137): error C2061: syntax error : identifier 'iterator'
1>          d:\github\jsondummy\src\json.hpp(4893) : see reference to class template instantiation 'nlohmann::basic_json<ObjectType,ArrayType,StringType,BooleanType,NumberIntegerType,NumberFloatType,AllocatorType>' being compiled
1>d:\github\jsondummy\src\json.hpp(137): error C2238: unexpected token(s) preceding ';'
1>d:\github\jsondummy\src\json.hpp(143): error C2061: syntax error : identifier 'const_iterator'
1>d:\github\jsondummy\src\json.hpp(143): error C2238: unexpected token(s) preceding ';'
1>d:\github\jsondummy\src\json.hpp(149): error C2061: syntax error : identifier 'reverse_iterator'
1>d:\github\jsondummy\src\json.hpp(149): error C2238: unexpected token(s) preceding ';'
1>d:\github\jsondummy\src\json.hpp(155): error C2061: syntax error : identifier 'const_reverse_iterator'
1>d:\github\jsondummy\src\json.hpp(155): error C2238: unexpected token(s) preceding ';'
1>d:\github\jsondummy\src\json.hpp(382): error C3646: 'noexcept' : unknown override specifier
1>d:\github\jsondummy\src\json.hpp(385): error C3646: 'noexcept' : unknown override specifier
1>d:\github\jsondummy\src\json.hpp(402): warning C4346: 'std::is_constructible<ObjectType<StringType,nlohmann::basic_json<ObjectType,ArrayType,StringType,BooleanType,NumberIntegerType,NumberFloatType,AllocatorType>,>::key_type,V::key_type>::value' : dependent name is not a type
1>          prefix with 'typename' to indicate a type
1>d:\github\jsondummy\src\json.hpp(402): error C2146: syntax error : missing ',' before identifier 'and'
1>d:\github\jsondummy\src\json.hpp(402): error C2065: 'and' : undeclared identifier
[...]

Issue #61, opened 2015-04-24
Complete brief documentation


nlohmann:
The following functions are not yet documented:

- [ ] type
- [ ] push_back
- [ ] operator+=
- [ ] swap
- [ ] find
- [ ] count
- [ ] dump
- [ ] parse
- [ ] operator==, operator!=, operator<, operator<=, operator>, operator>=
- [ ] std::swap
- [ ] std::hash
- [ ] operator<<, operator>>
- [ ] operator"" _json

nlohmann:
Closed this issue: The documentation is no longer maintained in the Wiki, but as Doxygen documentation here: http://nlohmann.github.io/json

PR #858, opened 2017-12-05
:wrench: Fix up a few more effc++ items

mattismyname:
test suite has these two failures, but the same tests fail even before this change:

21/70 Test #62: test-regression_all .................***Failed   12.56 sec
29/70 Test #61: test-regression_default .............***Failed   12.91 sec

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -3526,7 +3526,7 @@ class primitive_iterator_t
         return *this;
     }
 
-    primitive_iterator_t operator++(int)
+    primitive_iterator_t const operator++(int)
     {
         auto result = *this;
         m_it++;
@@ -3539,7 +3539,7 @@ class primitive_iterator_t
         return *this;
     }
 
-    primitive_iterator_t operator--(int)
+    primitive_iterator_t const operator--(int)
     {
         auto result = *this;
         m_it--;
@@ -3850,7 +3850,7 @@ class iter_impl
     @brief post-increment (it++)
     @pre The iterator is initialized; i.e. `m_object != nullptr`.
     */
-    iter_impl operator++(int)
+    iter_impl const operator++(int)
     {
         auto result = *this;
         ++(*this);
@@ -3893,7 +3893,7 @@ class iter_impl
     @brief post-decrement (it--)
     @pre The iterator is initialized; i.e. `m_object != nullptr`.
     */
-    iter_impl operator--(int)
+    iter_impl const operator--(int)
     {
         auto result = *this;
         --(*this);
@@ -4299,7 +4299,7 @@ class json_reverse_iterator : public std::reverse_iterator<Base>
     json_reverse_iterator(const base_iterator& it) noexcept : base_iterator(it) {}
 
     /// post-increment (it++)
-    json_reverse_iterator operator++(int)
+    json_reverse_iterator const operator++(int)
     {
         return static_cast<json_reverse_iterator>(base_iterator::operator++(1));
     }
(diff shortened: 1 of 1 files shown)


nlohmann:
The tests should succeed... Can you make sure you on the latest `develop` version?

theodelrieu:
What are the errors?

mattismyname:
Ok, my bad. The failures were with gcc 4.9.3. When run under 5.4.0-6ubuntu1~16.04.5 everything passes.

nlohmann:
Hm... GCC 4.9 should work as well. Could you post the errors?

nlohmann:
And which compiler created the `-Weffc++` warnings?

mattismyname:
Original effc++ warnings were from icc (ICC) 11.0 20081105. I'll dig them up and post them here.

Sorry, never used ctest -- how can I get it to print the specific errors? I just get very general message currently:

```
$ ctest -R test-regression_default
Test project /nfs/orto/home/mkgumbel/clt_dev02/json/build
    Start 61: test-regression_default
1/1 Test #61: test-regression_default ..........***Failed    6.87 sec

0% tests passed, 1 tests failed out of 1

Label Time Summary:
default    =   6.87 sec (1 test)

Total Test time (real) =   6.94 sec

The following tests FAILED:
         61 - test-regression_default (Failed)
Errors while running CTest

```
thanks

nlohmann:
You can use `--output-on-failure`.

src/json.hpp
```diff
@@ -6101,7 +6101,9 @@ class serializer
     @param[in] ichar  indentation character to use
     */
     serializer(output_adapter_t<char> s, const char ichar)
-        : o(std::move(s)), loc(std::localeconv()),
+        : o(std::move(s)),
+          number_buffer(),
```
nlohmann: What is the warning that led to adding `number_buffer()` here? The member is initialized in the definition:

```cpp
std::array<char, 64> number_buffer{{}};
```

src/json.hpp
```diff
@@ -8011,6 +8013,7 @@ class basic_json
         json_value(number_float_t v) noexcept : number_float(v) {}
         /// constructor for empty values of a given type
         json_value(value_t t)
+            : object(nullptr)
```
nlohmann: If this is required, then maybe we can remove the lines

```cpp
object = nullptr;  // silence warning, see #821
```

below which have been added when discussing #821.

gregmarr: Related: line 8073 is now unreachable due to the case added at line 8062.

nlohmann: That case was always unreachable and was only added to get the hash `961c151d2e87f2686a955a9be24d316f1362bf21` into binaries using the library.

gregmarr: Interesting.  Perhaps a comment is in order the.

nlohmann:
@mattismyname Could you comment on https://github.com/nlohmann/json/pull/858#pullrequestreview-81649476 and https://github.com/nlohmann/json/pull/858#pullrequestreview-81650164?

nlohmann:
@mattismyname Could you please comment on https://github.com/nlohmann/json/pull/858#pullrequestreview-81649476 and https://github.com/nlohmann/json/pull/858#pullrequestreview-81650164? I would like to have this PR merged for the upcoming release.

src/json.hpp
mattismyname: It was bad icc check -- I think this change is not really necessary.

src/json.hpp
mattismyname: I think this can be removed -- it was bad icc check.

mattismyname:
Sorry for slowness

src/json.hpp
nlohmann: Thanks for checking back. Could you revert this change please?

src/json.hpp
nlohmann: Thanks. Could you remove the mentioned line above please?

mattismyname:
I've pushed the requested fixes to mattismyname/json develop branch.

nlohmann:
Thanks a lot!

Issue #757, opened 2017-09-28
CBOR string


arnaudbrejeon:
I received some CBOR files that contain strings (major type 2) and not UTF-8 strings(major type 3).
It looks like parsing fails. Is it on purpose?

nlohmann:
This is currently not supported, as we only support UTF-8 as encoding. I we would accept byte strings, we would also need to support additional encodings.

While this would just be a programming exercise (it seems that only base64url, base64, base16, and CBOR are mentioned in RFC 7049), it is currently not a priority for this library. However, PRs are always welcome :-)

azadkuh:
ATM, it's possible to *misuse* the `json::to_cbor()` / `json::to_msgpack()` as:
```cpp
const std::string binary = load_some_binary_like_a_jpeg_file();
json src = {"data", binary};

const auto bin_serialization = json::to_cbor(src); // json::to_msgpack()
auto dest = json::from_cbor(bin_serialization);    // json::from_msgpack()
// now: dest["data"].get<std::string>() == binary
// if: serialization/deserialization by this library
```

it may sound weird, but as long as all modern `std:string` implementations can handle binary data as well as utf8 text, what if `to_cbor()/from_cbor()` consider all strings as byte strings (raw/binary)?
then both binary/text are covered in `msgpack`/`cbor` serialization.

nlohmann:
Binary strings are now there, see #1662.

Issue #601, opened 2017-06-01
Use of the binary type in CBOR and Message Pack


Type1J:
For types such as std::vector<uint8_t> the CBOR and Message Pack array type is currently used and each value is written as a numeric value, but this has a high overhead in output size (for byte sized value types) due to each value costing more than 1 byte in CBOR and (most of the time) in Message Pack.

I'd like to propose that the to_* functions for binary formats take an additional bool argument that causes array types that are known to be numeric and byte-sized to serialize using the binary string type of the respective format. The from_* functions should accept either the current style array of numeric types or the binary.

This proposal **_might_** suggest that the nlohmann::json C++ type be augmented with a bytearray discriminator in addition to the normal array discriminator. However there might be an easier way to know that the array is an array of numeric 8-bit values. To be clear, the JSON form would still be an array, so the discriminator would only be set to bytearray if the values given to the array were numeric and inside of the range [0, 255].

Thoughts?

nlohmann:
Could be related to #373.

nlohmann:
Could you please provide a concrete example?

Type1J:
I'm not sure if this is what you meant, but:

The Message Pack and CBOR formats both only have 1 typed array form: a binary string. Using it eliminates the per element size overhead that arrays carry for byte-sized data.

I have an app that has a 131072 byte array (128KiB). There's about 2KiB worth of other data in the object containing that array. It is sent over a binary websocket in Message Pack. If I serialize it using the reference Message Pack library and use the binary string type for that 1 value, and use the same values that this library uses for all other types, then I get a little over 130KiB. Serializing it with this library using Message Pack, I get a little under 260KiB, which is still much smaller and quicker to parse than the JSON version which has commas and 1 to 3 bytes per value (somewhere around 450KiB), but it could be 130KiB by using a small bit of hint information about the type. FYI, I have to send this data in Message Pack often, but periodically I send it to a JSON REST web service as well, so I need to be able to quickly convert between these formats, and this library thankfully offers me the capability to do that with minimal additional code, but I'd like to get my size down (AWS I/O costs per byte, and this data is sent many times per hour).

[...]

Issue #483, opened 2017-03-03
byte string support


izvit:
I was wondering if there were any plans for supporting CBOR byte strings (RFC 7049 type 2)?

tksatware:
JSON is a subset of CBOR and does not support binary data.
If you need to rely on binary data in a cbor stream, this library is not the right tool.

nlohmann:
RFC 7049 states in Sect. 4.1:

>    o  A byte string (major type 2) that is not embedded in a tag that
      specifies a proposed encoding is encoded in base64url without
      padding and becomes a JSON string.

I have to look deeper into this, but this does not sound like arbitrary binary data. What do you think @tksatware ?

tksatware:
CBOR provides a tag as a HINT how the binary data should be decoded. Example "this is a string, but base64 encoded".

I would not go down that route.

First, the json library would need to support the different decoding algorithms and second, converting between json/cbor would not be bijective as it is when using only JSON features.

The concept of tags in CBOR is also not available in JSON (as said, JSON is a subset).

nlohmann:
@izvit What's your opinion on this?

izvit:
In general, it's a perfectly valid argument; however, it's unclear to me what the actual motivation for any CBOR support in the library.  
If the use case is compatibility with other CBOR libraries,  as tksaware correctly pointed out,  the JSON subset alone would not be enough. 
I would imagine most developers choose a binary format for message size or performance reasons, and as far as I know, the CBOR subset that has been chosen has not been standardized, which means it would only be useful for project solely using this library.   Thus, adding support for any CBOR data type that could easily be converted to JSON via hints, and provide value to users seems like a reasonable option.

nlohmann:
Hm. I should have another look at Section 4 ("Converting Data between CBOR and JSON") of RFC 7049. It would be good to follow the standard.

tksatware:
The Standard is non-normative here. It does not comment on the bijectivity, it suggests to convert byte streams to base64 encoded JSON strings. The other way round, it has no determined outcome.

nlohmann:
@tksatware You are right - I read Section 4 and did not really felt the need to change the library's behavior. However, I need to improve the documentation on the supported subset of CBOR to manage the expectations in this regard.

izvit:
[...]

PR #862, opened 2017-12-08
Experimental support for binary type and CBOR (de)serialization

khabinov:
First of all huge thanks for creating this library.

I saw a few issues https://github.com/nlohmann/json/issues/757, https://github.com/nlohmann/json/issues/601 and https://github.com/nlohmann/json/issues/483 where people discuss if this library can support `CBOR`'s binary data type. Motivation for this pull request is to discuss how easy it would be to add support for such data type.

Binary data type is not natural for `JSON`, but many people have to deal with it, especially, when converting from `CBOR` to `JSON`.

In `C++`, though, we can represent contiguous binary arrays using `std::string`, `std::vector<uint8_t>`, `std::list<uint8_t>`, `char*` and whatnot. It could be up to the user how to represent and convert this type into `JSON`.

Right now the change is mostly about the API aspect (e.g. adding binary type), without actual implementation.

Comments are very welcome! Thanks!

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -127,10 +127,15 @@ namespace nlohmann
 template<typename = void, typename = void>
 struct adl_serializer;
 
+// TODO: Probably some wrapper around std::string
+class SomeBinaryType;
+
 // forward declaration of basic_json (required to split the class)
 template<template<typename, typename, typename...> class ObjectType = std::map,
          template<typename, typename...> class ArrayType = std::vector,
-         class StringType = std::string, class BooleanType = bool,
+         class StringType = std::string,
+         class BinaryType = SomeBinaryType,
+         class BooleanType = bool,
          class NumberIntegerType = std::int64_t,
          class NumberUnsignedType = std::uint64_t,
          class NumberFloatType = double,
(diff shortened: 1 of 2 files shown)


gregmarr:
I'm not sure about the rest of this, but `std::list<>` is a horribly wasteful container to use for storing bytes.  The `std::list<>` container is a node-based container.  It has at least two pointers per node, and with the padding, that's a distinct memory allocation of a minimum of 24 bytes for every byte of data on a 64 bit system.

Much better would be `std::vector<uint8_t>`, or maybe `std::deque<uint8_t>` if there's a lot of incremental additions at the end.

nlohmann:
I shall have a look at this after the 3.0.0 release next week.

khabinov:
@gregmarr: you're absolutely right. The point was to show how binary type can be added, so, if this idea gets approved, I can spend more time polishing the change. Thanks!

khabinov:
I removed my naive implementation in order to focus mostly on API side of the change.

nlohmann:
I think we should not extend the JSON value with a binary type, but we may think about a SAX or StAX like parsing Interface that has a callback for binary types.

nlohmann:
Related: #971

nlohmann:
This issue will be addressed with #971 with a callback for binary types. The JSON value itself will not be changed for this.

Thank you for the initial PR.

PR #866, opened 2017-12-09
Floating-point formatting (#360)

### LINKED ISSUE #360: Loss of precision when serializing <double>
It seems that precision is lost when serializing a double. I cannot say why since
std::numeric_limits<double>::digits10 should provide enough digits !?
but if I change that to:
std::numeric_limits<long double>::digits10 then I'm not loosing anything.
I'm NOT using and "long double" types. Only "double".

abolz:
Try to fix #360. `printf` doesn't have a "to shortest" mode and `std::to_chars` is not available yet. Use the Grisu2 algorithm for floating-point formatting which is a nice compromise between using `printf` with a precision of `digits10` and `max_digits10`, i.e. the resulting decimal representation is guaranteed to round-trip and is almost always the shortest such representation.

This PR is mostly to see if everything compiles fine and check that all tests are passing. The change is quite large. But maybe it serves as a starting point for an alternative to always use `max_digits10`.

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -6096,6 +6096,902 @@ class binary_writer
 // serialization //
 ///////////////////
 
+// Implements the Grisu2 algorithm for binary to decimal floating-point
+// conversion.
+//
+// This implementation is a slightly modified version of the reference
+// implementation by Florian Loitsch which can be obtained from
+// http://florian.loitsch.com/publications (bench.tar.gz)
+//
+// References:
+//
+// [1]  Loitsch, "Printing Floating-Point Numbers Quickly and Accurately with Integers",
+//      Proceedings of the ACM SIGPLAN 2010 Conference on Programming Language Design and Implementation, PLDI 2010
+// [2]  Burger, Dybvig, "Printing Floating-Point Numbers Quickly and Accurately",
+//      Proceedings of the ACM SIGPLAN 1996 Conference on Programming Language Design and Implementation, PLDI 1996
+//
+// Note:
+// dtoa::to_short_string should be replaced with std::to_chars if available...
+namespace dtoa {
+
+template <typename Target, typename Source>
+inline Target reinterpret_bits(Source source)
+{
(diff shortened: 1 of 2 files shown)


nlohmann:
What is the license of the included conversion code?

abolz:
The conversion code is based on the reference implementation by Florian Loitsch which is distributed under the MIT license. I added a NOTICE file containing the license but obviously missed to commit it. Is it ok to add a NOTICE file? Or add the license to then end of json.hpp?

nlohmann:
Thanks a lot for the PR. I am currently a bit torn about merging it for the 3.0.0 release, because it involves quite a lot code and I won't have time for a thorough review until then. Furthermore, it could benefit from #700 as we can move this code in a separate (development) header. On the other hand, it does change the behavior of the library (albeit for the better), so maybe that adding it after 3.0.0 would mean to quickly release a 4.0.0 version.

Any thoughts on this?

gregmarr:
I would say that this is a bug fix not affecting the public API, and thus fixing it in a point release would be fine under SemVer guidelines.
I agree that adding it after #700 would allow all of this code to be in a separate header, and thus be properly identified with its license.

abolz:
I would even say that that using `max_digits10` for formatting is the bug fix and then this PR would only be a small enhancement. (And I could actually live without it.)
I agree with both of you that adding/reconsidering this PR after #700 would be much nicer.

abolz:
Is there still some interest in this PR?

I'm asking because I tried to rebase this PR and failed miserably. Two times... :-) So if there is any interest, I'd like to close this and open a new PR (which seems a lot easier for me...)

nlohmann:
Yes, please! Now that the header is split, we can focus on adding such features.

abolz:
Ok. Where shall I put the implementation? I thought about something like a `to_chars.hpp` file in detail/conversions.

nlohmann:
Sounds good.

Issue #910, opened 2018-01-11
Installation path is incorrect


InFaMoUsZero:
Calling "make install" now installs json.hpp to "nlohmann/src/json.hpp".

No sure if this was intentional.

The issue is from the latest PR change [here](https://github.com/nlohmann/json/blob/b67e00b9b56a344bfdb3ce92bf1f6724db518c95/CMakeLists.txt#L24).

nlohmann:
I am not CMake expert - what should be the value in your opinion?

theodelrieu:
Oops I haven't tried that, I'll take a look this afternoon

nlohmann:
Fixed with merging #911.

@InFaMoUsZero Could you please check if this fixes the issue for you?

InFaMoUsZero:
@nlohmann Sure, I can test this on Tuesday and hopefully get this closed off.

InFaMoUsZero:
@nlohmann I can confirm this has now fixed the issue.

nlohmann:
Thanks for checking back!

PR #911, opened 2018-01-13
Fix/cmake install

theodelrieu:
This PR fixes the CMake installation directory issue described in #910. (Due to PR #700)

It also adds a new CMake option to let users install the non-amalgamated version of the library.
The `amalgamate` tool was moved to avoid it to be installed as well

## Pull request checklist


CMakeLists.txt
```diff
@@ -16,12 +16,12 @@ include(ExternalProject)
 ## OPTIONS
 ##
 option(JSON_BuildTests "Build the unit tests when BUILD_TESTING is enabled." ON)
+option(JSON_MultipleHeaders "Use non-amalgamated version of the library." OFF)
```
nlohmann: Why would anyone want this?

theodelrieu: To include the `json_fwd.hpp` file, or the `adl_serializer.hpp`. That's something I want to use at work, I like to only include what I need.

nlohmann: Ok.

nlohmann:
Thanks!

amilcarlucas:
Would be nice if you would include a bit of documentation on how to install this.
make install, does not seam to do it :)

nlohmann:
It's like any other CMake project:

```sh
mkdir build
cd build
mkdir local
cmake .. -DCMAKE_INSTALL_PREFIX=$(pwd)/local
make
make install
```

After that, you should have

```
local
├── include
│   └── nlohmann
│       └── json.hpp
└── lib
    └── cmake
        └── nlohmann_json
            ├── nlohmann_jsonConfig.cmake
            ├── nlohmann_jsonConfigVersion.cmake
            └── nlohmann_jsonTargets.cmake
```

amilcarlucas:
Thanks for the info.

PR #915, opened 2018-01-15
Floating-point formatting

### LINKED ISSUE #360: Loss of precision when serializing <double>

abolz:
Try to fix #360. printf doesn't have a "to shortest" mode and std::to_chars is not available yet. Use the Grisu2 algorithm for floating-point formatting which is a nice compromise between using printf with a precision of digits10 and max_digits10, i.e. the resulting decimal representation is guaranteed to round-trip and is almost always the shortest such representation.


develop/detail/serializer.hpp
```diff
@@ -636,14 +637,36 @@ class serializer
...
+
+    void dump_float(number_float_t x, std::true_type /*is_ieee_single_or_double*/)
+    {
+        char* begin = number_buffer.data();
+        char* end = ::nlohmann::detail::to_chars(begin, begin + number_buffer.size(), x);
+
+        o->write_characters(begin, static_cast<size_t>(end - begin));
+    }
+
+    void dump_float(number_float_t x, std::false_type /*is_ieee_single_or_double*/)
```
nlohmann: Just out of curiosity: why do we still need the snprintf approach?

src/json.hpp
```diff
@@ -7079,14 +8108,36 @@ class serializer
     void dump_float(number_float_t x)
     {
         // NaN / inf
-        if (not std::isfinite(x) or std::isnan(x))
```
nlohmann: Why can `NaN` be ignored here?

develop/detail/serializer.hpp
abolz: Because the implementation only works for IEEE single and double precision.
It should work in case `long double` is an alias for `double`, though.

src/json.hpp
abolz: Its actually not ignored: the `isnan` test is just redundant. `isfinite` return false for NaN

nlohmann:
Thanks a lot!

PR #918, opened 2018-01-16
add std::to_string method for json object

DELTA37:
Pull request for issue 916 

to_string function Implementation

### DIFF
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -15290,6 +15290,18 @@ inline nlohmann::json operator "" _json(const char* s, std::size_t n)
 }
 
 /*!
+@brief user-defined to_string function for JSON values
+
+This function implements a user-defined to_string  for JSON objects. 
+
+@param[in] j  a JSON object
+@return a std::string object
+*/
+inline std::string to_string(const nlohmann::json& j) {
+  return j.dump();
+}
+
+/*!
 @brief user-defined string literal for JSON pointer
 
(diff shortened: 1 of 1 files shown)


theodelrieu:
@nlohmann It would be awesome to get Travis/Appveyor to run the `make amalgamate` step automatically.

It seems to me that [Outcome](https://github.com/ned14/outcome) does something similar

nlohmann:
@theodelrieu I need to add `astyle` as dependency. I forgot this before...

theodelrieu:
I guess it is a good opportunity to do both!

src/json.hpp
```diff
@@ -15289,6 +15289,18 @@ inline nlohmann::json operator "" _json(const char* s, std::size_t n)
...
 
+/*!
+@brief user-defined to_string function for JSON values
+
+This function implements a user-defined to_string  for JSON objects. 
+
+@param[in] j  a JSON object
+@return a std::string object
+*/
+inline std::string to_string(const nlohmann::json& j) {
```
theodelrieu: Just thought about this, but I think you should make a generic overload for every `basic_json` type (i.e. a templated overload).

Without it, users who do not use the default json type won't be able to use `to_string`.

nlohmann:
@DELTA37 Could you please address the issues described in https://github.com/nlohmann/json/pull/918#pullrequestreview-89218566 ?

PR #925, opened 2018-01-19
Howto installation of json_fwd.hpp (fixes #923)

### LINKED ISSUE #923: json_fwd.hpp not installed
**Bug Report**

- What is the issue you have?

Not sure if this is by design or not, but I notice that when I perform 'cmake install', the json_fwd.hpp is not installed. 

Well, if I want to perform an in-source build (ExternalProject_Add()), as an example, it makes things difficult if their is no install procedure for the fwd declaration.

Ubuntu 17.10 and GCC v7.2

Released

zerodefect:
Updated README.md to indicate how installation of json_fwd.hpp can be achieved as part of install step. I'm hoping that will save developers a few minutes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ using json = nlohmann::json;
 
 to the files you want to use JSON objects. That's it. Do not forget to set the necessary switches to enable C++11 (e.g., `-std=c++11` for GCC and Clang).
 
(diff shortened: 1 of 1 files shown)


zerodefect:
fixes #923  

(Not really a 'fix' per se, but it just links the issues together)

Issue #938, opened 2018-01-27
Allow overriding JSON_THROW to something else than abort()


kaidokert:
**Feature Request**

- Describe the feature in as much detail as possible.

On an embedded/rtos system, the stack is compiled without rtti/exceptions, and I'm running http workers in a separate thread. I can deal with failures by basically tearing the thread down and restarting it, however, i don't want to call abort() specifically and unconditionally. Also it's nice to still be able to obtain the error code from the exception

- Include sample usage where appropriate.

passing a -DJSON_THROW=my_deal_with_json_error(x) at compile time would allow easy customization of how the error handling gets dealt with

The addition is as simple as
```
#if !defined(JSON_THROW)
#define JSON_THROW abort
#endif
```

nlohmann:
That seems like a reasonable extension, but maybe also for `JSON_TRY` and `JSON_CATCH`. What do you think?

kaidokert:
I thought about that and didn't immediately think of a valid use case, but maybe doing something like
```
#define JSON_TRY(x) my_save_last_try_point(..., ___FILE__, ___LINE__ )
```
could be useful in debugging scenarios. I'll add and make a PR in a bit

nlohmann:
Fixed with merged PR #940.

PR #940, opened 2018-01-28
Allow overriding THROW/CATCH/TRY macros with no-exceptions #938

kaidokert:

### DIFF
--- a/develop/detail/macro_scope.hpp
+++ b/develop/detail/macro_scope.hpp
@@ -46,6 +46,20 @@
     #define JSON_CATCH(exception) if(false)
 #endif
 
+// override exception macros
+#if defined(JSON_THROW_USER)
+    #undef JSON_THROW
+    #define JSON_THROW JSON_THROW_USER
+#endif
+#if defined(JSON_TRY_USER)
+    #undef JSON_TRY
+    #define JSON_TRY JSON_TRY_USER
+#endif
+#if defined(JSON_CATCH_USER)
+    #undef JSON_CATCH
+    #define JSON_CATCH JSON_CATCH_USER
+#endif
+
 // manual branch prediction
 #if defined(__clang__) || defined(__GNUC__) || defined(__GNUG__)
     #define JSON_LIKELY(x)      __builtin_expect(!!(x), 1)
--- a/src/json.hpp
+++ b/src/json.hpp
@@ -156,6 +156,20 @@ using json = basic_json<>;
     #define JSON_CATCH(exception) if(false)
 #endif
 
+// override exception macros
+#if defined(JSON_THROW_USER)
+    #undef JSON_THROW
+    #define JSON_THROW JSON_THROW_USER
+#endif
+#if defined(JSON_TRY_USER)
+    #undef JSON_TRY
+    #define JSON_TRY JSON_TRY_USER
+#endif
+#if defined(JSON_CATCH_USER)
+    #undef JSON_CATCH
+    #define JSON_CATCH JSON_CATCH_USER
+#endif
+
 // manual branch prediction
 #if defined(__clang__) || defined(__GNUC__) || defined(__GNUG__)
     #define JSON_LIKELY(x)      __builtin_expect(!!(x), 1)


nlohmann:
Just a question: is it an issue that the following lines are at the end of the amalgamated JSON header:

```cpp
#undef JSON_CATCH
#undef JSON_THROW
#undef JSON_TRY
```

kaidokert:
Hasn't been an issue for me at least, don't need the macros anywhere else, it's better to remove them

nlohmann:
I think it is good practice to undefine the macros you define yourself. So removing the undefs could be an issue.

kaidokert:
Understand .. then i guess the cleanest way to deal with it would be to do this entire dance:
```
 #if !defined(JSON_THROW)
    #define JSON_THROW(exception) std::abort()
 #else
   #define USER_JSON_THROW 1
#endif
.... // at the end
#if !defined(USER_JSON_THROW)
#undef JSON_THROW
#else
#undef USER_JSON_THROW
#endif
```
and repeat that for all 3 ?

nlohmann:
I think so. Better call it `JSON_THROW_USER_PROVIDED` rather than `USER_JSON_THROW` to have a consistent prefix.

kaidokert:
Updated - i think it need indentation now too, unless its too much commit noise

nlohmann:
If you ran `make amalgamate` (and had `astyle` installed), then indentation is already fixed. If not, then this may be an issue in a Travis job.

develop/detail/macro_scope.hpp
```diff
@@ -41,9 +41,21 @@
     #define JSON_TRY try
     #define JSON_CATCH(exception) catch(exception)
 #else
-    #define JSON_THROW(exception) std::abort()
-    #define JSON_TRY if(true)
-    #define JSON_CATCH(exception) if(false)
+    #if !defined(JSON_THROW)
+        #define JSON_THROW(exception) std::abort()
```
nlohmann: I just realized the code can be simplified:

```cpp
#if !defined(JSON_THROW)
    #define JSON_THROW(exception) std::abort()
    #define JSON_THROW_DEFAULT
#endif
```

and later

```cpp
#if defined(JSON_THROW_DEFAULT)
    #undef JSON_THROW
    #undef JSON_THROW_DEFAULT
#endif
```

What do you think?

kaidokert: Yes, this is better, less noise

kaidokert: But then it requires the _DEFAULT defines for exceptions-on version as well, ends up adding a bit of clutter

nlohmann: Right... But why not allowing to define `JSON_THROW` etc. while using exceptions?

kaidokert: The current version in repo does #undef for both exceptions-on and exceptions-off cases. If i rewrite with JSON_THROW_DEFAULT then to keep the same behaviour it would become ..

```
#if (defined(__cpp_exceptions) ...
    #define JSON_THROW(exception) throw exception
    #define JSON_THROW_DEFAULT //  1 extra line here
   ...
#else
   #if !defined(JSON_THROW)
    #define JSON_THROW(exception) std::abort()
    #define JSON_THROW_DEFAULT
  #endif
  ...
#endif
```
and later
```
#if defined(JSON_THROW_DEFAULT)
    #undef JSON_THROW
    #undef JSON_THROW_DEFAULT
#endif
```

If that is cleaner, I'll rewrite

nlohmann: Another idea:

```cpp
// allow to disable exceptions
    #define JSON_THROW(exception) throw exception
    #define JSON_TRY try
#else
    #define JSON_THROW(exception) std::abort()
    #define JSON_TRY if(true)
    #define JSON_CATCH(exception) if(false)
#endif

// override exception macros
#if define(JSON_THROW_USER)
    #define JSON_THROW JSON_THROW_USER
#endif
#if define(JSON_TRY_USER)
    #define JSON_TRY JSON_TRY_USER
#endif
#if define(JSON_CATCH_USER)
    #define JSON_CATCH JSON_CATCH_USER
#endif
```

That is, the user can define `JSON_THROW_USER` to override the behavior of `JSON_THROW`. No need to differentiate the exception/no exception case, and no need to remember whether we need to undef.

What do you think?

kaidokert: This is probably the most hygienic , although for maximum warning compile it should be 
```
#if define(JSON_THROW_USER)
    #undef JSON_THROW
    #define JSON_THROW JSON_THROW_USER
#endif
```
to avoid macro redefinition warnings

nlohmann: Right.

nlohmann:
Thanks!

PR #954, opened 2018-02-02
Add a note about installing the library with cget

pfultz2:
This adds a note about installing with [`cget` ](http://cget.readthedocs.io/en/latest/) since its already supported(due to supporting cmake installation out of the box).


pfultz2:
The failures on travis seems unrelated to this.

nlohmann:
@pfultz2 Could you address the requested changes, please?

pfultz2:
> "the latest" -> "the latest version"

This actually installs the latest from develop, so it is the latest version, but not the latest released version.

nlohmann:
I just made that comment because the previous sentence missed a word. Maybe "the latest development version"? Furthermore, there is a space missing after "flag".

pfultz2:
> I just made that comment because the previous sentence missed a word. Maybe "the latest development version"?

Sure, that works.

> Furthermore, there is a space missing after "flag".

Got it.

nlohmann:
Thanks!

Issue #953, opened 2018-02-02
Coveralls shows no coverage data


nlohmann:
Since the split into the new subdirectories, Coveralls displays no coverage data, see https://coveralls.io/github/nlohmann/json. I tried to fix the `.travis.yml` file, but probably overlooked something.

theodelrieu:
Weird, I remember battling against coveralls in #700 and making it work.

Could you try to set `--include include`?

I think it's because in my PR, `src` was containing split headers, and there was a `single_include` folder, that has been renamed to `src` after the merge.

nlohmann:
But the tests are compiled using the header from the `single_header` directory.

nlohmann:
And I think it is more handy to have the coverage for the single header file, because this is the main artifact of the repository.

theodelrieu:
But having it on the `include` folder is way better to know where the not covered code is. When I had to open `json.hpp` in coveralls it froze my browser so I'd go for multiple headers all day everyday.

Is there a Travis job that runs `make check-amalgamation`? If that's the case, I think there is no need to run coveralls on the single header

nlohmann:
Yes, there is a job calling `make check-amalgamation`, so we can actually let Coveralls work on the multiple-headers version.

nlohmann:
Coverage data are now shown for the single-header version. I am fine with this and would close this issue. For coverage of the multi-header version, there is the `make coverage` target which uses lcov locally.

To extend Travis/Coveralls to use the multi-header version, we need way to call cmake with `-DJSON_Coverage=ON` **and** `-DJSON_MultipleHeaders=ON` - so far I only managed to pass one flag via environment variable `CMAKE_OPTIONS` to the call to cmake in the `script` passage.

PR #957, opened 2018-02-04
Make coveralls watch the include folder

theodelrieu:
Should be enough to fix it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,7 +80,7 @@ matrix:
     env:
       - COMPILER=g++-4.9
       - SPECIAL=amalgamation
-      - CMAKE_OPTIONS=-DJSON_MultipleHeader=ON
+      - MULTIPLE_HEADERS=ON
     addons:
       apt:
         sources: ['ubuntu-toolchain-r-test']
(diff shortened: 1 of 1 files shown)


theodelrieu:
The latest Coveralls job passed: https://coveralls.io/jobs/33439123

Could you retry the Travis job that failed? Seems like a spurious failure...

theodelrieu:
Oops, seems my fix didn't fix anything after all...

nlohmann:
coveralls

This is how [Coveralls](https://coveralls.io/builds/15376538) looks like right now. The 0% at the top is strange - and this is also the number reported in this PR. Any idea how this can be fixed?

theodelrieu:
That's very strange, maybe because we give `include` instead of `include/nlohmann`, I'll give it a try.

theodelrieu:
I might become crazy with Coveralls... @nlohmann I remember I could try locally before, but the private repo's token has been moved from the repo files (which is good!)

Could you try to fix this one? Running Travis for hours to finally see a big fat **0.0%** is quite frustrating...

nlohmann:
I'll have a look.

nlohmann:
See comments in #953. With 2dda87c3b7440ceac5b7ffa93de5c9bc1e85a57f, coverage data should be generated again. I am currently not planning to extend this to the multi-header version myself, but I think this can be done with a PR, because Travis/Coveralls seems not to care about who is comitting.

theodelrieu:
It seems CMake does not go well with environment variables which contains spaces...

Hence the `MULTIPLE_HEADERS` I've added in the travis file

theodelrieu:
I think it's good now, finally!

.travis.yml
```diff
@@ -281,14 +282,16 @@ script:
 
   # make sure CXX is correctly set
   - if [[ "${COMPILER}" != "" ]]; then export CXX=${COMPILER}; fi
+  # by default, use the single-header version
+  - if [[ "${MULTIPLE_HEADERS}" == "" ]]; then export MULTIPLE_HEADERS=OFF; fi
```
nlohmann: I did not think of this trick. Well done!

PR #978, opened 2018-02-22
Fix `#include` brackets

aardvarkk:
Fix `#include` brackets so they correspond to local includes instead of system includes.

When building code relying on this library in Xcode, the build process throws an error (not a warning!) stating its inability to locate the presumed system includes. The `<>` angle brackets on `#include` indicate a system-level include, not a user-level include. By swapping to quotation marks for the `#include` we more accurately represent the content of the included files and Xcode builds without complaint!

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -45,28 +45,28 @@ SOFTWARE.
 #include <string> // string, stoi, to_string
 #include <utility> // declval, forward, move, pair, swap
 
-#include <nlohmann/json_fwd.hpp>
-#include <nlohmann/detail/macro_scope.hpp>
-#include <nlohmann/detail/meta.hpp>
-#include <nlohmann/detail/exceptions.hpp>
-#include <nlohmann/detail/value_t.hpp>
-#include <nlohmann/detail/conversions/from_json.hpp>
-#include <nlohmann/detail/conversions/to_json.hpp>
-#include <nlohmann/detail/input/input_adapters.hpp>
-#include <nlohmann/detail/input/lexer.hpp>
-#include <nlohmann/detail/input/parser.hpp>
-#include <nlohmann/detail/iterators/primitive_iterator.hpp>
-#include <nlohmann/detail/iterators/internal_iterator.hpp>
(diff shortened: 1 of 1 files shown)


nlohmann:
I cannot reproduce the issue. With Xcode Version 9.2 (9C40b), I can successfully build and run the multi-header version. Furthermore, the files in the `include/nlohmann/detail` folder also contain `<>`-style includes.

nlohmann:
I think @theodelrieu introduced the angle brackets, maybe he can comment on this.

theodelrieu:
The choice of angle brackets was more a matter of style than functionality.
I usually use quote includes when I don't include something from a top-level directory.

But I don't understand how it could fail to compile in the first place, the weirdest thing is that it works by only replacing the includes in `json.hpp`, and not in the entire project.

What Xcode version do you have? Travis runs on a bunch of them without any problem, so I'm suspecting your Xcode settings here.

aardvarkk:
Perhaps the difference is that I'm using the library from a `git clone` rather than having installed it with Homebrew? I think when I installed using Homebrew it was fine (since the system library directories are searched), but when I just cloned the directory into my project and added it to the header search paths, it wasn't able to find the internal library headers.

theodelrieu:
Could you post the directory structure after the git clone? Also what include path did you add?

nlohmann:
Independent of *why* the compilation failed with Xcode, reading https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html makes me thing using quotes instead of angle brackets is a better choice as searching the system header path makes no sense here.

theodelrieu:
The only real advantage is to prevent inclusion from a relative path, `#include "../myfile.hpp"` from within the library's code. It's more because of this that I chose to use angle-brackets, not because of "search system headers first".

If going with double quote includes, we shall pay attention to that.

nlohmann:
@theodelrieu You are right - we would need relative paths in the header files in the `detail` folder and below. This may bring errors down the road...

nlohmann:
I am closing this. It seems arbitrary that only changing the top-level includes to quotes would fix this.

PR #981, opened 2018-02-23
Make integration section concise

wla80:
Rewrite in concise sentences to improve the readability of the section

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ See the [contribution guidelines](https://github.com/nlohmann/json/blob/master/.
 
 ## Integration
 
-The single required source, file `json.hpp` is in the `single_include/nlohmann` directory or [released here](https://github.com/nlohmann/json/releases). All you need to do is add
+`json.hpp` is the single required file in `single_include/nlohmann` or [released here](https://github.com/nlohmann/json/releases). You need to add
 
 ```cpp
 #include <nlohmann/json.hpp>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,9 +64,9 @@ The single required source, file `json.hpp` is in the `single_include/nlohmann`
 using json = nlohmann::json;
 ```
 
-to the files you want to use JSON objects. That's it. Do not forget to set the necessary switches to enable C++11 (e.g., `-std=c++11` for GCC and Clang).
+to the files you want to use JSON objects and set the necessary switches to enable C++11 (e.g., `-std=c++11` for GCC and Clang).
```
nlohmann: "JSON objects" sounds odd here (I know it was in my original sentence...). "Object" already has a meaning in JSON - maybe "to the files you want to use JSON in"? Any ideas?

README.md
```diff
@@ -64,9 +64,9 @@ The single required source, file `json.hpp` is in the `single_include/nlohmann`
...
 
-You can further use file [`include/nlohmann/json_fwd.hpp`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json_fwd.hpp) for forward-declarations. The installation of json_fwd.hpp (as part of cmake's install step), can be achieved by setting `-DJSON_MultipleHeaders=ON`:
+You can install [`include/nlohmann/json_fwd.hpp`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json_fwd.hpp) by setting `-DJSON_MultipleHeaders=ON` as a part of cmake's install step for forward-declarations.
```
nlohmann: `-DJSON_MultipleHeaders=ON` is not for installing forward-declarations - this is just a side effect. I wanted to note there *exists* a forward-declaration header. And that it is already part of the CMake project with those flags.

nlohmann:
@wla80 Could you comment on the remarks above?

README.md
wla80: Thanks for the feedback. This sounds OK but how about "to the files you want to process JSON". Also, you had used "JSON object" multiple times in this doc.

README.md
wla80: Sorry about this. If this is the case, I will keep the original version.

README.md
nlohmann: "process JSON" is even better. I have to go through the README again to check where I used "JSON object" in a different sense than "a JSON value of type object"...

README.md
nlohmann: OK.

wla80:
Hi @nlohmann , I just uploaded the changes based on our earlier discussion. Please check and thank you

PR #984, opened 2018-02-26
Set stream flags in `input_stream_adapter`

### LINKED ISSUE #976: istream >> json  --- 1st character skipped in stream
BUG: The first character in the stream file is skipped.

I took the single include source and used it in some test code.
Source code posted about Feb 13
https://github.com/nlohmann/json/blob/develop/single_include/nlohmann/json.hpp

What I did amounted to:
```
// Extracted from my code
json j;
std::ifstream is("filename");
is >> j;
```
Tracing thru the source, I found this behaviour is the result of the code looking for a BOM at the start of the UTF-8 stream:
It fetches the first character. Since it is not part of the BOM, it performs:
```Line 1669		is.unget(); // no byte order mark; process as usual```

Which is what it should do BUT ...
when get() is called shortly after:
```
Line 3024    token_type scan()
			{
				// read next character and ignore whitespace
				do
				{
					get();
			...
```
the variable 'current' contains the second char, as if unget() did not happen.

Not sure as to why.  Repeated get() do get all subsequent chars one by one.

Since this was just a test, the workaround is to add in a return and start text on line two. Then it works fine. Actually, to verify the bug, placing ANY character in the first position shows that character is ignored.

General settings:
Using MSVC C++ V17 (v141)
MFC in shared DLL
EXE w/Unicode support. no CLR support.

Text file loaded, edited, saved in MSVC IDE

Tested on source uploaded 8 days ago and also reproduced in a version before that.

### PR DESCRIPTION
**Pull request by abolz**
Set stream flags in `input_stream_adapter` on various error conditions. Fixes #976.

For performance reasons do not use the `istream` interface but the underlying `streambuf` to get/unget characters. The behavior of this implementation differs slightly from the reference implementation (see below) in case a `streambuf` operation throws an exception (`badbit` is not turned on in the `istream`'s error state).

---
For reference, this class was used to generate test cases:
```c++
class input_stream_adapter : public input_adapter_protocol {
    std::istream& is;

public:
    explicit input_stream_adapter(std::istream& i) : is(i) {
        // skip byte order mark
        if (is.peek() == 0xEF) {
            is.ignore();
            if (is.peek() == 0xBB) {
                is.ignore();
                if (is.peek() == 0xBF) {
                    is.ignore();
                    return;
                }
                is.unget();
            }
            is.unget();
        }
    }

    traits::int_type get_character() override {
        if (is.good()) {
            const auto ch = is.peek();
            if (ch != traits::eof()) {
                is.ignore();
                return ch;
            }
        }
        return traits::eof();
    }

    void unget_character() override {
        is.unget();
    }
};
```

### DIFF
--- a/include/nlohmann/detail/input/input_adapters.hpp
+++ b/include/nlohmann/detail/input/input_adapters.hpp
@@ -48,76 +48,67 @@ struct input_adapter_protocol
 using input_adapter_t = std::shared_ptr<input_adapter_protocol>;
 
 /*!
-Input adapter for a (caching) istream. Ignores a UFT Byte Order Mark at
-beginning of input. Does not support changing the underlying std::streambuf
-in mid-input. Maintains underlying std::istream and std::streambuf to support
-subsequent use of standard std::istream operations to process any input
-characters following those used in parsing the JSON input.  Clears the
-std::istream flags; any input errors (e.g., EOF) will be detected by the first
-subsequent call for input from the std::istream.
+Input adapter for a (caching) istream.
+Ignores a UTF Byte Order Mark at beginning of input.
+
+Does not support changing the underlying std::streambuf in mid-input.
 */
 class input_stream_adapter : public input_adapter_protocol
 {
   public:
-    ~input_stream_adapter() override
-    {
-        // clear stream flags; we use underlying streambuf I/O, do not
-        // maintain ifstream flags
-        is.clear();
-    }
+    using traits_type = std::char_traits<char>;
 
     explicit input_stream_adapter(std::istream& i)
-        : is(i), sb(*i.rdbuf())
+        : is(i)
     {
-        // skip byte order mark
-        std::char_traits<char>::int_type c;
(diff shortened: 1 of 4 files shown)


trilogy-service-ro:

abolz:
The current state of this PR is a compromise between consistency[*] and performance. And it's the best I could come up with. (Since I won't have the time to look at this closer in the near future, I'll remove the "WIP" state and leave this PR as is.)

AFAIK, there is no way to use the underlying `streambuf` directly while providing the same semantics as the `istream` interface (since there is no way to set the stream state without throwing an exception) in case a `streambuf` operation throws an exception. But most importantly, the stream state is no longer unconditionally set to "good" and no characters are extracted from "bad" streams.

IIRC, using the `istream` interface degrades the performance by a factor of ~5. So this PR still uses the `streambuf` directly, but sets the stream state on errors (or if EOF is reached). I think the performance degradation is acceptable for the increased consistency with built-in stream extraction operators...

[*] As I understand it from reading the standard 🙈

PR #1001, opened 2018-03-09
Fix memory leak during parser callback

nlohmann:
Fixes a memory leak in the case when the parser callback function discards values. The reason was a forgotten destroy call.

Thanks to @ralfbielig for reporting and proposing a fix.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -953,7 +953,7 @@ class basic_json
         /// constructor for rvalue strings
         json_value(string_t&& value)
         {
-            string = create<string_t>(std::move(value));
+            string = create<string_t>(std::forward < string_t&& > (value));
         }
 
         /// constructor for objects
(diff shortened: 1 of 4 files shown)


include/nlohmann/json.hpp
```diff
@@ -953,7 +953,7 @@ class basic_json
         /// constructor for rvalue strings
         json_value(string_t&& value)
         {
-            string = create<string_t>(std::move(value));
+            string = create<string_t>(std::forward < string_t&& > (value));
```
theodelrieu: `std::move` was correct here (and in all the other places in this PR) since `value` is not a forwarding reference, but an rvalue reference.

The behaviour is the same in this case, but is quite confusing.

nlohmann:
Thanks, I’ll change that. I was totally confused that the memory leak went undetected for so long.

PR #1004, opened 2018-03-10
dump to alternative implementation of string, as defined in basic_json template

agrianius:
Alternative string implementation can be defined in basic_json template. These changes allows json::dump to work with the alternative string type.


nlohmann:
Could you please also provide a test case for this?

agrianius:
Sure. I think I can provide one.

PR #1005, opened 2018-03-10
dump to alternative string type, as defined in basic_json template

agrianius:
An alternative string implementation can be defined in basic_json template. These changes allow json::dump to work with the alternative string type.


agrianius:
fatal: unable to access 'https://github.com/nlohmann/json.git/': Could not resolve host: github.com
Error: Failed to download resource "nlohmann_json"
Failure while executing: git clone --depth 1 --branch develop https://github.com/nlohmann/json.git /Users/travis/Library/Caches/Homebrew/nlohmann_json--git
Error: Testing requires the latest version of nlohmann/json/nlohmann_json

Do you know how to restart it?

PR #1006, opened 2018-03-11
dump to alternative string type, as defined in basic_json template

agrianius:
An alternative string implementation can be defined in basic_json template. These changes allow json::dump to work with the alternative string type.


nlohmann:
Thanks a lot!

PR #1028, opened 2018-03-29
Added support for string_view in C++17

gracicot:
This patch adds support for `std::string_view` conversion (explicit and implicit). It also added construction from `std::string_view`.

I did that by following how many compatible array types were handled and adding a type trait that checks for compatible string types.

However, I fear the trait is kinda loose in which type is allows (see `is_compatible_string_type`). Tell me if it needs to be fixed!

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -70,6 +70,25 @@ void from_json(const BasicJsonType& j, typename BasicJsonType::string_t& s)
     s = *j.template get_ptr<const typename BasicJsonType::string_t*>();
 }
 
+template <
+    typename BasicJsonType, typename CompatibleStringType,
+    enable_if_t <
+        is_compatible_string_type<BasicJsonType, CompatibleStringType>::value and
+        not std::is_same<typename BasicJsonType::string_t,
+                         CompatibleStringType>::value and
+        std::is_constructible <
+            BasicJsonType, typename CompatibleStringType::value_type >::value,
+        int > = 0 >
+void from_json(const BasicJsonType& j, CompatibleStringType& s)
+{
+    if (JSON_UNLIKELY(not j.is_string()))
+    {
+        JSON_THROW(type_error::create(302, "type must be string, but is " + std::string(j.type_name())));
+    }
+
+    s = *j.template get_ptr<const typename BasicJsonType::string_t*>();
+}
+
 template<typename BasicJsonType>
 void from_json(const BasicJsonType& j, typename BasicJsonType::number_float_t& val)
 {
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -75,9 +75,7 @@ template <
     enable_if_t <
         is_compatible_string_type<BasicJsonType, CompatibleStringType>::value and
         not std::is_same<typename BasicJsonType::string_t,
-                         CompatibleStringType>::value and
-        std::is_constructible <
-            BasicJsonType, typename CompatibleStringType::value_type >::value,
+                         CompatibleStringType>::value,
         int > = 0 >
 void from_json(const BasicJsonType& j, CompatibleStringType& s)
 {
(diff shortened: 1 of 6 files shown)


nlohmann:
Hey @gracicot, thanks for the PR!

single_include/nlohmann/json.hpp
```diff
@@ -977,6 +996,25 @@ void from_json(const BasicJsonType& j, typename BasicJsonType::string_t& s)
...
+        not std::is_same<typename BasicJsonType::string_t,
+                         CompatibleStringType>::value and
+        std::is_constructible <
+            BasicJsonType, typename CompatibleStringType::value_type >::value,
+        int > = 0 >
+void from_json(const BasicJsonType& j, CompatibleStringType& s)
+{
+    if (JSON_UNLIKELY(not j.is_string()))
+    {
+        JSON_THROW(type_error::create(302, "type must be string, but is " + std::string(j.type_name())));
```
nlohmann: Is there a test case checking for this exception? If not, please add one.

nlohmann:
There is an issue with MSVC 2017 using `/std:c++latest`: https://ci.appveyor.com/project/nlohmann/json/build/2697/job/f7no7s3vjq8xddu9

gracicot:
Sorry for the delay. I haven't investigated the issue with visual studio yet, I will be soon.

nlohmann:
Any ideas on how to fix this?

gracicot:
I investigated and it lead me to think it's linked to #464 .

This is a MSVC bug, and If I look at [cppreference](http://en.cppreference.com/w/cpp/string/basic_string/operator%3D), an overload of `operator=` taking `T` is defined for `T` only if `T` is convertible to a `std::string_view`. While digging deeper, I see that it's [LWG 2946](https://cplusplus.github.io/LWG/issue2946) that Microsoft hasn't yet implemented. I reported a bug about it.

I haven't found a solution yet. Maybe disallow the implicit conversion until the fix for the issue is fixed? That would at least allow `get<std::string_view>`

nlohmann:
I think disabling it for MSVC for the time being could be a way to go.

gracicot:
So the bug preventing enabling this feature on msvc has been fixed: https://developercommunity.visualstudio.com/content/problem/243183/stdstring-should-not-have-a-operator-that-takes-a.html

It should be out in 15.8 (msvc 19.15?), so in 19.14 and older it is disabled, as discussed.

include/nlohmann/detail/conversions/to_json.hpp
```diff
@@ -51,6 +51,16 @@ struct external_constructor<value_t::string>
         j.m_value = std::move(s);
         j.assert_invariant();
     }
+
+	template<typename BasicJsonType, typename CompatibleStringType,
+		     enable_if_t<not std::is_same<CompatibleStringType, typename BasicJsonType::string_t>::value,
+		                 int> = 0>
```
theodelrieu: This SFINAE check is already done in the `to_json` method, maybe we can only do it once:

- Adding a `to_json(const BasicJson::string_t&)` overload (only `BasicJson::string_t&&` is handled right now)
- Then, in the `to_json` overload where the `std::is_constructible<...>` check is performed, construct a `string_t` and move it to the `external_constructor`.

Additionally, I'd rather keep `external_constructor` function count low (Maybe the `array_t` specialization could get cleaned a bit).

This is mostly a style preference though.

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -70,6 +70,25 @@ void from_json(const BasicJsonType& j, typename BasicJsonType::string_t& s)
...
 }
 
+template <
+    typename BasicJsonType, typename CompatibleStringType,
+    enable_if_t <
+        is_compatible_string_type<BasicJsonType, CompatibleStringType>::value and
+        not std::is_same<typename BasicJsonType::string_t,
+                         CompatibleStringType>::value and
+        std::is_constructible <
+            BasicJsonType, typename CompatibleStringType::value_type >::value,
```
theodelrieu: Is this check relevant?

gracicot: You are right, it's not necessary. I also missed a check elsewhere, let me fix that.

nlohmann:
I'm on holidays, but I will have a look at this PR once I am back at the end of June.

nlohmann:
Could you please have a look at the conflicts?

nlohmann:
Thanks a lot!

PR #1048, opened 2018-04-10
Misc cmake packaging enhancements

chuckatkins:
Some additions to the CMake package configs to be more consistent with common CMake packaging guidelines:

*  Use namespaces for imported targets
*  Allow package configs in build directory 
*  User-defined install location for package configs


nlohmann:
Thanks a lot!

Issue #960, opened 2018-02-06
We should remove static_asserts


theodelrieu:
Sorry for the click-bait title, but I'd like to have some feedbacks about this :)

I'll try to be very clear, since it's related to the heart of the `from/to_json` conversion system.

**DISCLAIMER: This post is long. Really.**

## Reminder of the initial problem
Since version 2.1.0 and the support of user-defined types,
we have a non-explicit templated constructor for `basic_json`.

This can be very dangerous:

```cpp
struct Innocent {};
struct Evil {
  // Don't try this at home
  template <typename T>
  Evil(T) {}
};

bool operator==(Evil const&, Evil const&) { return true; }

int main() {
  Innocent a, b;

  // Should never compile, right?
  // But it does, thanks to the Evil non-explicit templated constructor
  a == b;
}
```

I'll call it `Implicit Conversion Mania` (ICM).

We could not make the constructor `explicit`, it would have ruined the great syntax that one can use to construct `json` objects from other types.

So we had to add constraints to that constructor.

### The fix(?)

Ideally the constraint should be:

  * `void basic_json::JSONSerializer::to_json(basic_json&, T&&)` is valid

Note: `JSONSerializer` is the latest of `basic_json`'s template arguments, and defaults to `adl_serializer`.

### The workaround(?)
But there were more to add to prevent clashes with other constructors (e.g. the old `basic_json(std::istream&)`).
So we added those, and everything was looking good... Until compiling `unit-iterators1.cpp`.

There was a test on `primitive_iterator_t` which triggered the ICM.
After a rude fight with the error message, I found that Catch was doing some SFINAE checks to know if some type was printable.

I'll explain why that triggered the ICM on a later section.
At that point, I just wanted to make stuff work.

### The hack

After a bit of digging, I discovered that the only classes that were triggering the ICM were `basic_json` nested classes... (I couldn't find exactly why, I have to open a SO question about that).

So the ugly fix (which is still in the code today) was to add another constraint on the template constructor,
checking `T` against every `basic_json`'s nested class.

It's ugly, but it works well. So no problem right? Why should anyone care?

## The return of the ICM

A few days ago, one of my colleagues had a huge compiler error while writing some tests with `trompeloeil`.

That library was doing the same SFINAE checks that Catch is doing, which caused the compiler error that I've described ahead.

So I sighed.

And this time, the type that was in cause was: `boost::function<void (nlohmann::json const&)>`.

## WAT

[...]

PR #1067, opened 2018-04-24
Fix operator== overload ambiguity

th0br0:
Currently, `json` does not work well together with GoogleMock.
(e.g. `MOCK_METHOD0(sample, nlohmann::json(void));`)

This is because of an ambiguous overload which this PR fixes by disabling one of the two.
As `std::is_null_pointer` is only available from C++14, this uses the possible implementation from http://en.cppreference.com/w/cpp/types/is_null_pointer

Example error:
```
In file included from external/com_google_googletest/googlemock/include/gmock/gmock.h:58:
external/com_google_googletest/googlemock/include/gmock/gmock-actions.h:388:18: error: use of overloaded operator '==' is ambiguous (with operand types 'const internal::linked_ptr<ActionInterface<const basic_json<std::map, std::vector, basic_string<char, char_traits<char>, allocator<char> >, bool, long, unsigned long, double, std::allocator, adl_serializer> ()> >' and 'nullptr_t')
    return impl_ == nullptr && fun_ == nullptr;
           ~~~~~ ^  ~~~~~~~
SNIP

external/com_google_googletest/googletest/include/gtest/internal/gtest-linked_ptr.h:186:8: note: candidate function
  bool operator==(T* p) const { return value_ == p; }
       ^
json.h:15149:17: note: candidate function [with ScalarType = nullptr_t, $1 = 0]
    friend bool operator==(const_reference lhs, const ScalarType rhs) noexcept
                ^
json.h:15078:17: note: candidate function
    friend bool operator==(const_reference lhs, const_reference rhs) noexcept
```

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -5533,8 +5533,10 @@ class basic_json
     @brief comparison: equal
     @copydoc operator==(const_reference, const_reference)
     */
-    template<typename ScalarType, typename std::enable_if<
-                 std::is_scalar<ScalarType>::value, int>::type = 0>
+    template < typename ScalarType, typename std::enable_if <
+                   std::is_scalar<ScalarType>::value, int >::type = 0,
+               typename std::enable_if <
+                   std::is_same<std::nullptr_t, std::remove_cv<ScalarType>>::value, int >::type = 0 >
     friend bool operator==(const_reference lhs, const ScalarType rhs) noexcept
     {
         return (lhs == basic_json(rhs));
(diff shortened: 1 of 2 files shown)


nlohmann:
Is there a way to add a test for the change?

nlohmann:
@th0br0 Do you have an idea how to proceed with this PR?

nlohmann:
Any news on this?

nlohmann:
@th0br0 Any idea how to proceed here?

nlohmann:
@th0br0 As for a test case: couldn't you just see which call Google Mock made when you detected the error and use it as regression test for your fix?

theodelrieu:
I just stumbled on this issue, and it looks very similar to what I've described in #960.

Here, the compiler considers converting an internal GMock type to `json`, because `json` is inside its template arguments:

>const internal::linked_ptr<ActionInterface<const basic_json<...>>>

Since ADL looks at the template arguments too, so you have the same issue. I've explained in detail why it happens in #960.

nlohmann:
@th0br0 This issue could be fixed after merging #1228. Could you try using the latest `develop` version whether the issue is fixed now?

PR #1078, opened 2018-05-03
fix typo in README

martin-mfg:
The README says:
> "The library itself contains of a single header file ..."

I guess "consists of a" is the wording that was intended here. Alternatively, "contains a" might also be what was intended. Either way, I think both versions work equally well and thus choose the one that I believe was intended.


nlohmann:
Thanks!

Issue #1111, opened 2018-05-27
How to resolve  " undefined reference to `std::__throw_bad_cast()'"


TridentTD:
- What is the issue you have?

I can compile well on ESP8266's xtensa_gcc  compiler version5.1.0, 
however when I want to receive string by dump() function .
It occurs some errors. How to resolve it?

- Please describe the steps to reproduce the issue. Can you provide a small but working code example?

First I try this code 
```
json j = {  {"pi", 3.141}, {"happy", true}, {"name", "testtest"}, 
                    {"nothing", nullptr}, { "answer", { {"everything", 42} } },
                    {"list", {1, 0, 2}}, { "object", { {"currency", "USD"}, {"value", 42.99} } } };

Serial.println((float) j["pi"],3);     // print to Serial Monitor
```
The result  can print well on Serial Monitor -->  3.141   
However when I want to dump like the following.

```
std::string dump_string =  j.dump();
```

It occurs error like this

```
libstdc++.a(ostream-inst.o): In function `std::ostream& std::ostream::_M_insert<long>(long)':
libstdc++-v3/include/ostream:544: undefined reference to `std::__throw_bad_cast()'
libstdc++.a(wlocale-inst.o): In function `std::ctype<wchar_t> const& std::use_facet<std::ctype<wchar_t> >(std::locale const&)':
libstdc++-v3/include/bits/locale_facets_nonio.tcc:1173: undefined reference to `std::__throw_bad_cast()'
...
```

How to resolve   "undefined reference to `std::__throw_bad_cast()' " error ?

Thank you.

nlohmann:
As `wchar_t` are mentioned: the return type of `j.dump()` is `std::string` - I am not sure about `Serial.println`, but if it expects `wchar_t`, this may be a problem. Quick googling also found this issue: https://stackoverflow.com/questions/22263732/undefined-reference-to-stdbad-castbad-cast

Can you compile and execute the test suite?

TridentTD:
My Serial.println()  already know  std::string
![Serial_println_std_string](https://image.ibb.co/edoL8J/image.png)
  
But when all  j.dump() ,  it occurs  std::__throw_bad_cast()  error.
![dump_call](https://image.ibb.co/bxbhoJ/image.png)  
  
![dump_error](https://image.ibb.co/cLU5ay/image.png)  
  
How can I change about THROW or anything like that to null String when it can't casting?

nlohmann:
This seems to be an issue with the compiler.

[...]

PR #1117, opened 2018-05-31
remove stringstream dependency

TinyTinni:
Hi,
this PR replaces the  `stringstream` conversions with `snprintf` ones and removes all stringstream dependencies.
`snprintf` is already use in the project.

Issue reference: #1111

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -6,12 +6,10 @@
 #include <cmath> // ldexp
 #include <cstddef> // size_t
 #include <cstdint> // uint8_t, uint16_t, uint32_t, uint64_t
+#include <cstdio> // snprinft
 #include <cstring> // memcpy
-#include <iomanip> // setw, setfill
-#include <ios> // hex
 #include <iterator> // back_inserter
 #include <limits> // numeric_limits
-#include <sstream> // stringstream
 #include <string> // char_traits, string
 #include <utility> // make_pair, move
 
@@ -1665,9 +1663,9 @@ class binary_reader
     */
     std::string get_token_string() const
     {
-        std::stringstream ss;
-        ss << std::setw(2) << std::uppercase << std::setfill('0') << std::hex << current;
-        return ss.str();
+        char cr[3];
+        snprintf(cr, 3, "%.2X", current);
+        return std::string{cr};
     }
 
   private:
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -6,7 +6,7 @@
 #include <cmath> // ldexp
 #include <cstddef> // size_t
 #include <cstdint> // uint8_t, uint16_t, uint32_t, uint64_t
-#include <cstdio> // snprinft
+#include <cstdio> // snprintf
 #include <cstring> // memcpy
 #include <iterator> // back_inserter
 #include <limits> // numeric_limits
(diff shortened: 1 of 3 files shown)


include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -6,12 +6,10 @@
 #include <cmath> // ldexp
 #include <cstddef> // size_t
 #include <cstdint> // uint8_t, uint16_t, uint32_t, uint64_t
+#include <cstdio> // snprinft
```
gregmarr: typo: snprintf

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -1665,9 +1663,9 @@ class binary_reader
     */
     std::string get_token_string() const
     {
-        std::stringstream ss;
-        ss << std::setw(2) << std::uppercase << std::setfill('0') << std::hex << current;
-        return ss.str();
+        char cr[3];
+        snprintf(cr, 3, "%.2X", current);
+        return std::string{cr};
```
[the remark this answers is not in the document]
gregmarr: Except that C++11 made that undefined, and the non-const `data()` doesn't exist until C++17, so never mind.

include/nlohmann/detail/input/binary_reader.hpp
TinyTinni: fixed, thanks

include/nlohmann/detail/input/binary_reader.hpp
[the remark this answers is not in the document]
TinyTinni: Even with non-const `.data()` it is undefined.
> [The pointer is such that the range [data(); data() + size()) is valid](http://en.cppreference.com/w/cpp/string/basic_string/data)

The main problem is the null-termination of `snprintf`, which always results in a string with an +1 increased stringlength compared to the the string created before.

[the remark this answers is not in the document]
gregmarr: That's fine, `size()` is 2, so `cr[0]` to `cr[2]` is valid, and that's exactly what `snprintf` with a size of `3` will write to, including the nul terminator.  In any case, we can't use it, as we can only require C++11.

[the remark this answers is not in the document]
TinyTinni: Ah, I see. Scrolling down to the return value, the definition interval is closed 
> data() + i == &operator[](i) for every i in [0, size()]. | (since C++11)

while in the brief, the definition interval is only an open one which excludes [size()].

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -389,9 +386,9 @@ class serializer
 
                 case UTF8_REJECT:  // decode found invalid UTF-8 byte
                 {
-                    std::stringstream ss;
-                    ss << std::setw(2) << std::uppercase << std::setfill('0') << std::hex << static_cast<int>(byte);
-                    JSON_THROW(type_error::create(316, "invalid UTF-8 byte at index " + std::to_string(i) + ": 0x" + ss.str()));
+                    std::string sn(3, '\0');
+                    snprintf(&sn[0], sn.size(), "%.2X", byte);
```
tramseyer: How about changing this to:
`snprintf(&sn[0], sn.size(), "%.2X", static_cast<uint8_t>(s[i]));`

In order to avoid "‘%.2X’ directive output may be truncated writing between 2 and 8 bytes into a region of size 3 [-Wformat-truncation=]" (GCC 7.3.0).

TinyTinni: I don't see the difference.
`byte` is defined as
`const auto byte = static_cast<uint8_t>(s[i]);`
So it is 1 byte and the same, what you wrote.
Maybe a false positive in GCC?

tramseyer: I suppose it might be a false positive since there is no difference.
Maybe someone with the same GCC version can confirm this warning.

PR #1126, opened 2018-06-11
test (non)equality for alt_string implementation

agrianius:
Add a bit more tests for alt_string implementation

### DIFF
--- a/test/src/unit-alt-string.cpp
+++ b/test/src/unit-alt-string.cpp
@@ -33,6 +33,12 @@ SOFTWARE.
 #include <string>
 #include <utility>
 
+
+/* forward declarations */
+class alt_string;
+bool operator<(const char* op1, const alt_string& op2);
+
+
 /*
  * This is virtually a string class.
  * It covers std::string under the hood.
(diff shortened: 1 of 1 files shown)


test/src/unit-alt-string.cpp
```diff
@@ -60,17 +66,36 @@ class alt_string
     }
 
     template <typename op_type>
-    bool operator==(op_type&& op) const
+    typename std::enable_if< // disable for alt_string
+                !std::is_same<  alt_string,
+                                typename std::remove_reference<op_type>::type
+                             >::value,
+                bool>::type
```
theodelrieu: I think you can remove all of this and simply take the parameter as `const op_type&`.

agrianius: Yeap, you are right, thanx for the notice. I will make another request.

theodelrieu: No need to open a different PR by the way, pushing another commit will serve the same purpose.

agrianius: Oh, I did not know, thanx.

PR #1144, opened 2018-06-24
Fix links to cppreference named requirements (formerly concepts)

jrakow:
The link in https://github.com/nlohmann/json#stl-like-access is broken.
It points to https://en.cppreference.com/w/cpp/concept/ReversibleContainer.

"Concepts" have been renamed to "named requirements".
This is because P0898R3 Standard Library Concepts has been merged into
C++20: cplusplus/draft#2145.
Cppreference have moved their links accordingly.

This PR also links to HTTPS and not HTTP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,7 +288,7 @@ json j = json::parse(v);
 
 ### STL-like access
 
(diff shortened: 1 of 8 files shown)


nlohmann:
Thanks!

Issue #1148, opened 2018-06-27
cocoapod integration


graph:
Is this ok? If I do it can I push to this repo? I want to be able to more easily use this library on ios. Perhaps under the pod `njson`.

nlohmann:
That would be great, but I'd rather have a name like `nlohmann_json`, because this name is used with other package managers. See https://github.com/nlohmann/json#integration for an overview.

graph:
and published https://cocoapods.org/pods/nlohmann_json

It's easy to add additional authors, for now I just put you. I didn't put your email. If you want me to update the list in anyway let me know. It's also very easy to do so yourself in the podspec file. The programming language of the podspec is ruby.

nlohmann:
Thanks for taking care of this! Is PR #1150 needed after all if the library is already published?

graph:
your welcome. Theoretically PR is not needed. But it's good to keep all integration files in the main repo.

nlohmann:
I'd rather have the file in a repo controlled by the package maintainer (which I do not want to be). I cannot provide support for tools I do not use.

graph:
I understand no worries. It will probably be good to add to your readme something like

If you are using cocoapods, you can use the unofficial pod by adding `pod "nlohmann_json", '~>3.1.2'` to your podfile.

nlohmann:
Alright. Do you have a repository where you keep the pod? Then I can link it if someone has issues. Furthermore, I would need such an issue tracker to notify you about new releases.

graph:
Happy Canada day. I'll set something up next week. In the mean time I will enjoy long weekend :)

nlohmann:
I would also need a small running example (something demonstrating how to compile a small program using the library) for the pod.

graph:
Ok I added a sample project too. I have set it up on my bitbucket, as [nlohmann_json-cocoapod](https://bitbucket.org/benman/nlohmann_json-cocoapod/src/master/). People can submit issues there in the [issue tracker](https://bitbucket.org/benman/nlohmann_json-cocoapod/issues?status=new&status=open)

nlohmann:
Thanks a lot!

PR #1150, opened 2018-06-27
cocoapods integration

graph:
adds cocoapods integration so it can be used as a pod. See issue https://github.com/nlohmann/json/issues/1148

### DIFF
--- a/nlohmann_json.podspec
+++ b/nlohmann_json.podspec
@@ -1,35 +1,28 @@
-#
-#  Be sure to run `pod spec lint nlohmann_json.podspec' to ensure this is a
-#  valid spec and to remove all comments including this before submitting the spec.
-#
-#  To learn more about Podspec attributes see http://docs.cocoapods.org/specification.html
-#  To see working Podspecs in the CocoaPods repo see https://github.com/CocoaPods/Specs/
-#
+# upgrades how to:
+# 1. update s.version
+# 2. run `pod spec lint nlohmann_json.podspec`
+# 3. register a session with cocoapods
+# 4. run `pod trunk push ./`
 
+# see cocoapods documentation http://docs.cocoapods.org/specification.html for
+# more details.
 Pod::Spec.new do |s|
 
   # ―――  Spec Metadata  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
   s.name         = "nlohmann_json"
   s.version      = "3.1.2"
-  s.summary      = "JSON reader & writer for C++ optimized for ease of use and modern C++14"
+  s.summary      = "JSON for Modern C++"
 
 
   # CocoaPods strips indentation for description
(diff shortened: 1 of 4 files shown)


nlohmann:
Is it possible to put the source files in a different directory?

graph:
yes where do you want the dummy source files to be?

graph:
interesting dummy files aren't needed. I was using an older version of cocoapods. It seems cocoapods linter is passing with just header only.

nlohmann_json.podspec
```diff
@@ -0,0 +1,89 @@
...
+#  To learn more about Podspec attributes see http://docs.cocoapods.org/specification.html
+#  To see working Podspecs in the CocoaPods repo see https://github.com/CocoaPods/Specs/
+#
+
+Pod::Spec.new do |s|
+
+  # ―――  Spec Metadata  ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
+  s.name         = "nlohmann_json"
+  s.version      = "3.1.2"
+  s.summary      = "JSON reader & writer for C++ optimized for ease of use and modern C++14"
```
nlohmann: That's not really the summary of the library. Please use "JSON for Modern C++" or something. And the library also supports C++11.

nlohmann_json.podspec
```diff
@@ -0,0 +1,89 @@
...
+
+
+  # CocoaPods strips indentation for description
+  s.description  = <<-DESC
+    JSON library for c++, it can read & write, uses C++14 features to have
```
nlohmann: See previous comment.

nlohmann_json.podspec
```diff
@@ -0,0 +1,89 @@
...
+  #  of the authors are extracted from the SCM log. E.g. $ git log. CocoaPods also
+  #  accepts just a name if you'd rather not provide an email address.
+  #
+  #  Specify a social_media_url where others can refer to, for example a twitter
+  #  profile URL.
+  #
+
+
+  s.authors            = {
+    "Niels Lohmann" => "nosend@respectmyprivacy.com"
```
nlohmann: My email address is <mail@nlohmann.me>.

nlohmann_json.podspec
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  # ――― Platform Specifics ――――――――――――――――――――――――――――――――――――――――――――――――――――――― #
+  #
+  #  If this Pod runs only on iOS or OS X, then specify the platform and
+  #  the deployment target. You can optionally include the target after the platform.
+  #
+
+  s.platform     = :ios, "8.0"
```
nlohmann: macOS should also be fine, right?

graph: I only tested iOS, theoretically it should be fine on macOS, watch and all other platforms. I don't want to add untested platforms.

PR #1210, opened 2018-08-23
Parse error lines cols

efp:
Added line and column number to parse error

- defined input position struct in exceptions.hpp (maybe there is a better place?)
- used by lexer to record line / column (based on '\n')
- used by parse_error to report line / column

I removed the raw char count from the error message; column will work if there are no lines. Not sure if we might still want to include it.

The position is added as a public member of the exception; the client can catch & extract it if they want.

I left a parse_error::create function that only takes the char count; wasn't sure what to do with, e.g., the binary reader.

All tests passed / amalgamated.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/exceptions.hpp
+++ b/include/nlohmann/detail/exceptions.hpp
@@ -23,6 +23,9 @@ struct input_position {
         lines_read(lines),
         chars_read_this_line(chars_this_line)
     {}
+    
+    //! Converts to chars_read
+    operator size_t() {return chars_read; }
 } ;
 
 /*!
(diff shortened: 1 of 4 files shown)


nlohmann:
@efp Can you check the errors that Travis and AppVeyor detected?

efp:
Aha, I didn't run amalgamate before the unit tests. Have to update some call signatures in the tests.

efp:
Gah, I'm down to having to change all the unit test strings from parse error. 

I think it'd be better if the exception class stored the pieces of the error message... along with 'id', 'type', 'desc', etc., and tested against those, instead of comparing string literals. Then you could tweak the error messages without re-writing any unit tests.

To not have to deal with changing interfaces in sax_parse (I don't know what that is for) I made an explicit conversion from input_position to size_t (#chars read).

efp:
I feel like I don't know the use cases well enough to know how best to proceed.

Thus far, only parse errors thrown from parser.hpp will report the line col format (not from binary_reader or json_sax). That was all I needed for my use case--just reading a json text file. So one approach is to change the unit test strings just for those. I did one such case in the previous commit. 

Or perhaps we'd want all the errors to give line and col; in which case I'd need to change the call signatures from the other objects and change all the unit test strings. Most of them only test one line, so could be mass replaced, but some have '\n' characters in there.

In either case, a few additional tests to see the lines/cols get processed correctly would be in order.

nlohmann:
> Or perhaps we'd want all the errors to give line and col; in which case I'd need to change the call signatures from the other objects and change all the unit test strings. Most of them only test one line, so could be mass replaced, but some have '\n' characters in there.

Which errors do you mean?

nlohmann:
@efp What can we do to get this PR forward? Do you need any assistance?

efp:
Sorry, I got stuck on the question of what to do about the error message unit tests. Since *some* of them changed.

I'm just not familiar enough with all the use cases to know the best way to proceed. All I ever do is read a JSON text file. In which case, I like line * col of the error reported.

The *easiest* way to proceed would be to not change the error messages, but just let the user catch a parse_error and use the stored line & col information if they so please.

nlohmann:
I see. Would you mind if I would try to make a PR myself?

efp:
Gosh no, have at it.

nlohmann:
@efp I opened #1282 with your ideas. It would be great if you could have a look. Then maybe we can close this PR and focus on #1282.

Issue #1212, opened 2018-08-24
Compilation error when using MOCK_METHOD1 from GMock and nlohmann::json


urbanekpj:
- What is the issue you have?
Compilation error
- Please describe the steps to reproduce the issue. Can you provide a small but working code example?
code :
```
#include "stdafx.h"
#include <gmock/gmock.h>
#include <nlohmann/json.hpp>

using Json = nlohmann::json;
//using Json = std::wstring;
namespace tests {
/**/
class GMockTest {
public:
    GMockTest() = default;
    virtual ~GMockTest() = default;
    virtual void read(const Json& data){

    };
};

class GMockTestMock final : public GMockTest {
public:
    ~GMockTestMock() override{};

    MOCK_METHOD1(read, void(const Json&));
};

using ::testing::_;
using ::testing::AnyNumber;
using ::testing::Invoke;
TEST(RequestClient, read_test)
{
    auto io = std::make_shared<GMockTestMock>();
    EXPECT_CALL(*io, read(_)).WillRepeatedly(Invoke([&](const Json& data) {
    }));
}
}
```

- What is the expected behavior?
code compiled
- And what is the actual behavior instead?
compilation errors:

> 1> \x86-windows-static\include\gmock\gmock-actions.h(388): error C2666: 'testing::internal::linked_ptr<testing::ActionInterface<F>>::operator ==': 3 overloads have similar conversions
1>        with
1>        [
1>            F=void (const nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer> &)
1>        ]
1> \x86-windows-static\include\gtest\internal\gtest-linked_ptr.h(186): note: could be 'bool testing::internal::linked_ptr<testing::ActionInterface<F>>::operator ==(T *) const'
1>        with
1>        [
1>            F=void (const nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer> &),
1>            T=testing::ActionInterface<void (const nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer> &)>
1>        ]
[...]

Issue #958, opened 2018-02-04
operator T() considered harmful


theodelrieu:
# `operator T()` considered harmful

There's two subjects of importance that I'd like to tackle.
This is the first one, I'll open another issue quickly (hopefully).

## The problem

There's been several issues in the past related to the implicit conversion operator that the library provides.
The current Readme demonstrates the following use-case:

```cpp
// conversion: json -> person
ns::person p2 = j;
```

This will call `json::operator T()` with `T = ns::person`, so no problems here.

You'd expect the following code to always work too:

```cpp
ns::person p2;
p2 = j;
```

And it works! Sometimes.

If `ns::person` has a user-defined `operator=`, it will break with the following message:

```shell
t.cpp: In function ‘int main(int, const char**)’:
t.cpp:14:7: error: ambiguous overload for ‘operator=’ (operand types are ‘ns::person’ and ‘nlohmann::json {aka nlohmann::basic_json<>}’)
   p2 = j;
       ^
t.cpp:6:6: note: candidate: ns::person& ns::person::operator=(int)
   ns::person& operator=(int) { return *this; }
      ^~~~~~~~
t.cpp:3:8: note: candidate: constexpr ns::person& ns::person::operator=(const ns::person&)
 struct ns::person
        ^
t.cpp:3:8: note: candidate: constexpr ns::person& ns::person::operator=(ns::person&&)
```

Hard to understand that error, and it's not something that can be fixed by the library.
It's triggered by the compiler before any template machinery on our side.

Now with such code:

```cpp
void need_precise_measurement(Milliseconds);

// later on
json j = json::parse(config);
need_precise_measurement(j);
```

It can fail in two cases:

1. Someone adds an overload for `need_precise_measurement`. Compiler error.
2. Someone changes the type of the argument to `Nanoseconds`. Runtime error at best.

## Implicit conversions

Implicit conversions make the library pleasant to use. Especially when implicitely converting **to** JSON.
That's because we have a templated non-explicit constructor with lots of constraints to correctly handle user types.
There's also a single type to convert to: `nlohmann::json`, so everything works.

However, when converting implicitely **from** json, it's the complete opposite. If the compiler cannot decide which type it should implicit convert the `nlohmann::json` value to, you get a compiler error.

And it could happen in a codebase months after writing the code, (e.g. when adding an `operator=` to a type).

In the end the only consistent way to convert from a json value is to call `json.get<T>`.

## Proposed change

I propose to simply and completely remove `operator T()`.

[...]

Issue #561, opened 2017-04-19
from_json declaration order and exceptions


andrewbleonard:
I am running into an issue where if a from_json function I wrote does a conversion that uses another from_json function I wrote, whether or not I can catch exceptions from nlohmann::json depends on the order of the from_json functions.  Repro code below:

```cpp
#include <exception>
#include <fstream>
#include <iostream>

#include "json.hpp"

struct A {
    int value;
};

struct B {
    A a;
};

void from_json(const nlohmann::json& j, B& b) {
    b.a = j.at("test").get<A>();
}

// If this is moved before the from_json for B, I can catch the exception below
void from_json(const nlohmann::json& j, A& a) {
    a.value = j.at("value").get<int>();
}

int main() {
    try {
        std::ifstream stream("test.json");
        nlohmann::json j;
        stream >> j;
        B b = j;
    } catch (const std::exception& e) {
        std::cerr << "caught exception: " << e.what() << std::endl;
    }
    return 0;
}
```

My test.json file is below:
```cpp
{
    "test": {
        "value": "hello"
    }
}
```

In the above example, the output is:
```
libc++abi.dylib: terminating with uncaught exception of type std::domain_error: type must be number, but is string
Abort trap: 6
```

If I switch the order of the from_json functions, the output is as expected:
```
caught exception: type must be number, but is string
```

I am pretty new to C++11 and C++ exceptions, so sorry if the reason for this is obvious.

My platform/compiler is OSX 10.11.6 with AppleClang 8.0.0.8000042 and I tested this using json version 2.1.1.

nlohmann:
I can confirm you observation, but I also have no idea. Maybe a language lawyer like @gregmarr or @theodelrieu knows more about this.

theodelrieu:
I'm far away of being a language lawyer, I will take a look, this looks really weird

theodelrieu:
Well well well, there is something fishy, looks like a dark corner of C++.

I have no idea why, but it seems that during the evaluation of the first `from_json` method's body, the compiler instantiates a lot of templates, one being the `from_json_fn::operator()()`.

This method has a `noexcept(noexcept(to_json(bla, bla)))` construct. The thing is, there is no method `void to_json(json const& , B&)` defined, it is defined later in the translation unit. However, it stills compiles, no idea why.

I printed the value of the `noexcept(noexcept())` clause in the definition of `nlohmann::detail::from_json_fn::operator()()`, and the output is .... `true`.

[...]

PR #1228, opened 2018-09-06
Remove static asserts

theodelrieu:
This PR addresses issues #960.

There is some refactoring in `detail/meta`, the checks we perform were too brittle (I fixed some issues in the mean time).

At some point I'd like to iron out the quirks in `meta`, (e.g. `is_compatible_*` traits, etc...), but this is not the goal of this PR.

I will sum up the pros/cons of this PR that can be found in detail in #960:

Pros:
* `nlohmann::json` is now *more* SFINAE-correct. The library has better interoperability with other generic libraries. (Not 100% because of #958 though)
* No API breakage
* Correct fix for #1067 #1212, #561

Cons:
* No more `static_assert`, instead the compiler will yell that no constructor could be used.

* * *

## Pull request checklist


include/nlohmann/adl_serializer.hpp
```diff
@@ -20,8 +20,10 @@ struct adl_serializer
     @param[in,out] val  value to write to
     */
     template<typename BasicJsonType, typename ValueType>
-    static void from_json(BasicJsonType&& j, ValueType& val) noexcept(
-        noexcept(::nlohmann::from_json(std::forward<BasicJsonType>(j), val)))
+    static auto from_json(BasicJsonType&& j, ValueType& val) noexcept(
+        noexcept(::nlohmann::from_json(std::forward<BasicJsonType>(j), val))) -> decltype(
+            ::nlohmann::from_json(std::forward<BasicJsonType>(j), val)
+        )
```
gregmarr: You need to add a `return` here or it will be malformed if the function you call has a return type.

theodelrieu: In this case `::nlohmann::from_json` returns `void`, but I should add `, void()` in the decltype to be extra sure.

gregmarr: Why do you need the `decltype` at all?  Is that for SFINAE?  Is that not covered by the expression in the `noexcept`?

theodelrieu: I thought so at first, but doing a qualified call inside a `noexcept` breaks. The `decltype` is mandatory here.

gregmarr: Okay, thanks.

nlohmann:
To me, the PR is SFIENA magic... :-S

@gregmarr Do you think the PR is OK to merge?

(I do trust @theodelrieu, but I really fail to understand the individual changes...)

gregmarr:
Those things that I can follow look good.  I guess as long as the test suite is still passing, we should be good.

nlohmann:
Thanks a lot!

PR #1230, opened 2018-09-10
Move lambda out of unevaluated context

### LINKED ISSUE #1225: Lambda in unevaluated context
I've seen #705 but it doesn't appear to be fixed. I'm unfortunately not on bleeding edge (xtensa gcc 5.2.0) so can't use newer gcc versions where this has been fixed. Can we fix this? I can make a PR.

```
~/code/work/esp32-core/main/json/single_include/nlohmann/json.hpp: In constructor 'nlohmann::detail::input_adapter::input_adapter(IteratorType, IteratorType)':
~/code/work/esp32-core/main/json/single_include/nlohmann/json.hpp:2165:20: error: lambda-expression in unevaluated context
                    [&first](std::pair<bool, int> res, decltype(*first) val)
```

mandreyel:
Fixes #1225.

As per issue #1225, the lambda has been moved out of the assert to prevent the "lambda-expression in unevaluated context" error that is present with older compilers, and has been wrapped in an `#ifdef DEBUG` directive to keep production and debug builds the same.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/input/input_adapters.hpp
+++ b/include/nlohmann/detail/input/input_adapters.hpp
@@ -320,15 +320,18 @@ class input_adapter
                  int>::type = 0>
     input_adapter(IteratorType first, IteratorType last)
     {
+#ifndef NDEBUG
         // assertion to check that the iterator range is indeed contiguous,
         // see http://stackoverflow.com/a/35008842/266378 for more discussion
-        assert(std::accumulate(
-                   first, last, std::pair<bool, int>(true, 0),
-                   [&first](std::pair<bool, int> res, decltype(*first) val)
+        const auto is_contiguous = std::accumulate(
+                                       first, last, std::pair<bool, int>(true, 0),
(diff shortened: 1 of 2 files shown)


include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -320,15 +320,18 @@ class input_adapter
                  int>::type = 0>
     input_adapter(IteratorType first, IteratorType last)
     {
+#ifdef DEBUG
```
nlohmann: I am sorry, I wrote it wrong in the original issue. This should be `#ifndef NDEBUG`, see https://en.cppreference.com/w/cpp/error/assert.

mandreyel: Ah, that's what I thought, should have asked. Brb then!

nlohmann:
Thanks a lot!

mandreyel:
Thank *you* for the great lib!

PR #1245, opened 2018-09-18
Use a version check to provide backwards comatible CMake imported target names

### LINKED ISSUE #1243: Cmake target_link_libraries unable to find nlohmann_json since version 3.2.0
We updated from `3.1.2` to `3.2.0`. The update causes the `target_link_libraries` cmake command to not properly include `nlohmann_json`, while `find_package(nlohmann_json REQUIRED)` did not result in an error.

After some investigation, it seems that the issue was the addition of namespaces to target names which is a **backward incompatible** change.

chuckatkins:
Fixes #1243


nlohmann:
Thanks a lot for the quick reaction!

SylvainCorlay:
I like the fix. Did not know about PACKAGE_FIND_VERSION.

SylvainCorlay:
@nlohmann @chuckatkins this does not appear to work. People are complaining that they cannot build against master and getting.

```
CMake Error at nlohmann_jsonConfig.cmake:30 (add_library):
  add_library cannot create ALIAS target "nlohmann_json" because target
  "nlohmann_json::nlohmann_json" is IMPORTED.
Call Stack (most recent call first):
  CMakeLists.txt:33 (find_package)
```

I think that we should revert that change.

chuckatkins:
Bad on my part :-(.  I'll push a fix today, but also add a test to validate import configs to prevent future breakage.

chuckatkins:
@ericu65 this should be fixed by #1260

ericu65:
@chuckatkins Thanks. On first look this would seems to fix it. Looking forward to a merge.

PR #1249, opened 2018-09-19
#1179 Reordered the code. It seems to stop clang 3.4.2 in RHEL 7 from crash…

### LINKED ISSUE #1179: Compiler crash with old Clang
I am building the library using Clang 3.4.2 (I don't currently have access to a more recent version) and it is crashing the compiler. The crash goes away if I disable the specialization of std::swap at line 17194, and my app still seems to work properly, but I wanted to get a second opinion about how safe this patch actually is...

LEgregius:
Closes #1179.

#1179  I moved code around a bit because the swap specialization. I tried running the tests, but they wouldn't compile and link due to some issues with the way the libraries are linked on RHEL 7. I can try it on another computer if issues arise, but since there shouldn't be any change in functionality because code was simply moved around, I don't expect there to be any problems with tests.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -7691,19 +7691,6 @@ class basic_json
 // specialization of std::swap, and std::hash
 namespace std
 {
-/*!
-@brief exchanges the values of two JSON objects
-
-@since version 1.0.0
-*/
-template<>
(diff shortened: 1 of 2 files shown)


nlohmann:
Thanks a lot!

Issue #1244, opened 2018-09-18
BSON support


julian-becker:
It would be great to have BSON included support in this great JSON library! 

**Motivation:**
I recently encountered the need for it at work, where we are already using this library for JSON processing (for many diverse purposes). Recently, we have been developing applications leveraging the  [rosbridge](https://github.com/RobotWebTools/rosbridge_suite) (configured in BSON mode) for communication between devices. If we can cover the BSON support with the same JSON library we are already using this would be much preferred over introducing a new library into the mix.

**First Prototypical Implementation:**
Actually, I have already started a [quick, prototypical implementation](https://github.com/julian-becker/json/blob/feature/bson/test/src/unit-bson.cpp) for BSON serialization and deserialization, and if you feel BSON-support should become part of the mainline, it could be used as a starting point, and I'd be willing to help with the implementation.

**Open Questions:**
There is a certain fundamental format mismatch between JSON and BSON such that not all JSON values can be mapped 1:1 into BSON and vice versa. 
Before continuing with an implementation, it should be clarified how these conversions should be handled (e.g. BSON timestamps, BSON 128-bit wide floats, or "naked" JSON-values which are not objects)

nlohmann:
I had a look at BSON before, but struggled with the absence of a 1:1 mapping between JSON and BSON. Right now, we already have some issues with MessagePack and CBOR which both support binary values (see #1129 for a recent issue), but I think we can be pragmatic there. What I would like to avoid is adding support for very specific types such as timestamp or UUIDs and rather focus on those values supported by JSON.

I did not find the time to look at your implementation. Once I have a rough idea, we should continue the discussion. And of course, PRs are welcome.

nlohmann:
I head a look at the prototype - very nice! It's a pity that BSON uses little endian and the other formats big endian. So I guess we may need to add a parameter to the existing read/write number functions to avoid having two functions.

I think it would be best if you could open a PR. Then we can shift the discussion related concrete code examples to reviews there.

julian-becker:
Great, thanks for the welcoming feedback! I will do a little consolidation on the code and will open a PR within the next few days, and we can discuss the details there.

[...]

PR #1254, opened 2018-09-24
Experimental support for BSON (de)serialization

julian-becker:
This PR includes basic support for BSON (de)serialization, as proposed in #1244.

It includes support for BSON records and record-entries of the following types (c.f. [bsonspec](http://bsonspec.org/spec.html)):
- 0x01: double
- 0x02: string
- 0x03: document
- 0x04: array
- 0x08: boolean
- 0x0A: null
- 0x10: int32
- 0x12: int64

It presently does not include support for the following BSON entry types:
- 0x05: binary
- 0x06: undefined (deprecated)
- 0x07: ObjectId
- 0x09: UTC Date-Time
- 0x0B: Regular Expression
- 0x0C: DB Pointer
- 0x0D: JavaScript Code
- 0x0E: Symbol (deprecated)
- 0x0F: JavaScript code w/ scope
- 0x11: Timestamp
- 0x13: 128-bit decimal floating point
- 0x7F: Max Key
- 0xFF: Min Key

The tests for the supported conversions can be found in unit-bson.cpp.
Feedback and discussions welcome, in particular with respect to the parts of BSON which are presently not yet supported.

For the initial discussion of this change and easier review/smaller patch, I have restrained as much as possible from modifying existing code. In particular, I have restrained from refactorings in the `binary_reader` and `binary_writer`, which could be beneficial to improve separation of concerns (not mixing procedures targeting different file formats in a single class). 

 Let me know what you think and I will happily adjust the code where needed/desired.


julian-becker:
@nlohmann I'm done so far with the consolidation of the code. Feedback and ideas for improvement welcome. 
Also, regarding the support of the BSON-specific types not yet supported: If you want, I could extend the PR to include BSON --> JSON one-trip conversions for some of these (I'm guessing some of these could be simply mapped to a string, the timestamp could be just mapped to an integer).

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -120,6 +124,139 @@ class binary_reader
...
+            {
+                return true;
+            }
+            *out++ = static_cast<char>(current);
+        }
+
+        return true;
+    }
+
+    bool parse_bson_entries(bool is_array)
```
nlohmann: Please document the parameter and return value.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -120,6 +124,139 @@ class binary_reader
...
+
+            switch (entry_type)
+            {
+                case 0x01: // double
+                {
+                    double number;
+                    get_number<double, true>(number);
+                    sax->number_float(static_cast<number_float_t>(number), "");
+                }
+                break;
```
nlohmann: Please put the `break;` into the scope above. This affects all `switch` statements.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -120,6 +124,139 @@ class binary_reader
...
+                {
+                    parse_bson_internal();
+                }
+                break;
+                case 0x04: // array
+                {
+                    parse_bson_array();
+                }
+                break;
+            }
```
nlohmann: What about all the other values for `entry_type`?

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -676,26 +676,231 @@ class binary_writer
...
+    /*!
+    @param[in] j  JSON value to serialize
+    @pre       j.type() == value_t::object
+    */
+    void write_bson(const BasicJsonType& j)
+    {
+        switch (j.type())
+        {
+            default:
+                JSON_THROW(type_error::create(317, "JSON value cannot be serialized to requested format"));
```
nlohmann: Please add the type of the value to the exception text.

include/nlohmann/json.hpp
```diff
@@ -6590,6 +6590,26 @@ class basic_json
         binary_writer<char>(o).write_ubjson(j, use_size, use_type);
     }
 
+
+
+    static std::vector<uint8_t> to_bson(const basic_json& j)
```
nlohmann: Please add documentation.

include/nlohmann/json.hpp
```diff
@@ -6897,6 +6917,34 @@ class basic_json
         return res ? result : basic_json(value_t::discarded);
     }
 
+
+
+
+
+    static basic_json from_bson(detail::input_adapter&& i,
```
nlohmann: Please add documentation.

include/nlohmann/json.hpp
```diff
@@ -6897,6 +6917,34 @@ class basic_json
...
+                                const bool strict = true,
+                                const bool allow_exceptions = true)
+    {
+        basic_json result;
+        detail::json_sax_dom_parser<basic_json> sdp(result, allow_exceptions);
+        const bool res = binary_reader(detail::input_adapter(i)).sax_parse(input_format_t::bson, &sdp, strict);
+        return res ? result : basic_json(value_t::discarded);
+    }
+
+    template<typename A1, typename A2,
```
nlohmann: Please add documentation.

julian-becker: done

include/nlohmann/json.hpp
julian-becker: done

include/nlohmann/json.hpp
julian-becker: done

include/nlohmann/detail/input/binary_reader.hpp
julian-becker: added a `default:` clause which will raise a new parse_error.114: Unsupported BSON record type. Also improved the error handling for the other cases, where in most occasions, the return values were ignored before.

include/nlohmann/detail/input/binary_reader.hpp
julian-becker: done

include/nlohmann/detail/input/binary_reader.hpp
julian-becker: done

include/nlohmann/detail/output/output_adapters.hpp
```diff
@@ -18,6 +18,8 @@ template<typename CharType> struct output_adapter_protocol
 {
     virtual void write_character(CharType c) = 0;
     virtual void write_characters(const CharType* s, std::size_t length) = 0;
+    virtual void write_characters_at(std::size_t position, const CharType* s, std::size_t length) = 0;
```
nlohmann: I am a bit worried about this new function. So far, the output adapter interface was similar to an output iterator, but adding the `write_characters_at` function, it is more a random-access iterator. I understand that BSON needs to precede values with their byte size, but I fear that we may exclude output adapters in the future. If you have an idea how to realized writing BSON without `write_characters_at`, I would be happy.

If not we may keep it as is and cope with this in the future.

include/nlohmann/json.hpp
nlohmann: Could you try to be more verbose, like in the `static std::vector<uint8_t> to_cbor(const basic_json& j)` function? I know this is a lot of work, but such a documentation saves a lot of issues in the future.

include/nlohmann/detail/output/output_adapters.hpp
julian-becker: I see your point. I guess there are several options available (beside keeping it as is):
1. Buffer internally (e.g. using and `std::vector<•>`), fix the size in the buffer at the end and then flush the buffer to the output iterator.
2. Find a way to estimate the size beforehand.

Option 1 is probably easier to implement but I dislike it because it requires a potentially very large buffer only to be able to fix a few bytes. Option 2 would be preferable, but I'm not sure yet how to implement it (given my limited familiarity with this code-base). This OTOH has the disadvantage of requiring an extra traversal of a potentially very large structure.

**Question:** Is there presently a convenient way to run a visitor (to calculate the necessary size) over the JSON object tree? 

I will think on the second option some more...

include/nlohmann/json.hpp
julian-becker: OK, I will expand the documentation.

julian-becker: added additional information and details regarding the type mapping

include/nlohmann/detail/output/output_adapters.hpp
julian-becker: Reworked the implementation to precompute the size of the BSON output, thus avoiding the additional random-access output-iterator requirement.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -676,26 +676,231 @@ class binary_writer
...
+        return /*id*/ 1ul + name.size() + 1ul + embedded_document_size;
+    }
+
+    std::size_t write_bson_object_entry(const typename BasicJsonType::string_t& name, const BasicJsonType& j)
+    {
+        switch (j.type())
+        {
+            // LCOV_EXCL_START
+            default:
+                assert(false);
```
nlohmann: In case of a discarded value, we should not raise an assertion, but either do nothing or throw an exception.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -676,26 +676,231 @@ class binary_writer
...
+    }
+
+    /*!
+    @param[in] j  JSON value to serialize
+    @pre       j.type() == value_t::object
+    */
+    std::size_t write_bson_object(const BasicJsonType& j)
+    {
+        assert(j.type() == value_t::object);
+        auto document_size_offset = oa->reserve_characters(4ul);
```
nlohmann: Could you please document the "magic numbers" like `4ul` or `5ul` below? I think a line in the comment for the function would be sufficient to understand.

include/nlohmann/detail/output/output_adapters.hpp
nlohmann: Great!

include/nlohmann/detail/output/binary_writer.hpp
julian-becker: good point! I adjusted the code for this case and added a test case to check that `discarded` values are actually discarded.

julian-becker:
@nlohmann No worries, that's what a review is for and that's why it's useful. Your feedback is very welcome, and I will do my best to tidy things up.
Fixing things once they are broken in the field will be a lot more painful than just fixing it now ;-)

nlohmann:
@julian-becker I think the PR is ready to merge. But I may need to make some more adjustments, add some tests, documentation, and fuzzers to have BSON on par with the other binary formats. Would you be available to work on such related tasks?

julian-becker:
@nlohmann I will start by looking into some more thorough and intensive tests over the next couple of days ... After that, I could check the documentation.

nlohmann:
Alright - if you look at the CBOR test suite, you know what I mean. It would be great if some of the existing test files could be converted to BSON. And if a well-accepted BSON test suite exists, it would be great to have it in the project as well. I would still want to merge the branch once you have nothing more to add to the parser, because then Google's OSS Fuzz can already start searching for bugs.

julian-becker:
@nlohmann My ongoing efforts for more intense tests revealed one thing that may need addressing before merging: 
The handling of large `std::uint64_t` values which exceed `INT64_MAX`. Those cannot presently be serialized to BSON and deserialized again without loss in information. While the binary representation survives the roundtrip to BSON and back to JSON, the type-information cannot be restored. Presently, the implementation silently (may be considered a bug) translates those values to `std::int64_t` with the result that large positive `std::uint64_t` values end up being negative `std::int64_t` when reading back from BSON to JSON. 
Maybe we should throw an error instead that informs the client that the corresponding large unsigned value cannot be serialized to BSON. 
What is your opinion on this issue? If you think the presently implemented behavior is acceptable, feel free to merge, otherwise I can add corresponding additional error-reporting.

nlohmann:
So BSON is does not support integers larger than `INT64MAX`? This sounds similar to UBJSON where a similar situation is currently treated with an exception, see https://github.com/nlohmann/json/issues/1288#issuecomment-430293411.

julian-becker:
As far as I understand the spec, there is only a signed int32 (type 0x10), and a signed int64 (type 0x12). There is an uint64 (type 0x11), however, this is to be interpreted explicitly as a timestamp, so is not really an option. 
I think throwing a `json.exception.out_of_range.407` is reasonable here and we should try to make it consistent with the handling of such cases UBJSON. I will adjust the code accordingly.

julian-becker:
also: keys are encoded as c-strings, with the effect that code point U+0000 cannot be serialized. This is not yet caught at the moment and will result in incorrect results if U+0000 is contained. I should probably adjust the error-handling to catch & report this.

test/src/unit-bson.cpp
```diff
@@ -0,0 +1,1115 @@
...
+            // discarded values are not serialized
+            json j = json::value_t::discarded;
+            const auto result = json::to_bson(j);
+            CHECK(result.empty());
+        }
+
+        SECTION("null")
+        {
+            json j = nullptr;
+            REQUIRE_THROWS_AS(json::to_bson(j), json::type_error);
```
nlohmann: Could you please add a `CHECK_THROWS_WITH` clause that checks the string of the exception? Please do this for all exceptions in unit tests.

nlohmann: Please replace `json::type_error` with `json::type_error&` to avoid warnings. Please do this for all exceptions in unit tests.

test/src/unit-bson.cpp
```diff
@@ -0,0 +1,1115 @@
...
+    }
+
+    SECTION("keys containing code-point U+0000 cannot be serialized to BSON")
+    {
+        json j =
+        {
+            { std::string("en\0try", 6), true }
+        };
+        REQUIRE_THROWS_AS(json::to_bson(j), json::out_of_range);
+        CHECK_THROWS_WITH(json::to_bson(j), "[json.exception.out_of_range.409] BSON key cannot contain code point U+0000");
```
nlohmann: Is it possible to add an offset (i.e., the byte position) to simplify understanding and fixing the error?

julian-becker: Good point! I adjusted the code accordingly to report the position within the key-string.

test/src/unit-bson.cpp
julian-becker: done

julian-becker: done

test/src/unit-bson.cpp
```diff
@@ -48,46 +48,53 @@ TEST_CASE("BSON")
         SECTION("null")
         {
             json j = nullptr;
-            REQUIRE_THROWS_AS(json::to_bson(j), json::type_error);
+            REQUIRE_THROWS_AS(json::to_bson(j), json::type_error&);
+            CHECK_THROWS_WITH(json::to_bson(j), "[json.exception.type_error.317] JSON value of type 0 cannot be serialized to requested format");
```
nlohmann: I don't think "type 0" etc. is a helpful description for users of the library. Please use `type_name()` to explicitly name the type (null, boolean, etc.).

Furthermore, "to requested format" should be replaced by "BSON" to be explicit.

nlohmann:
I only had another small comment to the latest commit.

Furthermore, it would be great if you could add a BSON translation to the files

- test/data/json_testsuite/sample.json
- test/data/json_tests/pass3.json
- test/data/json.org/*.json

with a `.json.bson` extension. These files can then be used to test round-tripping (see the CBOR test suite). Furthermore, the `.bson` files can be used by the fuzz testers.

nlohmann:
@julian-becker I have some time to work on this this week. I branched your branch and make the changes there myself.

nlohmann:
I created branch https://github.com/nlohmann/json/compare/julian-becker-feature/bson and shall make more changes there.

julian-becker:
@nlohmann Sure, feel free to continue this work. Sorry that I have not found as much time as I had hoped to put into this. One thing I noticed though was that for large containers, things get rather slow... which I suspect is mostly due to the size-precomputation. Thinking about the present implementation some more, I came to realize that -- in the worst case -- it is of order O(n^2), for the case of a deeply nested object containing an object containing an object ...

nlohmann:
No worries - I thank you for your work!

We may speed up the code by adding a cache like `std::unordered_map<const json*, size_t>` to avoid re-calculation of sizes.

nlohmann:
There may be a problem in the current code:

I added the examples from http://bsonspec.org/faq.html as tests. The first example is OK, but the second example does not properly roundtrip: Parsing the BSON input, we get the correct JSON value. But serializing it again yields

```
  { '.', 0, 0, 0, 4, 'B', 'S', 'O', 'N', 0, '#', 0, 0, 0, 2, 0, 8, 0, 0, 0,
  'a', 'w', 'e', 's', 'o', 'm', 'e', 0, 1, 0, '3', '3', '3', '3', '3', '3', 20,
  '@', 16, 0, '?', 7, 0, 0, 0, 0 }
```

(46 characters)

However, the original input is
```
  { '1', 0, 0, 0, 4, 'B', 'S', 'O', 'N', 0, '&', 0, 0, 0, 2, '0', 0, 8, 0, 0,
  0, 'a', 'w', 'e', 's', 'o', 'm', 'e', 0, 1, '1', 0, '3', '3', '3', '3', '3',
  '3', 20, '@', 16, '2', 0, '?', 7, 0, 0, 0, 0 }
```

(49 characters)

Apparently, our encoding takes fewer bytes, but I have not yet understood why. Do you have an idea?

julian-becker:
Just pushed a fix for the array serialization.

nlohmann:
Thanks for the fix. As I already made a lot of changes in the https://github.com/nlohmann/json/compare/julian-becker-feature/bson branch, I see that I can take over the patch.

Once I did that, I propose to open a new PR based on that branch and close this PR.

nlohmann:
Closing this PR to avoid diverging too much from #1320. Future discussion should take place in #1320.

PR #212, opened 2016-02-20
support gcc 4.8.1

abc100m:
for issue #211 , support  gcc 4.8.1 since it already had fully support c++ 11 (though STL library are not yet)

PR #1257, opened 2018-09-25
GCC 48

henryiii:
This adds GCC 4.8 support, all tests pass on CentOS/RHEL 7. Previously discussed in #212.

**All tests pass on Travis, even on GCC 4.8!**


cjh1:
:+1:

include/nlohmann/json.hpp
```diff
@@ -4943,6 +4943,23 @@ class basic_json
         return {it, res.second};
     }
 
+    /// helper for insertion of an iterator (supports GCC 4.8+)
```
nlohmann: Could you please reference this PR in a comment, so one may find the rationale for this change quicker?

henryiii:
I missed the compiler testing list in the readme, added a line there too. The other version strings may be a bit out of date.

Also, I'm testing with the external GCC 4.8.5 package just like all the other GCC versions, but GCC 4.8.4 is the default compiler on Travis, so for the first time you could have a no extra packages GCC build of JSON on Travis.

henryiii:
You’ll also want to update the text in the little graphic on the main page where there’s a 3.2.1 release.

nlohmann:
Thanks so much for the effort!

PR #1260, opened 2018-09-26
Fix broken cmake imported target alias

chuckatkins:
This is a followup on #1245 in which the fix caused additional breakage.  This PR does a more complete implementation of the package configs with version checking.  It also adds a set of tests to make sure that other CMake projects can import and use the package config.


nlohmann:
I'm looking forward to having this issue fixed - I could then make the next release.

chuckatkins:
@nlohmann This should be good to go now

nlohmann:
Thanks for the fix!

PR #1265, opened 2018-09-29
Make nlohmann friendly for inclusion.

### LINKED ISSUE #1250: Including nlohmann the badwrong way.
I want to use your library in a game. I've added proper code to find and link to it, but I also want to allow for the possibility somebody might want to compile the game and not know how to (or even just not want to) find/install nlohmnann. 

So... instead of just dropping an ```include_directories``` I decided I'd try embedding your library as a submodule and just including your CMakeLists if a config module cannot be found. I am not entirely sure what nasty repercusions this could have but I figured it was better since it's still using your logic to define the target.

The only problem is, in doing this I ran into a million errors due to includes or subdirectories not existing (being called from your CMakeLists). I tracked this down to CMake populating essentially all variables (and also using it when not otherwise explicitly specified, such as in 

```add_subdirectory(test)```

with my project folder instead of yours. I then realized I could use ```CMAKE_CURRENT_LIST_DIR``` instead of ```CMAKE_CURRENT_SOURCE_DIR``` and ```PROJECT_SOURCE_DIR``` and it will work fine.

Since I'd rather not modify your files (Ideally I'd like to be able to easily update my included version from your releases)... do you think it might be possible to support what I'm doing by comparing ```CMAKE_CURRENT_LIST_DIR``` and ```CMAKE_CURRENT_SOURCE_DIR``` and using the former if they're not equal?

Lord-Kamina:

Changes that would fix #1250, while presumably not affecting normal usage in any way.

## Pull request checklist

## Please don't

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,17 +26,17 @@ set(NLOHMANN_JSON_CONFIG_INSTALL_DIR        "lib/cmake/${PROJECT_NAME}"
   CACHE INTERNAL "")
 set(NLOHMANN_JSON_INCLUDE_INSTALL_DIR       "include")
 set(NLOHMANN_JSON_TARGETS_EXPORT_NAME       "${PROJECT_NAME}Targets")
-set(NLOHMANN_JSON_CMAKE_CONFIG_TEMPLATE     "cmake/config.cmake.in")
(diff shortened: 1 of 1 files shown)


nlohmann:
It would be great of @chuckatkins could have a look at this and whether it is related to #1260.

nlohmann:
The PR is closed as I deleted the 3.2.0 release branch. Could you please make another PR for the `develop` branch? Sorry for the inconvenience.

PR #1266, opened 2018-09-30
Make nlohmann friendly for inclusion.

Lord-Kamina:

New PR for #1250 

The one change I haven't tested is the natvis bit because I don't have an MSVC system on hand. Other than that, I tried installing regularly, running the tests and including my way and it worked as expected.

## Pull request checklist

## Please don't

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,17 +26,17 @@ set(NLOHMANN_JSON_CONFIG_INSTALL_DIR        "lib/cmake/${PROJECT_NAME}"
   CACHE INTERNAL "")
 set(NLOHMANN_JSON_INCLUDE_INSTALL_DIR       "include")
 set(NLOHMANN_JSON_TARGETS_EXPORT_NAME       "${PROJECT_NAME}Targets")
-set(NLOHMANN_JSON_CMAKE_CONFIG_TEMPLATE     "cmake/config.cmake.in")
+set(NLOHMANN_JSON_CMAKE_CONFIG_TEMPLATE    "${CMAKE_CURRENT_LIST_DIR}/cmake/config.cmake.in")
 set(NLOHMANN_JSON_CMAKE_CONFIG_DIR          "${CMAKE_CURRENT_BINARY_DIR}")
 set(NLOHMANN_JSON_CMAKE_VERSION_CONFIG_FILE "${NLOHMANN_JSON_CMAKE_CONFIG_DIR}/${PROJECT_NAME}ConfigVersion.cmake")
 set(NLOHMANN_JSON_CMAKE_PROJECT_CONFIG_FILE "${NLOHMANN_JSON_CMAKE_CONFIG_DIR}/${PROJECT_NAME}Config.cmake")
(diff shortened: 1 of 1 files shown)


chuckatkins:
I'm a bit confused as to the problem this is trying to solve.  When would PROJECT_SOURCE_DIR not be the same as CMAKE_CURRENT_LIST_DIR in the top level CMake?  The usual way to add a third party CMake library to another CMake project is via `add_subdirectory`, not `include()`, although it shouldn't matter too much which way you do it.

We (at kitware) are currently using this as a third party library via find_package for external or add_subdirectory for internal, without issue.  So what doesn't work that this is trying to fix?  It wasn't really clear what doesn't work since you should be abel to do the following as is without changing anything:

`CMakeLists.txt`
```cmake
...
add_subdirectory(3rdParty)
...
add_subdirectory(src)
...
```
`3rdParty/CMakeLists.txt`
```cmake
...
find_package(nlohmann_json 3.2.0 QUIET)
if(nlohmann_json_FOUND)
  message(STATUS "Nlohmann JSON found externally.")
else()
  message(STATUS "Nlohmann JSON not found externally; will use included version.")
  add_subdirectory(json)
endif()
```
`3rdParty/json/CMakeLists.txt`
```cmake
# Set any project specific options for nlohmann_json
set(BUILD_TESTING OFF)
add_subdirectory(nlohmann_json)
```
`src/CMakeLists.txt`
```cmake
...
target_link_libraries(performous PRIVATE nlohmann_json::nlohmann_json)
...
```
`3rdParty/json/nlohmann_json` is then the source for this project as a git submodule.

This should work as a third party library infrastructure in general with each `3rdParty/foo` containing a `CMakeLists.txt` that sets the `foo` specific build options and adds the sub-directory for `3rdParty/foo/libfoo` which is the project's actual code.

chuckatkins:
> it's perhaps an unusual way to do things

Using `include` to add another project carries a slew of implications with it that break other CMake code in the project.  It pulls the included file in and processes it in the directory the include is called from rather than changing the working directory, so `CMAKE_CURRENT_SOURCE_DIR` and `CMAKE_CURRENT_BINARY_DIR` aren't valid in the included `CMakeLists.txt` file.  Similarly, since the `project()` command is executed in the included file anything after the include has the `PROJECT_*` variables set to the value from the included file instead of the current project:
```cmake
project(foo ...)
# PROJECT_<...> variables set correctly for the foo project
include(3rdParty/json/CMakeLists.txt)
# PROJECT_<...> variables now set to the values from nlohmann_json
```
There should probably be better documentation in CMake for some of these implications and to suggest not to add additional cmake projects in that way.  Using `add_subdirectory` keeps things scoped accordingly.

nlohmann:
I have little experience with CMake and trust @chuckatkins in this regard. To my understanding, the PR would add a fix to support a discouraged way of including the library. If this is the case, then I'd rather not have the PR merged, especially as #1260 added a comment https://github.com/nlohmann/json#integration on how to use CMake for the library.

PR #1269, opened 2018-10-01
Convert unit testing to use new APIs instead of deprecated ones.

chuckatkins:
Unit tests were still using deprecated APIs causing a lot of noise during compilation.  This converts the unit tets to use the new APIs for iterators and stream operators.


nlohmann:
I would still like to test the deprecated functions.

chuckatkins:
Sure thing.  I was cleaning up warnings so instead of changing the code here would you be okay with just disabling the deprecation warnings for these files?

chuckatkins:
Closing.  Will keep deprecated apis in testing but disable deprecation warnings for these files only in a new pr

PR #1271, opened 2018-10-02
Turn off additional deprecation warnings for GCC.

chuckatkins:
In follow up from the conversation in #1269, this adds the `-Wno-deprecate-definitions` flag to unit tests to allow them to test deprecated APIs without all the noisy compiler warnings.  This also refactors the setting of build properties for test targets to use `target_<...>` commands instead of `set_target_properties()`


nlohmann:
Thanks a lot!

PR #1282, opened 2018-10-07
Improve diagnostic messages

nlohmann:
This is a follow-up to #1210.

- added line and column information where possible
- improved error messages in case of unescaped control characters
- added context to error messages


efp:
I did a fresh checkout and it looks good to me.

PR #1305, opened 2018-10-17
Add Meson related info to README

koponomarenko:
Add more information about how one can get a Meson wrap file.


nlohmann:
Thanks!

Issue #1242, opened 2018-09-17
[Question] Access to end() iterator reference


loumalouomega:
How can I access to the reference of the .end() iterator when there is an assert that forces to avoid it?

Maybe the question could look a little bit silly but I have problems with iterators because of this, and of course commenting the assert is just a short term and dirty solution to the problem.

https://github.com/nlohmann/json/blob/186c747a19c56c3208e7539713607211b6059ceb/single_include/nlohmann/json.hpp#L5304

nlohmann:
Could you provide an example?

loumalouomega:
We are doing a migration from rapidjson to your library:

https://github.com/KratosMultiphysics/Kratos/blob/07929fd8063e68dd39747cb9b8c13f91e949e128/kratos/includes/kratos_parameters.h

We have a custom class that uses internally your library, so for that reason we create an iterator adaptor. During the construction of our begin() and end() methods we call to the operator*(), both to the end() and begin(), but the end() iterators give us the error due to the assert.

https://github.com/KratosMultiphysics/Kratos/blob/07929fd8063e68dd39747cb9b8c13f91e949e128/kratos/includes/kratos_parameters.h#L241

https://github.com/KratosMultiphysics/Kratos/blob/07929fd8063e68dd39747cb9b8c13f91e949e128/kratos/includes/kratos_parameters.h#L895

gregmarr:
It is illegal to dereference an `end` iterator, from ANY container.

nlohmann:
You can turn off assertions by defining the preprocessor macro `NDEBUG`, see https://en.cppreference.com/w/cpp/error/assert. But as @gregmarr described, you must never dereference a past-the-end iterator in the first place.

loumalouomega:
Thanks, I will do that, I close this, thanks

PR #1313, opened 2018-10-23
[Iterator] Adding the methods is_begin() and is_end() to the iter_impl

loumalouomega:
Hi,

I don't know if this is the proper way to proceed with a PR. Maybe I should create an issue first. I already created one some time ago https://github.com/nlohmann/json/issues/1242

After that I tried several things, the point is we create a iterator adaptor and we need to skip the end iterators during that creation process to avoid the assert.

 But after trying many things the only thing I could do to avoid the assert was creating this is_end() method.

Tell me if tis is not a proper solution to the problem or if you want me to implement a test for this methods that I added.

### DIFF
--- a/include/nlohmann/detail/iterators/iter_impl.hpp
+++ b/include/nlohmann/detail/iterators/iter_impl.hpp
@@ -209,6 +209,54 @@ class iter_impl
     }
 
   public:
+    /// return whether the iterator can be dereferenced
+    constexpr bool is_begin() const noexcept
+    {
+        assert(m_object != nullptr);
+
+        switch (m_object->m_type)
+        {
+            case value_t::object:
+            {
+                return m_it.object_iterator == m_object->m_value.object->begin();
+            }
+            case value_t::array:
+            {
+                return m_it.array_iterator == m_object->m_value.array->begin();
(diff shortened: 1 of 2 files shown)


nlohmann:
Why not check `it == j.begin()` instead of of `it.is_begin()`?

loumalouomega:
Where I execute the code I cannot access to the whole "json list"

Here for example: 
https://github.com/KratosMultiphysics/Kratos/blob/8a1f314b960f9d6fd08172a03c3fa7ce6983f8db/kratos/includes/kratos_parameters.h#L162

nlohmann:
I don't think this is a good addition as this is an extension to the usual iterator interface. If you need to know the container the iterator belongs to, then store a reference to it. As a PR may be the wrong place to discuss, please open an issue for this.

loumalouomega:
I will try what you say (that I already tried but didn't work), if it doesn't solve the problem I will open an issue

Thanks

loumalouomega:
Just to mention that I solved the problem, my error was that I was using a temporal container and the referenced value gave a memory error, now is properly solved. Sorry for the noise and thanks for the library

PR #1315, opened 2018-10-23
Properly convert constants to CharType

### LINKED ISSUE #1286: truncation of constant value in to_cbor()
Although [documentation](https://nlohmann.github.io/json/classnlohmann_1_1basic__json_a6defa7ec3d3ace8f713f001f720182d7.html#a6defa7ec3d3ace8f713f001f720182d7) for `static void to_cbor(const basic_json& j, detail::output_adapter<char> o)` says nothing I assumed that it has some purpose but f.e. following sample:
```
	  ::nlohmann::json jj;
	  std::vector<char> test_cbor_data;
	  ::nlohmann::json::to_cbor(jj, test_cbor_data);
```
generates lots of warnings like

> Warning	C4309	'static_cast': truncation of constant value	testapp	json.hpp	7731	

in class binary_writer
at lines like:
https://github.com/nlohmann/json/blob/e426219256e8011fc8e4dd0bc8f54c85034218fa/single_include/nlohmann/json.hpp#L7949

or

https://github.com/nlohmann/json/blob/e426219256e8011fc8e4dd0bc8f54c85034218fa/single_include/nlohmann/json.hpp#L7955-L7958

Basically the cause for that is `static_cast<CharType>(/*some integer value > 127*/)` with `CharType = char` leads to loss of integer value (sizeof(char) is 1 byte and it is signed, so value range is [-128;127] and it can't fit integer number > 127).

I don't know if it is an issue but that code smells.

nlohmann:
Fixes #1286.


jaredgrubb:
+1
I really don't like that code, but necessary evil I suppose :)

jaredgrubb:
Actually: do you think there should be some comments in this code to just link to this Issue or the SO post that explains why it's necessary? I suppose you could get some of that via GitHub, but that might get buried? very minor thing you can ignore.

oktonion:
@nlohmann That's it. Except if you are using it with signed chars (like 'l') then for the sake of micro optimization we could add some overload like

```cpp
    template < typename C = CharType,
               enable_if_t < std::is_signed<C>::value and std::is_signed<char>::value > * = nullptr >
    static constexpr CharType to_char_type(char x) noexcept
    {
        return x;
    }
```

Idk if it is so necessary tho.

@jaredgrubb I think that with some bit shift magic this could be done in less obvious but more compact style. But I prefer straightforward `memcpy` to be on the safe side.

nlohmann:
@oktonion The code you propose has the signature `CharType to_char_type(char x)`, whereas all the other functions have `CharType to_char_type(std::uint8_t x)`. Did I miss anything?

oktonion:
@nlohmann my bad, there is a working example:

```cpp
// Example program (I removed constexpr for the sake of std::cout calls)
#include <iostream>
#include <string>
#include <cstring>
    
    template<bool B, typename T = void>
    using enable_if_t = typename std::enable_if<B, T>::type;
    
    template < typename C = CharType,
               enable_if_t < std::is_signed<C>::value and std::is_signed<char>::value > * = nullptr >
    static CharType to_char_type(std::uint8_t x) noexcept
    {
        std::cout << "1" << std::endl;
        
        return *reinterpret_cast<char*>(&x);
    }
    template < typename C = CharType,
               enable_if_t < std::is_signed<C>::value and std::is_unsigned<char>::value > * = nullptr >
    static CharType to_char_type(std::uint8_t x) noexcept
    {
        std::cout << "2" << std::endl;
        
        static_assert(sizeof(std::uint8_t) == sizeof(CharType), "size of CharType must be equal to std::uint8_t");
        static_assert(std::is_pod<CharType>::value, "CharType must be POD");
        CharType result;
        std::memcpy(&result, &x, sizeof(x));
        return result;
    }
    template<typename C = CharType,
             enable_if_t<std::is_unsigned<C>::value>* = nullptr>
    static CharType to_char_type(std::uint8_t x) noexcept
    {
        std::cout << "3" << std::endl;
        
        return x;
    }
    
    template < typename InputCharType,
                typename C = CharType,
               enable_if_t < 
               std::is_signed<C>::value and 
               std::is_signed<char>::value and 
               std::is_same<char, typename std::remove_cv<InputCharType>::type>::value 
               > * = nullptr >
    static CharType to_char_type(InputCharType x) noexcept
    {
        std::cout << "4" << std::endl;
        
        return x;
    }

int main()
{
    to_char_type(0x08);
    to_char_type(-23);
    to_char_type(254);
    to_char_type('l');
}

```

so for `typedef unsigned char CharType;` the output is

> 3 
> 3 
> 3 
> 3

for `typedef char CharType;` the output is (note that '1' in 4th call is replaced with '4' overload)

> 1
> 1
> 1
> 4

the '4' is faster than '1' and more clear for compiler optimizer. It's micro optimization though.

jaredgrubb:
If you really want to stress-test your implementation, make sure to try all three ```char```, ```unsigned char```, and ```signed char``` (remember that these three types are _distinct_ types in C++, they are **not** aliases [eg, like how "signed int" and "int" are the same]); you could also force your compiler to flip the signedness of char (eg, -funsigned-char and -fsigned-char for GCC/clang).

nlohmann:
Thanks everyone!

PR #1320, opened 2018-10-25
Add BSON support

### LINKED ISSUE #1244: BSON support

nlohmann:
Based on #1254, closes #1244.


test/src/fuzzer-parse_bson.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+            // parse serialization
+            json j2 = json::from_bson(vec2);
+
+            // serializations must match
+            assert(json::to_bson(j2) == vec2);
+        }
+        catch (const json::parse_error&)
+        {
+            // parsing a CBOR serialization must not fail
```
julian-becker: minor: comment should probably mention BSON instead of CBOR

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -121,6 +125,216 @@ class binary_reader
...
+             @a element_type will give rise to a parse_error.114:
+             Unsupported BSON record type 0x...
+    @return whether a valid BSON-object/array was passed to the SAX parser
+    */
+    bool parse_bson_element_internal(const int element_type,
+                                     const std::size_t element_type_parse_position)
+    {
+        switch (element_type)
+        {
+            case 0x01: // double
```
julian-becker: maybe we should consider introducing an enum for the magic `case` values here, thereby avoiding the need for the comments to explain the purpose of each case.

nlohmann: We do the same with CBOR, etc., and I think it is OK to be explicit about the byte values.

nlohmann:
Thanks @julian-becker !

julian-becker:
Thanks, @nlohmann for creating and maintaining this awesome library, and for letting me contribute a small piece to it!

PR #1322, opened 2018-10-25
Enable string_view, const_char_star wrapper, for map_key_type

### LINKED ISSUE #1274: separate object_t::key_type from basic_json::key_type, and use an allocator which returns object_t::key_type
- What

I would like the key_type which is searched for: 
json j = { "hi", "there" };
j.at("hi") <--- that "hi" is a going to be object_t::key_type, I believe.

To be different from the internal key type
for instance object_t::key_type might be string_view,
but the parameter types, might still be string...  but most likely would be string_view (if the internal type is string_view).

- Why?

I deal with json which has many millions of objects containing the same keys.
I would like there to be an allocator type which allocates strings in a big set of strings.  And returns the string_view of that string, rather than a new string, (or a may even return a const char * of the string.c_str())

This is close to the way the jvm allocates strings.

In this way, I would save a lot of memory, and also, key/value lookups in maps would be faster.

timprepscius:
Closes #1274.

Hey there,

Unfortunately, I forked this string_view off of a merged erase_by_json_pointer + develop branch.  
I'm not sure if you can easily cherry pick the edits you would want or not.
If you need these changes forked off the develop branch pre-erase-by-json-pointer let me know, I think I can apply a diff of this vs merged-develop branch, and apply that to a pre-merged-erase-json-pointer new branch.  (but if I don't have to, etc, etc, etc)

In any case there is still work to be done.  
Mainly 
1. make the map_key_type -string_view/std::string/whatever part of the template parameters.
2. see if there is a better place to put those "to_map_key" "to_lookup_key" template functions..  And or, is it possible to make them do a better job of picking up from outside the amalgamated json.hpp file..  (see unit alt string tests).  Since you have dealt with this before, you probably would need far less time to figure this out.

But, this was quite a lot of work actually to get all the unit tests compiling and working, especially alt_string..

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/input/json_sax.hpp
+++ b/include/nlohmann/detail/input/json_sax.hpp
@@ -6,6 +6,7 @@
 
 #include <nlohmann/detail/input/parser.hpp>
 #include <nlohmann/detail/exceptions.hpp>
+#include <nlohmann/detail/json_string_view.hpp>
 
 namespace nlohmann
 {
@@ -21,6 +22,7 @@ input.
 template<typename BasicJsonType>
 struct json_sax
 {
+    using object_t = typename BasicJsonType::object_t;
     /// type for (signed) integers
     using number_integer_t = typename BasicJsonType::number_integer_t;
     /// type for unsigned integers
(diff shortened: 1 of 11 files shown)


timprepscius:
Hey there,
So this string_view modification is necessary for another project I'm working on.  Last time I tried to make a pull request I messed it up..  This time, I'll try to do it right.  I just want to check that the process for making a valid pull request.

1.  Fork the repository
2. Branch from the develop branch
3. Make modifications
4. Pull request from branch

I think last time I had already merged changes from a previous local change before making the branch, or something...

PR #1330, opened 2018-11-02
CMake: Optional Install if Embedded

ax3l:
When adding this library as embedded library with private "target link", e.g. only used inside private source files, the library does not need to be installed when the main project gets installed.

This adds an additional option `JSON_Install` similar to the test-build control switch in order to skip installing headers and CMake config files if requested.

Avoids using
```cmake
add_subdirectory(path/to/nlohmann_json EXCLUDE_FROM_ALL)
```

which has further side-effects:
https://cmake.org/cmake/help/v3.0/command/add_subdirectory.html

cc @chuckatkins do you have an opinion on this?


nlohmann:
Yes, I would also know what @chuckatkins thinks about this.

ax3l:
Any thoughts, @chuckatkins ?

ax3l:
ping @chuckatkins any idea?

t-b:
I just came up with the exact same patches. Should have checked the PRs first ;)

Can I do help with something to get this merged?

ax3l:
@t-b I am still wondering what other projects do, since this seems to be a common workflow that a simple `EXCLUDE_FROM_ALL` doesn't cut.

nlohmann:
Thanks a lot!

t-b:
@nlohmann Thanks for merging.

@ax3l According to https://stackoverflow.com/a/51746390 most projects add an option to skip the installation.

ax3l:
Good to know, so my intuition wasn't all off.

Thanks for merging!

PR #1357, opened 2018-11-14
make release: Generate a lightweight buildable zip file

mpoquet:
This commit adds the generation of a ``buildable-meson.zip`` file when calling ``make release``.
The generated zip file contains ``meson.build`` and the ``include`` and ``single_include`` directories.

The goal is to allow both package managers and Meson users to get a lightweight (~250 KB) buildable/installable source tree for each release — as the tarballs of the latest release were heavy (~110 MB).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -340,8 +340,12 @@ release:
 	zip -9 -r include.zip include/*
 	gpg --armor --detach-sig include.zip
 	mv include.zip include.zip.asc release_files
+	zip -9 -r buildable_meson.zip meson.build single_include/* include/*
+	gpg --armor --detach-sig buildable_meson.zip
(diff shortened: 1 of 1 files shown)


nlohmann:
Who would call `make release`?

mpoquet:
I was thinking of the person (script?) in charge of doing the releases of this project on GitHub.

I saw many assets for each released version [there](https://github.com/nlohmann/json/releases) and I thought they were generated by ``make release``.

nlohmann:
Oh, I forgot about this target - I am using a quite large Python script for releasing now. What would happen with the created archive then? Because I would not like to add it to the release, because it is already quite crowded and I do not want to take care about the Meson build myself. Any ideas?

PR #1391, opened 2018-12-10
 Added Support for Structured Bindings 

### LINKED ISSUE #1388: Adding Structured Binding Support
- Describe the feature in as much detail as possible.
Given the fact that the Json Library now implements `items()` function, utilising [Structured Bindings](https://en.cppreference.com/w/cpp/language/structured_binding) in this scenario would make the code much more simpler, easier to read and understand
Even though Structured Bindings were added in C++17, the components/engines which ensure it exists are present in C++11 and C++14 and adding support would not require any `#ifdef `

- Include sample usage where appropriate.
 A Sample usage of Structured Binding is

```
   // example for an object
   for (auto& [key,value] : j_object.items())
   {
      std::cout << "key: " << key << ", value: " << value << '\n';
   }
```

- Implementation
The reason I am writing it as an issue rather than as a pull request is because it makes one questionable change.
Most of the implementation is referenced from [TartanLlama's Blog
](https://blog.tartanllama.xyz/structured-bindings/)
In the Class `iteration_proxy_internal` we would have to add the function get for Structured Bindings Support.
Thanks for @theodelrieu 's recommendations which helped shorten the code

```
// Required for Structured Bindings support
template<std::size_t N, typename = typename std::enable_if<N >= 2>::type>
void get()const
{
   static_assert(false, "Has Max 2 Elements:- Key or Value");
}
template<std::size_t N, typename = typename  std::enable_if<N == 0>::type>
auto get() const -> decltype(key())
{
   return key();
}
template<std::size_t N, typename = typename std::enable_if<N == 1>::type>
auto get() const  -> decltype(value())
{
   return value();
}
```
This is not a major change and would cause no incompatibility.
Adding support for Structured Bindings however involves minor editing of the std namespace which may or may not be favoured by a few projects.

```
namespace std {
   template<std::size_t N>
   struct tuple_element<N, nlohmann::detail::iteration_proxy<nlohmann::json::iterator>::iteration_proxy_internal> {
      using type = decltype(std::declval<nlohmann::detail::iteration_proxy<nlohmann::json::iterator>::iteration_proxy_internal>().get<N>());
   };

   template<>
   struct tuple_size<nlohmann::detail::iteration_proxy<nlohmann::json::iterator>::iteration_proxy_internal>
      : std::integral_constant<std::size_t, 2> {};
}
```
This is where we face an issue.
In both these methods, iteration_proxy_internal needs to be accessible.
However, as we know, iterator_proxy_internal is defined as a `private` method in the `iterator_proxy` class.
Hence in order to add support for Structured Bindings, we would have to either make it public, find a hack or use something similar to `friend`
This is also the reason I decided to submit it as an issue rather than make a pull request.

I apologise for any and all mistakes committed by me while writing this.
I am quite new to GitHub and this is one of my first feature requests on a project

EDIT: Thanks to @theodelrieu for making recommendations which helped shorten the code

pratikpc:
Closes #1388.

```
   for (auto const& [key,value] : json.items())
   {
      std::cout << "key: " << key << ", value: " << value << '\n';
   }
```
For further details, read #1388 and https://blog.tartanllama.xyz/structured-bindings/

Structured Bindings help simplify writing code and using this library greatly especially given the way they are implemented in C++17 and the addition of the items function.
The method of implementation ensures that Structured Bindings implementation can also be added in projects using C++11 with only C++17 being able to reap the benefits.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -290,9 +290,9 @@ void to_json(BasicJsonType& j, typename BasicJsonType::object_t&& obj)
 template <
     typename BasicJsonType, typename T, std::size_t N,
     enable_if_t<not std::is_constructible<typename BasicJsonType::string_t,
-                const T (&)[N]>::value,
+                const T(&)[N]>::value,
                 int> = 0 >
-void to_json(BasicJsonType& j, const T (&arr)[N])
+void to_json(BasicJsonType& j, const T(&arr)[N])
 {
     external_constructor<value_t::array>::construct(j, arr);
 }
@@ -300,21 +300,21 @@ void to_json(BasicJsonType& j, const T (&arr)[N])
 template<typename BasicJsonType, typename... Args>
 void to_json(BasicJsonType& j, const std::pair<Args...>& p)
 {
-    j = {p.first, p.second};
+    j = { p.first, p.second };
 }
 
 // for https://github.com/nlohmann/json/pull/1134
-template<typename BasicJsonType, typename T,
-         enable_if_t<std::is_same<T, typename iteration_proxy<typename BasicJsonType::iterator>::iteration_proxy_internal>::value, int> = 0>
+template < typename BasicJsonType, typename T,
+           enable_if_t<std::is_same<T, iteration_proxy_value<typename BasicJsonType::iterator>>::value, int> = 0>
 void to_json(BasicJsonType& j, const T& b)
 {
-    j = {{b.key(), b.value()}};
+    j = { {b.key(), b.value()} };
 }
 
 template<typename BasicJsonType, typename Tuple, std::size_t... Idx>
 void to_json_tuple_impl(BasicJsonType& j, const Tuple& t, index_sequence<Idx...> /*unused*/)
 {
-    j = {std::get<Idx>(t)...};
+    j = { std::get<Idx>(t)... };
 }
 
 template<typename BasicJsonType, typename... Args>
--- a/include/nlohmann/detail/iterators/iteration_proxy.hpp
+++ b/include/nlohmann/detail/iterators/iteration_proxy.hpp
@@ -3,104 +3,105 @@
 #include <cstddef> // size_t
 #include <string> // string, to_string
 #include <iterator> // input_iterator_tag
+#include <tuple> // tuple_size, get, tuple_element
 
 #include <nlohmann/detail/value_t.hpp>
+#include <nlohmann/detail/meta/type_traits.hpp>
 
 namespace nlohmann
 {
 namespace detail
 {
+template <typename IteratorType> class iteration_proxy_value
+{
+  public:
+    using difference_type = std::ptrdiff_t;
+    using value_type = iteration_proxy_value;
+    using pointer = value_type * ;
+    using reference = value_type & ;
+    using iterator_category = std::input_iterator_tag;
+
+  private:
+    /// the iterator
+    IteratorType anchor;
+    /// an index for arrays (used to create key names)
+    std::size_t array_index = 0;
+    /// last stringified array index
+    mutable std::size_t array_index_last = 0;
+    /// a string representation of the array index
+    mutable std::string array_index_str = "0";
+    /// an empty string (to return a reference for primitive values)
+    const std::string empty_str = "";
+
+  public:
+    explicit iteration_proxy_value(IteratorType it) noexcept : anchor(it) {}
+
+    /// dereference operator (needed for range-based for)
+    iteration_proxy_value& operator*()
+    {
+        return *this;
+    }
+
+    /// increment operator (needed for range-based for)
+    iteration_proxy_value& operator++()
+    {
+        ++anchor;
+        ++array_index;
+
+        return *this;
+    }
+
+    /// equality operator (needed for InputIterator)
+    bool operator==(const iteration_proxy_value& o) const noexcept
+    {
+        return anchor == o.anchor;
+    }
+
+    /// inequality operator (needed for range-based for)
+    bool operator!=(const iteration_proxy_value& o) const noexcept
+    {
+        return anchor != o.anchor;
+    }
+
+    /// return key of the iterator
+    const std::string& key() const
+    {
+        assert(anchor.m_object != nullptr);
+
+        switch (anchor.m_object->type())
+        {
+            // use integer array index as key
+            case value_t::array:
+            {
+                if (array_index != array_index_last)
+                {
+                    array_index_str = std::to_string(array_index);
+                    array_index_last = array_index;
+                }
+                return array_index_str;
+            }
+
+            // use key from the object
+            case value_t::object:
+                return anchor.key();
+
+            // use an empty key for all primitive types
+            default:
+                return empty_str;
+        }
+    }
+
+    /// return value of the iterator
+    typename IteratorType::reference value() const
+    {
+        return anchor.value();
+    }
+};
+
 /// proxy class for the items() function
 template<typename IteratorType> class iteration_proxy
 {
   private:
-    /// helper class for iteration
-    class iteration_proxy_internal
-    {
-      public:
-        using difference_type = std::ptrdiff_t;
-        using value_type = iteration_proxy_internal;
-        using pointer = iteration_proxy_internal*;
-        using reference = iteration_proxy_internal&;
-        using iterator_category = std::input_iterator_tag;
-
-      private:
-        /// the iterator
-        IteratorType anchor;
-        /// an index for arrays (used to create key names)
-        std::size_t array_index = 0;
-        /// last stringified array index
-        mutable std::size_t array_index_last = 0;
-        /// a string representation of the array index
-        mutable std::string array_index_str = "0";
-        /// an empty string (to return a reference for primitive values)
-        const std::string empty_str = "";
-
-      public:
-        explicit iteration_proxy_internal(IteratorType it) noexcept : anchor(it) {}
-
-        /// dereference operator (needed for range-based for)
-        iteration_proxy_internal& operator*()
-        {
-            return *this;
-        }
-
-        /// increment operator (needed for range-based for)
-        iteration_proxy_internal& operator++()
-        {
-            ++anchor;
-            ++array_index;
-
-            return *this;
-        }
-
-        /// equality operator (needed for InputIterator)
-        bool operator==(const iteration_proxy_internal& o) const noexcept
-        {
-            return anchor == o.anchor;
-        }
-
-        /// inequality operator (needed for range-based for)
-        bool operator!=(const iteration_proxy_internal& o) const noexcept
-        {
-            return anchor != o.anchor;
-        }
-
-        /// return key of the iterator
-        const std::string& key() const
-        {
-            assert(anchor.m_object != nullptr);
-
-            switch (anchor.m_object->type())
-            {
-                // use integer array index as key
-                case value_t::array:
-                {
-                    if (array_index != array_index_last)
-                    {
-                        array_index_str = std::to_string(array_index);
-                        array_index_last = array_index;
-                    }
-                    return array_index_str;
-                }
-
-                // use key from the object
-                case value_t::object:
-                    return anchor.key();
-
-                // use an empty key for all primitive types
-                default:
-                    return empty_str;
-            }
-        }
-
-        /// return value of the iterator
-        typename IteratorType::reference value() const
-        {
-            return anchor.value();
-        }
-    };
-
     /// the container to iterate
     typename IteratorType::reference container;
 
@@ -110,16 +111,52 @@ template<typename IteratorType> class iteration_proxy
         : container(cont) {}
 
     /// return iterator begin (needed for range-based for)
-    iteration_proxy_internal begin() noexcept
+    iteration_proxy_value<IteratorType> begin() noexcept
     {
-        return iteration_proxy_internal(container.begin());
+        return iteration_proxy_value<IteratorType>(container.begin());
     }
 
     /// return iterator end (needed for range-based for)
-    iteration_proxy_internal end() noexcept
+    iteration_proxy_value<IteratorType> end() noexcept
     {
-        return iteration_proxy_internal(container.end());
+        return iteration_proxy_value<IteratorType>(container.end());
     }
 };
+// Structured Bindings Support
+// For further reference see https://blog.tartanllama.xyz/structured-bindings/
+// And see https://github.com/nlohmann/json/pull/1391
+template <std::size_t N, typename IteratorType, enable_if_t<N == 0, int> = 0>
+auto get(const nlohmann::detail::iteration_proxy_value<IteratorType>& i) -> decltype(i.key())
+{
+    return i.key();
+}
+// Structured Bindings Support
+// For further reference see https://blog.tartanllama.xyz/structured-bindings/
+// And see https://github.com/nlohmann/json/pull/1391
+template <std::size_t N, typename IteratorType, enable_if_t<N == 1, int> = 0>
+auto get(const nlohmann::detail::iteration_proxy_value<IteratorType>& i) -> decltype(i.value())
+{
+    return i.value();
+}
 }  // namespace detail
 }  // namespace nlohmann
+
+// The Addition to the STD Namespace is required to add
+// Structured Bindings Support to the iteration_proxy_value class
+// For further reference see https://blog.tartanllama.xyz/structured-bindings/
+// And see https://github.com/nlohmann/json/pull/1391
+namespace std
+{
+template <typename IteratorType>
+struct tuple_size<::nlohmann::detail::iteration_proxy_value<IteratorType>>
+            : std::integral_constant<std::size_t, 2> {};
+
+template <std::size_t N, typename IteratorType>
+struct tuple_element <
+    N, ::nlohmann::detail::iteration_proxy_value<IteratorType >>
+{
+    using type = decltype(
+                     get<N>(std::declval <
+                            ::nlohmann::detail::iteration_proxy_value<IteratorType >> ()));
+};
+}
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ benchmarks/files/numbers/*.json
 cmake-build-debug
 
 test/test-*
+/.vs
--- a/include/nlohmann/detail/iterators/iter_impl.hpp
+++ b/include/nlohmann/detail/iterators/iter_impl.hpp
@@ -17,24 +17,21 @@ namespace detail
 {
 // forward declare, to be able to friend it later on
 template<typename IteratorType> class iteration_proxy;
+template<typename IteratorType> class iteration_proxy_value;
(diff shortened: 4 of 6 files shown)


include/nlohmann/detail/iterators/iteration_proxy.hpp
```diff
@@ -99,11 +99,38 @@ template<typename IteratorType> class iteration_proxy
...
+        template<std::size_t N, typename = typename  std::enable_if<N == 0>::type>
+        auto get() const -> decltype(key())
+        {
+            return key();
+        }
+        template<std::size_t N, typename = typename std::enable_if<N == 1>::type>
+        auto get() const  -> decltype(value())
+        {
+            return value();
+        }
```
theodelrieu: I think it's better to make those non-member, to avoid users calling them (not even talking about the possible confusion with `json::get`)

pratikpc: That might not work.
Because the basic idea with Structured Bindings is that they return key and value
Key at Index 0
And Value at Index 1
At least when used with items.
Making them public/anything else would render them incompatible with Structured Bindings support in json.items

The issue was with iteration_proxy_internal being declared as private however, I have made it public so it should be compatible now

include/nlohmann/detail/iterators/iteration_proxy.hpp
```diff
@@ -99,11 +99,38 @@ template<typename IteratorType> class iteration_proxy
...
     };
 
     /// the container to iterate
     typename IteratorType::reference container;
 
+    // In order to Support Structured Bindings
+    // We would have to make iteration_proxy_internal
+    // Accessible to tuple_size and tuple_element in std namespace
+    // However, as iteration_proxy_internal is private,
+    // We will have to use a Hack to add Structured Bindings Support
```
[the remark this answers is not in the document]
pratikpc: I made the necessary changes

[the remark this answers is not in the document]
pratikpc: I do not have much of a clue how LGTM or AppVeyor work so do they restart with the new commit or would I have to create a new one??

include/nlohmann/detail/iterators/iteration_proxy.hpp
theodelrieu: About non-members, the prototype is the same, but it takes an argument, where the member-function takes none. It should work the same either way.

include/nlohmann/detail/iterators/iteration_proxy.hpp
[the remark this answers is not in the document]
theodelrieu: You do not have to create a new one IIRC

pratikpc:
```
   for (auto const& [key,value] : json.items())
   {
      std::cout << "key: " << key << ", value: " << value << '\n';
   }
```

Is probably same as

```
   for (auto const& elem : json.items())
   {
      auto const key = elem.get<0>();
      auto const value = elem.get<1>();
      std::cout << "key: " << key << ", value: " << value << '\n';
   }
```

include/nlohmann/detail/iterators/iteration_proxy.hpp
pratikpc: Not really sure.
Because as far as I know, the way Structured Bindings work is that they consider everything is a Tuple/Tuple Element.

For example Structured Binding related to the code

```
   for (auto const& [key,value] : json.items())
   {
      std::cout << "key: " << key << ", value: " << value << '\n';
   }

```
Is probably same as
```
   for (auto const& elem : json.items())
   {
      auto const key = elem.get<0>();
      auto const value = elem.get<1>();
      std::cout << "key: " << key << ", value: " << value << '\n';
   }
```

include/nlohmann/detail/iterators/iteration_proxy.hpp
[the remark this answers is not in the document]
pratikpc: @theodelrieu thanks!
That's actually a nice feature

theodelrieu:
In generic context, one would always use `std::get<0>(...)`, especially to avoid writing `j.template get<0>()`

include/nlohmann/detail/iterators/iteration_proxy.hpp
theodelrieu: From [cppreference](https://en.cppreference.com/w/cpp/language/structured_binding):

* `e.get<i>()`, if lookup for the identifier get in the scope of `E` by class member access lookup finds at least one declaration that is a function template whose first template parameter is a non-type parameter.

* Otherwise, `get<i>(e)`, where `get` is looked up by argument-dependent lookup only, ignoring non-ADL lookup.

Your assumption is correct when the first bullet point is matched, but when the second one is matched, code is equivalent to:

```cpp
for (auto const& elem : json.items())
{
    auto const key = std::get<0>(elem);
    auto const value = std::get<1>(elem);
    std::cout << "key: " << key << ", value: " << value << '\n';
}
```

pratikpc: > 
> 
> In generic context, one would always use `std::get<0>(...)`, especially to avoid writing `j.template get<0>()`

I apologise.
It seems I can push it out to std namespace
The code would look similar to

```
namespace std {

   // Required for Structured Bindings support
   template<std::size_t N, typename = typename std::enable_if<N >= 2>::type>
   void get(nlohmann::detail::iteration_proxy<nlohmann::json::iterator>::iteration_proxy_internal&)
   {
      static_assert(N >= 2, "Has Max 2 Elements:- Key or Value");
   }
   template<std::size_t N, typename = typename  std::enable_if<N == 0>::type>
   auto get(nlohmann::detail::iteration_proxy<nlohmann::json::iterator>::iteration_proxy_internal const& x)  -> decltype(x.key())
   {
      return x.key();
   }
   template<std::size_t N, typename = typename std::enable_if<N == 1>::type>
   auto get(nlohmann::detail::iteration_proxy<nlohmann::json::iterator>::iteration_proxy_internal const& x)   -> decltype(x.value())
   {
      return x.value();
   }

   template<std::size_t N>
   struct tuple_element<N, nlohmann::detail::iteration_proxy<nlohmann::json::iterator>::iteration_proxy_internal> {
      using type = decltype(get<N>(std::declval<nlohmann::detail::iteration_proxy<nlohmann::json::iterator>::iteration_proxy_internal const&>()));
   };

   template<>
   struct tuple_size<nlohmann::detail::iteration_proxy<nlohmann::json::iterator>::iteration_proxy_internal>
      : std::integral_constant<std::size_t, 2> {};
}
```

We still however would require iteration_proxy_internal to be public and json::iterator to be a parameter which doesn't seem to be so bad.

theodelrieu: What is the compiler error? Could you show the full snippet?

pratikpc: error C2672: 'get': no matching overloaded function found
  C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include\utility(482): note: see reference to class template instantiation 'std::tuple_element<0,_Tuple>' being compiled
          with
          [
              _Tuple=nlohmann::detail::iteration_proxy<nlohmann::detail::iter_impl<nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>>>::iteration_proxy_internal
          ]
JSONTest.cpp(10): note: see reference to class template instantiation 'std::tuple_element<0,const nlohmann::detail::iteration_proxy<nlohmann::detail::iter_impl<nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>>>::iteration_proxy_internal>' being compiled
include\nlohmann\json.hpp(20236): error C2784: 'tuple_element<_Idx,std::pair<_Ty1,_Ty2>>::type &std::get(std::pair<_Ty1,_Ty2> &) noexcept': could not deduce template argument for 'std::pair<_Ty1,_Ty2> &' from 'const nlohmann::detail::iteration_proxy<nlohmann::detail::iter_impl<nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>>>::iteration_proxy_internal'
  C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\include\utility(600): note: see declaration of 'std::get'

The error occurred in my test project

theodelrieu: Indeed, I forgot about that one, I guess you have to forward declare json::iterator then...

pratikpc: Nope the new commit fails.
get function probably can't properly lie within the std namespace
It fails with an error that it was unable to find get when a Structured Binding is used

pratikpc: I think we should just stick to the old commit that placed the get function in iterator_proxy

theodelrieu: I will take a look tomorrow, I think we're close

pratikpc: Auto Deduction of Return Types also fails on a few older compilers so decided to copy paste return types there from the original method.
The issue with this is that if key/value's types are changed, then get's types need to be changed manually
https://travis-ci.org/nlohmann/json/jobs/466093802

theodelrieu:
I just pushed a commit on my [fork](https://github.com/theodelrieu/json/tree/feature/structured_bindings) which is still a wip, but the basic use-case works (with non-member `get` functions).

I had to move out `iteration_proxy_internal` outside of `iteration_proxy` to be able to partially specialize `tuple_size` and `tuple_element` (otherwise, only `nlohmann::json` would support structured bindings, not different instances of `nlohmann::basic_json<...>`).

There might be some refactoring to be done with `iteration_proxy/_internal`, I'd like to know what @nlohmann thinks about it, I do not know that part of the code very well.

Also, I have not tested edge-cases yet...

@pratikpc What do you think about incorporating my changes in your PR? We can work out the remaining details if you want :)

pratikpc:
> @pratikpc What do you think about incorporating my changes in your PR? We can work out the remaining details if you want :)

That's a superb change.
Making it public and independent of iteration_proxy kills of all the problems we were facing.
I am all for it.

In the commit we can possibly push the get function to the std namespace as well

```
namespace std
{
   // structured bindings support
   template <std::size_t N, typename IteratorType, enable_if_t<N == 0, int> = 0>
   auto get(const nlohmann::detail::iteration_proxy_internal<IteratorType>& i) -> decltype(i.key())
   {
      return i.key();
   }

   template <std::size_t N, typename IteratorType, enable_if_t<N == 1, int> = 0>
   auto get(const nlohmann::detail::iteration_proxy_internal<IteratorType>& i) -> decltype(i.value())
   {
      return i.value();
   }
}
```

Although I still believe we need to know @nlohmann 's views regarding that.

pratikpc:
@theodelrieu the commit broke a Unit Test.
Specifically the test-regression

One way to correct it is to add a Constructor in Basic Json class which accepts iteration_proxy_internal as an argument
```
template<typename Iterator>
basic_json(detail::iteration_proxy_internal<Iterator> other) : basic_json{ other.value() }
{}
```

However, my belief is that this is a temporary fix

The error I faced was
`[2018-12-13 07:41:19] [build] /usr/include/c++/7/bits/stl_algo.h:5337:16: error: no match for ‘operator=’ (operand types are ‘std::insert_iterator<nlohmann::basic_json<> >’ and ‘nlohmann::detail::iteration_proxy_internal<nlohmann::detail::iter_impl<nlohmann::basic_json<> > >’)
`
Also faced [here at LGTM](https://lgtm.com/projects/g/nlohmann/json/logs/rev/pr-de6145de23d2ca77b3e54238002cba61a9b56205/lang:cpp/stage:Build%20merge%20(f789b0d))

theodelrieu:
Oops, my bad, I only tested `unit-items.cpp`......

>In the commit we can possibly push the get function to the std namespace as well

That's illegal unfortunately, the `get` methods are where they should be, i.e. in the same namespace as the class.

I will make the code work on my box and let you know when I push!

pratikpc:
> That's illegal unfortunately, the `get` methods are where they should be, i.e. in the same namespace as the class.

Oh.
I thought it would be fine if they were kept in std namespace.
Because Structured Bindings still works on GCC and MSVC
> 
> I will make the code work on my box and let you know when I push!
The unit-tests now no longer fail after adding the Constructor.
But yes, finding a more permanent solution would be a good idea for us all.

I would add a commit which shifts get back into the namespace as before

theodelrieu:
I just push-forced on my fork, the fix was in `conversions/to_json.hpp` as you can see.
I've added a test with a custom `basic_json` alias as well, if you have other tests in mind do not hesitate to add them :)

>I thought it would be fine if they were kept in std namespace.

You can see the rules from [cppreference](https://en.cppreference.com/w/cpp/language/extending_std)

pratikpc:
@theodelrieu I can't find the changes.
Did you make them at https://github.com/theodelrieu/json/blob/feature/structured_bindings/include/nlohmann/detail/conversions/to_json.hpp

theodelrieu:
It is still in namespace `detail`, so the current name does not annoy me that much. Need @nlohmann's input on this one.

I do think the PR is complete, could you squash your commits?

pratikpc:
@theodelrieu how does one squash a commit exactly???

theodelrieu:
First, do not forget to run `make amalgamate` and commit the result.

Once this is done, run `git rebase -i develop`. It will open a prompt.
Then, keep the first line as-is, and replace `pick` in all the others by `fixup`.
Save and quit, then `git push --force`.

pratikpc:
@theodelrieu thanks!!!

pratikpc:
@theodelrieu The following tests FAILED:
	 80 - test-unicode_all (Timeout)
On Appveyor
Does this mean we have to start the process again?

nlohmann:
I’ll restart.

nlohmann:
Appveyor seems to work now. :)

nlohmann:
I am traveling right now, so I won’t be able to check before next week. But anything that brings us closer toward support of structured bindings is greatly appreciated!

include/nlohmann/detail/iterators/iteration_proxy.hpp
```diff
@@ -110,16 +111,45 @@ template<typename IteratorType> class iteration_proxy
...
+}
+
+template <std::size_t N, typename IteratorType, typename std::enable_if<N == 1, int>::type = 0>
+auto get(const nlohmann::detail::iteration_proxy_internal<IteratorType>& i) -> decltype(i.value())
+{
+    return i.value();
+}
 }  // namespace detail
 }  // namespace nlohmann
+
```
nlohmann: Could you please add a comment why you are adding code to the `std` namespace?

include/nlohmann/detail/conversions/to_json.hpp
```diff
@@ -290,31 +290,31 @@ void to_json(BasicJsonType& j, typename BasicJsonType::object_t&& obj)
 template <
     typename BasicJsonType, typename T, std::size_t N,
     enable_if_t<not std::is_constructible<typename BasicJsonType::string_t,
-                const T (&)[N]>::value,
+                const T(&)[N]>::value,
```
nlohmann: Can you make sure you ran `make pretty` - the indentation and whitespace looks odd in the whole PR. This is not a blocker, just a nice-to-have.

include/nlohmann/detail/iterators/iteration_proxy.hpp
```diff
@@ -3,104 +3,105 @@
...
 #include <nlohmann/detail/value_t.hpp>
+#include <nlohmann/detail/meta/type_traits.hpp>
 
 namespace nlohmann
 {
 namespace detail
 {
-/// proxy class for the items() function
-template<typename IteratorType> class iteration_proxy
+template <typename IteratorType> class iteration_proxy_internal
```
nlohmann: Discussion: Maybe rename this to `iteration_proxy_value`?

include/nlohmann/detail/conversions/to_json.hpp
pratikpc: Sorry.
I was previously using MSYS 32-Bit which didn't have astyle installed so amalgamate worked without issues.
Switching to MSYS 64 Bit solves this issue.

include/nlohmann/detail/iterators/iteration_proxy.hpp
pratikpc: Yes.
That sounds like a nice enough name and does provide information regarding what `iteration_proxy_internal `exactly does.

include/nlohmann/detail/iterators/iteration_proxy.hpp
pratikpc: Would something like the following be fine?
```
// The Addition to the STD Namespace is required to add
// Structured Bindings Support to the iteration_proxy_value class
// For further reference see https://blog.tartanllama.xyz/structured-bindings/
```
Or would a TartanLlama link be inappropriate and I should add the less helpful cppreference link
https://en.cppreference.com/w/cpp/language/structured_binding

include/nlohmann/detail/iterators/iteration_proxy.hpp
theodelrieu: I'm ok with `iteration_proxy_value`

nlohmann: I also prefer `iteration_proxy_value` over `iteration_proxy_kv` or `iteration_proxy_key_value`.

pratikpc: @nlohmann I corrected that and updated the branch with another Squash of the commits.
Could you please check and see if they are proper?
One of my many doubts is [here](https://github.com/nlohmann/json/pull/1391#discussion_r242005973)

include/nlohmann/detail/iterators/iteration_proxy.hpp
nlohmann: The comment is fine. Maybe also add a link to this PR as reference.

include/nlohmann/detail/iterators/iteration_proxy.hpp
nlohmann: I just added an answer.

include/nlohmann/detail/conversions/to_json.hpp
nlohmann: The diff still looks weird, but that's OK for now.

include/nlohmann/detail/iterators/iteration_proxy.hpp
nlohmann: @pratikpc If you add this, I would be ready to merge.

pratikpc: @nlohmann I made the required modifications

nlohmann:
Thanks so much!

Issue #1370, opened 2018-11-22
Using this library without stream ?


dumarjo:
Hi,

I would like to use this library with our product based on Cortex-M MCU. This mcu asnot much RAM/Flash. The use of stream (fstream/streambuf) is not really an option. 

Is it possible to use that libray without using stream ? 

Regards

Jonathan

nlohmann:
Assuming you mean parsing: you can also call `json::parse` with a string or char pointer. If that’s not what you mean, please clarify.

dumarjo:
Hi,

I would like to use this library in 2 differents way. The first one is with the data binding. The second one is with the SAX parser (from a json file). 

The main idea is to be able to use the library with file access but not with fstream.  I see the Parse method that can accept begin iterator and end iterator. The only problem, is the iterators need to be random access, but for file reading, this is somewhat not efficient. 

The other problem with the json::parse with string or char * is that I cannot keep the whole json file in RAM. I need to parse it with Sax to get only the values. 

Regards

Jonathan

nlohmann:
I see. The requirement for `random_access_iterator` may have become unnecessary. We had a `unget` method in the past which was removed recently, so we could work with an input iterator. Let me check.

dumarjo:
Hi, 

Some other library  offer an interface for using FILE * too.  Maybe could an improvement to your library too. This is more a wish than a requirement. 

Jonathan

dumarjo:
Hi,
I have added a new input_adapter to not using the streams when working with files. I use a simple FILE * as input and everything seem to work as expected. I would like tos know if this feature is desired to be added to the library ? I can make a PR if any interrest in that feature. This will eneble low end device working fith file whitout the cost of the streams. 

Regards
Jonathan

nlohmann:
Yes, a PR would be really appreciated!

PR #1392, opened 2018-12-12
Feature/add file input adapter

dumarjo:
This pull request enable using stdio file acces to reduce at minimum the RAM usage. When using this library with arm-none-eabi-gcc, the streams are very heavy in RAM and Flash. This new adapter is not using the streams at all.  This fix the issue https://github.com/nlohmann/json/issues/1370

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/input/input_adapters.hpp
+++ b/include/nlohmann/detail/input/input_adapters.hpp
@@ -46,6 +46,10 @@ struct input_adapter_protocol
 /// a type to simplify interfaces
 using input_adapter_t = std::shared_ptr<input_adapter_protocol>;
 
+/*!
+Input adapter for stdio file access. This adapter read only 1 byte and do not use any
+ buffer. This adapter is a very low level adapter. This adapter
+*/
 class file_input_adapter : public input_adapter_protocol
 {
 public:
@@ -55,11 +59,7 @@ public:
 
     std::char_traits<char>::int_type get_character() noexcept override
     {
-        auto res = fgetc(const_cast<FILE *>(file));
-        if(res == EOF)
-            return std::char_traits<char>::eof();
-        else
-            return static_cast<std::char_traits<char>::int_type>(res);
+        return fgetc(const_cast<FILE *>(file));
     }
 private:
     /// the file pointer to read from
--- a/include/nlohmann/detail/input/input_adapters.hpp
+++ b/include/nlohmann/detail/input/input_adapters.hpp
@@ -10,6 +10,7 @@
 #include <string> // string, char_traits
 #include <type_traits> // enable_if, is_base_of, is_pointer, is_integral, remove_pointer
 #include <utility> // pair, declval
+#include <cstdio> //FILE *
 
 #include <nlohmann/detail/macro_scope.hpp>
 
@@ -44,6 +45,27 @@ struct input_adapter_protocol
 
 /// a type to simplify interfaces
 using input_adapter_t = std::shared_ptr<input_adapter_protocol>;
+
+class file_input_adapter : public input_adapter_protocol
+{
+public:
+    explicit file_input_adapter(const FILE *file)  noexcept
+            : file(file)
+    {}
+
+    std::char_traits<char>::int_type get_character() noexcept override
+    {
+        auto res = fgetc(const_cast<FILE *>(file));
+        if(res == EOF)
+            return std::char_traits<char>::eof();
+        else
+            return static_cast<std::char_traits<char>::int_type>(res);
+    }
+private:
+    /// the file pointer to read from
+    const FILE * file;
+};
+
 
 /*!
 Input adapter for a (caching) istream. Ignores a UFT Byte Order Mark at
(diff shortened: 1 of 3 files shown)


include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -45,6 +46,27 @@ struct input_adapter_protocol
 /// a type to simplify interfaces
 using input_adapter_t = std::shared_ptr<input_adapter_protocol>;
 
+/*!
+Input adapter for stdio file access. This adapter read only 1 byte and do not use any
+ buffer. This adapter is a very low level adapter. This adapter
+*/
```
jaredgrubb: This comment is incomplete. Please update the text here.

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -45,6 +46,27 @@ struct input_adapter_protocol
...
+        : file(file)
+    {}
+
+    std::char_traits<char>::int_type get_character() noexcept override
+    {
+        return fgetc(const_cast<FILE*>(file));
+    }
+  private:
+    /// the file pointer to read from
+    const FILE* file;
```
jaredgrubb: Change to ```FILE * file;``` and remove the "const_cast" and the const from the constructor. (Using const_cast in this way is undefined behavior and there's no reason to take the FILE* by const anyway).

test/src/unit-testsuites.cpp
```diff
@@ -384,6 +384,46 @@ TEST_CASE("json.org examples")
         json j;
         CHECK_NOTHROW(f >> j);
     }
+    SECTION("FILE 1.json")
+    {
+        auto f = fopen("test/data/json.org/1.json", "r");
```
jaredgrubb: This is a minor issue, but will avoid having to write "fclose" at the bottom, and technically also makes sure that things get cleaned up even if an exception gets thrown. (Not a serious issue for test code, but it's still nice to have unit tests always clean up their resources in all cases).

```
std::unique_ptr<FILE, decltype(&fclose)> f(fopen("test/data/json.org/1.json", "r"), &fclose);
```
You'll have to do ```j.parse(f.get())``` in the lines below.

There are ways to make this prettier if we want, but this is bare minimum I would recommend.

dumarjo: updated as per your comment

include/nlohmann/detail/input/input_adapters.hpp
dumarjo: updated as per your comment

include/nlohmann/detail/input/input_adapters.hpp
dumarjo: updated as per your comment

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -45,6 +46,27 @@ struct input_adapter_protocol
...
 
+/*!
+Input adapter for stdio file access. This adapter read only 1 byte and do not use any
+ buffer. This adapter is a very low level adapter.
+*/
+class file_input_adapter : public input_adapter_protocol
+{
+  public:
+    explicit file_input_adapter(FILE* file)  noexcept
+        : file(file)
```
nlohmann: Please do not use `file` for a member variable name and parameter name. Proposal: name member variable `m_file`.

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -45,6 +46,27 @@ struct input_adapter_protocol
...
 using input_adapter_t = std::shared_ptr<input_adapter_protocol>;
 
+/*!
+Input adapter for stdio file access. This adapter read only 1 byte and do not use any
+ buffer. This adapter is a very low level adapter.
+*/
+class file_input_adapter : public input_adapter_protocol
+{
+  public:
+    explicit file_input_adapter(FILE* file)  noexcept
```
nlohmann: Please use `std::FILE` rather than `FILE`.

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -287,7 +309,8 @@ class input_adapter
 {
   public:
     // native support
-
+    input_adapter(FILE* file)
```
nlohmann: As above: please use `std::FILE`.

test/src/unit-testsuites.cpp
```diff
@@ -384,6 +384,41 @@ TEST_CASE("json.org examples")
         json j;
         CHECK_NOTHROW(f >> j);
     }
+    SECTION("FILE 1.json")
+    {
+        std::unique_ptr<FILE, decltype(&fclose)> f(fopen("test/data/json.org/1.json", "r"), &fclose);
```
nlohmann: Please use `std::fopen`.

nlohmann: Please use `std::FILE`.

nlohmann: Please use `std::fclose`.

dumarjo:
I chnaged all the fclose, fopen FILE to be std::

nlohmann:
Thanks a lot!

PR #1411, opened 2019-01-01
Improve dump_integer performance

nickaein:
This PR improves the performance of `dump_integer` using a tuned int2ascii implementation. The technique is discussed in [Fastware talk by Andrei Alexandrescu](https://youtu.be/o4-CwDo2zpg?t=1680) and has been already incorporated in some of other well-known libraries like [fmtlib](https://github.com/fmtlib/fmt/blob/master/include/fmt/format.h#L954).

In this pull request:

1. The fast int2ascii implementation is implemented (adapted from fmtlib implementation)
2. A benchmark for small integers is added. Previous benchmarks for integer are biased to very large integers. Also, in the original talk Andrei argues that the performance gain is larger (~2x) for smaller integers compared to a Naive implementation. Therefore, I believe it is worth covering the small integers in a separate benchmark too since in most use cases in the real world the integers are not that large. This lets us to keep track of the performance for both small and larger integer values.
3. I couldn't find unit-tests for integer formatting that checks different number of digits. To bring the code coverage to 100%, I have added some unit tests too.

The benchmarks show an considerable performance improvement (~1.5x) on my VM linux. This could be different (and probably higher) in a non-VM environment. Here is the original performance of Dump* operations (Ubuntu 18.04 x64, Clang 6):

```
Run on (4 X 2601 MHz CPU s)
CPU Caches:
  L1 Data 32K (x1)
  L1 Instruction 32K (x1)
  L2 Unified 256K (x1)
  L3 Unified 6144K (x1)
Dump/jeopardy / -           310113722 ns  308222627 ns         45   162.468MB/s
Dump/jeopardy / 4           367662837 ns  367510814 ns         38   181.292MB/s
Dump/canada / -              13027324 ns   13027200 ns       1096   153.024MB/s
Dump/canada / 4              16393044 ns   16328034 ns        858   473.766MB/s
Dump/citm_catalog / -         1956277 ns    1951371 ns       7267   244.506MB/s
Dump/citm_catalog / 4         2648083 ns    2647870 ns       5007   622.081MB/s
Dump/twitter / -              1912959 ns    1912949 ns       7319    232.77MB/s
Dump/twitter / 4              2181181 ns    2171835 ns       6523   336.927MB/s
Dump/floats / -             107072183 ns  106987779 ns        128    166.43MB/s
Dump/floats / 4             117135279 ns  117133075 ns        119   192.724MB/s
Dump/signed_ints / -         72700522 ns   72323217 ns        192   268.724MB/s
Dump/signed_ints / 4         80542202 ns   80504633 ns        173   300.645MB/s
Dump/unsigned_ints / -       61854487 ns   61853677 ns        230   314.503MB/s
Dump/unsigned_ints / 4       69747684 ns   69491170 ns        202   348.555MB/s
Dump/small_signed_ints / -   31521619 ns   31516332 ns        438   223.579MB/s
Dump/small_signed_ints / 4   39854903 ns   39854301 ns        349   296.449MB/s
```

And this is the performance with fast int2ascii method (note the last six benchmarks):

```
Dump/jeopardy / -           286181485 ns  286013895 ns         48   175.083MB/s
Dump/jeopardy / 4           349974284 ns  349965305 ns         40   190.381MB/s
Dump/canada / -              13124489 ns   13047201 ns       1069   152.789MB/s
Dump/canada / 4              17334759 ns   17313192 ns        876   446.808MB/s
Dump/citm_catalog / -         1871325 ns    1871283 ns       7435   254.971MB/s
Dump/citm_catalog / 4         2462448 ns    2462047 ns       5452   669.033MB/s
Dump/twitter / -              1825330 ns    1817293 ns       7690   245.022MB/s
Dump/twitter / 4              2058264 ns    2057411 ns       6820   355.666MB/s
Dump/floats / -             107943518 ns  107941378 ns        129    164.96MB/s
Dump/floats / 4             118865187 ns  118097256 ns        117   191.151MB/s
Dump/signed_ints / -         42482922 ns   42401566 ns        329   458.355MB/s
Dump/signed_ints / 4         49180866 ns   49179644 ns        285   492.142MB/s
Dump/unsigned_ints / -       41778353 ns   41575626 ns        333   467.898MB/s
Dump/unsigned_ints / 4       48426047 ns   48365002 ns        288   500.807MB/s
Dump/small_signed_ints / -   25048480 ns   25048284 ns        554   281.313MB/s
Dump/small_signed_ints / 4   34388632 ns   34332533 ns        401   344.128MB/s
````

More detailed benchmark results (including gcc, msvc) is available here: https://docs.google.com/spreadsheets/d/1sGE3UvLeLT_KrnC_Ujvtl_NFXPIp28xFOe9u9-6GSU0/edit?usp=sharing

This is my first contribution so any feedback is welcome.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/output/serializer.hpp
+++ b/include/nlohmann/detail/output/serializer.hpp
@@ -528,6 +528,40 @@ class serializer
     }
 
     /*!
+    @brief count digits
+
+    Count the number of decimal (base 10) digits for an input unsigned integer.
+
+    @param[in] x  unsigned integer number to count its digits
+    @return    number of decimal digits
+    */
+    inline unsigned int count_digits(number_unsigned_t x) noexcept
+    {
+        unsigned int n_digits = 1;
+        for (;;)
+        {
+            if (x < 10)
+            {
+                return n_digits;
+            }
+            if (x < 100)
+            {
+                return n_digits + 1;
+            }
+            if (x < 1000)
+            {
+                return n_digits + 2;
+            }
+            if (x < 10000)
+            {
+                return n_digits + 3;
+            }
+            x = x / 10000u;
+            n_digits += 4;
+        }
+    }
+
+    /*!
     @brief dump an integer
 
     Dump a given integer to output stream @a o. Works internally with
@@ -542,6 +576,22 @@ class serializer
                  int> = 0>
     void dump_integer(NumberType x)
     {
+        static constexpr std::array<std::array<char, 2>, 100> digits_to_99
+        {
+            {
+                {'0', '0'}, {'0', '1'}, {'0', '2'}, {'0', '3'}, {'0', '4'}, {'0', '5'}, {'0', '6'}, {'0', '7'}, {'0', '8'}, {'0', '9'},
+                {'1', '0'}, {'1', '1'}, {'1', '2'}, {'1', '3'}, {'1', '4'}, {'1', '5'}, {'1', '6'}, {'1', '7'}, {'1', '8'}, {'1', '9'},
+                {'2', '0'}, {'2', '1'}, {'2', '2'}, {'2', '3'}, {'2', '4'}, {'2', '5'}, {'2', '6'}, {'2', '7'}, {'2', '8'}, {'2', '9'},
+                {'3', '0'}, {'3', '1'}, {'3', '2'}, {'3', '3'}, {'3', '4'}, {'3', '5'}, {'3', '6'}, {'3', '7'}, {'3', '8'}, {'3', '9'},
+                {'4', '0'}, {'4', '1'}, {'4', '2'}, {'4', '3'}, {'4', '4'}, {'4', '5'}, {'4', '6'}, {'4', '7'}, {'4', '8'}, {'4', '9'},
+                {'5', '0'}, {'5', '1'}, {'5', '2'}, {'5', '3'}, {'5', '4'}, {'5', '5'}, {'5', '6'}, {'5', '7'}, {'5', '8'}, {'5', '9'},
+                {'6', '0'}, {'6', '1'}, {'6', '2'}, {'6', '3'}, {'6', '4'}, {'6', '5'}, {'6', '6'}, {'6', '7'}, {'6', '8'}, {'6', '9'},
+                {'7', '0'}, {'7', '1'}, {'7', '2'}, {'7', '3'}, {'7', '4'}, {'7', '5'}, {'7', '6'}, {'7', '7'}, {'7', '8'}, {'7', '9'},
+                {'8', '0'}, {'8', '1'}, {'8', '2'}, {'8', '3'}, {'8', '4'}, {'8', '5'}, {'8', '6'}, {'8', '7'}, {'8', '8'}, {'8', '9'},
+                {'9', '0'}, {'9', '1'}, {'9', '2'}, {'9', '3'}, {'9', '4'}, {'9', '5'}, {'9', '6'}, {'9', '7'}, {'9', '8'}, {'9', '9'},
+            }
+        };
+
         // special case for "0"
         if (x == 0)
         {
@@ -549,28 +599,58 @@ class serializer
             return;
         }
 
-        const bool is_negative = std::is_same<NumberType, number_integer_t>::value and not (x >= 0);  // see issue #755
-        std::size_t i = 0;
+        // use a pointer to fill the buffer
+        auto buffer_ptr = begin(number_buffer);
 
-        while (x != 0)
-        {
-            // spare 1 byte for '\0'
-            assert(i < number_buffer.size() - 1);
+        const bool is_negative = std::is_same<NumberType, number_integer_t>::value and not(x >= 0); // see issue #755
+        number_unsigned_t abs_value;
 
-            const auto digit = std::labs(static_cast<long>(x % 10));
-            number_buffer[i++] = static_cast<char>('0' + digit);
-            x /= 10;
-        }
+        unsigned int n_chars;
 
         if (is_negative)
         {
-            // make sure there is capacity for the '-'
-            assert(i < number_buffer.size() - 2);
-            number_buffer[i++] = '-';
+            *buffer_ptr = '-';
+            abs_value = static_cast<number_unsigned_t>(0 - x);
+
+            // account one more byte for the minus sign
+            n_chars = 1 + count_digits(abs_value);
+        }
+        else
+        {
+            abs_value = static_cast<number_unsigned_t>(x);
+            n_chars = count_digits(abs_value);
         }
 
-        std::reverse(number_buffer.begin(), number_buffer.begin() + i);
-        o->write_characters(number_buffer.data(), i);
+        // spare 1 byte for '\0'
+        assert(n_chars < number_buffer.size() - 1);
+
+        // jump to the end to generate the string from backward
+        // so we later avoid reversing the result
+        buffer_ptr += n_chars;
+
+        // Fast int2ascii implementation inspired by "Fastware" talk by Andrei Alexandrescu
+        // See: https://www.youtube.com/watch?v=o4-CwDo2zpg
+        const auto buffer_end = buffer_ptr;
+        while (abs_value >= 100)
+        {
+            const auto digits_index = static_cast<unsigned>((abs_value % 100));
+            abs_value /= 100;
+            *(--buffer_ptr) = digits_to_99[digits_index][1];
+            *(--buffer_ptr) = digits_to_99[digits_index][0];
+        }
+
+        if (abs_value >= 10)
+        {
+            const auto digits_index = static_cast<unsigned>(abs_value);
+            *(--buffer_ptr) = digits_to_99[digits_index][1];
+            *(--buffer_ptr) = digits_to_99[digits_index][0];
+        }
+        else
+        {
+            *(--buffer_ptr) = static_cast<char>('0' + abs_value);
+        }
+
+        o->write_characters(number_buffer.data(), n_chars);
     }
 
     /*!
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -11411,6 +11411,40 @@ class serializer
     }
 
     /*!
+    @brief count digits
+
+    Count the number of decimal (base 10) digits for an input unsigned integer.
+
+    @param[in] x  unsigned integer number to count its digits
+    @return    number of decimal digits
+    */
+    inline unsigned int count_digits(number_unsigned_t x) noexcept
+    {
+        unsigned int n_digits = 1;
+        for (;;)
+        {
+            if (x < 10)
+            {
+                return n_digits;
+            }
+            if (x < 100)
+            {
+                return n_digits + 1;
+            }
+            if (x < 1000)
+            {
+                return n_digits + 2;
+            }
+            if (x < 10000)
+            {
+                return n_digits + 3;
+            }
+            x = x / 10000u;
+            n_digits += 4;
+        }
+    }
+
+    /*!
     @brief dump an integer
 
     Dump a given integer to output stream @a o. Works internally with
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by nickaein** [comment]
The Appveyor is failing with following message:

> Cannot assign job to clouds/groups 'pro, pro-backup'. All clouds within the group are either failing, or offline, or busy.

Apparently they are [experiencing some outage](https://status.appveyor.com/incidents/39tq9r92flzh) for now. Is there any way to re-run Appveyor CI for this PR after they resolve the issue? 

Its latest status is being reported here: https://status.appveyor.com/

**Comment by nlohmann** [comment]
Will do. Thanks for the PR! I am looking forward to having this performance improvement in the library!

**Comment by jaredgrubb** [comment]
This patch looks pretty good and seems reasonable.

My only concern is whether this patch would have any licensing implications. Was this patch developed from scratch or copied from somewhere? If it is copied from the video, was there anything in the video about the license that the code is released under? Just because it's on youtube doesn't necessarily make it public domain.

**Comment by nickaein** [comment]
I have rewritten the implementation with some optimizations along the way. The new implementation benchmark as following with `clang` compiler:

```
Dump/signed_ints / -         39328352 ns   39117135 ns        365   496.841MB/s
Dump/signed_ints / 4         46257733 ns   45992578 ns        299   526.245MB/s
Dump/unsigned_ints / -       36924171 ns   36911475 ns        385   527.022MB/s
Dump/unsigned_ints / 4       46389870 ns   45955745 ns        304   527.062MB/s
Dump/small_signed_ints / -   23804581 ns   23831036 ns        585   295.682MB/s
Dump/small_signed_ints / 4   34276080 ns   34128556 ns        414   346.184MB/s
```

Compared to implementation adapted from `fmtlib`, there is a ~5% performance improvement on average on benchmarks.

I've also tried this implementation with a larger lookup table with 1000 _per-calculated_ numbers (zero to 999) instead of 100 numbers (zero to 99) and it lead to a speedup of ~9% compared to `fmtlib` implementation.

Though I haven't ran benchmarks on other compilers, I hope there will be not a performance drop on them at least.

We have three options to proceed:

1. Use the current implementation in PR as it is. If there is a copyright concern on adapting code from `fmtlib` which has BSD license, this option is off the table.
2. Use the rewritten implementation which is similar to `fmtlib` implementation, but with a ~5% performance boost on `clang`. Though I have to test this implementation more to make sure it doesn't drop performance on other compilers and it also passes CI.
3. Use the rewritten implementation with bigger (1000 numbers) lookup table that gives us ~9% performance boost compared to `fmtlib` on `clang`. The tests required in option 2 must be done in this case too. More importantly, we also have to consider whether the larger lookup table (which lead to ~1KB increase in the object binary) worth its performance boost.

**Comment by nlohmann** [comment]
As for the copyright issue, I would like to watch the video myself. I hope to find time for this on the weekend.

**Comment by nlohmann** [comment]
(Any idea why the AppVeyor tests fail?)

**Comment by nickaein** [comment]
@nlohmann Thanks. I will wait for your response on how to proceed.

As for Appveyor, I mentioned earlier that there was downtime in their servers when I submitted the PR, hence the failure. Here is more detail: https://status.appveyor.com/incidents/39tq9r92flzh

Apparently there is no facility that automatically re-runs the CI. I wonder if it can be done manually? As workaround, I also can submit "noop" commit to trigger the CI.

**Comment by nlohmann** [comment]
In fact I restarted the jobs and now there are actual failures, see https://ci.appveyor.com/project/nlohmann/json/builds/21351818/job/4y0907w9i0ri4dxv.

**Comment by nlohmann** [comment]
(And BSD and MIT should be no problem IIRC)

**Comment by nlohmann** [comment]
Now I watched the video and I think there is no issue with copyrights. The video presents ideas and exemplifies them with code snippets. However, I did not find a source saying that it is OK to re-license BSD-licensed code under a MIT license.

**Comment by nickaein** [comment]
I didn't find any sources either. In such cases where two open-source projects have different (and perhaps incompatible) licensing, the safest approach seems to be incorporating the source license into the destination project. However, that's too much burden for us in this case.

[As mentioned eariler](https://github.com/nlohmann/json/pull/1411#issuecomment-451243254), I've already re-implemented the technique from scratch which actually has a slight performance improvement.

Is it fine to add this implementation as a commit on top of previous commits on this pull request? Or it's better to open a new pull request that drops [this commit](https://github.com/nlohmann/json/pull/1411/commits/9bd112e9ee47d7fb109065519f3656b06b221cf7) from history which in some parts contains code from fmtlib?

**Comment by nickaein** [comment]
I've push the new changes which are present in this commit: https://github.com/nlohmann/json/pull/1411/commits/7a6102f73955e0c0a06a8ca20003a7ee2dabce87. The other commits are untouched and same as before.

**Comment by nickaein** [comment]
> Why is that so? Why the function in serializer is not reused?

The only reason could be my limited knowledge of the codebase. Can you point out where is the implementation? I couldn't find such method in `serializer.hpp` except mine. Also since these methods are tuned, we might get an inferior performance with a equivalent but less performing implementation.

BTW thank you for the feedback! Please let me know if you have other suggestions.

**Comment by nlohmann** [comment]
I just ran the benchmarks myself and can confirm a significant performance improvement. Well done!

For the licensing part: We must not take code with a non-MIT license. This would make this library a time-bomb for future use. As long as the code was only motivated by the YouTube video, I see no problems. But anything from fmtlib should go out.

@nickaein Can you realize this?

**Comment by nickaein** [comment]
You're welcome! Thank you for this great library! I'm aware that the speed is not the top design goal so there could be the ways that improves the performance without affecting public API or causes a major impact on library design.

As for licensing, I understand the significance of possible license violation. The initial implementation was partially adapted from fmtlib (e.g. `static const digits` string). This choice was actually by random chance and for convenient, not because fmtlib is the only/first implementer of this technique or theirs has any particular significance.

Nevertheless, I have re-written the implementation. The current code isn't derived from fmtlib in any way (neither in its design nor in implementation details).

File: include/nlohmann/detail/output/serializer.hpp [hunk 1]
```diff
@@ -542,35 +576,78 @@ class serializer
...
+            abs_value = static_cast<number_unsigned_t>(0 - x);
+        }
+        else
         {
-            // spare 1 byte for '\0'
-            assert(i < number_buffer.size() - 1);
+            abs_value = static_cast<number_unsigned_t>(x);
+        }
+
+        auto n_digits = count_digits(abs_value);
```
nlohmann: Please make this `const auto`.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -542,35 +576,78 @@ class serializer
                  int> = 0>
     void dump_integer(NumberType x)
     {
+        static constexpr std::array<std::array<char, 3>, 100> digits_to_99
```
nlohmann: Why `std::array<char, 3>` and not `std::array<char, 2>`?

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -527,6 +527,40 @@ class serializer
...
 
+    /*!
+    @brief count digits
+
+    Count the number of decimal (base 10) digits for an input unsigned integer.
+
+    @param[in] x  unsigned integer number to count its digits
+    @return    number of decimal digits
+    */
+    inline unsigned int count_digits(number_unsigned_t x)
```
nlohmann: Please make this `noexcept`.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -542,35 +576,78 @@ class serializer
...
+        buffer_ptr += n_digits;
+
+        // Fast int2ascii implementation inspired by "Fastware" talk by Andrei Alexandrescu
+        // See: https://www.youtube.com/watch?v=o4-CwDo2zpg
+        auto buffer_end = buffer_ptr;
+        while (abs_value >= 100)
+        {
+            auto digits_index = static_cast<unsigned>((abs_value % 100));
+            abs_value /= 100;
+            *--buffer_ptr = digits_to_99[digits_index][1];
```
nlohmann: Please add parentheses for readability: `*(--buffer_ptr)`.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -542,35 +576,78 @@ class serializer
...
+            *--buffer_ptr = digits_to_99[digits_index][0];
         }
 
-        if (is_negative)
+        if (abs_value >= 10)
+        {
+            unsigned digits_index = static_cast<unsigned>(abs_value);
+            *--buffer_ptr = digits_to_99[digits_index][1];
```
nlohmann: Please add parentheses for readability: `*(--buffer_ptr)`.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -542,35 +576,78 @@ class serializer
...
+            *--buffer_ptr = digits_to_99[digits_index][0];
+        }
+        else
         {
-            // make sure there is capacity for the '-'
-            assert(i < number_buffer.size() - 2);
-            number_buffer[i++] = '-';
+            *--buffer_ptr = static_cast<char>('0' + abs_value);
```
nlohmann: Please add parentheses for readability: `*(--buffer_ptr)`.

include/nlohmann/detail/output/serializer.hpp
nickaein: Nice catch! I was experimenting with a larger lookup table (with 3x1000 entries). Forgot to revert it back to 2x100.

single_include/nlohmann/json.hpp
```diff
@@ -11425,35 +11459,78 @@ class serializer
...
+        // jump to the end to generate the string from backward
+        // so we later avoid reversing the result
+        buffer_ptr += n_digits;
+
+        // Fast int2ascii implementation inspired by "Fastware" talk by Andrei Alexandrescu
+        // See: https://www.youtube.com/watch?v=o4-CwDo2zpg
+        const auto buffer_end = buffer_ptr;
+        while (abs_value >= 100)
+        {
+            auto digits_index = static_cast<unsigned>((abs_value % 100));
```
nlohmann: Please make this `const auto`. (Sorry, I overlooked that before.)

single_include/nlohmann/json.hpp
```diff
@@ -11425,35 +11459,78 @@ class serializer
...
+            abs_value /= 100;
+            *(--buffer_ptr) = digits_to_99[digits_index][1];
+            *(--buffer_ptr) = digits_to_99[digits_index][0];
         }
 
-        if (is_negative)
+        if (abs_value >= 10)
+        {
+            unsigned digits_index = static_cast<unsigned>(abs_value);
```
nlohmann: Please make this `const auto`. (Sorry, I overlooked that before.)

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -542,35 +576,78 @@ class serializer
...
+            abs_value = static_cast<number_unsigned_t>(0 - x);
+        }
+        else
         {
-            // spare 1 byte for '\0'
-            assert(i < number_buffer.size() - 1);
+            abs_value = static_cast<number_unsigned_t>(x);
+        }
+
+        const auto n_digits = count_digits(abs_value);
```
nlohmann: Just an idea:

1. Increment `n_digits` by 1 in case of a negative number. Then it would always be the number of bytes that we are writing.
2. This could allow to refine the assertion below (because we know whether or not we need to take care about a minus sign).
3. We could also use `n_digits` as the last parameter of `write_characters` and do not need to calculate it from the iterators.

nickaein: Good idea! Done.

nickaein: BTW, I am applying these changes as a fixup to previous commits using `git rebase` (with eventual `git push -f`). Is this the preferred method to apply these fixup changes to a pull request? Or I should submit them in a new commit?

nlohmann: There is no preference for this yet. Just do whatever you like :)

jaredgrubb:
Looks good to me!

nlohmann:
Thanks a lot!

Issue #1374, opened 2018-11-25
appveyor x64 builds appear to be using Win32 toolset


rmawatson:
need -A x64 or -DCMAKE_GENERATOR_PLATFORM=x64

Issue #971, opened 2018-02-13
Add a SAX parser


nlohmann:
The library currently only supports DOM-like parsing. This does not scale when the input files are enormous (#927). I would like to discuss how a SAX-like parser could look like.

My proposal (heavily motivated by [RapidJSON](http://rapidjson.org/md_doc_sax.html)) is as follows:

```cpp
struct SAX
{
    // a null value was read
    bool null();

    // a boolean value was read
    bool boolean(bool);

    // an integer number was read
    bool number_integer(number_integer_t);

    // an unsigned integer number was read
    bool number_unsigned(number_unsigned_t);

    // a floating-point number was read
    // the string parameter contains the raw number value
    bool number_float(number_float_t, const std::string&);

    // a string value was read
    bool string(const std::string&);

    // the beginning of an object was read
    // binary formats may report the number of elements
    bool start_object(std::size_t elements);

    // an object key was read
    bool key(const std::string&);

    // the end of an object was read
    bool end_object();

    // the beginning of an array was read
    // binary formats may report the number of elements
    bool start_array(std::size_t elements);

    // the end of an array was read
    bool end_array();

    // a binary value was read
    // examples are CBOR type 2 strings, MessagePack bin, and maybe UBJSON array<uint8t>
    bool binary(const std::vector<uint8_t>& vec);

    // a parse error occurred
    // the byte position and the last token are reported
    bool parse_error(int position, const std::string& last_token);
};
```

Some remarks:

- All functions return a `bool`: `true` if the parser should continue or `false` if the parser should stop processing the input.
- The proposal covers parsing of JSON, but also of CBOR, MessagePack, and UBJSON. Therefore, it contains extensions like array or object sizes as well as a binary type which do not occur when parsing JSON.
- The idea is that the user would implement the above struct (we need to discuss whether we make all functions virtual, define a default implementation, etc.) and pass it to new parse functions, e.g. `void parse_json(SAX &sax);` or `void parse_ubjson(SAX &sax);`.

What do you think?

gregmarr:
>The idea is that the user would implement the above struct (we need to discuss whether we make all functions virtual, define a default implementation, etc.) and pass it to new parse functions, e.g. void parse_json(SAX &sax); or void parse_ubjson(SAX &sax);.

For "define a default implementation", were you thinking that the library would include the struct declaration in the headers, and the implementation in a separate .cpp file that could be used or replaced as needed?

theodelrieu:
[...]

PR #1414, opened 2019-01-06
Fix x64 target platform for appveyor

nickaein:
This pull request fixes https://github.com/nlohmann/json/issues/1374.

More details are explained in cmake documentation for Visual Studio [2015](https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2014%202015.html)/[2017](https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2015%202017.html)

Note that the value of `platform` specified in `appveyor.yml` doesn't always match the target platform string required by cmake (e.g. `x86` in appveyor.yml vs. ` cmake -A Win32`). Therefore, we set the target platform for x64 directly in the generator name instead of passing the `platform` variable to cmake.

I've already run Appveyor on my fork to make sure this change is effective: https://ci.appveyor.com/project/nickaein/nlohmann-json/builds/21408703

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,17 +26,17 @@ environment:
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
       platform: x64
       FLAGS: ""
-      GENERATOR: Visual Studio 14 2015
+      GENERATOR: Visual Studio 14 2015 Win64
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
       platform: x64
       FLAGS: ""
-      GENERATOR: Visual Studio 15 2017
+      GENERATOR: Visual Studio 15 2017 Win64
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
       platform: x64
       FLAGS: "/permissive- /std:c++latest /utf-8"
-      GENERATOR: Visual Studio 15 2017
+      GENERATOR: Visual Studio 15 2017 Win64
 
 init:
   - cmake --version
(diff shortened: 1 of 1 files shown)


nickaein:
CI job for Visual Studio 2017 x64 with flags `/permissive- /std:c++latest /utf-8` is failing with a *Stack overflow* SEGFAULT error for file `test/data/json_testsuite/sample.json`. There might be some bug in the test unit in handling this file.

nickaein:
After some digging, the issue seems to be caused by small stack size (1MB) on x64 platform on Windows. For demonstration, setting a higher stack size inside `test/CMakeLists.txt` fixes the issue:

```
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /F2000000")
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /STACK:2000000")
```

As default stack size is higher (usually 8MB) on Linux, the tests doesn't fail for Linux x64 builds.

It seems the root cause is the recursive implementation of [write_ubjson](https://github.com/nlohmann/json/blob/6f89613acde82b81e824af5e43be132369c02fbf/include/nlohmann/detail/output/binary_writer.hpp#L543) which is susceptible to stackoverflow for deeply-nested JSON objects. Should `write_ubjson` be rewritten as non-recursive? Is there any other concerns (.e.g. performance)?

nickaein:
@nlohmann What's your take on this? The PR fixes the build on x64 but the recursive implementation of [`write_ubjson`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/output/binary_writer.hpp#L549) causes stack overflow on Win x64 due to small default stack size on Win x64.

Will [adding SAX parser](https://github.com/nlohmann/json/issues/971) eventually fix this issue? UBJSON seems to be covered in the proposal, but the implementation of `write_ubjson` in `develop` branch is still recursive.

nickaein:
Good idea! This should be doable, I will give it a try.

nlohmann:
Thanks a lot!

PR #1428, opened 2019-01-13
buildsystem: fix support for cmake older than 3.8

yann-morin-1998:
In e8b6b7ad, we relaxed the cmake version check down to 3.1, the first
version to expose target_compile_features().

However, an error while testing that change improperly concluded that
the change was OK. While target_compile_features() was indeed intriduced
in cmake 3.1, the actual feature we use it to test, cxx_std_11, was
really introduced only with cmake-3.8, which explained the actual
version that was requested before e8b6b7ad.

As Patrick noted, however, we can still convince cmake-3.1 to test
for C++11, by testing for an actual feature introduced in C++11, like
testing for cxx_range_for, which will instruct cmake to set the
appropriate C++11 options for the current compiler, which for gcc would
be adding -std=gnu++11.

Reported-by: Patrick Boettcher <patrick.boettcher@posteo.de>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,7 @@ endif()
 ##
 add_library(${NLOHMANN_JSON_TARGET_NAME} INTERFACE)
 add_library(${PROJECT_NAME}::${NLOHMANN_JSON_TARGET_NAME} ALIAS ${NLOHMANN_JSON_TARGET_NAME})
-target_compile_features(${NLOHMANN_JSON_TARGET_NAME} INTERFACE cxx_std_11)
+target_compile_features(${NLOHMANN_JSON_TARGET_NAME} INTERFACE cxx_range_for)
 
 target_include_directories(
     ${NLOHMANN_JSON_TARGET_NAME}
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -23,5 +23,5 @@ file(COPY ${CMAKE_SOURCE_DIR}/../test/data/regression/floats.json
 
 # benchmark binary
 add_executable(json_benchmarks src/benchmarks.cpp)
-target_compile_features(json_benchmarks PRIVATE cxx_std_11)
+target_compile_features(json_benchmarks PRIVATE cxx_range_for)
 target_link_libraries(json_benchmarks benchmark ${CMAKE_THREAD_LIBS_INIT})
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -64,7 +64,7 @@ set_target_properties(catch_main PROPERTIES
     COMPILE_DEFINITIONS "$<$<CXX_COMPILER_ID:MSVC>:_SCL_SECURE_NO_WARNINGS>"
     COMPILE_OPTIONS "$<$<CXX_COMPILER_ID:MSVC>:/EHsc;$<$<CONFIG:Release>:/Od>>"
 )
-target_compile_features(catch_main PUBLIC cxx_std_11)
+target_compile_features(catch_main PUBLIC cxx_range_for)
 target_include_directories(catch_main PRIVATE "thirdparty/catch")
 
 # https://stackoverflow.com/questions/2368811/how-to-set-warning-level-in-cmake


yann-morin-1998:
@pboettch Could you please give your opinion on this? Thanks! ;-)

nlohmann:
I'd also like to know whether @pboettch is fine with the change. I am just concerned about the different semantics:

- In 3.4.x we had `-std=c++11` as explicit flag for C++11.
- Then we moved to `cxx_std_11` which is the same, but compiler-agnostics. It was only ugly, because it required CMake 3.8.
- Now we only check whether ranged for loops are supported. I am aware that this is a C++11 feature, but there were compilers supporting this, but only a subset of C++11. These compiler should be excluded by the compiler version check in the library, though.

We could switch to 

```cmake
if (${CMAKE_VERSION} VERSION_LESS "3.8.0") 
    target_compile_features(catch_main PUBLIC cxx_range_for)
else()
    target_compile_features(catch_main PUBLIC cxx_std_11)
endif()
```

to use `cxx_range_for` only where `cxx_std_11` is not available. Do you think this is nitpicky?

pboettch:
In the meantime I had a problem with another project with `cxx_range_for` vs. `cxx_std_11`.

The conclusion there was to force CMake 3.8 and leave `cxx_std_11`. Because we wanted to force C++11.

`cxx_range_for` requires the compiler to _at least_ activate C++11; with gcc-7 it will be C++14 and gcc-8 even C++17 (I believe) whereas `cxx_std_11` forces C++11 even on compilers defaulting to later C++-versions.

In addition setting `cxx_std_11` as `PUBLIC` on a library will make all dependents inherit the forced flag.

**Hmm. Maybe for this library using `cxx_range_for` as a "at least C++11 is needed" is the best option.** 

There is also the property `CXX_STANDARD_11` and `CXX_STANDARD_REQUIRED`, but target-properties are not transitive and they don't solve the 'C++11 or later'-idea.

```CMake
set_property(TARGET <tgt> PROPERTY CXX_STANDARD 11)
set_property(TARGET <tgt> PROPERTY CXX_STANDARD_REQUIRED ON)
```

yann-morin-1998:
@pboettch @nlohmann Sorry guys, but this is straddling into C++ and cmake
topics I am definitely neither accustomed to nor aware of, and is getting far
from my initial goal of making it work on older cmake.

If it were me, I would just revert my previous change to at least restore the
previous, **working** situation. Then, if people are still motivated by supporting
older cmake versions, a better solution can be devised; then.

nlohmann:
I agree - we should revert to the status quo, requiring 3.8.0 and using `cxx_std_11`. @yann-morin-1998 can you make a PR, pleae?

yann-morin-1998:
@nlohmann Sure, I'll send it in a moment.

iwanders:
I would really like to be able to use a cmake version that's less recent. Some of us are still on Ubuntu 16.04, which ships with CMake 3.5.1. Using this library on Xenial requires modifying the CMakeLists just to change that feature requirement on the target.

I've filed #1441, which has the `if` statement approach @nlohmann  suggested, please consider merging that workaround to support older versions of cmake.

iwanders:
Closed #1441 because of the point raised by @pboettch  that the if statement will cause different behaviour depending on the CMake version, which is undesirable.

I presume in this case that we weren't setting `cxx_std_11` to force C++11 was used even by compilers that support C++14 and C++17. Because if we explicitly don't want C++14 or C++17 the following point isn't worth raising:

> Hmm. Maybe for this library using cxx_range_for as a "at least C++11 is needed" is the best option.

Instead of picking `cxx_range_for`, why not use `cxx_auto_type`? That is supported by CMake [3.1](https://cmake.org/cmake/help/v3.1/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html) and both range based for and `auto` were introduced with C++11. While that technically doesn't ensure all the features this library requires are provided by the compiler, it may strike a balance between the CMake version requirement and the way we select the C++ version?

iwanders:
Oh, now I see what you mean. You are right, `cxx_range_for` is supported by CMake 3.1 just like `cxx_auto_type` >_<. My bad, sorry about that.

If deemed acceptable I'd be happy to file a PR that does:
```cmake
cmake_minimum_required(VERSION 3.1)
#...
target_compile_features(${NLOHMANN_JSON_TARGET_NAME} INTERFACE cxx_range_for)
```

nlohmann:
Fixed with #1441.

Issue #832, opened 2017-11-16
Stack-overflow (OSS-Fuzz 4234)


nlohmann:
```
Detailed report: https://oss-fuzz.com/testcase?key=5854339613589504

Project: json
Fuzzer: libFuzzer_json_parse_cbor_fuzzer
Job Type: libfuzzer_asan_json
Platform Id: linux

Crash Type: Stack-overflow
Crash Address: 0x7ffc55e22ce8
Crash State:
nlohmann::basic_json<std::__1::map, std::__1::vector, std::__1::basic_string<cha
nlohmann::detail::binary_reader<nlohmann::basic_json<std::__1::map, std::__1::ve
nlohmann::basic_json<std::__1::map, std::__1::vector, std::__1::basic_string<cha

Sanitizer: address (ASAN)

Regressed: https://oss-fuzz.com/revisions?job=libfuzzer_asan_json&range=201701031958:201701032147

Reproducer Testcase: https://oss-fuzz.com/download?testcase_id=5854339613589504

Issue filed automatically.

See https://github.com/google/oss-fuzz/blob/master/docs/reproducing.md for more information.

This bug is subject to a 90 day disclosure deadline. If 90 days elapse
without an upstream patch, then the bug report will automatically
become visible to the public.
```

[clusterfuzz-testcase-minimized-5854339613589504.dms.zip](https://github.com/nlohmann/json/files/1479599/clusterfuzz-testcase-minimized-5854339613589504.dms.zip)

Stack trace:
```
AddressSanitizer:DEADLYSIGNAL
--
  | =================================================================
  | ==1==ERROR: AddressSanitizer: stack-overflow on address 0x7ffc55e22ce8 (pc 0x000000515f22 bp 0x7ffc55e23530 sp 0x7ffc55e22cd0 T0)
  | SCARINESS: 10 (stack-overflow)
  | #0 0x515f21 in operator new(unsigned long) _asan_rtl_
  | #1 0x5a6470 in __allocate /usr/local/include/c++/v1/new:226:10
  | #2 0x5a6470 in allocate /usr/local/include/c++/v1/memory:1747
[...]

PR #1431, opened 2019-01-14
Iterative destruction of JSON containers (objects and arrays)

mistersandman:
Deeply nested JSON hierarchies may lead to stack overflows during destruction, since JSON containers call the destructor of their children recursively (see issue #832).

This change provides an iterative destruction of JSON containers, avoiding a deep call stack by maintaining an explicit stack on the heap.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -1002,7 +1002,7 @@ class basic_json
                     if ((t == value_t::object && !this->object->empty()) || (t == value_t::array && !this->array->empty()))
                     {
                         std::vector<std::pair<json_value*, value_t>> stack;
-                        stack.push_back(std::make_pair(this, t));
+                        stack.emplace_back(this, t);
                         while (!stack.empty())
                         {
                             json_value* value;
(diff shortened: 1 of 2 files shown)


nlohmann:
Thanks for the PR!

jaredgrubb:
I applaud your effort here and the PR looks correct. However I found it **very** difficult to audit and read -- which makes me worry that it will be difficult to maintain.

The hardest thing here is that there are many levels of indentation, and worse, two nested while loops. 

I think splitting out the loops into separate functions (and avoiding having to split the top-level switch-case) makes it much easier to audit and understand what it's doing.

Eg, a rough sketch:
```
               case value_t::object:
               {
                    if (!this->object->empty())
                    {
                         destroy_subelements_depth_first(this);
                    }

                    std::allocator_traits<decltype(alloc)>::destroy(alloc, object);
                    std::allocator_traits<decltype(alloc)>::deallocate(alloc, object, 1);
                    break;
                }
                //// Ditto for array ....
```

And then:
```
void destroy_subelements_depth_first(json_value *value)
{
    std::vector<std::pair<json_value*, value_t>> stack;
    stack.emplace_back(this, t);
    while (!stack.empty())
    {
        json_value* value;
        value_t type;
        std::tie(value, type) = stack.back();
        bool all_done;
        if (type == value_t::object)
        {
             all_done = destroy_object_subelements(stack, value);
        }
        else
        {
             all_done = destroy_array_subelements(stack, value);
        }
        if (all_done)
        {
             stack.pop_back();
        }
    }
}

mistersandman:
@jaredgrubb Thank you very much for your feedback and sorry for the late reply! I changed the code pretty much exactly according to your suggestion and added a few comments. It is indeed much better readable now 👍

Issue #837, opened 2017-11-20
json_pointer public push_back, pop_back


HugoLopata:
**Feature Request**
I am generating some json pointers. It would be handy using push_back instead of composing string with slashes, which is subsequently broken to vector of strings by json_pointer.

Access to internal data may be handy too:
`const vector<string>& json_pointer::impl() const;`

nlohmann:
I think this addition makes sense - especially as reference tokens are stored inside a vector internally.

nlohmann:
@HugoLopata I added the functions to a feature branch, see 516d4e67087d5997596d3dea7f81d4cbc60ef8a4. Is this what you had in mind?

HugoLopata:
Yes, exactly as expected. 
Please add note to the documentation about difference between constructor and push_back.
Push_back do not unescape and slash is simply slash. See an example.

```
json::json_pointer ptr1("/foo/bar~0baz");
json::json_pointer ptr2;
ptr2.push_back("/foo/bar~0baz");
std::cout << "ptr1: " << ptr1.to_string() << '\n'; // /foo/bar~0baz
std::cout << "ptr2: " << ptr2.to_string() << '\n'; // /~1foo~1bar~00baz
```

Consider also adding some examples with escape characters to the tests.

nlohmann:
Oh I did not think about escaping. What do you think? Should `push_back` escape slashes?

HugoLopata:
- Variant 1: Single level, unescaped - as proposed in patch. This behavior is little bit inconsistent. Push_back is single level operation. No escaping needed. -> Simple usage, i like it.
- Variant 2: Single level, escaped - if you use the same algorithm as in constructor, then slashes should be denied (user should use ~0 instead). Consistent with constructor, but user has to do escaping (some `public static string escape(const &string)` will be convenient then).
- Variant 3: Multilevel push_back. I don't like it.

theodelrieu:
I am not fond of `push_back`, it's way too associated with containers.

However I think adding both `operator/` and `operator/=` would be great.

Whether these escape or not should depend on the most intuitive/expected outcome. I've never used JSON pointers, so this should be agreed upon by experienced users.

Anyway we have to support both cases, so mimicking `std::filesystem::path` seems reasonable (i.e. having `operator+=` and `concat`)

Boost.Filesystem decided to not have an `operator+`, I haven't seen the exact rationale, but my guess would be that they feared a lot of misuses would have happened with it.

[...]

PR #1434, opened 2019-01-15
allow push_back() and pop_back() calls on json_pointer

pboettch:
Putting pop_back() to public and creating a trivial push_back()
method allows users of nlohmann::json_pointer to manipulate an
existing json-pointer by adding or removing keys at the end.

This is useful for traversing a JSON-instance and keeping track
of its "absolute path" at any moment.

In my case for a schema-validator error-handler.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/json_pointer.hpp
+++ b/include/nlohmann/detail/json_pointer.hpp
@@ -97,7 +97,6 @@ class json_pointer
         return res;
     }
 
-  private:
     /*!
     @brief remove and return last reference pointer
     @throw out_of_range.405 if JSON pointer has no parent
@@ -114,6 +113,16 @@ class json_pointer
         return last;
     }
 
+    /*!
+    @brief remove and return last reference pointer
+    @throw out_of_range.405 if JSON pointer has no parent
+    */
+    void push_back(const std::string& tok)
+    {
+        reference_tokens.push_back(tok);
+    }
+
+  private:
     /// return whether pointer points to the root document
     bool is_root() const noexcept
     {
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -11885,7 +11885,6 @@ class json_pointer
         return res;
     }
 
-  private:
     /*!
     @brief remove and return last reference pointer
     @throw out_of_range.405 if JSON pointer has no parent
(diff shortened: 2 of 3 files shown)


include/nlohmann/detail/json_pointer.hpp
```diff
@@ -114,6 +113,16 @@ class json_pointer
         return last;
     }
 
+    /*!
+    @brief remove and return last reference pointer
+    @throw out_of_range.405 if JSON pointer has no parent
+    */
+    void push_back(const std::string& tok)
+    {
+        reference_tokens.push_back(tok);
```
nlohmann: Don't you need to make sure that the token is valid wrt. to the JSON Pointer grammar. I think tildes need to be escaped somehow. You would just pass the given string to the tokens without check.

nlohmann:
FYI: We had some discussions on this topic some time ago: #837. It would be great if you could have a look at that discussion and see if you can comment on it.

include/nlohmann/detail/json_pointer.hpp
pboettch: No, escaping is not needed when pushing back a key-value here, because escaping is done in the to_string()-function. It is not the single key which is escaped in a json-pointer, but the json-pointer it its slashed representation.

Here we are adding an element which is a non-escaped key. As we get a non-escaped when calling pop_back.

I don't know if there are any constraints regarding JSON-keys except than being strings. If there are, this would need to be checked here.

nlohmann:
Thanks!

nlohmann:
@pboettch It's been a long time that I merged this, and I am currently preparing the 3.6.0 release and thus working on documentation for the new JSON Pointer functions. Though I approved the respective PRs, I would not propose some changes before making the actual release, as later interface changes are complicated.

- I would make `array_index()` private. I see little value in having this helper function called by the user.
- I find it confusing that `pop_back()` returns a value. I would rather have it `void`, and add a function `back()` to return the last reference token. This would make the interface more consistent to `std::vector`.
- I would rename `parent_pointer()` to `parent()`, but that is just a stylistic question.

@pboettch What are your thoughts? I do not expect any PRs; I can make the changes myself once we get an agreement.

Issue #1419, opened 2019-01-09
Segmentation fault (stack overflow) due to unbounded recursion


guidovranken:
The following causes a segmentation fault (stack overflow) due to an apparent unbounded recursion.

```cpp
#include <string>
#include "json.hpp"
int main(void)
{
    const std::string s(512000, '[');
    try {
        auto j = nlohmann::json::parse(s);
    } catch ( ... )  { } ;

    return 0;
}
```

nickaein:
Good catch! 

There is also a similar issue related to `write_ubjson` recursive implementation: https://github.com/nlohmann/json/pull/1414

The example you posted demonstrate another case of recursion. While I couldn't find an obvious recursive implementation in `parser` and `json_sax`, the behavior tells a case of stack smashing.

Re-writing an recursive implementation as a non-recursive is generally an improvement. However, there might be some considerations in revising these algorithms since they are fundamental part of the library. @nlohmann can give some insight on how to proceed.

nickaein:
The recursive happens during array deconstruction [here](https://github.com/nlohmann/json/blob/db53bdac1926d1baebcb459b685dcd2e4608c355/include/nlohmann/json.hpp#L1010). The saniziter

nlohmann:
Thanks for reporting!

The parser itself is not recursive - the stack is modeled by a `std::vector<bool>` as we only need to know the nesting of objects and arrays. The input itself is not the problem, but the way the library models values. The input opens and nests one `std::vector<basic_json>` with each opening `[`.

It would be great if we could find a better way for this. Or a way to detect the issue and throw an exception before it happens.

The issue was already detected by Google's OSS Fuzz, see #832. There, I already asked for some opinions on how to proceed. I would mark this issue as duplicate and invite for a discussion in #832.

nlohmann:
Fixed by merging #1436.

PR #1436, opened 2019-01-15
Prevent stackoverflow caused by recursive deconstruction

nickaein:
This pull request fixes https://github.com/nlohmann/json/issues/832, https://github.com/nlohmann/json/issues/1419, #1835. 

For a deeply-nested JSON object, the recursive implementation of [`json_value::destroy`](https://github.com/nlohmann/json/blob/db53bdac1926d1baebcb459b685dcd2e4608c355/include/nlohmann/json.hpp#L995) function causes stack overflow. 

This is problematic for users that wish to expose a public API since an attacker can crash their systems by sending a deeply nested JSON request (e.g. a request containing 512K bytes of `[`).

In this pull request:

* First, a regression test has been added for this particular issue. We should probably add more regression tests for more complex scenarios.

* [`json_value::destroy`](https://github.com/nlohmann/json/blob/db53bdac1926d1baebcb459b685dcd2e4608c355/include/nlohmann/json.hpp#L995) method has been modified to deconstruct children of the JSON value iteratively to prevent triggering a recursive deconstruction. The computational complexity and memory requirement of this approach is `O(n)` which is equivalent to previous recursive deconstruction. Also, `std::move` has been used to minimize the overhead of flattening original JSON object.

**One question:** The iterative deconstruction can be also implemented inside [`~basic_json::basic_json`](https://github.com/nlohmann/json/blob/db53bdac1926d1baebcb459b685dcd2e4608c355/include/nlohmann/json.hpp#L1904). I am not sure which place is more appropriate. I have currently implemented it inside `json_value::destory` merely because I find it easier. 

* * *

## Pull request checklist

## Please don't


nlohmann:
How is this PR related to #1431?

nickaein:
@nlohmann Sorry for the long delay. While this PR and #1431 are in the same spirit, I believe this PR is more concise and easier to read. Also, it is accompanied by a basic unit test to prevent regression.

Regarding the other PR, I don't quite understand the implementation thus I cannot verify its validity. Specifically, I am not sure why there is a `break` at https://github.com/nlohmann/json/pull/1431/files#diff-641eb804df1899b4e8277af5a52e6b7dR1020. As the `value` object might have many children of type object/array, we should not stop processing them after the first hit.

nlohmann:
@nickaein I just resurrected this from the dead and forgotten, as I found #1835. I'm curious whether your solution fixes this.

nickaein:
@nlohmann Yes, this will probably fix #1835. I will check it with [bad_json_parsers](https://github.com/lovasoa/bad_json_parsers) to make sure. Thanks for reminding me!

nickaein:
I can confirm it passes [bad_json_parsers](https://github.com/lovasoa/bad_json_parsers) without crashing on any depths (1 to 5,000,000) provided in the test.

nickaein:
Travis CI stalled on one of the jobs (https://travis-ci.org/nlohmann/json/jobs/609598763?utm_medium=notification&utm_source=github_status). Restarting it might help.

nickaein:
Once again, a timeout issue with CI :-). Travis CI [terminates `valgrind test-regression`](https://travis-ci.org/nlohmann/json/jobs/609598763#L2602) since it doesn't output anything for 10 minutes.

This is caused by creating a very deep JSON object in [unit-test](https://github.com/nlohmann/json/pull/1436/files#diff-fd33adc37899b833d4a2c69d0ef251c1R1712), that requires a lot of memory allocations hence imposing a heavy burden on valgrind to check for memory leaks.

To mitigate this problem, I moved this test to a separate unit-test (`unit-large_json.cpp`). If this doesn't help, the next step would be running tests with `travis_wait` as indicated in [Travis docs](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) to periodically output a message.

Meanwhile, breaking down `test-regression` into multiple unit-tests could be helpful too since it already takes a considerable amount of time to run with valgrind (~5min on my workstation).

nlohmann:
Thanks for taking care about this. I would have taken a more complex approach and would have used the parser's callback mechanism to produce some output to tame Travis. But your approach is much cleaner. Thanks a lot!

Sorry for not have merged this issue earlier in January. I have no idea why it slipped through. But I did remember it the moment I saw the stacktrace in #1835.

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [x] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Fixed a bug that lead to stack overflow for deeply nested JSON values (objects, array) by changing the implementation of the destructor from a recursive to an iterative approach. #832, #1419, #1835

* * *

nickaein:
No problem at all! I had forgotten about it too, thanks for reminding me :-)

> I would have taken a more complex approach and would have used the parser's callback mechanism

That's interesting, I didn't know that is possible.

BTW, there were more considerations on Travis build time which posted in #1836 but I doubt if that qualify for an issue.

ilyapopov:
There seems to be a problem with the implementation: 

Calling vector::reserve in a loop can be big pessimization.It effectively disables exponential resizing, and makes everything O(n^2).

See a benchmark here: http://quick-bench.com/KT304ZLCCZpN8sd1tVzrGaPkl4A

Another question is: an array is cleared after it is inserted to the stack, but not an object. I suppose it will cause only one level of recursion (because values are now moved-from), so maybe it is fine.

nickaein:
Thanks for the feedback.

> It effectively disables exponential resizing, and makes everything O(n^2).

Very good point! I will look into it. PRs are welcome too :-)

> Another question is: an array is cleared after it is inserted to the stack, but not an object. I suppose it will cause only one level of recursion (because values are now moved-from), so maybe it is fine.

That's correct. I can't remember why the object wasn't explicitly cleared. While it might not make any difference, it's nice to clear it for the sake of consistency at least.

nickaein:
While the use of `vector::reserve` is sub-optimal in this case, the current benchmarks doesn't show a performance regression. They don't aim to precisely measure creation/destruction though. Also the previous state (destruction through recursion) might also had some overhead that worked in our favor.

Nevertheless, we should get rid of `vector::reserve` here. In addition, having dedicated benchmarks for measure the cost of creation/destruction can be more telling and let us avoid regressions in future.

ilyapopov:
Current parsing benchmarks specifically exclude destruction time from measurements.

https://github.com/nlohmann/json/blob/411158d896c86e210b8fdea47a41522d4408380b/benchmarks/src/benchmarks.cpp#L22-L25

nickaein:
I modified the benchmarks to only measure `delete j` time. The results show that there isn't that much of difference with or without `reserve`. Actually, having `reserve` gives us a slightly better performance, probably because there are not too many re-allocations during the deconstruction.

The performance of recursive deconstruction is significantly higher, but I believe avoiding the stack-overflow worth this performance penalty in destruction.

---

**non-recursive with `vector::reverse`:**

```

2019-11-11 13:39:20
Running ./json_benchmarks
Run on (12 X 4700 MHz CPU s)
CPU Caches:
  L1 Data 32K (x6)
  L1 Instruction 32K (x6)
  L2 Unified 256K (x6)
  L3 Unified 12288K (x1)
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
--------------------------------------------------------------
Benchmark                       Time           CPU Iterations
--------------------------------------------------------------
ParseFile/jeopardy         216309 us     216310 us          6   244.931MB/s
ParseFile/canada             3581 us       3581 us        195   599.517MB/s
ParseFile/citm_catalog       1642 us       1642 us        424   1003.14MB/s
ParseFile/twitter             560 us        560 us       1245   1075.24MB/s
```

---

**non-recursive without `vector::reverse`:**

```
2019-11-11 13:45:43
Running ./json_benchmarks
Run on (12 X 4700 MHz CPU s)
CPU Caches:
  L1 Data 32K (x6)
  L1 Instruction 32K (x6)
  L2 Unified 256K (x6)
  L3 Unified 12288K (x1)
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
--------------------------------------------------------------
Benchmark                       Time           CPU Iterations
--------------------------------------------------------------
ParseFile/jeopardy         231977 us     231974 us          6   228.392MB/s
ParseFile/canada             3139 us       3139 us        223   683.868MB/s
ParseFile/citm_catalog       1696 us       1696 us        416   971.453MB/s
ParseFile/twitter             617 us        617 us       1114   976.121MB/s
```

---

**recursive:**
```
2019-11-11 13:36:07
Running ./json_benchmarks
Run on (12 X 4700 MHz CPU s)
CPU Caches:
  L1 Data 32K (x6)
  L1 Instruction 32K (x6)
  L2 Unified 256K (x6)
  L3 Unified 12288K (x1)
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
--------------------------------------------------------------
Benchmark                       Time           CPU Iterations
--------------------------------------------------------------
ParseFile/jeopardy         161115 us     161113 us         10   328.844MB/s
ParseFile/canada             1424 us       1424 us        489   1.47193GB/s
ParseFile/citm_catalog        426 us        426 us       1629   3.77215GB/s
ParseFile/twitter             234 us        234 us       2992   2.51837GB/s
```

ilyapopov:
I made a benchmark that demonstrates the issue. To avoid this being buried in the comments to a closed PR, I created a separate issue #1837

PR #1452, opened 2019-01-22
Pass null bytes to the lexer when it makes sense.

### LINKED ISSUE #1095: Null bytes in files are treated like EOF
- What is the issue you have?

When reading parsing JSON from a file, null bytes are treated as EOF. The reason for this is explicit code in the JSON lexer to treat `\0` as EOF to properly parse string literals.

Parse file https://github.com/nst/JSONTestSuite/blob/master/test_parsing/n_multidigit_number_then_00.json, e.g. by

```cpp
std::ifstream f("n_multidigit_number_then_00.json");
json::parse(f);
```

Parse error: unexpected null byte after reading `123`.

The file is successfully parsed as number `123`.

Apple LLVM version 9.1.0.

`develop`.

- If you experience a compilation error: can you [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests)?

N/A

FrancoisChabot:
Fixes #1095 

The key point is that unlike what the comment said, string literals never went to the null-terminated string constructor of input-adapter, but to the sized-array constructor.

This change primarily makes it so that a sized-array's final character is ignored if it happens to be a null character. Otherwise, everything else was already in place for the null-check in the lexer to be considered redundant.

* * *

## Pull request checklist

## Please don't


include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -399,7 +399,7 @@ class input_adapter
     /// input adapter for array
     template<class T, std::size_t N>
     input_adapter(T (&array)[N])
-        : input_adapter(std::begin(array), std::end(array)) {}
+        : input_adapter(std::begin(array), array[N - 1] == 0 ? std::prev(std::end(array)) : std::end(array)) {}
```
nlohmann: Shouldn't you make sure that `N` is > 0?

test/src/unit-class_parser.cpp
```diff
@@ -1620,7 +1620,7 @@ TEST_CASE("parser class")
 
         SECTION("from std::array")
         {
-            std::array<uint8_t, 5> v { {'t', 'r', 'u', 'e'} };
+            std::array<uint8_t, 4> v { {'t', 'r', 'u', 'e'} };
```
nlohmann: Could you please add another test case for `std::array<uint8_t, 5> v { {'t', 'r', 'u', 'e', 0} }`?

test/src/unit-deserialization.cpp
```diff
@@ -358,7 +358,7 @@ TEST_CASE("deserialization")
 
             SECTION("from std::array")
             {
-                std::array<uint8_t, 5> v { {'t', 'r', 'u', 'e'} };
+                std::array<uint8_t, 4> v { {'t', 'r', 'u', 'e'} };
```
nlohmann: Could you please add another test case for `std::array<uint8_t, 5> v { {'t', 'r', 'u', 'e', 0} };`?

test/src/unit-class_parser.cpp
FrancoisChabot: Just to make sure we are on the same page, the new test would ensure that `std::array<uint8_t, 5> v { {'t', 'r', 'u', 'e', 0} }` fails, right?

test/src/unit-deserialization.cpp
nlohmann: It should be fine if the parser has the `strict=true` option, which is the default. But with `strict=false` this should be OK. So maybe this should be two tests.

include/nlohmann/detail/input/input_adapters.hpp
FrancoisChabot: That check is a bit on the redundant side since it would only apply to explicitly zero-sized arrays `char data[0]`, but I might as well add it in.

test/src/unit-class_parser.cpp
FrancoisChabot: Done

test/src/unit-deserialization.cpp
FrancoisChabot: done

nlohmann:
Do you have an idea why the MSVC and GCC tests are failing?

nlohmann:
Any idea why the CI is failing?

nickaein:
Definition of raw arrays with zero-length like `uint8_t v[] = {};` probably falls into realm of undefined behavior as suggested in [this SO answer](https://stackoverflow.com/a/6180200/161640). 

[`std::array`](https://en.cppreference.com/w/cpp/container/array) supports zero-length array though. Can `std::array` be used in these unit-tests instead?

nlohmann:
> Defining raw arrays with zero-length like `uint8_t v[] = {};` probably falls into realm of undefined behavior as suggested in [this SO answer](https://stackoverflow.com/a/6180200/161640).
> 
> [`std::array`](https://en.cppreference.com/w/cpp/container/array) supports zero-length array though. Can `std::array` be used in these unit-tests instead?

Yes, of course. Thanks for not checking back earlier.

nlohmann:
@FrancoisChabot Sorry for not checking on this earlier. Could you have a look at  https://github.com/nlohmann/json/pull/1452#issuecomment-466198221? It would be great to have this merged.

PR #1454, opened 2019-01-24
JSON-pointer: add operator+() returning a new json_pointer

pboettch:
Adds the operator+() to the json_pointer-class.

Fixes some comment/documentations as well.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/json_pointer.hpp
+++ b/include/nlohmann/detail/json_pointer.hpp
@@ -114,12 +114,31 @@ class json_pointer
     }
 
     /*!
-    @brief remove and return last reference pointer
-    @throw out_of_range.405 if JSON pointer has no parent
+    @brief append a token at the end of the reference pointer
     */
     void push_back(const std::string& tok)
     {
         reference_tokens.push_back(tok);
+    }
+
+    /*!
+    @brief append a key-token at the end of the reference pointer and return a new json-pointer.
+    */
+    json_pointer operator+(const std::string& tok) const
+    {
+        auto ptr = *this;
+        ptr.push_back(tok);
+        return ptr;
+    }
+
+    /*!
+    @brief append a array-index-token at the end of the reference pointer and return a new json-pointer.
+    */
+    json_pointer operator+(const size_t& index) const
+    {
+        auto ptr = *this;
+        ptr.push_back(std::to_string(index));
+        return ptr;
     }
 
   private:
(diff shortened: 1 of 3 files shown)


pboettch:
Good point, at first I was convinced you're right, but reading the documentation `std::filesystem::operator/()` makes me now think otherwise:

`std::filesystem::path` does not have a `operator/()`. It is a global function in the `std::filesystem`-namespace which is used to 

> Concatenate(s) two path components using the preferred directory separator if appropriate (see operator/= for details).

The operator I'm adding is not doing that but it only adds 1 new token add the end of the pointer.

My C++-powers are not going far enough to see whether this can be easily achieved with such an operator as well (and keeping it simple). But I'm eager to learn.

garethsb-sony:
In my opinion, it does make sense to provide this feature via ``op/`` and ``op/=`` as @FrancoisChabot  suggested, like so:

```C++
class json_pointer
{
    // existing class members
    explicit json_pointer(const std::string& s = "")
    // ...
    std::string pop_back()
    // ...
    void push_back(const std::string& tok)
    // ...

    // concatenate two `json_pointer`s
    json_pointer& operator/=(const json_pointer& ptr); // effectively returns ptr.reference_tokens.insert(reference_tokens.end(), ptr.reference_tokens.begin(), ptr.reference_tokens.end()), *this
    friend json_pointer operator/(const json_pointer& lhs, const json_pointer& rhs); // effectively returns json_pointer(lhs) /= rhs

    // append a single unescaped `reference-token` to a `json_pointer`
    json_pointer& operator/=(std::string tok); // effectively returns this->push_back(tok), *this
    json_pointer& operator/=(std::size_t index); // effectively returns *this /= std::to_string(index)
    friend json_pointer operator/(const json_pointer& lhs, std::string tok); // effectively returns json_pointer(lhs) /= tok
    friend json_pointer operator/(const json_pointer& lhs, std::size_t index); // effectively returns json_pointer(lhs) /= index
};
```

The operation of removing the last *reference-token* is less common and ``back`` and ``pop_back`` serve this purpose, though adding a ``parent_pointer`` member function that returned a new ``json_pointer`` would mirror ``std::filesystem::path::parent_path`` nicely.

pboettch:
Thanks for suggesting this. @FrancoisChabot already convinced. A lack of time made me not getting forward. @garethsb-sony Feel free to get along with this PR with your suggestion.

PR #1469, opened 2019-01-31
Add operator/= and operator/ to construct a JSON pointer by appending two JSON pointers

garethsb-sony:
Add ``operator/=`` and ``operator/`` to construct a JSON pointer by appending two JSON pointers, as well as convenience ``op/=`` and ``op/`` to append a single unescaped token or array index. Design inspired by ``std::filesystem::path``, based on feedback on PR #1454.

* * *

## Pull request checklist

### DIFF
--- a/include/nlohmann/detail/json_pointer.hpp
+++ b/include/nlohmann/detail/json_pointer.hpp
@@ -77,6 +77,52 @@ class json_pointer
     }
 
     /*!
+    @brief append another JSON pointer at the end of this JSON pointer
+    */
+    json_pointer& operator/=(const json_pointer& ptr)
+    {
+        reference_tokens.insert(reference_tokens.end(), ptr.reference_tokens.begin(), ptr.reference_tokens.end());
+        return *this;
+    }
+
+    /// @copydoc push_back(std::string&&)
+    json_pointer& operator/=(std::string token)
+    {
+        push_back(std::move(token));
+        return *this;
+    }
+
+    /// @copydoc operator/=(std::string)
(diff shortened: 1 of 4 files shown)


pboettch:
FWIW: LGTM

garethsb-sony:
Thanks, @pboettch. Hopefully @nlohmann approves too. :-)

nlohmann:
Very nice!

garethsb-sony:
Thanks. How would you feel about any or all of:

* making ``is_root`` public, potentially with a rename to ``empty`` for consistency with ``std::filesystem::path`` (this would avoid users having to compare vs. default-constructed ``json_pointer`` to detect this)
* adding ``parent_pointer`` for consistency with ``std::filesystem::path::parent_path`` (parent of empty pointer is empty)

The fact that we have ``pop_back`` that returns a value (and no ``back``) bothers me a little, but that's probably not a change to make at this point. Likewise the inconsistency of the name of the private ``top`` member function (I believe only used by ``basic_json::patch``); it ought to be ``front`` for consistency if it were ever made public.

garethsb-sony:
@nlohmann I have now added ``parent_pointer`` and ``empty`` as described, and updated the unit tests. Do you think it would be possible to merge these few ``json_pointer`` changes sometime? Thanks.

garethsb-sony:
I think the reported Travis CI failures are not actually related to this PR, am I right?

nlohmann:
Thanks a lot, and thanks in particular for your patience!

Issue #1471, opened 2019-02-01
Can we have an `exists()` method?


p-i-:
I wish to determine whether a particular key exists.

This is one of those things it's reasonable to expect, and surprising that there is no easy / obvious / direct solution.

`is_null()` (rather dangerously IMHO) silently creates the key.

https://stackoverflow.com/questions/49850890/c-nlohmann-json-how-to-test-nested-object-is-existent-or-empty suggests `.count()` but again is not obvious how to apply this. 

I've seen other solutions suggested involving iterators.

Pretty sure I saw a nice solution the other day in an issue, but now I can't find it.

Why not simply have an `exists()` method?

The documentation is super helpful for ENcoding JSON, but lacking in code examples for _extracting from_ JSON.

nlohmann:
You could implement this yourself with

```cpp
bool exists(const json& j, const std::string& key)
{
    return j.find(key) != j.end();
}
```

gregmarr:
> is_null() (rather dangerously IMHO) silently creates the key.

No it does not.  There is no key involved in an `is_null()` call, so it can't create a key.

If you are referring to your `j["bar"].is_null()` test, then it is `j["bar"]` that creates the key, just like in `std::map`.  This is required in order to enable things like `j["bar"] = 1;`.

You will get the same result on the last line if you leave out the `.is_null()`.

    json j = {
        { "foo", 1 }
    };

    std::cout << j["bar"] << std::endl;

    std::cout << j << std::endl;
    // {"bar":null,"foo":1}

gregmarr:
The answer to your question is the same as the answer to this question: https://stackoverflow.com/questions/1939953/how-to-find-if-a-given-key-exists-in-a-c-stdmap

ammaratef45:
> You could implement this yourself with
> 
> ```c++
> bool exists(const json& j, const std::string& key)
> {
>     return j.find(key) != j.end();
> }
> ```

@nlohmann I think it's better to be available too.
It's the way we think of JSON using (we search for the key and act differently when it's found or not found)
If everyone would implement that so you should add it (I guess)

hydratim:
Make a PR. 
Why put all the pressure on the maintainer? It's an easy feature for anyone to contribute.

nlohmann:
[...]

PR #1474, opened 2019-02-10
Implement contains() member function

nickaein:
This pull request is intended to tackle #1471.

This adds `contains()` method which similar to [`std::map::contains()`](https://en.cppreference.com/w/cpp/container/map/contains) available in C++20. While there could be various APIs and choices for such function, this PR tries to be aligned with STL counterpart. 

Having such method, the user can check the existence of a key in a JSON object more concisely:

```
if(j.contains("foo"))
{
    // "foo" key exists
}
else
{
    // "foo" key not exists	
}
```

`std::map::contains()` has been proposed in [P0458R2](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0458r2.html) and is merged to the [latest C++20 draft N4800](https://github.com/cplusplus/draft/blob/master/papers/n4800.pdf) (as now). Boost library [has already adopted it](https://www.boost.org/doc/libs/master/doc/html/container/release_notes.html#container.release_notes.release_notes_boost_1_69_00) since then.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -3957,6 +3957,39 @@ class basic_json
         return is_object() ? m_value.object->count(std::forward<KeyT>(key)) : 0;
     }
 
+    /*!
+    @brief check the existence of an element in a JSON object
+
+    Check whether an element exists in a JSON object with key equivalent to
+    @a key. If the element is not found or the JSON value is not an object,
+    false is returned.
+
+    @note This method always returns false when executed on a JSON type
+          that is not an object.
+
+    @param[in] key key value to check its existence.
+
+    @return true if an element with specified @a key exists. If no such
+    element with such key is found or the JSON value is not an object,
+    false is returned.
+
+    @complexity Logarithmic in the size of the JSON object.
+
+    @since version 3.6.0
+    */
+    template<typename KeyT>
+    bool contains(KeyT&& key) const
+    {
+        if (is_object())
+        {
+            return (m_value.object->find(std::forward<KeyT>(key)) != m_value.object->end());
+        }
+        else
+        {
+            return false;
+        }
+    }
+
     /// @}
 
 
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -16456,6 +16456,39 @@ class basic_json
         return is_object() ? m_value.object->count(std::forward<KeyT>(key)) : 0;
     }
 
+    /*!
+    @brief check the existence of an element in a JSON object
+
+    Check whether an element exists in a JSON object with key equivalent to
+    @a key. If the element is not found or the JSON value is not an object,
+    false is returned.
+
+    @note This method always returns false when executed on a JSON type
+          that is not an object.
+
+    @param[in] key key value to check its existence.
+
+    @return true if an element with specified @a key exists. If no such
+    element with such key is found or the JSON value is not an object,
+    false is returned.
+
+    @complexity Logarithmic in the size of the JSON object.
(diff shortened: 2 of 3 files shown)


include/nlohmann/json.hpp
```diff
@@ -3957,6 +3957,39 @@ class basic_json
...
+    @complexity Logarithmic in the size of the JSON object.
+
+    @since version 3.6.0
+    */
+    template<typename KeyT>
+    bool contains(KeyT&& key) const
+    {
+        if (is_object())
+        {
+            return (m_value.object->find(key) != m_value.object->end());
```
nlohmann: I'm not sure, but shouldn't this be `find(std::forward(key))`?

nlohmann:
Very nice - I only have a small question, see above.

include/nlohmann/json.hpp
nickaein: Good point! Done.

nlohmann:
Thanks!

nickaein:
I initially assumed this will be available for the next minor release (`3.6.0`) since it's an backward-compatible API change. So I put `@since version 3.6.0` in the documentation. Sorry that I forgot to mention it.

Will this addition be available in `3.5.1` and should I open another PR to fix the documentation?

include/nlohmann/json.hpp
```diff
@@ -3957,6 +3957,39 @@ class basic_json
...
+    template<typename KeyT>
+    bool contains(KeyT&& key) const
+    {
+        if (is_object())
+        {
+            return (m_value.object->find(std::forward<KeyT>(key)) != m_value.object->end());
+        }
+        else
+        {
+            return false;
```
jaredgrubb: Should this throw?

jaredgrubb:
If I ask if a number contains a key, I think it should throw.
Even more so, if I ask an *array* if it contains a value, getting false back feels odd.

include/nlohmann/json.hpp
nickaein: It seems [std::map::find cannot be marked as noexcept](https://stackoverflow.com/a/34639172/161640). I am not sure if we can guarantee that conditions not happening in our case.

nickaein:
What would a better behavior is such cases is debatable. I tried to keep it on par with the already existing [`json::find()`](https://github.com/nickaein/nlohmann-json/blob/develop/include/nlohmann/json.hpp#L3903). For non-object types, it returns `std::end()` which is the same return value if the type is object and the key is not found.

Essentially `json.contains(x)` is equivalent to `json.find(x) != end()` and I believe this is probably expected behavior. It might be confusing that `json.find(x)` (and `json.count(x)`, etc) doesn't throw for non-objects but `json::contains(x)` does.

jaredgrubb:
Those are good arguments and, with that justification, I have to reluctantly agree. :)

In another world I would try to argue that they all should throw, but the cure is probably worse than the cause at this point.

nickaein:
@nlohmann What's your take on https://github.com/nlohmann/json/pull/1474#issuecomment-463180891

nlohmann:
I already adjusted the target version to 3.6.0 if this is what you mean.

nickaein:
Yes Thanks. The timeline still says

> [nlohmann](/nlohmann) added this to the [Release 3.5.1](/nlohmann/json/milestone/28) milestone [9 days ago](#event-2136514712)

but I can now see the sidebar has the correct milestone.

nlohmann:
I renamed the milestone which apparently does not change past log messages. :)

PR #1488, opened 2019-02-21
Make library buildable with -Werror=effc++

koalo:
Adding initialization of several members to avoid the
error: ... should be initialized in the member initialization list

Deleting copy constructors and assignment operators for classes
with pointer data member to avoid the
error: ... has pointer data members but does not override ...

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -52,6 +52,16 @@ class binary_reader
         assert(ia);
     }
 
+    ~binary_reader() = default;
+
+    // delete copy constructor and assignment operator since
+    // exclusive access to the input adapter is assumed
+    binary_reader(const binary_reader&) = delete;
+    binary_reader& operator=(const binary_reader&) = delete;
+
+    binary_reader(binary_reader&&) = default;
+    binary_reader& operator=(binary_reader&&) = default;
+
     /*!
     @param[in] format  the binary format to parse
     @param[in] sax_    a SAX event processor
--- a/include/nlohmann/detail/input/input_adapters.hpp
+++ b/include/nlohmann/detail/input/input_adapters.hpp
@@ -58,6 +58,14 @@ class file_input_adapter : public input_adapter_protocol
         : m_file(f)
     {}
 
+    // this class is only a wrapper around the file pointer and does
+    // not own it, so the defaults are safe
+    ~file_input_adapter() = default;
+    file_input_adapter(const file_input_adapter&) = default;
+    file_input_adapter& operator=(const file_input_adapter&) = default;
+    file_input_adapter(file_input_adapter&&) = default;
+    file_input_adapter& operator=(file_input_adapter&&) = default;
+
     std::char_traits<char>::int_type get_character() noexcept override
     {
         return std::fgetc(m_file);
--- a/include/nlohmann/detail/input/json_sax.hpp
+++ b/include/nlohmann/detail/input/json_sax.hpp
@@ -158,6 +158,16 @@ class json_sax_dom_parser
         : root(r), allow_exceptions(allow_exceptions_)
     {}
 
+    ~json_sax_dom_parser() = default;
+
+    // delete copy constructor and assignment operator since
+    // exclusive access to the JSON value to be manipulated is assumed
+    json_sax_dom_parser(const json_sax_dom_parser&) = delete;
+    json_sax_dom_parser& operator=(const json_sax_dom_parser&) = delete;
+
+    json_sax_dom_parser(json_sax_dom_parser&&) = default;
+    json_sax_dom_parser& operator=(json_sax_dom_parser&&) = default;
+
     bool null()
     {
         handle_value(nullptr);
@@ -306,7 +316,7 @@ class json_sax_dom_parser
     /// the parsed JSON value
     BasicJsonType& root;
     /// stack to model hierarchy of values
-    std::vector<BasicJsonType*> ref_stack;
+    std::vector<BasicJsonType*> ref_stack {};
     /// helper to hold the reference for the next object element
     BasicJsonType* object_element = nullptr;
     /// whether a syntax error occurred
@@ -333,6 +343,16 @@ class json_sax_dom_callback_parser
     {
         keep_stack.push_back(true);
     }
+
+    ~json_sax_dom_callback_parser() = default;
+
+    // delete copy constructor and assignment operator since
+    // exclusive access to the JSON value to be manipulated is assumed
+    json_sax_dom_callback_parser(const json_sax_dom_callback_parser&) = delete;
+    json_sax_dom_callback_parser& operator=(const json_sax_dom_callback_parser&) = delete;
+
+    json_sax_dom_callback_parser(json_sax_dom_callback_parser&&) = default;
+    json_sax_dom_callback_parser& operator=(json_sax_dom_callback_parser&&) = default;
 
     bool null()
     {
@@ -611,11 +631,11 @@ class json_sax_dom_callback_parser
     /// the parsed JSON value
     BasicJsonType& root;
     /// stack to model hierarchy of values
-    std::vector<BasicJsonType*> ref_stack;
+    std::vector<BasicJsonType*> ref_stack {};
     /// stack to manage which values to keep
-    std::vector<bool> keep_stack;
+    std::vector<bool> keep_stack {};
     /// stack to manage which object keys to keep
-    std::vector<bool> key_keep_stack;
+    std::vector<bool> key_keep_stack {};
     /// helper to hold the reference for the next object element
     BasicJsonType* object_element = nullptr;
     /// whether a syntax error occurred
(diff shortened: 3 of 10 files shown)


include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -52,6 +52,10 @@ class binary_reader
         assert(ia);
     }
 
+    // delete because of pointer members
+    binary_reader(binary_reader&) = delete;
+    binary_reader& operator=(binary_reader&) = delete;
+
```
jaredgrubb: Caveat: I'm not up to speed on this file's implementation :)

Two things:
1. The comment doesn't feel right. This isn't non-copyable because it has pointers (that justification would only apply if the pointers were *owned* by this class [aka, the destructor deletes them], but this class has no destructor).  Isn't it more that it "consumes" the inputs and expects that it has exclusive operation on the input stream. Ie, if you copy this class and then interleave calls to each copy, there is nothing coherent that comes from that. Therefore there is an implication of "exclusive access", not that there are pointers.
2. Your patch changes the ABI here because declaring the copy constructors removes the move constructors, and I believe this class should still be movable. Since this patch is all about following Effective rules, we should follow Rule of 5 and declare all the special functions explicitly here.

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -58,6 +58,16 @@ class file_input_adapter : public input_adapter_protocol
...
+    file_input_adapter(const file_input_adapter& other) noexcept
+        : m_file(other.m_file)
+    {}
+
+    file_input_adapter& operator=(const file_input_adapter& other) noexcept
+    {
+        m_file = other.m_file;
+        return *this;
+    }
+
```
jaredgrubb: 1. Why not just "=default" for both of these?
2. Same comment as above: we should always follow the Rule of 5.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -611,11 +619,11 @@ class json_sax_dom_callback_parser
...
     BasicJsonType& root;
     /// stack to model hierarchy of values
-    std::vector<BasicJsonType*> ref_stack;
+    std::vector<BasicJsonType*> ref_stack{};
     /// stack to manage which values to keep
-    std::vector<bool> keep_stack;
+    std::vector<bool> keep_stack{};
     /// stack to manage which object keys to keep
-    std::vector<bool> key_keep_stack;
+    std::vector<bool> key_keep_stack{};
```
nickaein: Bracket isn't necessary for types with default constrcutor like `std::string` and `std::vector`, correct? Is there an argument of consistency in initialization?

theodelrieu:
I do not think we should strive to please -Weffc++, especially if that means changing some perfectly valid code and potentially introduce bugs in the meantime.

You can take a look at this [SO comment](https://stackoverflow.com/a/11529328), one of GCC devs explains briefly why he never uses it.

To me this should only be applied to your own code.
To workaround the error you could simply include the library with `-isystem`.

koalo:
I fully agree that we should never make the code quality worse or forfeit efficiency just to please some overcautious static analysis tool!

Even though the commit message suggests that I am only interested in getting rid of the warnings, this is only one goal. The primary goal of the patch is to be more explicit and/or consistent at some places (that - admittedly - got pointed out by `-Weffc++`). I was also motivated to submit this patch by the mention of  `-Weffc++` in the main `README.md`.

Of course, I do not expect that you apply this patch if it changes the behaviour of the code in any way or makes the code quality worse. But if you are just unsure if it might change the behaviour of the code, I would say this is a sign that the code should be more explicit!

This holds for the additional (or deleted) constructors, but also for the addition of the `{}`. There should be no change in the behaviour of the code, but it indicates that we are fully aware that we want the default constructor. Since there are already some `{}` in the code I do not see a reason for not applying them more widespread. But I am also fine with removing them from the code, again.

theodelrieu:
>Since there are already some `{}` in the code I do not see a reason for not applying them more widespread.

IIRC the only time we add brackets is on types that must be value-initialized (e.g. integers).
I do not see the point to add them to `vector` variables to be honest, it is quite clear that the default constructor will be called and thus they will be empty.

For constructors, I'm not against your change, deleting the copy constructors here make sense.

include/nlohmann/detail/input/input_adapters.hpp
nlohmann: Does `-Weffc++` really force defaulting everything?

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -306,7 +316,7 @@ class json_sax_dom_parser
     /// the parsed JSON value
     BasicJsonType& root;
     /// stack to model hierarchy of values
-    std::vector<BasicJsonType*> ref_stack;
+    std::vector<BasicJsonType*> ref_stack {};
```
nlohmann: This should not be required. Please remove.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -611,11 +631,11 @@ class json_sax_dom_callback_parser
...
```
nlohmann: This should not be required. Please remove.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -611,11 +631,11 @@ class json_sax_dom_callback_parser
...
     /// stack to manage which values to keep
-    std::vector<bool> keep_stack;
+    std::vector<bool> keep_stack {};
```
nlohmann: This should not be required. Please remove.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -611,11 +631,11 @@ class json_sax_dom_callback_parser
...
     /// stack to manage which object keys to keep
-    std::vector<bool> key_keep_stack;
+    std::vector<bool> key_keep_stack {};
```
nlohmann: This should not be required. Please remove.

include/nlohmann/detail/input/lexer.hpp
```diff
@@ -1483,7 +1483,7 @@ class lexer
     bool next_unget = false;
 
     /// the start position of the current token
-    position_t position;
+    position_t position {};
```
nlohmann: This should not be required. Please remove.

include/nlohmann/detail/iterators/iter_impl.hpp
```diff
@@ -607,7 +607,7 @@ class iter_impl
     /// associated JSON instance
     pointer m_object = nullptr;
     /// the actual iterator of the associated instance
-    internal_iterator<typename std::remove_const<BasicJsonType>::type> m_it;
+    internal_iterator<typename std::remove_const<BasicJsonType>::type> m_it {};
```
nlohmann: This should not be required. Please remove.

test/src/unit-alt-string.cpp
```diff
@@ -149,7 +149,7 @@ class alt_string
     }
 
   private:
-    std::string str_impl;
+    std::string str_impl {};
```
nlohmann: This should not be required. Please remove.

test/src/unit-class_parser.cpp
```diff
@@ -125,7 +125,7 @@ class SaxEventLogger
         return false;
     }
 
-    std::vector<std::string> events;
+    std::vector<std::string> events {};
```
nlohmann: This should not be required. Please remove.

test/src/unit-deserialization.cpp
```diff
@@ -123,7 +123,7 @@ struct SaxEventLogger : public nlohmann::json_sax<json>
         return false;
     }
 
-    std::vector<std::string> events;
+    std::vector<std::string> events {};
```
nlohmann: This should not be required. Please remove.

test/src/unit-regression.cpp
```diff
@@ -124,8 +124,13 @@ struct nocopy
 
 struct Data
 {
-    std::string a;
-    std::string b;
+    Data() {}
```
nlohmann: Maybe default?

test/src/unit-regression.cpp
```diff
@@ -124,8 +124,13 @@ struct nocopy
...
+
+    Data(std::string a, std::string b)
+        : a(std::move(a)), b(std::move(b)) {}
+
+    std::string a {};
```
nlohmann: This should not be required. Please remove.

test/src/unit-regression.cpp
```diff
@@ -124,8 +124,13 @@ struct nocopy
...
+    std::string b {};
```
nlohmann: This should not be required. Please remove.

nlohmann:
FYI: I am currently working on compiler and linter warnings myself. Maybe I can fix this issue as well.

nlohmann:
@koalo The develop branch now builds without `-Weffc++` warnings (tried with gcc (GCC) 9.0.1 20190313). I played around with clang-tidy, and I fixed these warnings along the way. It would be great if you could double-check whether these fixed work for you, too.

koalo:
@nlohmann I builds without warnings now, thanks a lot! Also, good to see that you have added all the `{}` now that were rejected in my patch ;-)

Issue #670, opened 2017-07-29
Operator= ambiguous in C++1z and GCC 7.1.1


nick-prat:
With the new GCC 7.1.1 there is a bug with reassigning to strings.

In this code
```cpp
#include <string>
#include <iostream>
#include <json.hpp>

using json = nlohmann::json;

int main() {
    json js;
    std::string name = "";
    name = js["name"];
}
```
it produces the error
```
In file included from /usr/include/c++/7.1.1/string:52:0,
                 from json_error.cc:1:
/usr/include/c++/7.1.1/bits/basic_string.h: In instantiation of ‘std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::_If_sv<_Tp, std::__cxx11::basic_string<_CharT, _Traits, _Alloc>&> std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::operator=(_Tp) [with _Tp = nlohmann::basic_json<>; _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>; std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::_If_sv<_Tp, std::__cxx11::basic_string<_CharT, _Traits, _Alloc>&> = std::__cxx11::basic_string<char>&]’:
json_error.cc:10:21:   required from here
/usr/include/c++/7.1.1/bits/basic_string.h:764:28: error: call of overloaded ‘assign(nlohmann::basic_json<>&)’ is ambiguous
  { return this->assign(__sv); }
                            ^
/usr/include/c++/7.1.1/bits/basic_string.h:1313:7: note: candidate: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>& std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::assign(const std::__cxx11::basic_string<_CharT, _Traits, _Alloc>&) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]
       assign(const basic_string& __str)
       ^~~~~~
/usr/include/c++/7.1.1/bits/basic_string.h:1329:7: note: candidate: std::__cxx11::basic_string<_CharT, _Traits, _Alloc>& std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::assign(std::__cxx11::basic_string<_CharT, _Traits, _Alloc>&&) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]
       assign(basic_string&& __str)
[...]

PR #1496, opened 2019-02-25
fix GCC 7.1.1 - 7.2.1 on CentOS

lieff:
Closes https://github.com/nlohmann/json/issues/670

I've checked also gcc 5.5 and 8.1 and seems fix do not breaks anything.


lieff:
Also may be cleanup redundant check (defined(_MSC_VER) and _MSC_VER <= 1914) ?
```
#ifndef _MSC_VER  // fix for issue #167 operator<< ambiguity under VS2015
                   and not std::is_same<ValueType, std::initializer_list<typename string_t::value_type>>::value
#if defined(JSON_HAS_CPP_17) && defined(__GNUC__)
                   and not std::is_same<ValueType, typename std::string_view>::value
#endif
#endif
```

nlohmann:
Thanks!

PR #1517, opened 2019-03-14
Fix setting the C++ standard for Travis builds

thomasschaub:
Travis did not build with C++14 and C++17, but silently used the default version. Also the tests failed with C++17.

Before this patch, when passing `CMAKE_CXX_FLAGS` to CMake, the compiler would be invoked as follows: `/usr/bin/g++-7 ... -std=c++1z ... -std=gnu++11`, where `c++1z` was set via `CMAKE_CXX_FLAGS` and `gnu++11` was derived from a `target_compile_features` directive. The second flag overwrites the first, so whatever was set using `CMAKE_CXX_FLAGS` gets ignored.

The test was straight up failing and I implemented a new one. I came up with the test case from the [(missing) declaration of operator std::string_view](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json.hpp#L2863) in `basic_json`.


test/CMakeLists.txt
```diff
@@ -104,6 +106,20 @@ foreach(file ${files})
       CATCH_CONFIG_FAST_COMPILE
       $<$<CXX_COMPILER_ID:MSVC>:_SCL_SECURE_NO_WARNINGS>
     )
+    if (${CMAKE_VERSION} VERSION_LESS "3.8.0")
+        if (JSON_ForceCpp14)
+            target_compile_features(${testcase} PRIVATE cxx_binary_literals)
+        endif()
+        if (JSON_ForceCpp17)
+        message(FATAL_ERROR "Cannot force C++17 with this CMake version, you need at least version 3.8")
+        endif()    else()
```
pboettch: else should be on a new-line here, I think.

thomasschaub: True, sorry for the noise!

test/src/unit-conversions.cpp
```diff
@@ -480,8 +494,7 @@ TEST_CASE("value conversion")
 #if defined(JSON_HAS_CPP_17)
         SECTION("std::string_view")
         {
-            std::string_view s = j;
-            CHECK(json(s) == j);
+            static_assert(!is_implicitly_convertible<decltype(j), std::string_view>::value);
```
nlohmann: Why did you change the existing test?

test/src/unit-conversions.cpp
```diff
@@ -51,6 +51,20 @@ using nlohmann::json;
     #include <string_view>
 #endif
 
+template<typename From, typename To>
```
nlohmann: Why did you add this test?

test/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ option(JSON_Sanitizer "Build test suite with Clang sanitizer" OFF)
 option(JSON_Valgrind "Execute test suite with Valgrind" OFF)
 option(JSON_NoExceptions "Build test suite without exceptions" OFF)
 option(JSON_Coverage "Build test suite with coverage information" OFF)
+option(JSON_ForceCpp14 "Use C++14 standard" OFF)
```
nlohmann: This flag is not used in Travis or AppVeyor, right?

.travis.yml
```diff
@@ -212,7 +212,7 @@ matrix:
     compiler: gcc
     env:
       - COMPILER=g++-8
-      - CXXFLAGS=-std=c++17
+      - CMAKE_OPTIONS=-DJSON_ForceCpp17=ON
```
nlohmann: Should this flag also be used in AppVeyor?

test/CMakeLists.txt
thomasschaub: True, I will remove it.

.travis.yml
thomasschaub: The AppVeyor cfg currently uses `/std:c++latest`. Imo the new flag is more descriptive, but maybe I'm biased. :-) Shall I change the AppVeyor cfg as well? I think there's nothing to be fixed, but it would be more consistent, I guess.

test/src/unit-conversions.cpp
thomasschaub: From the PR description:

The test was straight up failing and I implemented a new one. I came up with the test case from the [(missing) declaration of operator std::string_view](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json.hpp#L2863) in `basic_json`.

Would you like me to elaborate on anything here?

test/CMakeLists.txt
nlohmann: Ok.

.travis.yml
nlohmann: Right, then I think we should leave it as is.

test/src/unit-conversions.cpp
nlohmann: Could we add a check to only have the test if C++17 support is used?

thomasschaub: There's a ` #if defined(JSON_HAS_CPP_17)` in line 494 (line 480 in the old version).

test/CMakeLists.txt
```diff
@@ -104,6 +105,21 @@ foreach(file ${files})
       CATCH_CONFIG_FAST_COMPILE
       $<$<CXX_COMPILER_ID:MSVC>:_SCL_SECURE_NO_WARNINGS>
     )
+    if (${CMAKE_VERSION} VERSION_LESS "3.8.0")
+        if (JSON_ForceCpp14)
```
nlohmann: That option does not exist (any more).

test/CMakeLists.txt
```diff
@@ -104,6 +105,21 @@ foreach(file ${files})
...
+            target_compile_features(${testcase} PRIVATE cxx_binary_literals)
+        endif()
+        if (JSON_ForceCpp17)
+            message(FATAL_ERROR "Cannot force C++17 with this CMake version, you need at least version 3.8")
+        endif()
+    else()
+        if (JSON_ForceCpp14)
```
nlohmann: That option does not exist (any more).

test/src/unit-conversions.cpp
nlohmann: Ok. But why is the whole string_view test gone now?

nlohmann:
Any idea why the AppVeyor tests fail? (Travis is always a bit buggy - I keep restarting the failing tests until they work...)

thomasschaub:
Regarding the failing builds:

is_implicity_convertible does not seem to work on MSVC, I will look into that. Here's a [reduced example](https://gcc.godbolt.org/z/8QlUZx), MSVC says it is convertible, GCC says it isn't.

The Clang 7 test installs g++6, presumably to get libstdc++. However, that doesn't ship with `string_view`. I could replace that with, e.g., `libc++-7-dev`?!

nlohmann:
I think we should be able to detect whether there is a `string_view` header.

thomasschaub:
I can introduce `JSON_HAS_STRING_VIEW` and define it in terms of `__has_include`, iirc libfmt does something similar.

test/src/unit-conversions.cpp
nlohmann: With the `JSON_HAS_STRING_VIEW` macro, is it possible to bring back the test

```cpp
SECTION("std::string_view")
{
    std::string_view s = j;
    CHECK(json(s) == j);
}
```

thomasschaub:
I finally got the checks to pass, is everybody happy with the current state? :-)

nlohmann:
I still have one question: 
https://github.com/nlohmann/json/pull/1517#discussion_r268421691

test/src/unit-conversions.cpp
thomasschaub: I removed the template magic that would prevent the implicit conversion to `string_view` and adjusted the test accordingly. I'm still a bit worried that there was a good reason why the implicit conversion was disabled, though.

gregmarr: I know that there has been discussion about removing all the implicit conversions.  I don't remember if it was completed or not.  This might have been part of that.

nlohmann: No, the implicit conversions have not yet been removed. The test above worked so far, and if we merge this PR, it would be necessary that it still works as we otherwise would break existing implementations.

PR #1523, opened 2019-03-18
Fix inconsistent of type alias 

### LINKED ISSUE #1507: Inconsistent use of type alias.
Inside `basic_json` the alias `reference` and `const_reference` are used in inconsistently, for example the method `iteration_proxy` uses `const_reference` as a parameter type, while `push_back` uses `const basic_json&`.
Which one should be used? I think the non-alias are better and i would like make a PR if you agree.

Sorry about my horrible english.

DodoMomo:
Fixes #1507

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -126,10 +126,10 @@ and `from_json()` (@ref adl_serializer by default)
 - Library-wide
  - [EqualityComparable](https://en.cppreference.com/w/cpp/named_req/EqualityComparable):
    JSON values can be compared with `==`, see @ref
-   operator==(const_reference,const_reference).
+   operator==(const basic_json&,const basic_json&).
  - [LessThanComparable](https://en.cppreference.com/w/cpp/named_req/LessThanComparable):
    JSON values can be compared with `<`, see @ref
-   operator<(const_reference,const_reference).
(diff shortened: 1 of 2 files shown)


nlohmann:
Sorry, I messed up the current develop branch. I shall commit a fix in a second.

nlohmann:
Please rebase to 37a72dac48035e12df13c923e19f10cbe8a22f5f.

nlohmann:
FYI:
https://github.com/nlohmann/json/issues/1507#issuecomment-473994405

include/nlohmann/json.hpp
```diff
@@ -2912,7 +2912,7 @@ class basic_json
     written using `at()`. It also demonstrates the different exceptions that
     can be thrown.,at__size_type}
     */
-    reference at(size_type idx)
+    basic_json& at(size_type idx)
```
nlohmann: I think the change from `reference` to `basic_json&` is not improving readability. Unless there are good reasons, I would not approve this PR.

DodoMomo:
So what do you think is the right procedure for type alias? Only parts of the STL-like interface? Every place where the it refers to the element type?

nlohmann:
I would say each occurrence of (const) `basic_json&` in the interface should be replaced by (const) `reference`.

PR #1559, opened 2019-04-04
Feat/explicit conversion operator

theodelrieu:
Here it is!

Thanks to @gracicot, who found a good solution to #958, I'm finally attempting to solve it.

This PR introduces a CMake option: `JSON_ExplicitConversions`, defaulted to `OFF`.

To help users move away from the implicit conversion operator, a warning will be issued each time it is used. This warning can be silenced by users defining `JSON_USE_IMPLICIT_CONVERSIONS`.

On the other hand, defining `JSON_USE_EXPLICIT_CONVERSIONS` will very likely break code, but will ease the transition to the next major version, where only explicit conversion operator will be kept.

Meanwhile, I've removed every implicit conversion from the library's code, including tests.
It might be a good idea to add a short `ifdef`ed test case, which tests implicit conversions (to avoid duplicating a lot of test code).

With explicit conversions, we can also remove the constraints that are present today, e.g. forbidding conversion from `char`, `std::initializer_list<char>`. Note that this has not been done in this PR.

I think this change is more suited to be released in a minor version, so we might release 3.6.2 first?

## Pull request checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -145,6 +145,11 @@ matrix:
   - os: osx
     osx_image: xcode12
 
+  - os: osx
+    osx_image: xcode12
+    env:
+      - IMPLICIT_CONVERSIONS=OFF
+
   # Linux / GCC
 
   - os: linux
@@ -198,6 +203,16 @@ matrix:
   - os: linux
     compiler: gcc
     env: COMPILER=g++-9
+    addons:
+      apt:
+        sources: ['ubuntu-toolchain-r-test']
+        packages: ['g++-9', 'ninja-build']
+
+  - os: linux
+    compiler: gcc
+    env:
+      - COMPILER=g++-9
+      - IMPLICIT_CONVERSIONS=OFF
     addons:
       apt:
         sources: ['ubuntu-toolchain-r-test']
@@ -315,10 +330,12 @@ script:
   - if [[ "${COMPILER}" != "" ]]; then export CXX=${COMPILER}; fi
   # by default, use the single-header version
   - if [[ "${MULTIPLE_HEADERS}" == "" ]]; then export MULTIPLE_HEADERS=OFF; fi
+  # by default, use implicit conversions
+  - if [[ "${IMPLICIT_CONVERSIONS}" == "" ]]; then export IMPLICIT_CONVERSIONS=ON; fi
 
   # compile and execute unit tests
   - mkdir -p build && cd build
-  - cmake .. ${CMAKE_OPTIONS} -DJSON_MultipleHeaders=${MULTIPLE_HEADERS} -DJSON_BuildTests=On -GNinja && cmake --build . --config Release
+  - cmake .. ${CMAKE_OPTIONS} -DJSON_MultipleHeaders=${MULTIPLE_HEADERS} -DJSON_ImplicitConversions=${IMPLICIT_CONVERSIONS} -DJSON_BuildTests=On -GNinja && cmake --build . --config Release
   - ctest -C Release --timeout 2700 -V -j
   - cd ..
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ endif ()
 option(JSON_BuildTests "Build the unit tests when BUILD_TESTING is enabled." ON)
 option(JSON_Install "Install CMake targets during install step." ON)
 option(JSON_MultipleHeaders "Use non-amalgamated version of the library." OFF)
+option(JSON_ImplicitConversions "Enable implicit conversions" ON)
 
 ##
 ## CONFIGURATION
@@ -59,6 +60,12 @@ if (${CMAKE_VERSION} VERSION_LESS "3.8.0")
 else()
     target_compile_features(${NLOHMANN_JSON_TARGET_NAME} INTERFACE cxx_std_11)
 endif()
+
+target_compile_definitions(
+    ${NLOHMANN_JSON_TARGET_NAME}
+    INTERFACE
+    JSON_USE_IMPLICIT_CONVERSIONS=$<BOOL:${JSON_ImplicitConversions}>
+)
 
 target_include_directories(
     ${NLOHMANN_JSON_TARGET_NAME}
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,7 @@ environment:
       platform: x86
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
+      CMAKE_OPTIONS: ""
       GENERATOR: Visual Studio 14 2015
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
@@ -14,6 +15,7 @@ environment:
       platform: x86
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
+      CMAKE_OPTIONS: ""
       GENERATOR: Visual Studio 15 2017
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
@@ -21,6 +23,7 @@ environment:
       platform: x86
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
+      CMAKE_OPTIONS: ""
       GENERATOR: Visual Studio 16 2019
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
@@ -28,6 +31,7 @@ environment:
       platform: x64
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
+      CMAKE_OPTIONS: ""
       GENERATOR: Visual Studio 16 2019
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
@@ -36,6 +40,7 @@ environment:
       platform: x86
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
+      CMAKE_OPTIONS: ""
       GENERATOR: Ninja
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
@@ -44,6 +49,7 @@ environment:
       platform: x86
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
+      CMAKE_OPTIONS: ""
       GENERATOR: Ninja
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
@@ -51,6 +57,7 @@ environment:
       platform: x86
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
+      CMAKE_OPTIONS: ""
       GENERATOR: Visual Studio 14 2015
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
@@ -59,6 +66,7 @@ environment:
       name: with_win_header
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
+      CMAKE_OPTIONS: ""
       GENERATOR: Visual Studio 14 2015
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
@@ -66,6 +74,7 @@ environment:
       platform: x86
       CXX_FLAGS: "/permissive- /std:c++latest /utf-8"
       LINKER_FLAGS: ""
+      CMAKE_OPTIONS: ""
       GENERATOR: Visual Studio 15 2017
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
@@ -73,6 +82,15 @@ environment:
       platform: x86
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
+      CMAKE_OPTIONS: "-DJSON_ImplicitConversions=OFF"
+      GENERATOR: Visual Studio 16 2019
+
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+      configuration: Release
+      platform: x64
+      CXX_FLAGS: ""
+      LINKER_FLAGS: ""
+      CMAKE_OPTIONS: ""
       GENERATOR: Visual Studio 16 2019
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
@@ -80,6 +98,7 @@ environment:
       platform: x64
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
+      CMAKE_OPTIONS: ""
       GENERATOR: Visual Studio 14 2015
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
@@ -87,6 +106,7 @@ environment:
       platform: x64
       CXX_FLAGS: "/permissive- /std:c++latest /Zc:__cplusplus /utf-8 /F4000000"
       LINKER_FLAGS: "/STACK:4000000"
+      CMAKE_OPTIONS: ""
       GENERATOR: Visual Studio 15 2017
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
@@ -94,6 +114,7 @@ environment:
       platform: x64
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
+      CMAKE_OPTIONS: ""
       GENERATOR: Visual Studio 16 2019
 
 init:
@@ -112,7 +133,7 @@ before_build:
   # for with_win_header build, inject the inclusion of Windows.h to the single-header library
   - ps: if ($env:name -Eq "with_win_header") { $header_path = "single_include\nlohmann\json.hpp" }
   - ps: if ($env:name -Eq "with_win_header") { "#include <Windows.h>`n" + (Get-Content $header_path | Out-String) | Set-Content $header_path }
-  - if "%GENERATOR%"=="Ninja" (cmake . -G "%GENERATOR%" -DCMAKE_BUILD_TYPE="%configuration%" -DCMAKE_CXX_FLAGS="%CXX_FLAGS%" -DCMAKE_EXE_LINKER_FLAGS="%LINKER_FLAGS%" -DCMAKE_IGNORE_PATH="C:/Program Files/Git/usr/bin" -DJSON_BuildTests=On) else (cmake . -G "%GENERATOR%" -A "%GENERATOR_PLATFORM%" -DCMAKE_CXX_FLAGS="%CXX_FLAGS%" -DCMAKE_EXE_LINKER_FLAGS="%LINKER_FLAGS%" -DCMAKE_IGNORE_PATH="C:/Program Files/Git/usr/bin" -DJSON_BuildTests=On)
+  - if "%GENERATOR%"=="Ninja" (cmake . -G "%GENERATOR%" -DCMAKE_BUILD_TYPE="%configuration%" -DCMAKE_CXX_FLAGS="%CXX_FLAGS%" -DCMAKE_EXE_LINKER_FLAGS="%LINKER_FLAGS%" -DCMAKE_IGNORE_PATH="C:/Program Files/Git/usr/bin" -DJSON_BuildTests=On "%CMAKE_OPTIONS%") else (cmake . -G "%GENERATOR%" -A "%GENERATOR_PLATFORM%" -DCMAKE_CXX_FLAGS="%CXX_FLAGS%" -DCMAKE_EXE_LINKER_FLAGS="%LINKER_FLAGS%" -DCMAKE_IGNORE_PATH="C:/Program Files/Git/usr/bin" -DJSON_BuildTests=On "%CMAKE_OPTIONS%")
 
 build_script:
   - cmake --build . --config "%configuration%"
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -128,7 +128,7 @@ void from_json(const BasicJsonType& j, EnumType& e)
 
 // forward_list doesn't have an insert method
 template<typename BasicJsonType, typename T, typename Allocator,
-         enable_if_t<std::is_convertible<BasicJsonType, T>::value, int> = 0>
+         enable_if_t<is_getable<BasicJsonType, T>::value, int> = 0>
 void from_json(const BasicJsonType& j, std::forward_list<T, Allocator>& l)
 {
     if (JSON_HEDLEY_UNLIKELY(!j.is_array()))
@@ -145,7 +145,7 @@ void from_json(const BasicJsonType& j, std::forward_list<T, Allocator>& l)
 
 // valarray doesn't have an insert method
 template<typename BasicJsonType, typename T,
-         enable_if_t<std::is_convertible<BasicJsonType, T>::value, int> = 0>
+         enable_if_t<is_getable<BasicJsonType, T>::value, int> = 0>
 void from_json(const BasicJsonType& j, std::valarray<T>& l)
 {
     if (JSON_HEDLEY_UNLIKELY(!j.is_array()))
@@ -153,7 +153,11 @@ void from_json(const BasicJsonType& j, std::valarray<T>& l)
         JSON_THROW(type_error::create(302, "type must be array, but is " + std::string(j.type_name())));
     }
     l.resize(j.size());
-    std::copy(j.begin(), j.end(), std::begin(l));
+    std::transform(j.begin(), j.end(), std::begin(l),
+                   [](const BasicJsonType & elem)
+    {
+        return elem.template get<T>();
+    });
 }
 
 template<typename BasicJsonType, typename T, std::size_t N>
--- a/include/nlohmann/detail/macro_scope.hpp
+++ b/include/nlohmann/detail/macro_scope.hpp
@@ -284,3 +284,13 @@
 #define NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(Type, ...)  \
     void to_json(nlohmann::json& j, const Type& t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__)) } \
     void from_json(const nlohmann::json& j, Type& t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_FROM, __VA_ARGS__)) }
+
+#ifndef JSON_USE_IMPLICIT_CONVERSIONS
+    #define JSON_USE_IMPLICIT_CONVERSIONS 1
+#endif
+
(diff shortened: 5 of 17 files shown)


theodelrieu:
Where is the LGTM config file?

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -153,3 +153,14 @@
     basic_json<ObjectType, ArrayType, StringType, BooleanType,             \
     NumberIntegerType, NumberUnsignedType, NumberFloatType,                \
     AllocatorType, JSONSerializer>
+
+#ifndef JSON_USE_EXPLICIT_CONVERSIONS
+    #define JSON_EXPLICIT
+    #ifndef JSON_USE_IMPLICIT_CONVERSIONS
+        #warning Implicit conversions are being deprecated and will be \
```
gracicot: I realized that the `#warning` statement is not standard, and not supported on msvc. There is some solution like `#pragma message()` but it's not a real warning.

I think we could use `[[deprecated("...")]]` or the msvc non standard `__declspec(deprecated("..."))` for earlier versions in that case.

theodelrieu: I've added JSON_DEPRECATED_MSG, and tried building the tests by removing the `Wno-deprecated` warnings, it works \o/

gracicot: Thank you, the patch is great now. I would just wait for @nlohmann to do a review too and merge it!

nlohmann: I like the idea of having a switch for the conversions, but I am still not convinced that implicit conversions should be deprecated.

nlohmann:
> Where is the LGTM config file?

There is no config file. Here is a link to the log: https://lgtm.com/projects/g/nlohmann/json/logs/rev/pr-e7a2ac9be19e57d05404374d62bf6a8ab3794bfe/lang:cpp/stage:Build%20merge_d94187b754143b1064d1f7bcd3000e3db0419555. Strange that the test fails.

include/nlohmann/detail/macro_scope.hpp
theodelrieu: The main reason is because it is broken today. The pile of issues related to #958 is a good indicator of that. Not to mention the **why** is quite hard to grasp.

The required changes are quite minimal as well:

```cpp
std::string s;
s = j;
// becomes
std::string s;
j.get_to(s);

std::string s = j;
// becomes
std::string s(j);
```

theodelrieu:
But where do you define how to build the library?

include/nlohmann/detail/macro_scope.hpp
gracicot: After thinking about it I think @nlohmann is right on not deprecate *right now* (I still think it should eventually). The compile switch must be there for a while before any deprecation, and the documentation should also rely on explicit conversion and it should recommend enabling the switch.

There is non-trivial cases that must be refactored when changing from implicit to explicit. Take this for example:
```c++
struct special_type {
    int a, b;
    std::string c, d;
};

special_type{value1, value2, value3, value4}; // works with implicit only
```

I really don't think a lot of users are relying on this, but it's still an annoyance. If we can recommend using the compiler switch for a while, warn users that the implicit operator will be deprecated in the future then the transition will be smooth.

The only way to recreate the behaviour above is to create a thin wrapper that allows casts:

```c++
struct implicit {
    json::value value;
    template<typename T, typename = decltype(std::declval<json::value>().get<T>())>
    operator T () {
        return value.get<T>();
    }
};

special_type{implicit{value1}, implicit{value2}, implicit{value3}, implicit{value4}};
```

I would like to see the implicit operator be deprecated too but I also agree that even though the change in the syntax is minimal, its use is widespread so it must be done very gradually.

theodelrieu:
Ok, the LGTM build uses default CMake arguments, therefore it builds with the implicit conversions.
All the other builds use explicit conversions.

@nlohmann do you wish to do like with `JSON_MultipleHeaders`? i.e. define a variable in the CI files.

theodelrieu:
It would be nice to merge this PR and discuss what is the long-term plan after.

theodelrieu:
@nlohmann Shall we merge this PR now that 3.7.0 has been released?

gracicot:
Since tackling the conversion operator is being considered for version 4.0.0, I'd vote to add the deprecation back when `JSON_ExplicitConversions` is off.

If we want to use explicit conversion on the next major version, deprecation and the switch is the first step to transition the community.

gregmarr:
@theodelrieu All of the whitespace and line wrapping changes in the first commit makes it really hard to see what's changed, even with the "ignore whitespace" option.  They also broke the formatting of some of the tables in the comments.  https://github.com/nlohmann/json/pull/1559/files#diff-641eb804df1899b4e8277af5a52e6b7dR335

theodelrieu:
@gregmarr damnit, I'll run amalgamate in a first commit

CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ endif ()
 option(JSON_BuildTests "Build the unit tests when BUILD_TESTING is enabled." ON)
 option(JSON_Install "Install CMake targets during install step." ON)
 option(JSON_MultipleHeaders "Use non-amalgamated version of the library." OFF)
+option(JSON_ExplicitConversions "Enable explicit conversion operator" OFF)
```
nlohmann, line 27: I'm not sure about the wording. Isn't the flag more about *disabling* implicit conversions? At least this would be the difference between the current code and when the flag is set.

.travis.yml
```diff
@@ -318,7 +318,7 @@ script:
 
   # compile and execute unit tests
   - mkdir -p build && cd build
-  - cmake .. ${CMAKE_OPTIONS} -DJSON_MultipleHeaders=${MULTIPLE_HEADERS} -DJSON_BuildTests=On -GNinja && cmake --build . --config Release
+  - cmake .. ${CMAKE_OPTIONS} -DJSON_ExplicitConversions=ON -DJSON_MultipleHeaders=${MULTIPLE_HEADERS} -DJSON_BuildTests=ON -GNinja && cmake --build . --config Release
```
nlohmann, line 321: Why do you set the flag for all Travis tests? Wouldn't it be better to have it set for one test - just like what we do with disabled exceptions?

Makefile
```diff
@@ -71,7 +71,7 @@ check:
 coverage:
 	rm -fr build_coverage
 	mkdir build_coverage
-	cd build_coverage ; cmake .. -GNinja -DCMAKE_BUILD_TYPE=Debug -DJSON_Coverage=ON -DJSON_MultipleHeaders=ON
+	cd build_coverage ; cmake .. -GNinja -DCMAKE_BUILD_TYPE=Debug -DJSON_Coverage=ON -DJSON_MultipleHeaders=ON -DJSON_ExplicitConversions=ON
```
nlohmann, line 74: Same question - why to set this flag for coverage?

appveyor.yml
```diff
@@ -112,7 +112,7 @@ before_build:
...
   - ps: if ($env:name -Eq "with_win_header") { $header_path = "single_include\nlohmann\json.hpp" }
   - ps: if ($env:name -Eq "with_win_header") { "#include <Windows.h>`n" + (Get-Content $header_path | Out-String) | Set-Content $header_path }
-  - if "%GENERATOR%"=="Ninja" (cmake . -G "%GENERATOR%" -DCMAKE_BUILD_TYPE="%configuration%" -DCMAKE_CXX_FLAGS="%CXX_FLAGS%" -DCMAKE_EXE_LINKER_FLAGS="%LINKER_FLAGS%" -DCMAKE_IGNORE_PATH="C:/Program Files/Git/usr/bin" -DJSON_BuildTests=On) else (cmake . -G "%GENERATOR%" -A "%GENERATOR_PLATFORM%" -DCMAKE_CXX_FLAGS="%CXX_FLAGS%" -DCMAKE_EXE_LINKER_FLAGS="%LINKER_FLAGS%" -DCMAKE_IGNORE_PATH="C:/Program Files/Git/usr/bin" -DJSON_BuildTests=On)
+  - if "%GENERATOR%"=="Ninja" (cmake . -G "%GENERATOR%" -DCMAKE_BUILD_TYPE="%configuration%" -DCMAKE_CXX_FLAGS="%CXX_FLAGS%" -DCMAKE_EXE_LINKER_FLAGS="%LINKER_FLAGS%" -DCMAKE_IGNORE_PATH="C:/Program Files/Git/usr/bin" -DJSON_BuildTests=On -DJSON_ExplicitConversions=On) else (cmake . -G "%GENERATOR%" -A "%GENERATOR_PLATFORM%" -DCMAKE_CXX_FLAGS="%CXX_FLAGS%" -DCMAKE_EXE_LINKER_FLAGS="%LINKER_FLAGS%" -DCMAKE_IGNORE_PATH="C:/Program Files/Git/usr/bin" -DJSON_BuildTests=On -DJSON_ExplicitConversions=On)
```
nlohmann, line 115: Same question as above - why setting the flag as default for the CIs?

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -144,16 +143,20 @@ void from_json(const BasicJsonType& j, std::forward_list<T, Allocator>& l)
...
+template<typename BasicJsonType, typename T>
+auto from_json(const BasicJsonType& j, std::valarray<T>& l) -> decltype(j.template get<T>(), void())
 {
     if (JSON_HEDLEY_UNLIKELY(!j.is_array()))
     {
         JSON_THROW(type_error::create(302, "type must be array, but is " + std::string(j.type_name())));
     }
     l.resize(j.size());
-    std::copy(j.begin(), j.end(), std::begin(l));
+    std::transform(
```
nlohmann, line 154: What's the benefit of using transform rather than copy?

include/nlohmann/json.hpp
```diff
@@ -319,12 +332,16 @@ class basic_json
...
     key         | description
     ----------- | ---------------
-    `compiler`  | Information on the used compiler. It is an object with the following keys: `c++` (the used C++ standard), `family` (the compiler family; possible values are `clang`, `icc`, `gcc`, `ilecpp`, `msvc`, `pgcpp`, `sunpro`, and `unknown`), and `version` (the compiler version).
-    `copyright` | The copyright line for the library as string.
-    `name`      | The name of the library as string.
-    `platform`  | The used platform as string. Possible values are `win32`, `linux`, `apple`, `unix`, and `unknown`.
-    `url`       | The URL of the project as string.
-    `version`   | The version of the library. It is an object with the following keys: `major`, `minor`, and `patch` as defined by [Semantic Versioning](http://semver.org), and `string` (the version string).
+    `compiler`  | Information on the used compiler. It is an object with the
+    following keys: `c++` (the used C++ standard), `family` (the compiler family;
```
nlohmann, line 336: This looks broken.

include/nlohmann/json.hpp
```diff
@@ -1049,7 +1069,9 @@ class basic_json
                     object = nullptr;  // silence warning, see #821
                     if (JSON_HEDLEY_UNLIKELY(t == value_t::null))
                     {
-                        JSON_THROW(other_error::create(500, "961c151d2e87f2686a955a9be24d316f1362bf21 3.8.0")); // LCOV_EXCL_LINE
+                        JSON_THROW(other_error::create(500,
+                                                       "961c151d2e87f2686a955a9be24d316f136"
```
nlohmann, line 1073: Please revert this change - the hash should be in one literal.

include/nlohmann/json.hpp
```diff
@@ -1262,14 +1286,21 @@ class basic_json
...
     ------------------ | ----------- | ------------------ | -------------------
-    parse_event_t::object_start | the parser read `{` and started to process a JSON object | depth of the parent of the JSON object | a JSON value with type discarded
-    parse_event_t::key | the parser read a key of a value in an object | depth of the currently parsed JSON object | a JSON string containing the key
-    parse_event_t::object_end | the parser read `}` and finished processing a JSON object | depth of the parent of the JSON object | the parsed JSON object
-    parse_event_t::array_start | the parser read `[` and started to process a JSON array | depth of the parent of the JSON array | a JSON value with type discarded
-    parse_event_t::array_end | the parser read `]` and finished processing a JSON array | depth of the parent of the JSON array | the parsed JSON array
-    parse_event_t::value | the parser finished reading a JSON value | depth of the value | the parsed JSON value
-
-    @image html callback_events.png "Example when certain parse events are triggered"
+    parse_event_t::object_start | the parser read `{` and started to process a
```
nlohmann, line 1289: Please revert the whitespace changes in comments.

include/nlohmann/json.hpp
```diff
@@ -1336,8 +1367,7 @@ class basic_json
 
     @since version 1.0.0
     */
-    basic_json(const value_t v)
-        : m_type(v), m_value(v)
+    basic_json(const value_t v) : m_type(v), m_value(v)
```
nlohmann, line 1370: Style: please revert this change.

include/nlohmann/json.hpp
```diff
@@ -1360,8 +1390,7 @@ class basic_json
 
     @since version 1.0.0
     */
-    basic_json(std::nullptr_t = nullptr) noexcept
-        : basic_json(value_t::null)
+    basic_json(std::nullptr_t = nullptr) noexcept : basic_json(value_t::null)
```
nlohmann, line 1393: Style: please revert this change.

include/nlohmann/json.hpp
```diff
@@ -3215,18 +3306,25 @@ class basic_json
...
+     (defined(_MSC_VER) && _MSC_VER >= 1910 && _MSC_VER <= 1914))
+            && !std::is_same<ValueType, typename std::string_view>::value
 #endif
-                   && detail::is_detected<detail::get_template_function, const basic_json_t&, ValueType>::value
-                   , int >::type = 0 >
-    operator ValueType() const
+            && detail::is_detected<detail::get_template_function,
+                                   const basic_json_t&, ValueType>::value,
+            int >::type = 0 >
+    JSON_EXPLICIT operator ValueType() const
```
nlohmann, line 3327: As this is the only place where `JSON_EXPLICIT` is actually used, please revert all the whitespace/formatting changes to have a better overview what actually changed.

include/nlohmann/json.hpp
```diff
@@ -3793,20 +3898,23 @@ class basic_json
             const auto it = find(key);
             if (it != end())
             {
-                return *it;
+                return it->template get<ValueType>();
```
nlohmann, line 3901: Why this change?

test/src/unit-udt.cpp
```diff
@@ -311,30 +311,6 @@ TEST_CASE("basic usage" * doctest::test_suite("udt"))
             person_json["name"].get_to(name).m_val = "new name";
             CHECK(name.m_val == "new name");
         }
-
```
nlohmann, line 314: I do not think this test should be removed. Can't we detect the flag?

CMakeLists.txt
gregmarr: The technical side is making the conversion operator explicit.  If you're looking at the option from the POV of how it changes behavior, then yes it is disabling implicit conversions, but if you are looking at the effect the option has on the code, then making the conversion operator explicit makes sense.  (This assumes that this is all that this option does.  If it's doing more than that, then it may make sense to word it the other way.)

include/nlohmann/detail/conversions/from_json.hpp
gregmarr: copy requires implicit conversions

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -144,16 +143,20 @@ void from_json(const BasicJsonType& j, std::forward_list<T, Allocator>& l)
...
+        j.begin(), j.end(), std::begin(l),
```
gregmarr, line 155: Can we use `std::begin` and `std::end` on `j` as well for consistency?

include/nlohmann/json.hpp
gregmarr: This code was using the implicit conversion operator.

CMakeLists.txt
gregmarr: The next question is what happens to this option in the next major version where we can make this the default?  Does it become "Enable implicit conversion operator"?

nlohmann: Hm. To me, an explicit conversion seems less a feature than an implicit one. So I would rather expect a switch controlling whether or not I would like to have implicit conversions. So for the time being,

```cmake
option(JSON_ImplicitConversions "Enable implicit conversion operators" ON)
```

theodelrieu: @nlohmann The current behavior is broken, I wouldn't reason in terms of which one is more a feature, but rather as a new behavior

test/src/unit-udt.cpp
theodelrieu: It's been a while since I've worked on this PR, I'll try to make both cohabit. That's also why I've put `JSON_ExplicitConversions=ON` everywhere.

include/nlohmann/detail/conversions/from_json.hpp
theodelrieu: I don't see the need to do that, since `j` is of type `json`, thus we know its member functions (c.f. `is_array` used a few lines above).

Also, I think this change is outside the scope of this PR.

test/src/unit-udt.cpp
theodelrieu: One annoying thing is that code using implicit conversions will not build when explicit conversions are enabled, I'd like to find a way to avoid sprinkling `ifdef` everywhere in the tests.

theodelrieu:
Should be good (bis)

theodelrieu:
@nlohmann @gregmarr Some changes in the tests are necessary to make the code compile/work on both implicit/explicit conversions builds.

Containers having a `std::initializer_list` cannot be constructed with braces (i.e. `{json}`), because it will fail in implicit conversion builds (but will work in explicit conversion builds).

CMakeLists.txt
gregmarr: Ah, I wasn't thinking of the possibility of having the "use CMake" option being different than the "just include the header" option.  That would make it so that we don't have to change the name of the CMake option in the next major version.

include/nlohmann/detail/conversions/from_json.hpp
gregmarr: It just seems jarring to mix member and non-member in the same function call.  If we weren't already changing this line, then I wouldn't bring it up.  In any case, not something for which I would hold up the PR.

include/nlohmann/json.hpp
```diff
@@ -3793,7 +3794,7 @@ class basic_json
             const auto it = find(key);
             if (it != end())
             {
-                return *it;
+                return static_cast<ValueType>(*it);
```
gregmarr, line 3797: Curious if there is a technical reason that we use `static_cast<ValueType>` here but `get<ValueType>` in other places.

include/nlohmann/json.hpp
```diff
@@ -8333,8 +8334,8 @@ class basic_json
             }
 
             // collect mandatory members
-            const std::string op = get_value("op", "op", true);
-            const std::string path = get_value(op, "path", true);
+            const std::string op(get_value("op", "op", true));
```
gregmarr, line 8337: Is there a technical reason for this change (3 lines)?

theodelrieu: The previous makes use of implicit conversions, which does not build when using the `explicit` operator.

test/src/unit-cbor.cpp
```diff
@@ -1011,21 +1011,21 @@ TEST_CASE("CBOR")
                     SECTION("0 (0 00000 0000000000)")
                     {
                         json j = json::from_cbor(std::vector<uint8_t>({0xf9, 0x00, 0x00}));
-                        json::number_float_t d = j;
+                        json::number_float_t d{j};
```
gregmarr, line 1014: Is there a technical reason for this change, and if so, is this something that users of the library will have to do as well if they enable explicit conversions?

gregmarr: I assume this is the same answer as the other one.  In that case, we should probably add something to the docs/README about the error that can happen and how to resolve it.

include/nlohmann/json.hpp
gregmarr: Ah, fun with different initialization syntaxes.

include/nlohmann/json.hpp
theodelrieu: good catch

test/src/unit-cbor.cpp
theodelrieu: The previous line simply does not compile when using explicit conversions. There are several ways to fix it, the easiest is the one I wrote, but you can be really explicit and use `get`, or even `static_cast`.

theodelrieu:
Ok, those macOS fails are due to libcxx 5.0.2 `std::string` hidden constructor, taking a `__self_view` (i.e. `std::string_view`, even when not using C++17).

Thus testing if a json object is explicitly convertible to a `std::string` will fail to compile. I removed the `is_explicitly_convertible_to` trait, and replaced it with a `is_getable` trait.

This trait checks that `j.get<T>()` does not yield a compiler error. It should be used as a SFINAE condition wherever `get` is used (instead of `std::is_convertible`, or `has_from_json`).

theodelrieu:
@nlohmann In the last run, the single fail is due to a GitHub outage, can you rerun the failed job? Or do I have to trigger a build?

nlohmann:
Done.

theodelrieu:
@nlohmann Seems good finally :pray:

test/src/unit-conversions.cpp
```diff
@@ -734,235 +542,430 @@ TEST_CASE("value conversion")
...
+        SECTION("bool")
+        {
+            bool b = j.get<bool>();
+            CHECK(json(b) == j);
         }
 
         SECTION("exception in case of a non-number type")
         {
-            CHECK_THROWS_AS(json(json::value_t::null).get<json::number_integer_t>(),
+            CHECK_THROWS_AS(json(json::value_t::null).get<json::boolean_t>(),
```
nlohmann, line 922: Why these changes to from `number_integer_t` to `boolean_t`?

include/nlohmann/detail/meta/type_traits.hpp
```diff
@@ -94,6 +94,12 @@ using get_template_function = decltype(std::declval<T>().template get<U>());
 template<typename BasicJsonType, typename T, typename = void>
 struct has_from_json : std::false_type {};
 
+template <typename BasicJsonType, typename T>
```
nlohmann, line 97: Could you please add the rationale why this function is needed rather than `std::is_constructible`?

test/src/unit-conversions.cpp
```diff
@@ -189,48 +189,6 @@ TEST_CASE("value conversion")
         }
     }
 
-    SECTION("get an object (implicit)")
```
nlohmann, line 192: Maybe we should add more `#ifdef` blocks rather than reordering the file?

CMakeLists.txt
nlohmann: I'm still struggling with the wording. To me, the switch "enable explicit conversion" sounds like an additional feature. Instead, it breaks existing code, because the assumed implicit conversions do not work any longer. If the code already only used explicit conversions, then flipping the switch does not change anything.

test/src/unit-conversions.cpp
theodelrieu: What is the benefit?

include/nlohmann/detail/meta/type_traits.hpp
theodelrieu: I've added the rationale [here](https://github.com/nlohmann/json/pull/1559#issuecomment-661111824)

test/src/unit-conversions.cpp
nlohmann: The benefit is to maintain the order of the tests.

test/src/unit-conversions.cpp
theodelrieu: Oops, must be a leftover from last year

theodelrieu:
@nlohmann About your last comment: I can swap the CMake option definition

include/nlohmann/detail/meta/type_traits.hpp
theodelrieu: Do you want me to add it in the code?

nlohmann: Yes, please. Such things are so easy forgotten...

theodelrieu:
@nlohmann GitHub was unreachable again, sigh. Can you restart the job please? :pray: (https://travis-ci.org/github/nlohmann/json/jobs/710669524#L5359)

theodelrieu:
@nlohmann It seems appveyor is stuck as well 🤔

nlohmann:
No, it's just slow...

CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ endif ()
 option(JSON_BuildTests "Build the unit tests when BUILD_TESTING is enabled." ON)
 option(JSON_Install "Install CMake targets during install step." ON)
 option(JSON_MultipleHeaders "Use non-amalgamated version of the library." OFF)
+option(JSON_ImplicitConversions "Enable implicit conversions" ON)
```
aokellermann, line 27: Maybe put period at end of description for consistency's sake.

gracicot:
I wonder if this can be removed to re-enable conversions to `std::string_view` once we switch to explicit conversions:
```
    && !std::is_same<ValueType, typename std::string_view>::value
```
I think it was disabled because of a problem with string views and visual studio, but I'm not sure. If someone can find the issue/pull request, that would be awesome.

nlohmann:
@gracicot Was this changed in this PR?

gracicot:
> @gracicot Was this changed in this PR?

No.

nlohmann:
> No. The pull request mentioned that this hasn't been done for`std::initializer_list` but I was wondering if `std::string_view` was worth to do. Right now `json_value.get<std: string_view>()` works, but not the conversion and `is_gettable` returns false for it.

I see. Let's handle this in a follow-up issue. This PR has been open for too long...

nlohmann:
Thank you all so much!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [x] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Add option to switch off implicit conversions. #958 #1559

* * *

theodelrieu:
@gracicot I think all these type of explicit SFINAE disabling can be removed when using explicit conversions. IIRC there is one for `char` as well somewhere.

gracicot:
Thank you so much! 🎉

tlemo:
Thanks @theodelrieu for driving this!

Issue #1536, opened 2019-03-21
Releases 3.6.0 and 3.6.1 don't build on conda / windows


SylvainCorlay:
- What is the issue you have?

I am the maintainer of the nlohman_json package on conda-forge. The package for the latest releases don't build on windows (VS 2015)

- Please describe the steps to reproduce the issue. Can you provide a small but working code example?

   The logs for the failing builds are available

   - here: (version 3.6.1) https://github.com/conda-forge/nlohmann_json-feedstock/pull/7
   - here: (version 3.6.0) https://github.com/conda-forge/nlohmann_json-feedstock/pull/6

- Which compiler and operating system are you using? Is it a [supported compiler](https://github.com/nlohmann/json#supported-compilers)?

MSVC / Windows / (VS 2015)

- Did you use a released version of the library or the version from the `develop` branch?

Released version.

- If you experience a compilation error: can you [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests)?

The compilation error occurs while building the tests, as per the logs above.

nlohmann:
That is odd, because the MSVC 2015 build runs successfully at our AppVeyor, see https://ci.appveyor.com/project/nlohmann/json/builds/23234117.

SylvainCorlay:
Note that this only occurs with 3.6.x. Previous versions of `nlohmann_json` up until 3.5.x have been packaged succesfully on conda-forge.

nlohmann:
The compilation error is in file `unit-allocator.cpp`. The changes there toward 3.6.0 came with #1492. I'm no expert in MSVC. Maybe @ stac47 or @theodelrieu have an idea on this, as they also proposed the fixes for GCC 9.

theodelrieu:
It would be great to look at the `xmemory` file to see what the call is. But my guess is that `my_allocator` only defines some requirements of [`Allocator`](https://en.cppreference.com/w/cpp/named_req/Allocator), it seems to miss constructors at least.

Maybe `using std::allocator<T>::allocator` will be enough. If not, adding the constructors should do the trick.

I would suggest removing the inheritance from `std::allocator` and rewrite it from scratch though.

nlohmann:
@theodelrieu Good point, I'll try the `using std::allocator<T>::allocator`.

nlohmann:
@SylvainCorlay Could you try whether baaa2a4d0ffe210b32405e76fc007c46ecdc33ff builds?

mistersandman:
[...]

Issue #1531, opened 2019-03-20
Failed to build with <Windows.h>


ilgiz-badamshin:
Related to #1514 

- What is the issue you have?
Failed to build if <Windows.h> is included:
> C4003: not enough arguments for function-like macro invocation 'max'

Code: 
`assert(p2 <= std::numeric_limits<std::uint64_t>::max() / 10);`
should be: 
`assert(p2 <= (std::numeric_limits<std::uint64_t>::max)() / 10);`

nlohmann:
Thanks for reporting! Could you please check if 9d6ab9014f59e345701a3906d1defc80c4506d10 fixes the issue?

ilgiz-badamshin:
> Build succeeded.

Thanks.

nlohmann:
* * *

### 🔖 Release item

This issue/PR will be part of the next release of the library. This template helps preparing the release notes.

#### Type

- [ ] ✨ New Feature
- [x] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Fixed a compilation error when `<Windows.h>` was included. #1531

* * *

dvj:
If you want to try to avoid this type of regression in the future, a test that includes `windows.h` (a la https://github.com/CLIUtils/CLI11/pull/145/files#diff-da661fc0dc16940197d2112712cbf2bb) could be added as well!

PR #1570, opened 2019-04-17
Regression tests for MSVC

### LINKED ISSUE #1543: Add regression tests for MSVC
We recently had two regressions in MSVC:

- In #1531, we forgot to put parentheses around a `min` call. The problem is that we don't have a unit test that includes `Windows.h`.
- In #1536, the test suite did not build in Debug mode. The problem is that all tests in AppVeyor seem to be compiled in Release mode.

It would be great if someone with experience with MSVC could add regression tests for these issues.

### LINKED ISSUE #1608: is_nothrow_copy_constructible fails for json::const_iterator on MSVC2015 x86 Debug build
- What is the issue you have?

The following check in `unit-concepts` fails on MSVC 2015 x86 Debug mode ([log](https://ci.appveyor.com/project/nlohmann/json/builds/24717829/job/79u9kecc4wxyf928#L1495)):

```
CHECK(std::is_nothrow_copy_constructible<json::const_iterator>::value)
```

By borrowing [build flags](https://ci.appveyor.com/project/nlohmann/json/builds/24717829/job/79u9kecc4wxyf928#L426), I have reproduced it here: https://godbolt.org/z/MOkndK

This unit test should pass.

The unit-test fails merely because of x86 debug build on VS2015.

- Which compiler and operating system are you using? 

This only fails on MSVC 2015 (msvc 19.0), x86 target in debug mode.

**Speculation**

This bug slipped under the radar since our Debug mode for x86 was incorrectly configured and was actually building in Release mode https://github.com/nlohmann/json/issues/1543. This issue is a blocker for PR https://github.com/nlohmann/json/pull/1570 which aims to bring back the Debug build for x86 MSVC 2015.

By playing with build flags in Godbolt, it seems that the presence of [`/MDd`](https://docs.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=vs-2015) flag causes this error. This could be similar to #1536 where the C++ compiler on VS2015 added some definitions and symbols to binary in debug mode. This error only happens on VS2015 (msvc 19.0) Debug build (where `/MDd` is present) and doesn't happen in Release build and/or later compiler versions.

nickaein:
Fixes #1543 (and avoid regression to issues like #1531 and #1536).
Fixes #1608.

This pull request:

1. Adds Debug builds on Appveyor for VS2015 and VS2017.

2. Add a separate build job that first injects `#include <Windows.h>` line to the top of `json.hpp` header. This build job helps to detect whether the presence of `Windows.h` could have any undersiable side-effects.

**On point 2:** Currently the unit-test for case 2 doesn't do much and only can catch compile errors. Also, I tested it on my `json.hpp` with some unfixed `numeric_limits<>::max()` and `numeric_limits<>min()` functions and it only caused compiler errors on Debug builds. Probably on Release builds, the compiler eliminates the codes related to unfixed macro definitions. Therefore, the unit-test only can catch compiler *errors* on Debug mode.

A more thorough approach is that instead of adding a unit-test with `#include <Windows.h>` on top, we add a separate build configuration in which the `#include <Windows.h>` gets  injected to the top of `json.hpp`. In this build, all codes and unit-tests will be complied and checked in presence of `Windows.h`.

I have implemented and tested the second approach on my fork but and wanted to double check and get some feedback before adding it to the PR.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,6 +25,7 @@ environment:
       GENERATOR: Ninja
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+      configuration: Release
       COMPILER: mingw
       platform: x86
       CXX_FLAGS: ""
@@ -32,12 +33,14 @@ environment:
       GENERATOR: Ninja
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+      configuration: Release
       platform: x86
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
       GENERATOR: Visual Studio 14 2015
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+      configuration: Release
       platform: x86
       name: with_win_header
       CXX_FLAGS: ""
@@ -45,30 +48,35 @@ environment:
       GENERATOR: Visual Studio 14 2015
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+      configuration: Release
       platform: x86
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
       GENERATOR: Visual Studio 15 2017
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+      configuration: Release
       platform: x86
       CXX_FLAGS: "/permissive- /std:c++latest /utf-8"
       LINKER_FLAGS: ""
       GENERATOR: Visual Studio 15 2017
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+      configuration: Release
       platform: x64
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
       GENERATOR: Visual Studio 14 2015 Win64
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+      configuration: Release
       platform: x64
       CXX_FLAGS: ""
       LINKER_FLAGS: ""
       GENERATOR: Visual Studio 15 2017 Win64
 
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+      configuration: Release
       platform: x64
       CXX_FLAGS: "/permissive- /std:c++latest /utf-8 /F4000000"
       LINKER_FLAGS: "/STACK:4000000"
@@ -77,8 +85,6 @@ environment:
 init:
   - cmake --version
   - msbuild /version
-  # set the default build configuration to Release
-  - if "%configuration%"=="" (set configuration="Release")
 
 install:
   - if "%COMPILER%"=="mingw" appveyor DownloadFile https://github.com/ninja-build/ninja/releases/download/v1.6.0/ninja-win.zip -FileName ninja.zip
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -37,6 +37,13 @@ environment:
       LINKER_FLAGS: ""
       GENERATOR: Visual Studio 14 2015
 
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+      platform: x86
+      name: with_win_header
+      CXX_FLAGS: ""
+      LINKER_FLAGS: ""
+      GENERATOR: Visual Studio 14 2015
+
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
       platform: x86
       CXX_FLAGS: ""
@@ -81,6 +88,9 @@ install:
   - if "%COMPILER%"=="mingw" g++ --version
 
 before_build:
+  # for with_win_header build, inject the inclusion of Windows.h to the single-header library
+  - ps: if ($env:name -Eq "with_win_header") { $header_path = "single_include\nlohmann\json.hpp" }
+  - ps: if ($env:name -Eq "with_win_header") { "#include <Windows.h>`n" + (Get-Content $header_path | Out-String) | Set-Content $header_path }
   - cmake . -G "%GENERATOR%" -DCMAKE_BUILD_TYPE="%configuration%" -DCMAKE_CXX_FLAGS="%CXX_FLAGS%" -DCMAKE_EXE_LINKER_FLAGS="%LINKER_FLAGS%" -DCMAKE_IGNORE_PATH="C:/Program Files/Git/usr/bin"
 
 build_script:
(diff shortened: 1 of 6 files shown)


nlohmann:
The AppVeyor tests seem to be broken. Any idea on this?

nickaein:
There were two errors:

1. The `test-unicode_all` was timing out due to slower binary in Debug mode. I've pushed a fix to increase the timeout for this particular unit-test on Debug build.

2. The second error, which is still present, is `unit-concepts` failing on this check:

```
CHECK(std::is_nothrow_copy_constructible<json::const_iterator>::value)
```

By borrowing [build flags for VS2015-Debug](https://ci.appveyor.com/project/nlohmann/json/builds/24006441/job/njal7ju4syjt4f50?fullLog=true#L424) it can be reproduced: https://godbolt.org/z/MOkndK

It seems that [`/MDd`](https://docs.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=vs-2015) flag causes this error. This could be similar to #1536 where the C++ compiler on VS2015 added some definitions in debug mode. This error only happens on VS2015 (msvc 19.0) Debug build (where `/MDd` is present) and doesn't happen in Release build and/or later compiler versions.

That's how far I was be able to get. Maybe @mistersandman or @theodelrieu could shed some light on what is causing `json::const_iterator` to fail `is_nothrow_copy_constructible` criteria when we are building on debug mode (msvc 19.0).

nickaein:
Also, do you have any suggestion on the choices in https://github.com/nlohmann/json/pull/1570#issue-271436407 to have a build with `Windows.h`?

appveyor.yml
```diff
@@ -48,6 +77,8 @@ environment:
 init:
   - cmake --version
   - msbuild /version
+  # set the default build configuration to Release
+  - if "%configuration%"=="" (set configuration="Release")
```
nlohmann: Shouldn't we make the CMake build type explicit in the build matrix above? Just by browsing through the code, I saw the `Debug` entries and asked myself what build type is used if they are not explicitly mentioned.

test/CMakeLists.txt
```diff
@@ -133,6 +133,12 @@ foreach(file ${files})
     )
     set_tests_properties("${testcase}_all" PROPERTIES LABELS "all")
 
+    # Increase timeout for test-unicode_all on Debug build
```
nlohmann: As this only affects a certain configuration at AppVeyor, I would propose to leave the CMakeLists.txt unchanged, but add a parameter to the `ctest` call like `ctest --timeout 3600`.

nickaein: Sure! It just increases the timeout for all unit-tests to 1 hour. Is it OK?

nlohmann: Yes - so far, all but the Unicode test are usually fast. (And since this affects AppVeyor only, I do not care the timeout is ever reached.)

appveyor.yml
nickaein: Done.

test/CMakeLists.txt
nickaein: Done.

**Sidenote:** As `Debug` build for MSVC is now working, it takes a long time for `test-unicode_all` to finish (probably since debug binaries are much slower on MSVC and also AppVeyor _crawls_ to run). I set it to ~30min intially, but even that wasn't enough. Now it is set to 1 hour.

nickaein:
@nlohmann: Hey Niels! No problem at all. This was blocked until yesterway anyway.

The build seems to be passing. While reviewing, you might take a closer look at  https://github.com/nlohmann/json/commit/4c23af159b99c592b6d54d64ff79b5543f39c4c0 which is intended to define copy-constructor for `const_iterator` explicitly. Is it OK?

include/nlohmann/detail/iterators/iter_impl.hpp
```diff
@@ -118,6 +118,13 @@ class iter_impl
           to iterator is not defined.
     */
 
+    /* We had to explicitly define the copy constructor for the const
+       case to circumvent a bug on msvc 2015 debug build.
+       More info: https://github.com/nlohmann/json/issues/1608
+     */
+    iter_impl(const iter_impl<const BasicJsonType>& other) noexcept
```
nlohmann: Could you please add some documentation like from the constructor below and add your note there as `@note` annotation?

nickaein: Sure! Done.

Let me know if you had any other suggestions on the wording, etc.

nlohmann:
Now the latest CI build at AppVeyor is failing again... Either we set up the timeout or we only use the small test suite in the Debug build. What do you think?

nickaein:
In [the last succesful CI build](https://ci.appveyor.com/project/nlohmann/json/builds/25468760), MSVC debug builds has taken ~1 hour. It seems we are almost exceeding the [1 hour time restriction](https://www.appveyor.com/pricing/) of AppVeyor.

As you suggested, we can skip time-consuimg test(s) for MSVC debug, at least `test-unicode --all` which seems to be the worst offender. Since they still will be run in other build jobs (including MSVC 2015/2017 Release mode), I think this would be a good compromise between keeping the integerity of test and speeding up (and fixing!) the CI.

test/CMakeLists.txt
```diff
@@ -127,11 +127,18 @@ foreach(file ${files})
     )
     set_tests_properties("${testcase}_default" PROPERTIES LABELS "default")
 
-    add_test(NAME "${testcase}_all"
-      COMMAND ${testcase} ${DOCTEST_TEST_FILTER} --no-skip
-      WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
-    )
-    set_tests_properties("${testcase}_all" PROPERTIES LABELS "all")
+    # Skip test-unicode_all on MSVC Debug build
```
nlohmann: Let's not do this in the CMake file, but rather in the AppVeyor YAML using the ctest command line options.

nickaein: Totally agree and done.

Sorry about that. I wasn't aware of `--exclude-regex` option for `ctest`. Hope this sorts it out.

nlohmann:
Now I just wait until the CI jobs run through and we're set. Thanks!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [x] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Add regression tests for MSVC.
- Fix MSVC Debug build.

* * *

nickaein:
Thanks!

nlohmann:
Thank you!

Issue #916, opened 2018-01-15
Feature: to_string(const json& j);


DELTA37:
It is very helpful to add method to_string(const json& j);

theodelrieu:
Agreed, would you like to open a PR? This one seems quite easy to implement :)

nlohmann:
What would be the expected result? `dump()`?

theodelrieu:
That's what I would expect, do you agree @DELTA37?

cronnosli:
I guess that .dump() is already enough, unless the intent is to replace .dump()!

theodelrieu:
The use-case I see would be something like the following:
 
```cpp
// in some generic code
template <typename T>
void doStuff(T const& value) {
    using std::to_string;
    
    auto const str = to_string(value);
    // process the string
}
```

Other than that, I agree that `dump()` is pretty much enough.

DELTA37:
Yes, i guess to add "to_string()" method which will do the same as dump()
It is a standard method to extend std::to_string for user-defined class
And it is very helpful to do it with json

That is a discussion https://stackoverflow.com/questions/33399594/making-a-user-defined-class-stdto-stringable

// Sorry for my English

DELTA37:
https://github.com/nlohmann/json/pull/918
Attach pill request

Macr0Nerd:
@nlohmann is this still available? It doesn't seem terribly important but it was marked as good first issue so :)

nlohmann:
@Macr0Nerd Yes, it is! There were open issues in PR #918, so this issue was never fixed.

Macr0Nerd:
Alright! So @nlohmann all I need to is read a function that takes the JSON key, and pulls a deserializes the data into a key, correct? I’m still relatively green to C++ but I’ll give this my best!

nlohmann:
No, a `json` value is given and `to_string` should return its serialization.

Macr0Nerd:
So essentially just do what .dump() does and return the entire given JSON object passed as a string? I apologize if I am asking to many questions, I just want to confirm that i correctly do it.

Macr0Nerd:
I think i see what you want (and based on the previous responses). It seems that you want to add an argument dependent lookup like [this comment](https://stackoverflow.com/a/33399769/5859623) from the SO post from above. As it wasn't requested, I wouldn't overload the `<<` operator but is this what you want @nlohmann?

[...]

PR #1585, opened 2019-04-26
Added to_string and added basic tests

Macr0Nerd:
I created the `to_string` function for the `nlohmann` namespace and JSON. It uses the template
```cpp
template <class T>
inline std::string to_string(const T& j){
    return j.dump();
}
```
to easily accept any JSON or basic_json data type. This is in reference to Issue #916 to fully add `to_string`. 

**Note**: As with all additives to the std namespace using ADL, it is required to declare `using std::to_string` otherwise the feature won't work.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -7965,6 +7965,21 @@ class basic_json
 
     /// @}
 };
+
+/*!
+@brief user-defined to_string function for JSON values
+
+This function implements a user-defined to_string  for JSON objects.
+
+@param[in] j  a JSON object
+@return a std::string object
+*/
+
+NLOHMANN_BASIC_JSON_TPL_DECLARATION
+std::string to_string(const NLOHMANN_BASIC_JSON_TPL& j)
+{
+    return j.dump();
+}
 } // namespace nlohmann
 
 ///////////////////////
--- a/test/src/unit-serialization.cpp
+++ b/test/src/unit-serialization.cpp
@@ -173,4 +173,19 @@ TEST_CASE("serialization")
             test("\xE1\x80\xE2\xF0\x91\x92\xF1\xBF\x41", "\\ufffd" "\\ufffd" "\\ufffd" "\\ufffd" "\x41");
         }
     }
+
+    SECTION("to_string")
+    {
+        auto test = [&](std::string const & input, std::string const & expected)
+        {
+            using std::to_string;
+            json j = input;
+            CHECK(to_string(j) == "\"" + expected + "\"");
+        };
+
+        test("{\"x\":5,\"y\":6}", "{\\\"x\\\":5,\\\"y\\\":6}");
+        test("{\"x\":[10,null,null,null]}", "{\\\"x\\\":[10,null,null,null]}");
+        test("test", "test");
+        test("[3,\"false\",false]", "[3,\\\"false\\\",false]");
+    }
 }
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -20766,6 +20766,21 @@ class basic_json
 
     /// @}
 };
+
+/*!
+@brief user-defined to_string function for JSON values
+
+This function implements a user-defined to_string  for JSON objects.
+
+@param[in] j  a JSON object
+@return a std::string object
+*/
+
+NLOHMANN_BASIC_JSON_TPL_DECLARATION
+std::string to_string(const NLOHMANN_BASIC_JSON_TPL& j)
+{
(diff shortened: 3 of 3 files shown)


include/nlohmann/json.hpp
```diff
@@ -7965,6 +7965,20 @@ class basic_json
...
+@brief user-defined to_string function for JSON values
+
+This function implements a user-defined to_string  for JSON objects.
+
+@param[in] j  a JSON object
+@return a std::string object
+*/
+
+template <class T>
+std::string to_string(const T& j){
```
theodelrieu: It'd be better to constrain `T` to be a `basic_json<...>`.

You can use the ugly macro defined in `macro_scope.hpp` for that matter.

test/src/unit-serialization.cpp
```diff
@@ -173,4 +173,19 @@ TEST_CASE("serialization")
             test("\xE1\x80\xE2\xF0\x91\x92\xF1\xBF\x41", "\\ufffd" "\\ufffd" "\\ufffd" "\\ufffd" "\x41");
         }
     }
+
+    SECTION("to_string")
+    {
+        auto test = [&](std::string const & input, std::string const & expected)
+        {
+            using std::to_string;
```
theodelrieu: You should not have to use that for that specific test. The `using std::thing` is only necessary in generic code.

Macr0Nerd:
@theodelrieu I’ll change it accordingly but I used the `std::to_string` because to the best of my knowledge `to_string` can’t be overwritten or specialized so that’s how you define the ADL. I could easily just use the `nlohmann::to_string` but I thought that we wanted this to allow input into `std::to_string`.

theodelrieu:
>to the best of my knowledge `to_string` can’t be overwritten or specialized so that’s how you define the ADL.

That's not exactly it. The details are explained in [this paper](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4381.html), but here's a short exlanation:

There is a flaw in the design of customization points. The wanted behavior would be to always call `std::to_string`, which would pick ADL customization points, if any, and then consider the overloads in `std`.

This will be fixed in C++20 with `ranges::begin`, `ranges::end`, etc... Note that `nlohmann::to_json` and `nlohmann::from_json` follow the same design.

Meanwhile, we must rely on using unqualified calls:

```cpp
// Pre C++20
template <typename T> void f(T const& container) {
  using std::begin;
  // If a free function begin exists in T's namespace, it will be picked up
  // If not, std::begin will be used as a fallback 
  auto it = begin(container);
}

// C++20
template <typename T> void f(T const& container) {
  // ranges::begin has the same behavior than the pre-C++20 mechanism.
  auto it = ranges::begin(container);
}
```

Anyway, the `using std::thing` idiom is only necessary in generic code, and it's up to users to write it.

Macr0Nerd:
Ah alright thank you. I’ll change the test statements, but I’m going to leave `to_string` as is (except for changing to the ugly macro).

theodelrieu:
Oh, I don't know why, but my brain made me believe you had put `using std::to_string` in the library's code... It's fine to let it in the tests :)

Macr0Nerd:
Ahh ok I understand! It’s all good, lord knows I would’ve actually put it in the code xD.

Macr0Nerd:
@theodelrieu how do I use the ugly macro. I’m a bit confused on that front. Once I get that implemented I’ll close this and make a new PR

theodelrieu:
No need to open a new PR, you can just rebase/squash your commits instead.

Here is an example taken from the code:

```cpp
NLOHMANN_BASIC_JSON_TPL_DECLARATION
struct is_basic_json<NLOHMANN_BASIC_JSON_TPL> : std::true_type {};
```

so basically you want:

```cpp
NLOHMANN_BASIC_JSON_TPL_DECLARATION
std::string to_string(const NLOHMANN_BASIC_JSON_TPL &j) { /* ... */ }
```

Macr0Nerd:
@theodelrieu I updated and squashed and rebased. I apologize for the very last commit, rebasing caused some conflicts. But they are all resolved and ready to go!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [x] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Added a function `to_string` to allow for generic conversion to strings. #916

#### Contributor entry

- [Gabe Ron](https://github.com/Macr0Nerd) implemented the `to_string` method.

* * *

PR #1606, opened 2019-05-22
add bazel build

horance-liu:
User can depends nlohmann json directly using bazel, but this commit don't support test all cases using bazel.


nlohmann:
Thanks for the PR! Could you also add a small paragraph to the README file how to use the package?

README.md
```diff
@@ -147,6 +147,84 @@ endif()
 
 `thirdparty/nlohmann_json` is then a complete copy of this source tree.
 
+### Bazel
```
nlohmann: Thanks for the extensive example! What I had in mind, though, was a paragraph like for the package managers in the "Package Managers" section below. Just enough to let people knowing Bazel understand what they need to do to use this library.

WORKSPACE
```diff
@@ -0,0 +1 @@
+workspace(name = "nlohmann_json")
```
nlohmann: Is this file really needed in this repository?

nlohmann:
@horance-liu Can you please have a look at the changes I requested?

BUILD.bazel
```diff
@@ -0,0 +1,13 @@
+licenses(["notice"])  # 3-Clause BSD
+
+exports_files(["LICENSE.MIT"])
+
+cc_library(
+    name = "json",
+    hdrs = glob([
+        "include/**/*.hpp",
```
remyabel: From the [docs](https://docs.bazel.build/versions/master/be/functions.html#glob): "This pattern should be avoided if possible, as it can reduce build incrementality and therefore increase build times." A better approach would be to generate the list of sources and headers from CMake (that way, you don't need to repeat all of it in Bazel), i.e: https://github.com/googleapis/google-cloud-cpp/blob/8e8dd3d83ec27b7a5176f9fe43cdd93aa5ace3dc/cmake/CreateBazelConfig.cmake#L43

README.md
remyabel: Since these instructions pertain more to developers, it would probably be better to just download the single header release, not build the entire thing from source. Like so:

```
        http_file(
            name = "com_github_nlohmann_json_single_header",
            urls = [
                "https://github.com/nlohmann/json/releases/download/v3.6.1/json.hpp",
            ],
            sha256 = "blah_blah_blah",
        )
```

README.md
```diff
@@ -147,6 +147,84 @@ endif()
...
+int main(int argc, char** argv) {
+  nlohmann::json obj = {
+    {"bazel", "https://bazel.build"},
+    {"cmake", "https://cmake.org/"},    
+  };
+  std::cout << obj.dump(4) << std::endl;
+}
+```
+
+NOTE: you should use double quotation marks instead of angle brackets. You can see details at [horance-liu/nlohmann_json_demo](https://github.com/horance-liu/nlohmann_json_demo) project.
```
remyabel: I looked at the demo and didn't see any details about why quotation marks are needed here. Can you explain?

nlohmann:
@remyabel Thanks for your comments! As the PR got stale and you seem to be familiar with the topic: would you be able to take over the PR or create a new one with all comments (including yours) addressed?

WORKSPACE
Vertexwahn: Yes it is - otherwise, this is not recognized by Bazel - I named the file `WORKSPCE.bazel` (which is an alternative to WORKSPACE only) to make it more clear that this is a Bazel thing -> see my PR: https://github.com/nlohmann/json/pull/3709

BUILD.bazel
Vertexwahn: I changed this in https://github.com/nlohmann/json/pull/3709

PR #1643, opened 2019-06-16
Fix json.hpp compilation issue with other typedefs with same name (Issue #1642)

### LINKED ISSUE #1642: json.hpp compilation issue with other typedefs with same name
The issue is similar to #606 and #1530: a compilation error involving `json.hpp`, with `<` being interpreted as a template instead of "less than"

nlohmann-json

Essentially, other typedefs with the same name as the ones used by the library are taken as reference when templating, which happens often under `using namespace`

This is confirmed to happen with GCC 7.4.0 on Ubuntu and GCC 9.1.0 on macOS Mojave, compiling against the version from `apt-get` and `brew` respectively.

A fix would be wrapping the comparisons with parentheses.

Also discovered by @edo9300 and @DyXEL

kevinlul:
While in our case, only the string case was affected, to be safe and allow the library to work with other unforeseen cases, all of the cases have been wrapped with parentheses.

Thank you @DyXEL and @edo9300

Fixes #1642 

* * *

## Pull request checklist

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -5746,25 +5746,25 @@ class basic_json
                     return (*lhs.m_value.array) < (*rhs.m_value.array);
 
                 case value_t::object:
-                    return *lhs.m_value.object < *rhs.m_value.object;
+                    return (*lhs.m_value.object) < (*rhs.m_value.object);
 
                 case value_t::null:
                     return false;
 
                 case value_t::string:
-                    return *lhs.m_value.string < *rhs.m_value.string;
+                    return (*lhs.m_value.string) < (*rhs.m_value.string);
 
                 case value_t::boolean:
-                    return lhs.m_value.boolean < rhs.m_value.boolean;
+                    return (lhs.m_value.boolean) < (rhs.m_value.boolean);
 
                 case value_t::number_integer:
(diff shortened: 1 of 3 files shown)


kevinlul:
Do you mean adding a test case to `tests/` for this behaviour?

nlohmann:
Yes. Best to `unit-regression.cpp` if this is possible.

nlohmann:
@kevinlul Let me know if you need help on the test case.

kevinlul:
It seems like `boolean` is already a definition somewhere under VS2015, including Windows headers?

nlohmann:
Indeed - the error only occurs in the configuration where we include `windows.h` to detect exactly these kind of issues... I think it would be best not to mention `boolean` in the regression test.

nlohmann:
Thanks!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [x] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Fixed an issue where typedefs with certain names yielded a compilation error #1642.

* * *

Issue #1129, opened 2018-06-12
CBOR byte string support


Lusitanian:
Currently, the CBOR feature of the library only supports writing strings out as UTF-8 and doesn't support byte arrays as described in the CBOR standard. I'd like to implement COSE on top of this library, but in order to do that, byte string support is needed.

I'm happy to submit a pull request for this, but I wanted to check in advance if this is something you'd be open to

nlohmann:
Related: #757, #587.

Lusitanian:
my thought was creation of some byte string type, leaving existing treatment of strings alone. i'll have time this week to show in more detail

nlohmann:
So you mean the byte string's bytes should just be parsed and stored in a JSON string? What about encodings? Please also have a look at the issues I referenced above.

nifker:
@nlohmann How about adding the suffix _BIN or _BINARY to the key string and the string then contains the byte string to a base64 converted string?

nlohmann:
What about an array of binary strings? What about an array of values of which some are binary strings? How would roundtripping work?

nlohmann:
Any news on this?

nlohmann:
@Lusitanian @nifker Any input on this?

nifker:
My approach would be still using a prefix/suffix for the key and for the data itself base32768 or base64 - but I dont what performance base32768 has or usable it is.
An example would look like this:
> BINARY_binarystring: "媒腻㐤┖ꈳ埳"

> binarystring_BINARY: "XmFFzQtc"

The prefix/suffix is trimmed from the key string and the resulting string is used as key in CBOR. 
The value is just decoded from base64/base32768 back into a binary string and will be the value in CBOR.

nlohmann:
I do not like this approach, as it relies on the convention to use and respect such prefixes/suffixes.

nifker:
Couldn't imagine how else you want to detect that.

nlohmann:
Any opinion on this?

alexkarpenko:
How about adding a data type for binary blobs?

```cpp
std::vector<uint8_t> myData{0,1,2,3,4,5,6};
json j = {
  {"binary", json::bin(myData.begin(), myData.end())}
};

auto v_msgpack = json::to_msgpack(j);
json j2 = json::from_msgpack(v_msgpack);

bool is_binary = j2["binary"].is_binary(); // true
[...]

PR #1662, opened 2019-07-05
Add binary type support to all binary file formats, as well as an internally represented binary type

OmnipotentEntity:
This pull request completely implements binary type serialization and deserialization for all formats that support it.  And it adds an internally represented binary type for JSONs (non-standard and serializable by default only to binary formats).  The internally represented binary type also carries along an optional binary subtype, which is used by certain binary formats (msgpack, bson)

UBJSON does not support binary types, and they recommend serialization to a list of numbers, which is what is implemented by this pull request.  This has the side effect of not being roundtrippable.  Similarly, cbor does not support binary subtypes, so the subtype field for cbor is also not roundtrippable.  Otherwise, the implementation should be roundtrippable on all inputs.

This pull request addresses #1129.

(Code coverage 100% is checked, because I think I covered everything, I don't know if we're using coveralls anymore, and I don't know how to trigger a recheck.  Please trigger a recheck or tell me how, and if my code coverage is lacking I'll add more tests.)

~~deserialization.  Currently no tests and completely untested other than
for basic compilation~~

~~This is not a complete pull request as it stands, I have made no attempt to test or amalgamate the code yet.  This is currently in the state it is in because I'd like feedback before going farther.  This pull request is intended to address #1129 .~~

~~Code is thus far only tested as compiling using gcc 7.4.0 (however, because much of the code is templated, and I don't have any tests that actually calls the code, there are undoubtedly numerous compilation issues just waiting to say hi just as soon as I get around to it.)~~

~~Is this approach decent?  Do you have any specific requests as far as implementation is concerned?  Is there anything I'm doing that I ought not?  Have I missed anything large?~~

* * *

## Pull request checklist

## Please don't


nlohmann:
Thank you very much! On first browsing, this looks all very good! Unfortunately, I won't be able to have a deeper look at this until next week.

OmnipotentEntity:
That's totally understandable. Clearly I'm OK with things taking a long while. ;)

I'll patiently await your feedback before continuing.

OmnipotentEntity:
If you don't have anything specific to critique about my general approach, then I'll keep at it for now.  Let me know though.

nlohmann:
Please go ahead! The approach appears sound and has great potential!

include/nlohmann/json_fwd.hpp
```diff
@@ -31,6 +31,7 @@ template<template<typename U, typename V, typename... Args> class ObjectType =
          class NumberIntegerType = std::int64_t,
          class NumberUnsignedType = std::uint64_t,
          class NumberFloatType = double,
+         class BinaryType = std::vector<std::uint8_t>,
```
tambry: Maybe use `std::byte` when C++17? That type was added exactly for such purposes.  
Or at least consider a possible define for this.

nlohmann: I’d like to keep it uint8, because this is also use in the binary readers and writers. Maybe a template type could help in the future.

OmnipotentEntity: I have been avoiding using C++14 and beyond in this code base because the library is advertised as working on very old versions of GCC (like 4.8), otherwise, there's also a [[maybe_unused]] attribute flag in C++17 that I want to use as well to silence some spurious warnings about casting (but are actually unused variables.)

OmnipotentEntity:
Does anyone have any good test cases for the various binary types?  I'd rather not haaaaave to make them all by hand if I don't need to. ;)

include/nlohmann/json_fwd.hpp
nlohmann: For attributes, I recently started using [Hedley](https://nemequ.github.io/hedley/) which allows a portable way to use all kind of attributes.

nlohmann:
> Does anyone have any good test cases for the various binary types? I'd rather not haaaaave to make them all by hand if I don't need to. ;)

I can add examples.

Meanwhile, the build is red. It seems that you are using macros like `JSON_UNLIKELY`. This macro is now called `JSON_HEDLEY_UNLIKELY` as we now use Hedley under the hood. Can you please update?

nlohmann:
Do you have an idea why the tests are failing?

OmnipotentEntity:
Yeah, sorry I buried the reason in a commit message.  When I make the json it calls the json array constructor instead of the json binary constructor, because the template match is better somehow.  I was contemplating a solution, but I think the best course of action is to simply remove the unadorned binary ctor so you can't accidentally call it, and then make static functions that return a constructed json of a specific type.  Something like `json::make_binary(BinaryType foo)`, but I haven't gotten around to it yet, because my irl job has been busy.

OmnipotentEntity:
I have a few cbor data tests for binary.  But no messagepack, ubjson, or bson.  Soliciting contributions to make sure my code doesn't suck. ;)

nlohmann:
Thanks for the effort! I will have a look at this during the weekend!

nlohmann:
Any idea how I can add tests to this PR? Should I make a copy of the branch and go ahead from there?

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -112,10 +112,10 @@
              template<typename, typename...> class ArrayType,              \
              class StringType, class BooleanType, class NumberIntegerType, \
              class NumberUnsignedType, class NumberFloatType,              \
-             template<typename> class AllocatorType,                       \
+             class BinaryType, template<typename> class AllocatorType,     \
```
nlohmann: Just to be sure: adding the `BinaryType` not to the end of the list would be a breaking change, right?

test/src/unit-cbor.cpp
```diff
@@ -36,6 +36,7 @@ using nlohmann::json;
 #include <fstream>
 #include <sstream>
 #include <iomanip>
+#include <iostream>
```
nlohmann: There should not be output in the unit tests.

OmnipotentEntity: oof ouch, my plebian printf debugging has been revealed.

include/nlohmann/detail/macro_scope.hpp
OmnipotentEntity: Yes, I can move it to the back of the list though, it was honestly something I hadn't considered.

OmnipotentEntity:
Typically the flow in Github is you make a fork of my branch of your fork, submit a pull request to me, I merge it there, and then you merge it here.  But that's complicated and kind of dumb tbh.

In regular git, you'd just merge my diff into a different branch, change that branch, and then merge that branch into the standard branch all locally.

One thing that I did to fix the tests, which probably should be discussed a bit, is I added text serialization (but not deserialization) to binary values (as 'b[]').  This is because the testing harness really, really wants to print the value of json, and printing the value of the json is the same as serializing it.  So it was throwing when it contained a binary value.  I wanted to ask if I should add a deserialization or wrap the serialization code in a `ifdef` that only is compiled when running tests.

OmnipotentEntity:
It looks like the failing travis test is a result of a failed apt-get command.  Strange.

include/nlohmann/detail/macro_scope.hpp
nlohmann: No worries. But let's move it to the back not to break existing code. We can move it to a better position once we go for 4.0.0.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -236,6 +237,50 @@ class serializer
                 return;
             }
 
+            case value_t::binary:
```
nlohmann: We may want to control this behavior similar to the `error_handler_t` in https://github.com/nlohmann/json#character-encoding. Just going for a string like this should not be the only option.

include/nlohmann/json.hpp
```diff
@@ -2540,7 +2755,7 @@ class basic_json
     @brief get a value (explicit)
 
     Explicit type conversion between the JSON value and a compatible value
-    which is [CopyConstructible](https://en.cppreference.com/w/cpp/named_req/CopyConstructible)
+    which is [Copynonstructible](https://en.cppreference.com/w/cpp/named_req/CopyConstructible)
```
nlohmann: Typo.

nlohmann:
> It looks like the failing travis test is a result of a failed apt-get command. Strange.

I restarted the builds. Travis is flaky sometimes.

nlohmann:
> Typically the flow in Github is you make a fork of my branch of your fork, submit a pull request to me, I merge it there, and then you merge it here. But that's complicated and kind of dumb tbh.
> 
> In regular git, you'd just merge my diff into a different branch, change that branch, and then merge that branch into the standard branch all locally.
> 
> One thing that I did to fix the tests, which probably should be discussed a bit, is I added text serialization (but not deserialization) to binary values (as 'b[]'). This is because the testing harness really, really wants to print the value of json, and printing the value of the json is the same as serializing it. So it was throwing when it contained a binary value. I wanted to ask if I should add a deserialization or wrap the serialization code in a `ifdef` that only is compiled when running tests.

I'll try the fork of your branch tomorrow and open a PR then.

include/nlohmann/detail/output/serializer.hpp
OmnipotentEntity: I'm attempting to do this by adding a parameter, just like the character encoding.  However, the testing harness simply calls `operator<<`, it doesn't call dump directly, and I don't want to override `operator<<`'s just to make this call work.  I can adjust the reaction to some `iomanip` flags, but I feel like that would have unintended consequences or be otherwise surprising, and I do not want to add a bool to the json data structure to store just this toggle, seems inelegant.  Any ideas?

OmnipotentEntity: `public static` member to toggle it?

OmnipotentEntity: For now, I'm using a public static member.  This isn't very nice because it introduces global state into serialization.  And I'm open to other ideas.

OmnipotentEntity:
I'm intending on spending part of Sunday working on this.  As far as I can tell all that is left is the test cases for msgpack, bson, and ubjson.  Please let me know if this is correct.  (And what you think of the serialization toggle.)

nlohmann:
@OmnipotentEntity Sorry for not answering earlier. I will try to check this during the weekend.

nlohmann:
I am really sorry. Something unexpected came up and I won't be able to work in this the next days. :-/

OmnipotentEntity:
No worries, I understand life happens, and this is quite a large change that touches many, many different systems.  I'll try to get in some test cases for the other binary data types today, and then I'll wait for a review.

include/nlohmann/detail/output/serializer.hpp
nlohmann: Where is that static member?

nlohmann: (Forget it - I found it.)

include/nlohmann/json.hpp
```diff
@@ -1922,6 +2101,18 @@ class basic_json
     // object inspection //
     ///////////////////////
 
+    /*!
```
nlohmann: This should not stay like this. If we do want to control the behavior of `dump`, we need to add another parameter.

include/nlohmann/detail/output/serializer.hpp
nlohmann: `operator<<` can only be controlled in a very limited way - right now, only the level of indentation can be controlled via `std::setw`. This is fine, because it is only a convenience way of calling the `dump` function with more detailed parameters. We should not add any API just to allow all options from `dump` to also work with `operator<<`.

include/nlohmann/json.hpp
```diff
@@ -1458,6 +1553,78 @@ class basic_json
...
+    @brief explicitly create a binary array from an already constructed copy of
+    its base type
+
+    Creates a JSON binary array value from a given `binary_t`. Because binary
+    values are not part of the JSON specification, this is not the constructor
+    you want unless you specifically want to serialize to a binary format.
+
+    @note Note, this function exists because of the difficulty in correctly
```
nlohmann: This comment is very defensive. I think it would be OK to mention that binary data is allowed in formats like CBOR, MessagePack, etc., and this constructor is meant to create binary values to be serialized using these formats.

include/nlohmann/json.hpp
```diff
@@ -1977,11 +2168,11 @@ class basic_json
 
         if (indent >= 0)
         {
-            s.dump(*this, true, ensure_ascii, static_cast<unsigned int>(indent));
+            s.dump(*this, true, ensure_ascii, static_cast<unsigned int>(indent), serialize_binary);
```
nlohmann: The `serialize_binary` parameter should be part of the signature of `dump`. I think a `bool` may be insufficient, because we may want to have similar behavior as `error_handler` controls for invalid UTF-8:

- strict - throw on binary values
- ignore - ignore binary values (i.e., do not serialize them)
- replace - dump with some kind of serialization

I'm not sure whether the exact same `error_handler_t` enum should be used, but something like this could do the job.

OmnipotentEntity:
I promise I'm still alive.  I've just been ridiculously busy over the past month or so.  I'll get to this ASAP.

nlohmann:
I have to apologize for not being able to add test cases...

OmnipotentEntity:
No worries.  I have two coding jobs now, so getting to this as a hobby has been a little touch and go lately.  I'll squeeze this in when I can.

OmnipotentEntity:
Ok, I think I'll have some time to work on this tomorrow.  Before I do that though, I had a question/concern.

The entire reason I added serialization of binary to the `dump` function and the global serialization flag to tell `operator<<` that it's ok to serialize binary is because the testing harness requires the ability to serialize an object via the `operator<<`, and as near as I can tell that's a hard requirement.  If the `operator<<` throws because binary data exists then there's no way for testing to complete.  So while I completely agree with you in principle that `operator<<` should only be a convenience wrapper around `dump` that does not need and should not attempt to emulate all its options, this one option seems more or less required (if only for testing).  Can you think of a better way to get around this?  Do you know of a way to make the testing harness call `dump` directly?

OmnipotentEntity:
Well... I've made a right mess of this.  Attempting to rebase and squash

OmnipotentEntity:
I don't know why Code Inspector is flagging me for those lines.  They're not part of this commit?

OmnipotentEntity:
Back to a single job.  Can do a final push soon and get this complete.

@nlohmann can you let me know what you think should be done about the problems with testing and serialization?

nlohmann:
Thanks for checking back!

I'll have a look at this.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -75,6 +76,14 @@ struct json_sax
     */
     virtual bool string(string_t& val) = 0;
 
+    /*!
+    @brief a binary string was read
+    @param[in] val  binary value
+    @return whether parsing should proceed
+    @note It is safe to move the passed binary.
+    */
+    virtual bool binary(binary_t& val) = 0;
```
nlohmann: As binary formats may know the size of a binary value in advance, we may want to pass it to `binary` just as we do with arrays and objects. What do you think?

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -236,6 +240,55 @@ class serializer
...
+            {
+                if (!serialize_binary)
```
nlohmann: Replace `!` with `not`.

include/nlohmann/json.hpp
nlohmann: This variable really needs to go. We can control `dump` with a parameter, and `operator<<` was never designed to have the same amount of control as `dump`.

include/nlohmann/json.hpp
```diff
@@ -2540,7 +2768,7 @@ class basic_json
     @brief get a value (explicit)
 
     Explicit type conversion between the JSON value and a compatible value
-    which is [CopyConstructible](https://en.cppreference.com/w/cpp/named_req/CopyConstructible)
+    which is [Copynonstructible](https://en.cppreference.com/w/cpp/named_req/CopyConstructible)
```
nlohmann: `Copynonstructible` -> `CopyConstructible`

nlohmann:
Hi @OmnipotentEntity, sorry for the delay. I now have some time to check this PR. I think it's 90% there, but I have a conceptual question:

Should be somewhere store the type of the binary value? What I mean: the binary formats usually have more then one binary type (like MessagePacks `bin` and `ext` types). If we could mark the type of the value somewhere, then we could even parse more types to binary types and just denote the type. The type could be as simple as an int (or `uint8_t`).

OmnipotentEntity:
I thought about it and I intentionally left it out due to the differences between formats.  I didn't want the design of that portion to be a blocker on the binary type.  For instance, if load a BSON of type 2, then serialize it to MsgPack it would naively want to serialize it as an ext with... I guess a type of 2.  But it would be more accurate to serialize it as a bin (ie, you can't just pass along the number or use a single sentinel value to determine between ext and bin serialization).  The mapping of binary types to number for msgpack is user defined, and bson also has user defined types. 

So ultimately you'd need some sort of translation between the file formats, and some user support in defining user types and user type translations between the file formats (because user type ranges are mutually exclusive between these two types!) I thought it best as a first pass to avoid the complexity between these two types and simply do a minimal subset.  (Also MsgPack uses an `int8_t` while bson uses a `uint8_t`.)  If you're worried about incompatible serialization between versions, then I can add a `std::optional<uint8_t>` (if that's not too new) to the file format and populate it with the the correct subtype, if present.  And allow the user to change it at their option.

include/nlohmann/json.hpp
OmnipotentEntity: Right, again I agree that the `serialize_binary` parameter is a janky hack.  Again, I only put it in the make the test harness work.  Otherwise when you call:

    CHECK(result == expected);

You get a unhandled exception because `CHECK` wants to dump the object using `operator<<`, and I don't know how to override this behavior or if it's even possible to override this behavior.  But it doesn't seem like it is, the offending line seems to be found [here](https://github.com/onqtam/doctest/blob/master/doctest/parts/doctest_fwd.h#L1994)? __VA_ARGS__ is passed via operator<< to the expression decomposer?

I'm sure I'm misunderstanding this code.  But the long story short is, when I tried it previously without the janky hack I was simply unable to test the code at all!  It's ugly, and it's awful, and I really, really hate it, but I cannot figure out a way around it.

OmnipotentEntity: Actually, I retract all of this.  This must have only been a problem with the old test suite.  Doctest seems to handle this just fine now.  I shall remove this flag, but leave in the `dump` parameter.

include/nlohmann/detail/input/json_sax.hpp
OmnipotentEntity: Perform byte by byte operations on binary formats?  I mean, it's not an awful idea.  But because an array and an object can contain arbitrary JSON objects, but binary data cannot, it's not nearly as useful.

nlohmann: What I meant is that, for instance, CBOR usually knows the number of bytes in a type (unless you parse until a `break`). So if `0x58 0x1f` is read, then we know it's byte string with 31 bytes. Instead of "just" putting the 31 bytes into a `std::vector<std::uint8_t>` and call `binary(val)`, we could call `binary(val, 31)` to help the consumer. But this idea is bogus. I thought about how we process arrays with a `start_array()` call. This is not the case here - whoever has access to `val` can call `val.size()` and get all the information.

Sorry for the confusion.

nlohmann:
I did not even think about translations between formats...

I just thought about two things:

1. In case there are different binary types in one format, there is currently no way to distinguish them once they are parsed to `basic_json`. This means we actually lose information. And this would mean that we would not be able to write code that can react on any input, but only for some previously agreed use cases. Of course, binary data is even more complicated, because it does not just fit built-in data types...
2. In case we could be able to differentiate binary types, we could be generous and map all remaining unsupported types to binary since we can denote their "original" type somewhere. As a result, we would not need to throw, but rather let the user decide later on how to transform their data in the desired form.

You are right - the numbers would not be portable between formats. We could extend this by an enum denoting the format (we already have this) and the number. So we would parse the CBOR input

```
0x42 0x01 0x02
```

into a byte vector `0x01 0x02` together with the information "this was CBOR" and "this was major type 2 (byte string)".

Having an API like this would also allow users to decide how binary data would be serialized.

Am I overengineering?!

OmnipotentEntity:
Sorry about the delay.  Holidays.

I agree that we can track the data source and binary subtype if it exists.  However, I'm having some difficulty figuring out what that will look like.

So, using the existing `value_t` enum to track this data is not tenable.  The data types are implementation defined in places, and mapping an `uint8_t` into a `uint8_t` with certain values already defined impossible without data loss.

One method of exposing this information without adding a field to the basic_json type that's only used for binary values, is to expose it only via the sax interface, (perhaps by passing an optional output parameter, though that's not very modern C++), and allow the user to parse it to a custom type based on the subtype during the parsing step.  However, this would mean that handling the value properly from within the basic_json type is impossible.

Another possibility would be to, instead of using `std::vector<uint8_t>` as the default binary type, use a `std::tuple<std::optional<uint8_t>, std::vector<uint8_t>>`.  However, that would greatly reduce the number of types that would casually template match such a type.  It's possible to define a struct instead that contains the type and vector.  However, again, this forces a rather restrictive structure on the data type.

It would also be possible to break the template types up into two types, and have an internal struct or tuple representation of both (ie, enforce the association between the types internally and hide the details).  So we'd have a `binary_type` as `std::vector<uint8_t>` and `binary_subtype_type` as a `uint8_t` and internally we'd have a `struct {binary_type bin; binary_subtype_type bin_sub;};` or something similar.

As far as the source is concerned, I don't see where this information is currently tracked.  (Forgive me if that's not what you intended to imply.)  We can store this data similarly to the subtype.

There's also the question of whether the subtype type should even be exposed as a template type.  I'm leaning towards no personally, because all of the types that track the subtype seem to use a `uint8_t` value, so there's no real reason to expose it?

OmnipotentEntity:
Ok, this should be everything required to support subtypes.  I could make `_subtype` a `std::optional`, but that would require dragging in C++17, which is probably not desirable right now (or ever?)  Let me know what you think of the approach, and I'll start writing test cases around them.  Thanks!

OmnipotentEntity:
@nlohmann any feedback on the `_subtype`?

OmnipotentEntity:
msgpack tests updated, and bugs found and fixed.

Pending bson and ubjson tests.

nlohmann:
I can build the code locally, but Travis and AppVeyor seem to have big problems. But I do get this warnings:

```
[24/85] Building CXX object test/CMakeFiles/test-udt.dir/src/unit-udt.cpp.o
In file included from ../test/src/unit-udt.cpp:32:
../include/nlohmann/json.hpp: In instantiation of 'static ReferenceType nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>::get_ref_impl(ThisType&) [with ReferenceType = const nlohmann::basic_json<>::internal_binary_t&; ThisType = const nlohmann::basic_json<>; ObjectType = std::map; ArrayType = std::vector; StringType = std::__cxx11::basic_string<char>; BooleanType = bool; NumberIntegerType = long long int; NumberUnsignedType = long long unsigned int; NumberFloatType = double; AllocatorType = std::allocator; JSONSerializer = nlohmann::adl_serializer; BinaryType = std::vector<unsigned char>]':
../include/nlohmann/json.hpp:3129:43:   required from 'ReferenceType nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>::get_ref() const [with ReferenceType = const nlohmann::basic_json<>::internal_binary_t&; typename std::enable_if<(std::is_reference<ReferenceType>::value && std::is_const<typename std::remove_reference<_SrcTuple>::type>::value), int>::type <anonymous> = 0; ObjectType = std::map; ArrayType = std::vector; StringType = std::__cxx11::basic_string<char>; BooleanType = bool; NumberIntegerType = long long int; NumberUnsignedType = long long unsigned int; NumberFloatType = double; AllocatorType = std::allocator; JSONSerializer = nlohmann::adl_serializer; BinaryType = std::vector<unsigned char>]'
../include/nlohmann/json.hpp:1488:56:   required from 'nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>::basic_json(const BasicJsonType&) [with BasicJsonType = nlohmann::basic_json<>; typename std::enable_if<(nlohmann::detail::is_basic_json<BasicJsonType>::value && (! std::is_same<nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>, BasicJsonType>::value)), int>::type <anonymous> = 0; ObjectType = std::map; ArrayType = std::vector; StringType = std::__cxx11::basic_string<char>; BooleanType = bool; NumberIntegerType = long long int; NumberUnsignedType = long long unsigned int; NumberFloatType = double; AllocatorType = std::allocator; JSONSerializer = another_adl_serializer; BinaryType = std::vector<unsigned char>]'
../test/src/unit-udt.cpp:714:26:   required from here
../include/nlohmann/json.hpp:2772:21: warning: returning reference to temporary [-Wreturn-local-addr]
 2772 |             return *ptr;
      |                     ^~~
[84/85] Building CXX object test/CMakeFiles/test-regression.dir/src/unit-regression.cpp.o
In file included from ../test/src/unit-regression.cpp:37:
../include/nlohmann/json.hpp: In instantiation of 'static ReferenceType nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>::get_ref_impl(ThisType&) [with ReferenceType = const nlohmann::basic_json<>::internal_binary_t&; ThisType = const nlohmann::basic_json<>; ObjectType = std::map; ArrayType = std::vector; StringType = std::__cxx11::basic_string<char>; BooleanType = bool; NumberIntegerType = long long int; NumberUnsignedType = long long unsigned int; NumberFloatType = double; AllocatorType = std::allocator; JSONSerializer = nlohmann::adl_serializer; BinaryType = std::vector<unsigned char>]':
../include/nlohmann/json.hpp:3129:43:   required from 'ReferenceType nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>::get_ref() const [with ReferenceType = const nlohmann::basic_json<>::internal_binary_t&; typename std::enable_if<(std::is_reference<ReferenceType>::value && std::is_const<typename std::remove_reference<_SrcTuple>::type>::value), int>::type <anonymous> = 0; ObjectType = std::map; ArrayType = std::vector; StringType = std::__cxx11::basic_string<char>; BooleanType = bool; NumberIntegerType = long long int; NumberUnsignedType = long long unsigned int; NumberFloatType = double; AllocatorType = std::allocator; JSONSerializer = nlohmann::adl_serializer; BinaryType = std::vector<unsigned char>]'
../include/nlohmann/json.hpp:1488:56:   required from 'nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>::basic_json(const BasicJsonType&) [with BasicJsonType = nlohmann::basic_json<>; typename std::enable_if<(nlohmann::detail::is_basic_json<BasicJsonType>::value && (! std::is_same<nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>, BasicJsonType>::value)), int>::type <anonymous> = 0; ObjectType = my_workaround_fifo_map; ArrayType = std::vector; StringType = std::__cxx11::basic_string<char>; BooleanType = bool; NumberIntegerType = long long int; NumberUnsignedType = long long unsigned int; NumberFloatType = double; AllocatorType = std::allocator; JSONSerializer = nlohmann::adl_serializer; BinaryType = std::vector<unsigned char>]'
../test/src/unit-regression.cpp:1623:23:   required from here
../include/nlohmann/json.hpp:2772:21: warning: returning reference to temporary [-Wreturn-local-addr]
 2772 |             return *ptr;
      |                     ^~~
../include/nlohmann/json.hpp: In instantiation of 'static ReferenceType nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>::get_ref_impl(ThisType&) [with ReferenceType = const nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char>, bool, long long int, long long unsigned int, double, std::allocator, ns::foo_serializer, std::vector<unsigned char> >::internal_binary_t&; ThisType = const nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char>, bool, long long int, long long unsigned int, double, std::allocator, ns::foo_serializer, std::vector<unsigned char> >; ObjectType = std::map; ArrayType = std::vector; StringType = std::__cxx11::basic_string<char>; BooleanType = bool; NumberIntegerType = long long int; NumberUnsignedType = long long unsigned int; NumberFloatType = double; AllocatorType = std::allocator; JSONSerializer = ns::foo_serializer; BinaryType = std::vector<unsigned char>]':
../include/nlohmann/json.hpp:3129:43:   required from 'ReferenceType nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>::get_ref() const [with ReferenceType = const nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char>, bool, long long int, long long unsigned int, double, std::allocator, ns::foo_serializer, std::vector<unsigned char> >::internal_binary_t&; typename std::enable_if<(std::is_reference<ReferenceType>::value && std::is_const<typename std::remove_reference<_SrcTuple>::type>::value), int>::type <anonymous> = 0; ObjectType = std::map; ArrayType = std::vector; StringType = std::__cxx11::basic_string<char>; BooleanType = bool; NumberIntegerType = long long int; NumberUnsignedType = long long unsigned int; NumberFloatType = double; AllocatorType = std::allocator; JSONSerializer = ns::foo_serializer; BinaryType = std::vector<unsigned char>]'
../include/nlohmann/json.hpp:1488:56:   required from 'nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>::basic_json(const BasicJsonType&) [with BasicJsonType = nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char>, bool, long long int, long long unsigned int, double, std::allocator, ns::foo_serializer, std::vector<unsigned char> >; typename std::enable_if<(nlohmann::detail::is_basic_json<BasicJsonType>::value && (! std::is_same<nlohmann::basic_json<ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType, NumberFloatType, AllocatorType, JSONSerializer, BinaryType>, BasicJsonType>::value)), int>::type <anonymous> = 0; ObjectType = std::map; ArrayType = std::vector; StringType = std::__cxx11::basic_string<char>; BooleanType = bool; NumberIntegerType = long long int; NumberUnsignedType = long long unsigned int; NumberFloatType = double; AllocatorType = std::allocator; JSONSerializer = nlohmann::adl_serializer; BinaryType = std::vector<unsigned char>]'
../test/src/unit-regression.cpp:1634:29:   required from here
../include/nlohmann/json.hpp:2772:21: warning: returning reference to temporary [-Wreturn-local-addr]
```

OmnipotentEntity:
I've seen those errors.  However those lines are not part of my diff, as I mentioned all of the way [up here](https://github.com/nlohmann/json/pull/1662#issuecomment-544296954).  So I don't know why it's flagging it on this pull rather than in general.

OmnipotentEntity:
UBJSON tests added.

OmnipotentEntity:
BSON Tests added, code is rebased and squashed.  This should be ready to merge pending review @nlohmann

OmnipotentEntity:
Latest pushes are to try to fix visual studio.  It seems that it doesn't like that `internal_binary_t` doesn't have a default ctor defined, despite not requiring one in this case.

OmnipotentEntity:
The remaining errors in Appveyor seem to be related to some code in test/src/unit-conversions.cpp where `operator=` is ambiguous when attempting to assign to/from.  I don't know which lines, because it doesn't tell me the line number.  It is only occurring in VS 2019, which makes me think it has something to do with the lines:

```
#if defined(JSON_HAS_CPP_17)
        SECTION("std::string_view")
        {
            std::string_view s = j.get<std::string_view>();
            CHECK(json(s) == j);
        }
#endif
```

However, I have an empty diff against this file, and my code shouldn't be affecting string handling or assignments.  Any ideas?

OmnipotentEntity:
Latest push is an attempt to eliminate any possibility that new code is causing the Visual Studio 2019 issue.  I've disallowed any attempt to implicitly and explicitly convert `internal_binary_t` to any type except for (optionally cv-qualified) `BinaryType`.

OmnipotentEntity:
Latest push is the same code, just a rebase onto the latest head to get #2008 and hopefully fix the VS2019 issue in appveyor I've been seeing.

OmnipotentEntity:
@nlohmann I think this pull request is finally ready.

nlohmann:
> @nlohmann I think this pull request is finally ready.

Cool! Thanks for keeping working on this! I will have a look soon!

test/src/unit-ubjson.cpp
```diff
@@ -905,6 +910,231 @@ TEST_CASE("UBJSON")
...
+                    std::vector<uint8_t> expected;
+                    expected.push_back(static_cast<uint8_t>('['));
+                    if (N != 0)
+                    {
+                        expected.push_back(static_cast<uint8_t>('$'));
+                        expected.push_back(static_cast<uint8_t>('U'));
+                    }
+                    expected.push_back(static_cast<uint8_t>('#'));
+                    expected.push_back(static_cast<uint8_t>('i'));
+                    expected.push_back(static_cast<uint8_t>(N));
```
nlohmann, line 935:
```suggestion
                for (std::size_t N = 0; N <= 127; ++N)
                {
                    CAPTURE(N)

                    // create JSON value with byte array containing of N * 'x'
                    const auto s = std::vector<std::uint8_t>(N, 'x');
                    json j = json::binary_array(s);

                    // create expected byte vector
                    std::vector<std::uint8_t> expected;
                    expected.push_back(static_cast<std::uint8_t>('['));
                    if (N != 0)
                    {
                        expected.push_back(static_cast<std::uint8_t>('$'));
                        expected.push_back(static_cast<std::uint8_t>('U'));
                    }
                    expected.push_back(static_cast<std::uint8_t>('#'));
                    expected.push_back(static_cast<std::uint8_t>('i'));
                    expected.push_back(static_cast<std::uint8_t>(N));
```

test/src/unit-ubjson.cpp
```diff
@@ -905,6 +910,231 @@ TEST_CASE("UBJSON")
...
+                    json j = json::binary_array(s);
+
+                    // create expected byte vector
+                    std::vector<uint8_t> expected;
+                    expected.push_back(static_cast<uint8_t>('['));
+                    expected.push_back(static_cast<uint8_t>('$'));
+                    expected.push_back(static_cast<uint8_t>('U'));
+                    expected.push_back(static_cast<uint8_t>('#'));
+                    expected.push_back(static_cast<uint8_t>('U'));
+                    expected.push_back(static_cast<uint8_t>(N));
```
nlohmann, line 983:
```suggestion
                for (std::size_t N = 128; N <= 255; ++N)
                {
                    CAPTURE(N)

                    // create JSON value with byte array containing of N * 'x'
                    const auto s = std::vector<std::uint8_t>(N, 'x');
                    json j = json::binary_array(s);

                    // create expected byte vector
                    std::vector<uint8_t> expected;
                    expected.push_back(static_cast<std::uint8_t>('['));
                    expected.push_back(static_cast<std::uint8_t>('$'));
                    expected.push_back(static_cast<std::uint8_t>('U'));
                    expected.push_back(static_cast<std::uint8_t>('#'));
                    expected.push_back(static_cast<std::uint8_t>('U'));
                    expected.push_back(static_cast<std::uint8_t>(N));
```

test/src/unit-ubjson.cpp
```diff
@@ -905,6 +910,231 @@ TEST_CASE("UBJSON")
...
+
+                    // create expected byte vector
+                    std::vector<uint8_t> expected(N + 7, 'x');
+                    expected[0] = '[';
+                    expected[1] = '$';
+                    expected[2] = 'U';
+                    expected[3] = '#';
+                    expected[4] = 'I';
+                    expected[5] = static_cast<uint8_t>((N >> 8) & 0xFF);
+                    expected[6] = static_cast<uint8_t>(N & 0xFF);
```
nlohmann, line 1024:
```suggestion
                    const auto s = std::vector<std::uint8_t>(N, 'x');
                    json j = json::binary_array(s);

                    // create expected byte vector
                    std::vector<std::uint8_t> expected(N + 7, 'x');
                    expected[0] = '[';
                    expected[1] = '$';
                    expected[2] = 'U';
                    expected[3] = '#';
                    expected[4] = 'I';
                    expected[5] = static_cast<std::uint8_t>((N >> 8) & 0xFF);
                    expected[6] = static_cast<std::uint8_t>(N & 0xFF);
```

test/src/unit-ubjson.cpp
```diff
@@ -905,6 +910,231 @@ TEST_CASE("UBJSON")
...
+                    // roundtrip only works to an array of numbers
+                    json j_out = s;
+                    CHECK(json::from_ubjson(result) == j_out);
+                    CHECK(json::from_ubjson(result, true, false) == j_out);
+                }
+            }
+
+            SECTION("N = 256..32767")
+            {
+                for (size_t N :
```
nlohmann, line 1005:
```suggestion
                for (std::size_t N :
```

test/src/unit-ubjson.cpp
```diff
@@ -905,6 +910,231 @@ TEST_CASE("UBJSON")
...
+                    // roundtrip only works to an array of numbers
+                    json j_out = s;
+                    CHECK(json::from_ubjson(result) == j_out);
+                    CHECK(json::from_ubjson(result, true, false) == j_out);
+                }
+            }
+
+            SECTION("N = 32768..2147483647")
+            {
+                for (size_t N :
```
nlohmann, line 1042:
```suggestion
                for (std::size_t N :
```

test/src/unit-ubjson.cpp
```diff
@@ -905,6 +910,231 @@ TEST_CASE("UBJSON")
...
+                    std::vector<uint8_t> expected(N + 9, 'x');
+                    expected[0] = '[';
+                    expected[1] = '$';
+                    expected[2] = 'U';
+                    expected[3] = '#';
+                    expected[4] = 'l';
+                    expected[5] = static_cast<uint8_t>((N >> 24) & 0xFF);
+                    expected[6] = static_cast<uint8_t>((N >> 16) & 0xFF);
+                    expected[7] = static_cast<uint8_t>((N >> 8) & 0xFF);
+                    expected[8] = static_cast<uint8_t>(N & 0xFF);
```
nlohmann, line 1063:
```suggestion
                    const auto s = std::vector<std::uint8_t>(N, 'x');
                    json j = json::binary_array(s);

                    // create expected byte vector
                    std::vector<std::uint8_t> expected(N + 9, 'x');
                    expected[0] = '[';
                    expected[1] = '$';
                    expected[2] = 'U';
                    expected[3] = '#';
                    expected[4] = 'l';
                    expected[5] = static_cast<std::uint8_t>((N >> 24) & 0xFF);
                    expected[6] = static_cast<std::uint8_t>((N >> 16) & 0xFF);
                    expected[7] = static_cast<std::uint8_t>((N >> 8) & 0xFF);
                    expected[8] = static_cast<std::uint8_t>(N & 0xFF);
```

test/src/unit-ubjson.cpp
```diff
@@ -905,6 +910,231 @@ TEST_CASE("UBJSON")
...
+                    json j_out = s;
+                    CHECK(json::from_ubjson(result) == j_out);
+                    CHECK(json::from_ubjson(result, true, false) == j_out);
+                }
+            }
+
+            SECTION("Other Serializations")
+            {
+                const size_t N = 10;
+                const auto s = std::vector<uint8_t>(N, 'x');
```
nlohmann, line 1082:
```suggestion
                const std::size_t N = 10;
                const auto s = std::vector<std::uint8_t>(N, 'x');
```

test/src/unit-ubjson.cpp
```diff
@@ -905,6 +910,231 @@ TEST_CASE("UBJSON")
...
+                SECTION("No Count No Type")
+                {
+                    std::vector<uint8_t> expected;
+                    expected.push_back(static_cast<uint8_t>('['));
+                    for (size_t i = 0; i < N; ++i)
+                    {
+                        expected.push_back(static_cast<uint8_t>('U'));
+                        expected.push_back(static_cast<uint8_t>(0x78));
+                    }
+                    expected.push_back(static_cast<uint8_t>(']'));
```
nlohmann, line 1094:
```suggestion
                    expected.push_back(static_cast<std::uint8_t>('['));
                    for (std::size_t i = 0; i < N; ++i)
                    {
                        expected.push_back(static_cast<std::uint8_t>('U'));
                        expected.push_back(static_cast<std::uint8_t>(0x78));
                    }
                    expected.push_back(static_cast<std::uint8_t>(']'));
```

test/src/unit-ubjson.cpp
```diff
@@ -905,6 +910,231 @@ TEST_CASE("UBJSON")
...
+                    expected.push_back(static_cast<uint8_t>('['));
+                    expected.push_back(static_cast<uint8_t>('#'));
+                    expected.push_back(static_cast<uint8_t>('i'));
+                    expected.push_back(static_cast<uint8_t>(N));
+
+                    for (size_t i = 0; i < N; ++i)
+                    {
+                        expected.push_back(static_cast<uint8_t>('U'));
+                        expected.push_back(static_cast<uint8_t>(0x78));
+                    }
```
nlohmann, line 1121:
```suggestion
                    std::vector<std::uint8_t> expected;
                    expected.push_back(static_cast<std::uint8_t>('['));
                    expected.push_back(static_cast<std::uint8_t>('#'));
                    expected.push_back(static_cast<std::uint8_t>('i'));
                    expected.push_back(static_cast<std::uint8_t>(N));

                    for (size_t i = 0; i < N; ++i)
                    {
                        expected.push_back(static_cast<std::uint8_t>('U'));
                        expected.push_back(static_cast<std::uint8_t>(0x78));
                    }
```

include/nlohmann/json.hpp
```diff
@@ -3555,15 +3799,126 @@ class basic_json
...
+
+    @return a mutable reference to the subtype indicating bool of the binary
+    JSON value
+
+    @complexity Constant.
+
+    @since version 3.7.0
+    */
+
+    bool& has_subtype()
```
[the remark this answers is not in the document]
OmnipotentEntity: It's useful for changing whether or not this binary data has a subtype.  For instance, if I read in a BSON (which always has a subtype), and I want to serialize to msgpack's bin (rather than the ext type), I would have to set the `has_subtype` value to false.

include/nlohmann/json.hpp
```diff
@@ -3555,15 +3799,126 @@ class basic_json
...
+
+    @return a mutable reference to the numerical subtype of the binary JSON
+    value
+
+    @complexity Constant.
+
+    @since version 3.7.0
+    */
+
+    uint8_t& subtype()
```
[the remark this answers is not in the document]
OmnipotentEntity: Similarly, if I read in a BSON with (for instance) a binary subtype of 0xFF (user defined), this maps to msgpack's predefined subtype value of "timestamp" but it's not actually a timestamp.  So in order to properly serialize to msgpack you'll have to translate the subtype to the range 0-0x7F (user defined types in msgpack).

test/src/unit-msgpack.cpp
```diff
@@ -1115,6 +1120,261 @@ TEST_CASE("MessagePack")
...
+                        {
+                            256u, 999u, 1025u, 3333u, 2048u, 65535u
+                        })
+                {
+                    CAPTURE(N)
+
+                    // create JSON value with string containing of N * 'x'
+                    const auto s = std::vector<uint8_t>(N, 'x');
+                    json j = json::binary_array(s);
+                    uint8_t subtype = 42;
```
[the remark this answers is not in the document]
OmnipotentEntity: Did you mean `std::uint8_t` rather than `std::unint8_t**`?

include/nlohmann/json.hpp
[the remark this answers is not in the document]
nlohmann: Thanks for the hint, now I understand.

Then I think it would be nicer to have an API that abstracts away from the simulated optional. As I understand, we would need the following functions:

```cpp
bool has_subtype() const noexcept;
void set_subtype(std::uint8_t subtype) noexcept;
void clear_subtype() noexcept;
std::size_t get_subtype() const noexcept;
```

The latter function could then return `std::size_t(-1)` as value for "no subtype" (just like `std::string::npos`).

What do you think?

include/nlohmann/json.hpp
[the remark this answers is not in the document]
nlohmann: See https://github.com/nlohmann/json/pull/1662#discussion_r407525936.

include/nlohmann/json.hpp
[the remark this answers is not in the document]
OmnipotentEntity: Sounds good to me.  I'll implement that.

test/src/unit-deserialization.cpp
```diff
@@ -76,6 +76,21 @@ struct SaxEventLogger : public nlohmann::json_sax<json>
         return true;
     }
 
+    bool binary(std::vector<std::uint8_t>& val)
+    {
+        std::string binary_contents = "binary(";
+        std::string commaSpace = "";
```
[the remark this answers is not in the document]
OmnipotentEntity: Resolved in 38f8e86

test/src/unit-class_parser.cpp
```diff
@@ -77,6 +77,21 @@ class SaxEventLogger
         return true;
     }
 
+    bool binary(std::vector<std::uint8_t>& val)
+    {
+        std::string binary_contents = "binary(";
+        std::string commaSpace = "";
```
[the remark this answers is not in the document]
OmnipotentEntity: Resolved in 38f8e86

include/nlohmann/json.hpp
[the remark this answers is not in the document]
OmnipotentEntity: Implemented in 38f8e86

include/nlohmann/json.hpp
```diff
@@ -2017,18 +2224,19 @@ class basic_json
     string_t dump(const int indent = -1,
                   const char indent_char = ' ',
                   const bool ensure_ascii = false,
-                  const error_handler_t error_handler = error_handler_t::strict) const
+                  const error_handler_t error_handler = error_handler_t::strict,
+                  const bool serialize_binary = false) const
```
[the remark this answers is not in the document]
OmnipotentEntity: Resolved in d0dbc84

include/nlohmann/json.hpp
```diff
@@ -818,6 +821,85 @@ class basic_json
...
+    struct internal_binary_t : public BinaryType
+    {
+        using BinaryType::BinaryType;
+        internal_binary_t() : BinaryType() {}
+        internal_binary_t(BinaryType const& bint) : BinaryType(bint) {}
+        internal_binary_t(BinaryType&& bint) : BinaryType(std::move(bint)) {}
+
+        // TOOD: If minimum C++ version is ever bumped to C++17, this field
+        // deserves to be a std::optional
+        uint8_t _subtype = 0;
```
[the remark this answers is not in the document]
OmnipotentEntity: Resolved in 38f8e86

include/nlohmann/json.hpp
```diff
@@ -818,6 +821,85 @@ class basic_json
...
+        bool _has_subtype = false;
```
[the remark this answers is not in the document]
OmnipotentEntity: Resolved in 38f8e86

test/src/unit-msgpack.cpp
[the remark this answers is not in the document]
OmnipotentEntity: Assumed you meant `std::uint8_t` resolved in 38f8e86

OmnipotentEntity:
I guess Code Inspector is tired of this pull request lol

include/nlohmann/json.hpp
```diff
@@ -3555,15 +3804,128 @@ class basic_json
...
+    @sa @ref has_subtype() -- returns whether or not the binary value has a
+    subtype
+
+    @since version 3.8.0
+    */
+    void clear_subtype() noexcept
+    {
+        if (is_binary())
+        {
+            m_value.binary->has_subtype = false;
```
nlohmann, line 3893: I would additionally set the subtype back to its initial value `0` to remove some WTF moments in debugging.

OmnipotentEntity: This shouldn't actually matter, as the only way (through this interface) to set the subtype is via the `set_subtype` command.  But I've resolved this in d767c1c7

OmnipotentEntity:
Good morning.  I've squashed the commits.

nlohmann:
Thanks a lot! I am not waiting for AppVeyor and will merge this as soon as all CI is green.

nlohmann:
THANKS!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [x] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Added binary type. #1662 #1129
* * *

mdenna-nviso:
@OmnipotentEntity thanks a lot for this great contribution!

PR #1690, opened 2019-07-29
Add CI build testing JSON as a bazel dependency.

remyabel:
In order to make the Bazel build actually useful, it should build the unit tests (future PR) or test that the library can be used as a dependency in a Bazel project (this PR).

To test it locally:

```shell
mkdir /tmp/build_output
docker run -e USER="1000" -u="1000" -v $PWD:$PWD -v /tmp/build_output:/tmp/build_output -w $PWD l.gcr.io/google/bazel:0.28.1 --output_user_root=/tmp/build_output build --test_output=errors --verbose_failures=true --keep_going -- //...:all
```

Repeating my previous comments:

Note that the WORKSPACE file is required, otherwise Bazel will complain.

My previous approach was to list the include files in CMakeLists.txt then export it to Bazel. However, since we never use those include files anywhere else, I cut out the middleman. ~Developers will still have to remember to update nlohmann_json.bzl if a new header file is added.~ I was wrong about not using recursive glob. Re-reading the docs, it appears that it is talking about the pattern of combining recursive globbing with exclude, not recursive globbing in general.

* * *

## Pull request checklist

## Please don't


WORKSPACE
```diff
@@ -0,0 +1 @@
+workspace(name = "com_github_nlohmann_json")
```
nlohmann: Is this file really required in this repository?

BUILD
```diff
@@ -0,0 +1,13 @@
+package(default_visibility = ["//visibility:public"])
```
nlohmann: Is this file really required in this repository?

.travis.yml
```diff
@@ -320,7 +333,7 @@ script:
   - $CXX --version
 
   # compile and execute unit tests
-  - mkdir -p build && cd build
+  - mkdir -p unit-test-build && cd unit-test-build
```
nlohmann: Why changing the name of the directory here?

README.md
```diff
@@ -148,6 +148,34 @@ endif()
 
 `thirdparty/nlohmann_json` is then a complete copy of this source tree.
 
+### Bazel
+
+In order to use this project in Bazel, the simplest approach is to download the
```
nlohmann: Is such a long explanation required? The other build managers are only described briefly as the packages should follow the package manager's defaults.

WORKSPACE
remyabel: Unfortunately. See this error:

```
ERROR: The 'build' command is only supported from within a workspace (below a directory having a WORKSPACE file).
```

.travis.yml
remyabel: I think because on MacOS (the build which was failing) there was a case sensitivity naming conflict. The file "BUILD" was conflicting with "build". I don't think you can rename `BUILD` but I will look into it.

BUILD
remyabel: I guess it isn't "required" per se. As I explained in the PR synopsis and other comments, the Bazel build is fairly useless in/of itself because there is nothing to "compile" for only headers. Therefore this PR tests that the library can be used as a dependency in a Bazel project. Without this `BUILD` file, the compilation will fail. If you're not a fan of that, then this PR can be closed since it's the crux of it.

WORKSPACE
nlohmann: Then I would rather have the bazel build in a different repository, because this seems rather intrusive. Is this possible?

BUILD
nlohmann: See https://github.com/nlohmann/json/pull/1690#discussion_r310347274

.travis.yml
nlohmann: I see.

WORKSPACE
remyabel: That is probably already covered by the original OP's demo project. Closing this.

PR #1694, opened 2019-07-31
release: add singleinclude and meson.build to include.zip

### LINKED ISSUE #1672: include.zip should contain meson.build
In order to make it easier to use nlohmann_json everywhere, it would be helpful to provide a unified meson interface to the project. Currently, the choices are:
- acquire the entire repository, which is very large, and use the meson.build as a subproject
- use the meson wrapdb to acquire just the header includes. The meson wrapdb is out of date and the wrap itself sucks -- it uses its own meson.build which doesn't include the fix from #1463
It's probably fair to say that relying on some thirdparty provider will not be a great way to stay up to date.
- download include.zip, vendor it into the other project's repository, and manually add the directory to the include paths.

It would be nice to see include.zip more closely resemble the expected use of this repository, except without the large files that aren't needed when simply using it in another project. The simple solution would be to change these lines:
https://github.com/nlohmann/json/blob/cf8251eb544dcaba780b12a1da90afb791d31395/Makefile#L594-L599

And add `$(AMALGAMATED_FILE) meson.build` to the zipped files.

This should not have any practical effect other than to enable first-class meson wrap support. No one currently using include.zip should notice the very small extra meson.build file, and the additional single_include/ file should be negligible. (784K vs. 1.5M vs. 435M)
Sadly, meson does not have a way to check if a directory exists before declaring an include_directories on it...

eli-schwartz:
This serves as a minimal release-only way to embed json into a project. Add meson support to this directly, to make it usable standalone as ameson subproject.

Fixes #1672

## Pull request checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,9 @@ endif()
 
 :beer: If you are using OS X and [Homebrew](http://brew.sh), just type `brew tap nlohmann/json` and `brew install nlohmann-json` and you're set. If you want the bleeding edge rather than the latest release, use `brew install nlohmann-json --HEAD`.
 
-If you are using the [Meson Build System](http://mesonbuild.com), then you can get a wrap file by downloading it from [Meson WrapDB](https://wrapdb.mesonbuild.com/nlohmann_json), or simply use `meson wrap install nlohmann_json`.
+If you are using the [Meson Build System](http://mesonbuild.com), add this source tree as a [meson subproject](https://mesonbuild.com/Subprojects.html#using-a-subproject). You may also use the `include.zip` published in this project's [Releases](https://github.com/nlohmann/json/releases) to reduce the size of the vendored source tree. Alternatively, you can get a wrap file by downloading it from [Meson WrapDB](https://wrapdb.mesonbuild.com/nlohmann_json), or simply use `meson wrap install nlohmann_json`. Please see the meson project for any issues regarding the packaging.
+
+The provided meson.build can also be used as an alternative to cmake for installing `nlohmann_json` system-wide in which case a pkg-config file is installed. To use it, simply have your build system require the `nlohmann_json` pkg-config dependency. In Meson, it is preferred to use the [`dependency()`](https://mesonbuild.com/Reference-manual.html#dependency) object with a subproject fallback, rather than using the subproject directly.
 
 If you are using [Conan](https://www.conan.io/) to manage your dependencies, merely add `jsonformoderncpp/x.y.z@vthiery/stable` to your `conanfile.py`'s requires, where `x.y.z` is the release version you want to use. Please file issues [here](https://github.com/vthiery/conan-jsonformoderncpp/issues) if you experience problems with the packages.
 
--- a/Makefile
+++ b/Makefile
@@ -594,7 +594,7 @@ ChangeLog.md:
 release:
 	rm -fr release_files
 	mkdir release_files
-	zip -9 --recurse-paths -X include.zip $(SRCS)
+	zip -9 --recurse-paths -X include.zip $(SRCS) $(AMALGAMATED_FILE) meson.build
 	gpg --armor --detach-sig include.zip
 	mv include.zip include.zip.asc release_files
 	gpg --armor --detach-sig $(AMALGAMATED_FILE)


eli-schwartz:
Still waiting for a review. ;)

nlohmann:
Before I merge this, I need to understand the role of this change. I am not using Meson myself, so I do not know whether it is actually needed. The README states:

> If you are using the Meson Build System, then you can get a wrap file by downloading it from Meson WrapDB, or simply use meson wrap install nlohmann_json.

Why do we need this change here in this repository?

eli-schwartz:
That is covered by the second point from #1672

nlohmann:
I see. Though it may be unwise to rely on third parties for Meson, I will also not be able to maintain the package. Could you add another sentence to the README describing how to still use the library with Meson if you do not use the provided package?

eli-schwartz:
I added a couple blurbs to the README, including a mention of pkg-config. Since meson's documentation exposes a preference for pkg-config, and the meson documentation on subprojects makes reference to pkg-config dependencies with subproject *fallbacks*, meson users will likely be interested in looking for pkg-config support. So the pkg-config blurb lets them know that pkg-config support can be available, and thus nlohmann_json is eligible to use that dependency fallback stuff.

But from a simple perspective, it is as simple as "use this source tree" or "use the include.zip which is equal to this subtree minus a lot of non-release files".

eli-schwartz:
Okay, force pushed to include Table of Contents links to the new sections I added.

README.md
```diff
@@ -73,6 +75,10 @@ to the files you want to process JSON and set the necessary switches to enable C
 
 You can further use file [`include/nlohmann/json_fwd.hpp`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json_fwd.hpp) for forward-declarations. The installation of json_fwd.hpp (as part of cmake's install step), can be achieved by setting `-DJSON_MultipleHeaders=ON`.
 
+### Pkg-Config
```
nlohmann: Please move this paragraph to the "Meson" entry in the "Package Managers" section.

README.md
```diff
@@ -17,7 +17,9 @@
 
 - [Design goals](#design-goals)
 - [Integration](#integration)
+  - [Pkg-Config](#pkg-config)
```
nlohmann: Please remove.

README.md
```diff
@@ -17,7 +17,9 @@
...
   - [CMake](#cmake)
+  - [Meson](#meson)
```
nlohmann: Please remove.

README.md
eli-schwartz: It should be distinct from meson, because you can use the pkg-config file from a Makefile too, not just from a meson.build file -- but sure, it can be moved to the package managers section.

eli-schwartz:
@nlohmann What happened?

README.md
```diff
@@ -150,9 +153,11 @@ endif()
 
 ### Package Managers
 
+Regardless of the build system you use, when installing this library using the provided meson.build, a pkg-config file is installed. To use it, simply have your build system require the `nlohmann_json` pkg-config dependency.
```
nlohmann: Please move this paragraph to the Meson paragraph.

README.md
nlohmann: This paragraph is empty now, right?

README.md
eli-schwartz: I would just like to reiterate that I believe this to be untrue, since pkg-config has nothing to do with whether your project uses meson (it only depends on whether nlohmann_json was installed using meson).

I believe moving the paragraph will be confusing to non-cmake users, but if you wish I can indeed move the paragraph...

README.md
eli-schwartz: Oops.

README.md
nlohmann: But if the installation of the pkg-config file depends on installing the library with meson, then why should this not be moved below?

eli-schwartz: I'd have liked to separate the concerns of "how do I install nlohmann_json" vs. "how do I use the installed version".

Either way, I've updated the docs as you've requested.

eli-schwartz:
Okay, hopefully this should be good for now.

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- add singleinclude and meson.build to include.zip

* * *

nlohmann:
Thanks!

PR #1701, opened 2019-08-08
cmake: Generate and install a pkg-config file

palmer-dabbelt:
I'm trying to build mtxclient, which expects a nlohmann_json pkg-config
file to exist in the system.  The meson builds install a pkg-config
file, but the cmake builds don't.  This adds a pkg-config file to the
cmake builds that is functionally equivalent to the one generated from
meson.

Signed-off-by: Palmer Dabbelt <palmer@dabbelt.com>


palmer-dabbelt:
@nlohmann is this the right place to send pull requests of this sort?

nlohmann:
Thanks for the PR! The code looks good, but could you please also add a note to the README how to use the library with pkgconfig?

PR #1722, opened 2019-08-23
Fix int64 min issue

### LINKED ISSUE #1708: Error in dump() with int64_t minimum value
There is a segfault when calling `dump()` on an `nlohmann::json` node containing an `int64_t` value set to `std::numeric_limits<int64_t>::min()`. Any other `int64_t` value (for example, the minimum value + 1, or the maximum value) seems to work fine. Other int or double values (`uint64_t`, for example) also seem to have no issues.

#### Please describe the steps to reproduce the issue. Can you provide a small but working code example?
For this example, the the entire nlohmann JSON repo is included in the folder `json`.
main.cpp:
```
#include <stdio.h>
#include <string>
#include <limits>

#include "json/single_include/nlohmann/json.hpp"

int main(int argc, char *argv[])
{
    nlohmann::json node_integer_limits = {{"int64_min", std::numeric_limits<int64_t>::min()}};

    std::string test_comp1 = std::to_string(std::numeric_limits<int64_t>::min());
    std::string test_comp2 = node_integer_limits["int64_min"].dump();
    printf("String difference: %d\n", test_comp1.compare(test_comp2));
    std::string test_node = node_integer_limits.dump();

    printf("Finished!\n");
}
```

Makefile:
```
OBJECTS = main.o

CPPFLAGS := -std=gnu++14 -O2 -g

$(OBJECTS): CPPFLAGS := $(CPPFLAGS)

tests: $(OBJECTS)
    g++ $(CPPFLAGS) -o tests $(OBJECTS)

main.o: main.cpp json/single_include/nlohmann/json.hpp
    g++ $(CPPFLAGS) -c main.cpp

.PHONY: clean
clean:
    rm -rf tests $(OBJECTS)
```

Calling `dump()` returns a string version of the value.

See above. The error code is:
```
Segmentation fault (core dumped)
```

In gdb:
```
(gdb) run
Starting program: /<full path removed>/tests

Program received signal SIGSEGV, Segmentation fault.
nlohmann::detail::serializer<nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer> >::dump_integer<unsigned long, 0> (
    x=<optimized out>, this=<optimized out>) at json/single_include/nlohmann/json.hpp:14306
14306           o->write_characters(number_buffer.data(), n_chars);
(gdb) bt
#0  nlohmann::detail::serializer<nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer> >::dump_integer<unsigned long, 0> (
    x=<optimized out>, this=<optimized out>) at json/single_include/nlohmann/json.hpp:14306
#1  nlohmann::detail::serializer<nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer> >::dump (
    this=this@entry=0x7fffffffde40, val=..., pretty_print=pretty_print@entry=false, ensure_ascii=<optimized out>, indent_step=indent_step@entry=0, current_indent=<optimized out>) at json/single_include/nlohmann/json.hpp:13894
#2  0x0000000000407c14 in nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer>::dump (this=0x61fd20,
    indent=indent@entry=-1, indent_char=indent_char@entry=32 ' ', ensure_ascii=ensure_ascii@entry=false, error_handler=error_handler@entry=nlohmann::detail::error_handler_t::strict) at json/single_include/nlohmann/json.hpp:16390
#3  0x000000000040188f in main (argc=<optimized out>, argv=<optimized out>) at main.cpp:12
```

The expected output is:
```
String difference: 0
Finished!
```

#### Which compiler and operating system are you using? Is it a [supported compiler](https://github.com/nlohmann/json#supported-compilers)?
All of this is on the same Ubuntu 16.04 machine. The issue occurs when compiling g++-5 (`(Ubuntu 5.5.0-12ubuntu1~16.04) 5.5.0 20171010`) for C++11, C++14, or C++17. The issue only occurs when compiling with -O2 or higher. Both g++-6 (`(Ubuntu 6.5.0-2ubuntu1~16.04) 6.5.0 20181026`) and g++-7 (`(Ubuntu 7.4.0-1ubuntu1~16.04~ppa1) 7.4.0`) compile and run with no issues across all optimization levels.

#### Did you use a released version of the library or the version from the `develop` branch?
Tried with the `develop` branch as well as versions 3.6.1 and 3.7.0. The behavior was the same in all cases.

There were no compilation errors. All tests passed.

#### Notes
The issue might be caused by line 14267 of `json.hpp` on the `develop` branch (https://github.com/nlohmann/json/blob/develop/single_include/nlohmann/json.hpp#L14267).

t-b:
Close #1708.

For some gcc version (Ubuntu 5.5.0-12ubuntu1~16.04) the existing code
crashes when the minimum value of int64_t is outputted.

Rewrite the code to be less complicated to avoid it. This partially
reverts what was done in 546e2cb (rotating_light fixed some warnings,
2019-03-13).

It might be the case that this reintroduces the warning removed earlier. If that is the case, I will rewrite the code using templates to conditionally remove the sign for the negative case only.

Do we need static assertions which ensure that every positive number_integer_t fits into a number_unsigned_t?

### DIFF
--- a/include/nlohmann/detail/output/serializer.hpp
+++ b/include/nlohmann/detail/output/serializer.hpp
@@ -630,7 +630,7 @@ class serializer
         if (is_negative)
         {
             *buffer_ptr = '-';
-            abs_value = static_cast<number_unsigned_t>(std::abs(static_cast<std::intmax_t>(x)));
+            abs_value = remove_sign(x);
 
             // account one more byte for the minus sign
             n_chars = 1 + count_digits(abs_value);
@@ -809,6 +809,32 @@ class serializer
 
         state = utf8d[256u + state * 16u + type];
         return state;
+    }
+
+    /*
+     * Overload to make the compiler happy while it is instantiating
+     * dump_integer for number_unsigned_t.
+     * Must never be called.
+     */
+    number_unsigned_t remove_sign(number_unsigned_t x)
+    {
+        assert(false); // LCOV_EXCL_LINE
+        return x; // LCOV_EXCL_LINE
+    }
+
+    /*
+     * Helper function for dump_integer
+     *
+     * This function takes a negative signed integer and returns its absolute
+     * value as unsigned integer. The plus/minus shuffling is necessary as we can
+     * not directly remove the sign of an arbitrary signed integer as the
+     * absolute values of INT_MIN and INT_MAX are usually not the same. See
+     * #1708 for details.
+     */
+    inline number_unsigned_t remove_sign(number_integer_t x) noexcept
+    {
+        assert(x < 0 and x < (std::numeric_limits<number_integer_t>::max)());
+        return static_cast<number_unsigned_t>(-(x + 1)) + 1;
     }
 
   private:
--- a/test/src/unit-serialization.cpp
+++ b/test/src/unit-serialization.cpp
@@ -189,3 +189,20 @@ TEST_CASE("serialization")
         test("[3,\"false\",false]", "[3,\\\"false\\\",false]");
     }
 }
+
+TEST_CASE_TEMPLATE("serialization for extreme integer values", T, int32_t, uint32_t, int64_t, uint64_t)
+{
+    SECTION("minimum")
+    {
+        constexpr auto minimum = (std::numeric_limits<T>::min)();
+        json j = minimum;
+        CHECK(j.dump() == std::to_string(minimum));
+    }
+
+    SECTION("maximum")
+    {
+        constexpr auto maximum = (std::numeric_limits<T>::max)();
+        json j = maximum;
+        CHECK(j.dump() == std::to_string(maximum));
+    }
+}
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -14264,7 +14264,7 @@ class serializer
         if (is_negative)
         {
             *buffer_ptr = '-';
-            abs_value = static_cast<number_unsigned_t>(std::abs(static_cast<std::intmax_t>(x)));
+            abs_value = remove_sign(x);
 
             // account one more byte for the minus sign
             n_chars = 1 + count_digits(abs_value);
@@ -14443,6 +14443,32 @@ class serializer
 
         state = utf8d[256u + state * 16u + type];
         return state;
+    }
+
+    /*
+     * Overload to make the compiler happy while it is instantiating
+     * dump_integer for number_unsigned_t.
+     * Must never be called.
+     */
+    number_unsigned_t remove_sign(number_unsigned_t x)
+    {
+        assert(false); // LCOV_EXCL_LINE
+        return x; // LCOV_EXCL_LINE
+    }
+
+    /*
+     * Helper function for dump_integer
+     *
+     * This function takes a negative signed integer and returns its absolute
+     * value as unsigned integer. The plus/minus shuffling is necessary as we can
+     * not directly remove the sign of an arbitrary signed integer as the
+     * absolute values of INT_MIN and INT_MAX are usually not the same. See
+     * #1708 for details.
+     */
+    inline number_unsigned_t remove_sign(number_integer_t x) noexcept
+    {
+        assert(x < 0 and x < (std::numeric_limits<number_integer_t>::max)());
+        return static_cast<number_unsigned_t>(-(x + 1)) + 1;
     }
 
   private:
--- a/test/src/unit-regression.cpp
+++ b/test/src/unit-regression.cpp
@@ -1802,6 +1802,13 @@ TEST_CASE("regression tests")
         json j = json::parse("[-9223372036854775808]");
         CHECK(j.dump() == "[-9223372036854775808]");
     }
+
+    SECTION("issue #1708 - minimum value of int64_t can be outputted")
+    {
+        constexpr auto smallest = (std::numeric_limits<int64_t>::min)();
+        json j = smallest;
+        CHECK(j.dump() == std::to_string(smallest));
+    }
 }
 
 #if not defined(JSON_NOEXCEPTION)


jaredgrubb:
This patch invokes Undefined Behavior. You aren't allowed to take the negative of "min int". Eg, for Int8, the negative of -128 is +128, which is not representable in Int8. This is UB in C++.

jaredgrubb:
Maybe you could do something like `static_cast<number_unsigned_t>(-(x+1)) + 1 `?

t-b:
@jaredgrubb Thanks for the review.

Good point.

Your suggestion is pretty close to the old solution prior to 546e2cb. I've used that one, but also moved the code into functions to avoid the warning which 546e2cb wanted to fix.

Let's see what CI says.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -811,6 +811,17 @@ class serializer
...
     }
 
+    inline number_unsigned_t remove_sign(number_unsigned_t x) noexcept
+    {
+        return x;
+    }
+
+    inline number_unsigned_t remove_sign(number_integer_t x) noexcept
+    {
+        assert(x < std::numeric_limits<number_integer_t>::max());
```
jaredgrubb: This assert doesn't make sense. Any value of type `number_integer_t` will be smaller than (or equal to) its max.

Even if it's a typo (and should be `number_unsigned_t`), I still am not sure what this is guarding. Are you trying to avoid that `sizeof(number_unsigned_t) < sizeof(number_integer_t)`?

t-b: @jaredgrubb

I'm adding 1 to x in the line below, and that does only work if x < max.

jaredgrubb: Right, but this function only gets called if the value is negative. I think you shouldn't have factored out this as a function because that fact gets lost.

t-b: If I don't factor it out the compilers complain because I'm taking the negative of an unsigned value.

See

```
C:\projects\json\single_include\nlohmann/json.hpp(14268): warning C4146: unary minus operator applied to unsigned type, result still unsigned [C:\projects\json\test\test-algorithms.vcxproj]
  C:\projects\json\single_include\nlohmann/json.hpp(13895): note: see reference to function template instantiation 'void nlohmann::detail::serializer<nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>>::dump_integer<unsigned __int64,0>(NumberType)' being compiled
          with
          [
              NumberType=unsigned __int64
          ]

```

from https://ci.appveyor.com/project/nlohmann/json/builds/26925012/job/5jasuk4xo67e7o39.

I can add another assertion with `x < 0`;

t-b: @jaredgrubb

I've rewritten the code a bit so that it is less confusing.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -811,6 +811,18 @@ class serializer
...
 
+    number_unsigned_t remove_sign(number_unsigned_t x)
+    {
+        assert(false); // LCOV_EXCL_LINE
+        return x; // LCOV_EXCL_LINE
+    }
+
+    inline number_unsigned_t remove_sign(number_integer_t x) noexcept
+    {
+        assert(x < 0 && x < (std::numeric_limits<number_integer_t>::max)());
```
nlohmann: This library uses alternative operator keywords. Please change to:

```suggestion
        assert(x < 0 and x < (std::numeric_limits<number_integer_t>::max)());
```

test/src/unit-serialization.cpp
```diff
@@ -189,3 +189,20 @@ TEST_CASE("serialization")
         test("[3,\"false\",false]", "[3,\\\"false\\\",false]");
     }
 }
+
+TEST_CASE_TEMPLATE("serialization for extreme integer values", T, int32_t, uint32_t, int64_t, uint64_t)
+{
+    SECTION("minimum")
+    {
+        static constexpr auto minimum = (std::numeric_limits<T>::min)();
```
nlohmann: Out of curiosity: why is this variable static?

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -811,6 +811,18 @@ class serializer
         return state;
     }
 
+    number_unsigned_t remove_sign(number_unsigned_t x)
+    {
+        assert(false); // LCOV_EXCL_LINE
+        return x; // LCOV_EXCL_LINE
+    }
+
+    inline number_unsigned_t remove_sign(number_integer_t x) noexcept
```
nlohmann: Could you please add a comment describing the necessity of this function and what the assertion and transformation in the return statement means? Maybe (also) add a link #1708.

include/nlohmann/detail/output/serializer.hpp
t-b: Done

test/src/unit-serialization.cpp
t-b: That is a left over from the first version which was only `static const`. Thanks for spotting. From my understanding there should be no benefit from `static constexpr` compared to `constexpr`.

include/nlohmann/detail/output/serializer.hpp
t-b: Done.

t-b:
@nlohmann Thanks for the review.

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [x] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Fix serialization to work with minimal/maximal values.

* * *

nlohmann:
Thanks a lot!

Issue #1759, opened 2019-09-19
Fixes Cppcheck warnings


Xav83:
Hi @nlohmann and other mainteners,

First of all, I must say that I love your work on this librairy.
It feels really easy to use, and to read code using this library ! ❤️ 

I have created this issue to talk about the warnings that [cppcheck](http://cppcheck.sourceforge.net/) found on this project.
I'm currently experimenting with both this library and cppcheck on some little side projects. And doing so, I've ended up with cppcheck  warnings me about some elements inside the json library.

So the purpose of this issue is to talk about the warnings from cppcheck, to address them when they are relevant, so that other people which are going to have both this library and cppcheck]  in their project, as little as warnings as possible.

Here is the list of the warnings I've got for now:
```
(1)[include\nlohmann\json.hpp:5076]: (style) Variable 'm_value.object->emplace' is reassigned a value before the old one has been used.
(2)[include\nlohmann\detail\conversions\to_chars.hpp:585]: (style) The expression 'kAlpha >= -60' is always true.
(3)[include\nlohmann\detail\conversions\to_chars.hpp:586]: (style) The expression 'kGamma <= -32' is always true.
(4)[include\nlohmann\detail\output\serializer.hpp:726]: (style) The expression 'thousands_sep != '\0'' is always false because 'thousands_sep' and ''\0'' represent the same value.
(5)[include\nlohmann\detail\output\serializer.hpp:736]: (style) The expression 'decimal_point != '\0'' is always false because 'decimal_point' and ''\0'' represent the same value.
(6)[include\nlohmann\detail\input\binary_reader.hpp:650]: (style) Unsigned expression 'mant' can't be negative so it is unnecessary to test it.
(7)[include\nlohmann\json.hpp:837]: (style) Local variable object shadows outer function
(8)[include\nlohmann\detail\input\input_adapters.hpp:339]: (style) Class 'input_adapter' has a constructor with 1 argument that is not explicit.
(9)[include\nlohmann\detail\input\input_adapters.hpp:342]: (style) Class 'input_adapter' has a constructor with 1 argument that is not explicit.
(10)[include\nlohmann\detail\input\input_adapters.hpp:346]: (style) Class 'input_adapter' has a constructor with 1 argument that is not explicit.
(11)[include\nlohmann\detail\input\input_adapters.hpp:349]: (style) Class 'input_adapter' has a constructor with 1 argument that is not explicit.
[...]

PR #1760, opened 2019-09-19
Cppcheck fixes

Xav83:
This Pull Request corrects two of the warnings reported by cppcheck, mentionned in #1759 

A warning about an unnecessary check in the first commit, and a warning about a code enhancement in the second warning, by using the std::accumulate function instead of a raw loop.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -861,13 +861,12 @@ class binary_writer
     */
     static std::size_t calc_bson_array_size(const typename BasicJsonType::array_t& value)
     {
-        std::size_t embedded_document_size = 0ul;
         std::size_t array_index = 0ul;
 
-        for (const auto& el : value)
+        const std::size_t embedded_document_size = std::accumulate(std::begin(value), std::end(value), 0ul, [&array_index](std::size_t result, const typename BasicJsonType::array_t::value_type & el)
         {
-            embedded_document_size += calc_bson_element_size(std::to_string(array_index++), el);
-        }
+            return result + calc_bson_element_size(std::to_string(array_index++), el);
+        });
 
         return sizeof(std::int32_t) + embedded_document_size + 1ul;
     }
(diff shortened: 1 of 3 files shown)


include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -861,13 +861,10 @@ class binary_writer
...
-        std::size_t embedded_document_size = 0ul;
-        std::size_t array_index = 0ul;
-
-        for (const auto& el : value)
+        const std::size_t embedded_document_size = std::accumulate(std::begin(value), std::end(value), [array_index = 0ul](std::size_t result, const auto & el)
         {
-            embedded_document_size += calc_bson_element_size(std::to_string(array_index++), el);
-        }
+            return result + calc_bson_element_size(std::to_string(array_index++), el);
+        });
```
jaredgrubb: Doesn't the lambda need to be mutable (so array_index can mutate)? If so, I'm curious how this passes unit tests; I would hope this would get checked?

Xav83: Indeed I've forgot to make it mutable !
This is weird, I run the commands describe in the [README](https://github.com/nlohmann/json#execute-unit-tests) and everything passed on my machine. At least the CI catched it

Xav83: I may have runned the tests before using the command `make amalgamate`. So this is why the tests passed on my machine before 😂 
I remove the indentifier initializer and the auto keyword in the lambda parameter since those are c++14 features, and if I understand well, this project is in c++11

Xav83:
I don't understand why the [Travis job with Xcode 9.3 fails](https://travis-ci.org/nlohmann/json/jobs/587402575#L2302), but I found a [ninja issue](https://github.com/ninja-build/ninja/issues/1595) similar to the error message.

And errors found by Code Inspector aren't referencing the code I modified. So, should I fix them or should they be fixed in another Pull Request ?

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Fixed some Cppcheck warnings.

* * *

nlohmann:
Thanks a lot!

PR #1765, opened 2019-09-28
Allow items() to be used with custom string

crazyjul:
When using an alternative string, the iteration_proxy_value is not compatible, as it returns const std::string &. I replaced it by the type given string.

I made a function to abstract std::to_string, but I don't really like the way it's done and would like your opinion.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/iterators/iteration_proxy.hpp
+++ b/include/nlohmann/detail/iterators/iteration_proxy.hpp
@@ -12,6 +12,10 @@ namespace nlohmann
 {
 namespace detail
 {
+void int_to_string( std::string& target, int value )
+{
+    target = std::to_string(value);
+}
 template <typename IteratorType> class iteration_proxy_value
 {
   public:
@@ -20,6 +24,7 @@ template <typename IteratorType> class iteration_proxy_value
     using pointer = value_type * ;
     using reference = value_type & ;
     using iterator_category = std::input_iterator_tag;
+    using string_type = typename std::remove_cv< typename std::remove_reference<decltype( std::declval<IteratorType>().key() ) >::type >::type;
 
   private:
     /// the iterator
@@ -29,9 +34,9 @@ template <typename IteratorType> class iteration_proxy_value
     /// last stringified array index
     mutable std::size_t array_index_last = 0;
     /// a string representation of the array index
-    mutable std::string array_index_str = "0";
+    mutable string_type array_index_str = "0";
     /// an empty string (to return a reference for primitive values)
-    const std::string empty_str = "";
+    const string_type empty_str = "";
 
   public:
     explicit iteration_proxy_value(IteratorType it) noexcept : anchor(it) {}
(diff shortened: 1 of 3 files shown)


jaredgrubb:
FWIW, looks good to me! Thanks for the tests too :)

t-b:
I dont' think the introduction of a custom `int_to_string` function is a good solution. Why not call it `to_string` and let the compiler choose the correct implementation from either the one in `std` namespace or a user implementation in its own namespace found via ADL? 

https://stackoverflow.com/a/50555639 has an example.

crazyjul:
It's not exactly the same here. The example shown is about creating a to_string for other types. Here, I need a to_string that returns other type of string, the alt string.

There's no easy way to select function on return value in c++. It tried something like this : 

``` 
using namespace std;
(static_cast<string_type ( int ) >( to_string ))( array_index )
```

But it doesn't find the to_string for alt_string, as the iterator is in a detail namespace.

If you have ideas to fix this, I'm open

jaredgrubb:
As @crazyjul says, ADL only kicks on _argument types_, not on _return types_ as you suggest. Therefore the custom type has to be one of the args. I don't see any other way to do it than to make it `void bikeshed_name(custom_string_type&, int)`. I think the proposed patch is the right solution.

t-b:
@crazyjul @jaredgrubb Thanks for the answers. Yes indeed ADL does not help here.

But still forcing the user to implement this function does not sound like the right approach. But I'm also running out of ideas how to do it better.

crazyjul:
There two other solutions : 
- force the alt_string to have a copy constructor / operator to std::string
- go to char * with `std::to_string( row ).c_str()`
Both those solution would add a memory allocation, but would remove the need for that extra function.

There's other solutions requiring some templates, that would allow user to choose that default mode, but be able to override with a function of his own. I can make some proposition, but I need to experiment first

t-b:
@crazyjul Let's hear first what @nlohmann thinks of the current state before you start putting more effort into it.

jaredgrubb:
What if you do:
```
template <typename T>
void int_to_string(T& str, int value) {
    str = std::to_string(value);
}

// ...

      if (array_index != array_index_last)
      {
              using detail::int_to_string;
              int_to_string( array_index_str, array_index );
              array_index_last = array_index;
      }
```
That provides a default implementation that only requires that the custom string type is convertible from `std::string`, but also provides a ADL-searchable customization point.

crazyjul:
Done ;-)

nlohmann:
Thanks!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [x] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- `items()` function can be used with custom string types.

* * *

PR #1779, opened 2019-10-07
test/CMakeLists.txt: Use an explicit list instead of GLOB

### LINKED ISSUE #1686: Consider listing files explicitly instead of using GLOB
From the [docs](https://cmake.org/cmake/help/latest/command/file.html):

> Note: We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild. 

The only small downside is that developers will have to remember to add any new files to the sources/headers list.

t-b:
Using GLOB is slow and considered bad practice.

From https://cmake.org/cmake/help/latest/command/file.html:

> We do not recommend using GLOB to collect a list of source files from
> your source tree. If no CMakeLists.txt file changes when a source is
> added or removed then the generated build system cannot know when to ask
> CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on
> all generators, or if a new generator is added in the future that cannot
> support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS
> works reliably, there is still a cost to perform the check on every
> rebuild.

Close #1686.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -101,7 +101,49 @@ endif()
 # one executable for each unit test file
 #############################################################################
 
-file(GLOB files "src/unit-*.cpp")
+set(files
+    src/unit-algorithms.cpp
+    src/unit-allocator.cpp
+    src/unit-alt-string.cpp
+    src/unit-bson.cpp
+    src/unit-capacity.cpp
+    src/unit-cbor.cpp
+    src/unit-class_const_iterator.cpp
+    src/unit-class_iterator.cpp
+    src/unit-class_lexer.cpp
+    src/unit-class_parser.cpp
+    src/unit-comparison.cpp
+    src/unit-concepts.cpp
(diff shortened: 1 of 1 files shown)


nlohmann:
Thansk!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Avoid usage of `GLOB` in CMake

* * *

Issue #1798, opened 2019-10-16
Always compile tests with all warnings enabled and error out on warnings


t-b:
I think we should patch test/CMakeLists.txt to always use `-Wall -Werror` on *nix and the corresponding MSVC flags. It is way too easy to overlook a warning in the CI output as #1797 has shown.

@nlohmann Thoughts?

nlohmann:
I compile the code with maximal warnings when I prepare a release. For CI, such a step would be nice - PRs welcome!

amiremohamadi:
@nlohmann 
I'll be happy if I can work on it.
I've a question! Should only related flags be added to the `test/CMakeLists.txt`?

t-b:
@amiremohamadi I'm not sure what you mean by "related flags" but I would start with `-Wall -Werror` for gcc/clang and `/W4 /WX` MSVC. Adding it to test/CmakeLists.txt makes most sense.

Mike-Bal:
In principle it is a nice goal, but it's always hard to retrofit `-Werror` / `/WX` into a project that runs CI on a wide variety of compilers and hasn't used those flags before. 

Often you don't have all compilers locally and have to debug through CI cycles and in some cases it is also not clear how to best fix a warning. Particularly problematic: If someone wants to add a new compiler to the CI, is it that person's responsibility to fix all new warnings that that compiler spits out before the PR can be merged?

Maybe a less ambitious first step would be to ensure that the user-visible headers don't produce any warnings (e.g. create a separate test that includes the headers and not much more and compile that single test with `-Werror` / `/WX`. From that, iteratively add the flag to more and more tests. In total that is more work, but has the advantage that you can merge intermediate steps and don't get regressions from new pull requests on the way.

nlohmann:
I can understand your point, but so far, any warning that did come up eventually lead to an issue here. And if we would be able to detect such a warning in the CI, we have the opportunity to fix it immediately rather than with a delay. That said, I'd rather have such an ambitious step and deal with the problems later than to start every second release notes with notes on fixed warnings which slipped through the release process as, for instance, I do not use MSVC myself. Does this make sense?

Mike-Bal:
Sure, I haven't actually looked into how many warnings there are currently generated - I just saw that a lot of CI tests started to fail and unless I'm missing something, there is at least one false positive on Appveyor about an unused parameter that is actually used (https://ci.appveyor.com/project/nlohmann/json/builds/28915988/job/jsk75q0pso682j5a). 

[...]

PR #1804, opened 2019-10-20
force to always compile with -Wall and -Werror in gcc/clang (and /W4,…

amiremohamadi:
Refer to #1798 some changes on `test/CMakeLists.txt` to always use -Wall, -Werror on GCC/Clang and the /WX, /W4 flags on MSVC.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -79,13 +79,16 @@ endif()
 target_include_directories(doctest_main PRIVATE "thirdparty/doctest")
 
 # https://stackoverflow.com/questions/2368811/how-to-set-warning-level-in-cmake
-if(MSVC)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     # Force to always compile with W4
     if(CMAKE_CXX_FLAGS MATCHES "/W[0-4]")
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -79,22 +79,31 @@ endif()
...
+    # Force to always compile with WX
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /WX")
+
     # Disable warning C4389: '==': signed/unsigned mismatch
 	# Disable warning C4309: 'static_cast': truncation of constant value
 	# Disable warning C4566: character represented by universal-character-name '\uFF01' cannot be represented in the current code page (1252)
 	# Disable warning C4996: 'nlohmann::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::adl_serializer>::operator <<': was declared deprecated
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4389 /wd4309 /wd4566 /wd4996")
 
 	# https://github.com/nlohmann/json/issues/1114
```
jaredgrubb: This comment is duplicated (and not in the right place here).

PR #1810, opened 2019-10-23
appveyor.yml: Compile and test with latest version for __cplusplus macro on MSVC

### LINKED ISSUE #1695: /Zc:__cplusplus leads to C2416
OS: Windows 7
Compiler: MSVC 2017 (15.9.11)

If project uses `/Zc:__cplusplus` json lib v3.7.0 fails to compile with this:
```
C2416: attribute 'deprecated' cannot be applied in this context
```

Looks like MSVC doesn't like this attribute applied to operator>> definition.

t-b:
According to [1] the C++ macro __cplusplus has always the value 199711L
even if you enable later version of C++ support via /std.

Use one job with /Zc:__cplusplus defined so that we know that it
compiles.

[1]: https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=vs-2019

Close #1695.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -85,7 +85,7 @@ environment:
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
       configuration: Release
       platform: x64
-      CXX_FLAGS: "/permissive- /std:c++latest /utf-8 /F4000000"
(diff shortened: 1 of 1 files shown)


t-b:
@nlohmann Done.

t-b:
@nlohmann See https://github.com/nlohmann/json/issues/1695#issuecomment-549056685 for a minimal example showing the problem. I'll will not find the time to fix that soonish.

PR #1846, opened 2019-11-19
Fix msvc warnings

Mike-Bal:
This PR:

1.  Changes the accumulation variable for size computations in binar_writer from unsiged long to size_t [*]
2. Remove two warning suppressions in msvc for warnings that seem to no longer appear anyway (waiting for CI results to verify [EDIT: Can confirm no new warnings appeared in CI])

 *) In the x64 platofrm on windows `std::size_t` (64 bit unsigned integer) is not the same as an unsigned long (32bit unsigned integer on all platforms targets). This leads to a (correct) warning about possible loss of data. I'm not sure if there is an actual situation, where this variable can overflow, but the warning itself needs to be fixed anyway and I believe the intent was to use a variable of type `std::size_t` (which is the same as unsigned long on almost all other platforms) anyway.

[EDIT: ] If preferred, I can split up the PR into one for 1. and 2.

* * *

## Pull request checklist

### DIFF
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -863,7 +863,7 @@ class binary_writer
     {
         std::size_t array_index = 0ul;
 
-        const std::size_t embedded_document_size = std::accumulate(std::begin(value), std::end(value), 0ul, [&array_index](std::size_t result, const typename BasicJsonType::array_t::value_type & el)
+        const std::size_t embedded_document_size = std::accumulate(std::begin(value), std::end(value), std::size_t(0), [&array_index](std::size_t result, const typename BasicJsonType::array_t::value_type & el)
         {
             return result + calc_bson_element_size(std::to_string(array_index++), el);
         });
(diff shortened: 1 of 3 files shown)


nlohmann:
Thanks!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [x] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- An MSVC warning has been fixed which no longer needs to be suppressed when compiling the test suite. #1846

* * *

Mike-Bal:
Thanks for the quick merge. 

Just a comment on the release item description: The warning that was fixed and the warnings that got suppressed where independent of each other i.e. the warnings about possible loss of data popped up even with the suppression still active and removing the suppression would not pop up any new warnings either way. As the suppressions only existed in unit tests, I don't think they need to be mentioned in the release item description.

Sorry for putting them into one PR.

nlohmann:
I see. What would you propose as alternative items for the release notes?

Mike-Bal:
Would have to look at your previous notes for reference (i.e. not sure, what the typical level of detail is).

Probably something like either "Fix warning about possible loss of data on windows x64" or "Fix integer type mismatch in `binary_writer` size computation on windows x64"

Issue #1749, opened 2019-09-12
[C++17] Allow std::optional to convert to nlohmann::json


Treeston:
This is what I assume to be a fairly minor change to the library's logic, but would represent a large improvement in usability.

## Feature request

C++17 adds `std::optional<T>`. Let `std::optional<T>` convert to `nlohmann::json`.

If `std::optional<T>` is empty, I would expect this to result in `null` - if `std::optional<T>`, I would expect it to result in the contained value converted.

## Use case example

The change would simplify (especially nested) uses of `std::optional`.

For example, imagine a simple key-value map, where each integer key can optionally have a localized string representation for each of 8 locales. The client dynamically selects the "best" available locale depending on user preferences.

In server code, one could express this as a `std::vector<std::array<std::optional<std::string>, 8>>`.
Currently, this type does not convert to `nlohmann::json` because `std::optional<std::string>` does not - resulting in needing to allocate a bunch of temporaries so the inner conversions can be done "manually".

Allowing `std::optional` to convert to `nlohmann::json` would simplify this logic greatly.

jaredgrubb:
It's a neat idea, though this library is still intended as a C++11 library. I wonder at what point that could change to allow things like this?

nlohmann:
The library defines a macro `JSON_HAS_CPP_17` to detect C++17. With that, the feature could be realized. The required code should look similar to #760.

jnhyatt:
In addition, I believe the library would benefit from the ability to return a `std::optional` from a getter method -- this way, client programs could avoid having to catch exceptions to detect a missing value, etc.:
```c++
std::string name = j["name"].get_optional<std::string>().value_or("default-name");
```
Qt uses something similar in their config file interface. This makes it very easy to save a default json file on a load failure -- you deserialize a class from json with default values using `std::optional` and then serialize the resulting class back to json; if there were no changes to the class, the json should be identical, otherwise missing data will be added to the output json.

nlohmann:
Wouldn't this be similar to [`j.value("name", "default-name")`](https://nlohmann.github.io/json/classnlohmann_1_1basic__json_adcf8ca5079f5db993820bf50036bf45d.html#adcf8ca5079f5db993820bf50036bf45d)?

jnhyatt:
Yes, it would be very similar in most cases. I admit I hadn't searched the library exhaustively for such a feature before having made this suggestion.

djcb:
[...]

PR #1872, opened 2019-12-17
Use JSON_HAS_CPP_17 only after it has been defined

remedi:
Fix for Visual Studio "error C2039: 'optional': is not a member of 'std'" #1749

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -13,14 +13,6 @@
 #include <utility> // pair, declval
 #include <valarray> // valarray
 
-#ifdef JSON_HAS_CPP_17
-    #if __has_include(<optional>)
-        #include <optional>
-    #elif __has_include(<experimental/optional>)
-        #include <experimental/optional>
-    #endif
-#endif
-
 #include <nlohmann/detail/exceptions.hpp>
 #include <nlohmann/detail/macro_scope.hpp>
 #include <nlohmann/detail/meta/cpp_future.hpp>
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -10,14 +10,6 @@
 #include <valarray> // valarray
 #include <vector> // vector
 
-#ifdef JSON_HAS_CPP_17
-    #if __has_include(<optional>)
-        #include <optional>
-    #elif __has_include(<experimental/optional>)
-        #include <experimental/optional>
-    #endif
-#endif
-
 #include <nlohmann/detail/iterators/iteration_proxy.hpp>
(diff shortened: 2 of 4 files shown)


remedi:
I seem to have messed up the amalgamation. I will try to figure out how the files are combined so that the JSON_HAS_CPP_17 is used only after language standard detection in single_include json.hpp.

remedi:
Okay now the amalgamation issue is fixed. I am not sure if this a proper place for including, but I moved the #include <optional> to macro_scope.hpp just after the language standard detection.

nlohmann:
Let's wait for the CI to become green.

remedi:
Sorry for re-triggering the CI. Fixed committer email.

nlohmann:
There still seems to be an issue in the latest AppVeyor builds.

remedi:
Yes it is the conversion issue that I mentioned in #1749.

> However, now there is another error in unit-conversions.cpp lines 1624 and 1633 where there is
> no suitable conversion from nlohmann::json to std::vector containing std::optional (1624) or
> std::map containing std::optional (1633). I am not familiar with the code base so I am unsure
> where exactly the conversion should happen.

I can try looking into this myself today, but I am not sure if I can find the correct place. What I think is missing is implicit conversion from `nlohmann::json` to a type that can be passed as an argument to `std::vector<std::optional>` and `std::map<std::string, std::optional>` constructors.

The compiler errors are:
`C:\projects\json\test\src\unit-conversions.cpp(1624): error C2440: '<function-style-cast>': cannot convert from 'nlohmann::json' to 'std::vector<std::optional<int>,std::allocator<_Ty>>'`
and
`C:\projects\json\test\src\unit-conversions.cpp(1633): error C2440: '<function-style-cast>': cannot convert from 'nlohmann::json' to 'std::map<std::string,std::optional<int>,std::less<_Kty>,std::allocator<std::pair<const _Kty,_Ty>>>'`

I was kind of hoping that you would immediately know what would be required.

nlohmann:
I see. I did not have a look at the error message, but just hoped that the PR would fix them ;-)

I have no idea what the issue is. I am not expert with MSVC, and since GCC and Clang compile the code, it seems OK. However, I think merging your changes into the feature branch is a good idea anyway since it fixes the issue with the macro definition.

remedi:
Thanks for merging the PR. If you mean that Travis compiles the code on GCC and Clang you would be mistaken since looking at the build logs only 2 test cases (270 assertions) are being run for test-conversions.exe for all builds regardless of the compiler or options used.

When JSON_HAS_CPP_17 is defined there should be 3 test cases (298 assertions). I think there is something wrong with the language standard detection for GCC and Clang so JSON_HAS_CPP_17 is always undefined therefore the problematic unit test code does not seem to be compiled on GCC or Clang at all.

PR #1876, opened 2019-12-20
Ignores some cppcheck warnings

Xav83:
This Pull Request adds comments made so that, when using `inline-suppr` option, cppcheck will ignore the warnings about not using explicit keyword when a constructor has one parameter.

It addresses some of the warnings in #1759 

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/input/input_adapters.hpp
+++ b/include/nlohmann/detail/input/input_adapters.hpp
@@ -336,22 +336,28 @@ class input_adapter
   public:
     // native support
     JSON_HEDLEY_NON_NULL(2)
+    // cppcheck-suppress noExplicitConstructor
     input_adapter(std::FILE* file)
         : ia(std::make_shared<file_input_adapter>(file)) {}
     /// input adapter for input stream
+    // cppcheck-suppress noExplicitConstructor
     input_adapter(std::istream& i)
         : ia(std::make_shared<input_stream_adapter>(i)) {}
 
     /// input adapter for input stream
+    // cppcheck-suppress noExplicitConstructor
     input_adapter(std::istream&& i)
         : ia(std::make_shared<input_stream_adapter>(i)) {}
 
+    // cppcheck-suppress noExplicitConstructor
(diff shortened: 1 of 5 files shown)


Xav83:
Hi @nlohmann 😃 

First of all, thanks for your interest in this Pull Request. 👍 

I agree with you that this is not something really elegant.
But if someone uses cppcheck on a project including the json library, he should not have those warnings too (which is my personal objective, with the issue #1759).

There are other ways to suppress warnings in cppcheck, detailed in the [manual](http://cppcheck.sourceforge.net/manual.pdf).

We can suppress each warnings in the command line and given this line to the cppcheck users.
But the more warnings will be ignored, the less this command line will be easy to read, or to integrate in a project using the json library and cppcheck.

We can create a text file, or a xml file in which the ignored warning will be detailed and given to cppchek, so that it won't trigger them. If this solution is used, then, it should be well described in the README, to make it easy for the cppcheck users to integrate this file in their workflow(since they can integrate it from conan, or as a cmake external project, for example).

I would happily modify this PR to use another solution. 😃 
But I will need you, @nlohmann , or someone else to validate which solution to integrate.

Thank you in advance 🙂

trraney328:
@Xav83 I went ahead and reviewed it

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -374,6 +380,7 @@ class input_adapter
                  std::is_integral<typename std::remove_pointer<CharT>::type>::value and
                  sizeof(typename std::remove_pointer<CharT>::type) == 1,
                  int>::type = 0>
+    // cppcheck-suppress noExplicitConstructor
     input_adapter(CharT b)
         : input_adapter(reinterpret_cast<const char*>(b),
                         std::strlen(reinterpret_cast<const char*>(b))) {}
```
trraney328, line 386: this is the reason why

trraney328 (commented):
you may have to fix line 386

Issue #1939, opened 2020-02-12
Compile warning on architectures that are not x86


tasleson:
- What is the issue you have?

Compiling library results in a compiler ~error~ warning which using the latest compilers on non-x86 architectures using the gcc compiler included in fedora 32 aka. rawhide.

- Please describe the steps to reproduce the issue. Can you provide a small but working code example?

Simply try to compile the library.

- What is the expected behavior?

Compile success

- And what is the actual behavior instead?

Compile ~error~ warning
```
./json.hpp:22700:42:   required from here
./json.hpp:8494:24: error: comparison is always true due to limited range of data type [-Werror=type-limits]
 8494 |             if ('\x00' <= c and c <= '\x1F')
      |                 ~~~~~~~^~~~
```

- Which compiler and operating system are you using? Is it a [supported compiler](https://github.com/nlohmann/json#supported-compilers)?

Not listed on the supported compilers.  Fedora rawhide, which I believe is using GCC 10.0.1.

- Did you use a released version of the library or the version from the `develop` branch?

Used latest release.

- If you experience a compilation error: can you [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests)?

I can for x86, but not easily for other architectures at the moment.

nlohmann:
The example is a warning that only becomes in array with `-Werror`. Does the code compile if you do not treat the warning as error?

nlohmann:
(I am also not sure all non-x86 machines have this error, because we can compile successfully on ARM, see https://github.com/nlohmann/json#supported-compilers)

nlohmann:
Thanks for checking back. I'll have a look at #1940.

nlohmann:
I tried to reproduce the problem locally and compiled the library successfully with

- `-funsigned-char -Werror=type-limits`
- `-fsigned-char -Werror=type-limits`

I got no error in either setting. Any ideas?

tasleson:
Here is the actual builds that failed: https://koji.fedoraproject.org/koji/taskinfo?taskID=41468709

Details of the command line option for the failure, snippet from [here](https://kojipkgs.fedoraproject.org//work/tasks/8741/41468741/build.log)
```
[...]

PR #1940, opened 2020-02-12
Specify signed char

tasleson:
Addresses https://github.com/nlohmann/json/issues/1939

x86 architecture default to signed characters.  Other architectures
default to unsigned characters.  When building on non-x86 arch we
get the following error

```
./json.hpp:22700:42:   required from here
./json.hpp:8494:24: error: comparison is always true due to limited
                           range of data type [-Werror=type-limits]
 8494 |             if ('\x00' <= c and c <= '\x1F')
      |                 ~~~~~~~^~~~
```
Signed-off-by: Tony Asleson <tasleson@redhat.com>

### DIFF
--- a/include/nlohmann/detail/input/lexer.hpp
+++ b/include/nlohmann/detail/input/lexer.hpp
@@ -1492,7 +1492,7 @@ scan_number_done:
     position_t position {};
 
     /// raw input token string (for error messages)
-    std::vector<char> token_string {};
+    std::vector<signed char> token_string {};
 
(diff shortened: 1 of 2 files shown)


jaredgrubb:
Hmm, not sure if that's the best fix. Maybe change the condition to `if (static_cast<unsigned char>(c) <= '\x1F')`?

nlohmann:
We may use SFINAE to decide between a `signed char` and an `unsigned char` version. Changing everything internally to `signed char` seems odd.

Issue #1911, opened 2020-01-21
Compile warnings on MSVC 14.2


paamand:
After upgrading from Visual Studio 17 to 19 i get the following compile warnings:

`Warning	C26451	Arithmetic overflow: Using operator '-' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '-' to avoid overflow (io.2).`
**.\nlohmann\detail\conversions\to_chars.hpp	993**
(also line 997, 1007, 1009, 1017, 1021, 1036 and 1038)

`Warning	C26444	Avoid unnamed objects with custom construction and destruction (es.84).`
**.\nlohmann\detail\input\json_sax.hpp	443**

`Warning	C28020	The expression '0<=_Param_(1)&&_Param_(1)<=400-1' is not true at this call.`
**.\nlohmann\detail\output\serializer.hpp	804**

`Warning	C26451	Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2).`
**.\nlohmann\detail\output\serializer.hpp	810**

`Warning	C26495	Variable 'nlohmann::basic_json<std::map,std::vector,std::basic_string<char,std::char_traits<char>,std::allocator<char> >,bool,__int64,unsigned __int64,double,std::allocator,nlohmann::adl_serializer>::m_type' is uninitialized. Always initialize a member variable (type.6).`
**.\nlohmann\json.hpp	1777**

dota17:
Bad bot

dota17:
1. C26495 in json.hpp
> Warning C26495 Variable 'nlohmann::basic_json<std::map,std::vector,std::basic_string<char,std::char_traits<char>,std::allocator<char> >,bool,__int64,unsigned __int64,double,std::allocator,nlohmann::adl_serializer>::m_type' is uninitialized. Always initialize a member variable (type.6).
.\nlohmann\json.hpp 1777

https://github.com/nlohmann/json/blob/973c52dd4a9e92ede5ca4afe8b3d01ef677dc57f/include/nlohmann/json.hpp#L6474
Unfortunately, I didn't got this warning in my VS2019. And i think the value `m_type` has been initialized.

2. C26444 in json_sax.hpp
> Warning C26444 Avoid unnamed objects with custom construction and destruction (es.84).
.\nlohmann\detail\input\json_sax.hpp 443

I think this is a C26444 false positive.

[...]

PR #1967, opened 2020-02-29
Fix C26451 warnnings in to_chars.hpp

dota17:
Refer to #1911 

> Warning C26451 Arithmetic overflow: Using operator '-' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '-' to avoid overflow (io.2).
.\nlohmann\detail\conversions\to_chars.hpp 993
(also line 997, 1007, 1009, 1017, 1021, 1036 and 1038

In `to_chars.hpp`, there were asserts before operator `-`. 
This PR will disable warning C26451.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/conversions/to_chars.hpp
+++ b/include/nlohmann/detail/conversions/to_chars.hpp
@@ -990,11 +990,11 @@ inline char* format_buffer(char* buf, int len, int decimal_exponent,
         // digits[000]
         // len <= max_exp + 2
 
-        std::memset(buf + k, '0', static_cast<size_t>(n - k));
+        std::memset(buf + k, '0', static_cast<size_t>(n) - static_cast<size_t>(k));
         // Make it look like a floating-point number (#362, #378)
         buf[n + 0] = '.';
         buf[n + 1] = '0';
-        return buf + (n + 2);
+        return buf + (static_cast<size_t>(n) + 2);
     }
 
     if (0 < n and n <= max_exp)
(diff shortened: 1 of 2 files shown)


nlohmann:
Is there now way to fix the warning?

dota17:
> Honestly: Can't we just use the original code and suppress the warning from the msvc analyzer? Those casts add absolutely nothing in terms of correctness, readability or safety to the code.

> What MSVC complains about is that computation is done in 32 bits first and then the result is widened to 64bits (a.k.a. std::size_t). Such a warning makes sense in circumstances where people expect the result of a computation might need more than 32 bits and thus assign the result to a 64 bit variable, but forget that the computation itself (and resulting overflow) still happens in 32 bits.

From https://github.com/nlohmann/json/pull/1969#issuecomment-601123017, i agree with @MBalszun .
There were asserts before operator `-` so that we can makesure that  the computation itself (and resulting overflow) still happens in 32 bits.

But if you dont want this, i will take a new version out.

MBalszun:
Assuming this warning needs to be tackled at all and not suppressed, wouldn't it be easiest to just turn k and n into variables of `std::size_t` or `std::ptrdiff_t` if they need to be signed instead of casting them to those types all over the code?

dota17:
> Assuming this warning needs to be tackled at all and not suppressed, wouldn't it be easiest to just turn k and n into variables of `std::size_t` or `std::ptrdiff_t` if they need to be signed instead of casting them to those types all over the code?

Good idea.
I will try and test.

dota17:
There are some errors in CI. And i tried to fix it.
However, the errors in #2004, #2002 are same as this issue, especially #2004 just add a yml file and don't change any code.

dota17:
Ready to go.

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Fix a compiler warning in MSVC. #1967 #1911

* * *

nlohmann:
Thanks!

Issue #859, opened 2017-12-05
code analysis: potential bug?


aj664:
warning C28020: The expression '0<=_Param_(1)&&_Param_(1)<=64-1' is not true at this call.

from the your  .h file:  the line the above comment is pointing at is the  number_buffer[i++  line.

            // make sure the number has been processed completely
            assert(x == 0);

            if (is_negative)
            {
                // make sure there is capacity for the '-'
                assert(i < number_buffer.size() - 2);
                number_buffer[i++] = '-';
            }

this code analysis tool gives off some rather strange false positives, so it might be nothing..

nlohmann:
The function dumps an integer of type `int64_t` or `uint64_t` to a string. The largest value for `uint64_t` is `18446744073709551615`(20 chars), the smallest value for `int64_t` is `-9223372036854775808` (also 20 chars), and the largest value for `int64_t` is `9223372036854775807` (19 chars). So after adding a null byte, we need at most 21 characters for `number_buffer` which has a size of 64 chars.

I think this is a false positive.

PR #1969, opened 2020-03-03
fix warnings in serializer.hpp for VS2019

dota17:
Refer to #1911 
fix warnings in serializer.hpp for VS2019

### DIFF
--- a/include/nlohmann/detail/output/serializer.hpp
+++ b/include/nlohmann/detail/output/serializer.hpp
@@ -807,7 +807,9 @@ class serializer
                 ? (byte & 0x3fu) | (codep << 6u)
                 : (0xFFu >> type) & (byte);
 
-        state = utf8d[256u + state * 16u + type];
+        std::size_t index = 256u + static_cast<size_t>(state) * 16u + static_cast<size_t>(type);
+        assert(0 <= index and index < 400);
+        state = utf8d[index];
         return state;
     }
 
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -14552,7 +14552,9 @@ class serializer
                 ? (byte & 0x3fu) | (codep << 6u)
                 : (0xFFu >> type) & (byte);
 
-        state = utf8d[256u + state * 16u + type];
+        std::size_t index = 256u + static_cast<size_t>(state) * 16u + static_cast<size_t>(type);
+        assert(0 <= index and index < 400);
+        state = utf8d[index];
         return state;
     }
 


include/nlohmann/detail/output/serializer.hpp
```diff
@@ -801,13 +801,14 @@ class serializer
...
 
+        assert(0 <= byte and byte < 400);
         const std::uint8_t type = utf8d[byte];
 
         codep = (state != UTF8_ACCEPT)
                 ? (byte & 0x3fu) | (codep << 6u)
                 : (0xFFu >> type) & (byte);
 
-        state = utf8d[256u + state * 16u + type];
+        state = utf8d[256 + (std::uint64_t)state * 16 + (std::uint64_t)type];
```
MBalszun: Casting those values to 64 bit unsigned ints to index a 400 element std::array doesn't make much sense for 32 bit platforms and will almost certainly trigger a narrowing warning there. 

Also, why are you turning the integral constants into signed integers (just curious)

nlohmann: This is indeed really weird. Could you comment on this?

dota17: @MBalszun , I will check it for 32 bit paltforms. I just forgot the 32 bit platforms, which i use rarely.

MBalszun: I guess the "right" thing would be to cast the values to std::size_t, although I find it somewhat annoying that perfectly fine, safe and readable code needs to be uglified to such a degree (and potentially even become less efficient) just to shut up a compiler warning (or is it a static analysis warning?) .

dota17: @MBalszun  Had been changed to `std::size_t`

Mike-Devel: Just to be on the safe side, you probably also want the litterals to remain unsigned (i.e. `256u` instead of `256`) It shouldn't make any difference, but I wouldn't be suprised if some compiler/analysis tool would start complaining about the mix of signed and unsigned types.

MBalszun: Just to be on the safe side, you probably want to keep the litterals unsigned too (i.e. `256u` instead of `256`) it should not make any difference, but I wouldn't be suprised if some compiler or analysis tool complains about mixing signed and unsigned types.

dota17: Just to shut up a compiler warning for a single line is not the best way, i think.
In some company, compiler warnings must be solved, so  i understand why people figure these warnings out. They have no choice.

MBalszun: Of course the next issue is that a static analysis tool that checks on core-guideline violations will complain about the c-style cast. This is going to be a neverending story. :(
@nlohmann:  Whats this project's policy about c-style casts?

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -801,13 +801,14 @@ class serializer
...
 
+        assert(0 <= byte and byte < 400);
         const std::uint8_t type = utf8d[byte];
 
         codep = (state != UTF8_ACCEPT)
                 ? (byte & 0x3fu) | (codep << 6u)
                 : (0xFFu >> type) & (byte);
 
-        state = utf8d[256u + state * 16u + type];
+        state = utf8d[256 + (std::size_t)state * 16 + (std::size_t)type];
```
nlohmann: Please use `static_cast`.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -801,13 +801,14 @@ class serializer
             }
         };
 
+        assert(0 <= byte and byte < 400);
```
nlohmann: As `byte` is unsigned, the check whether it is `>= 0` is redundant, right?

MBalszun:
> As byte is unsigned, the check whether it is >= 0 is redundant, right?

@nlohmann , @dota17: 

`byte < 400` is redundant too, because a `std::uint8_t` can be at most 255.

If anything, the expression `256 + (std::size_t)state * 16u + (std::size_t)type` should be bounds checked. Honestly: Can't we just use the original code and suppress the warning from the msvc analyzer? Those casts add absolutely nothing in terms of correctness, readability or safety to the code. 

What MSVC complains about is that computation is done in 32 bits first and then the result is widened to 64bits (a.k.a. std::size_t). Such a warning makes sense in circumstances where people expect the result of a  computation might need more than 32 bits and thus assign the result to a 64 bit variable, but forget that the computation itself (and resulting overflow) still happens in 32 bits. 

But here the result must be less than 400 (the size of the array) anyway and it doesn't matter if computation happens in 32 or 64 bits. At most, an argument can be made that computation has to happen in something bigger than 8 bits, but this is anyway happening as per the language rules no matter the input types (any operator if a small integral type gets promoted to int or unsigned int).

dota17:
> If anything, the expression 256 + (std::size_t)state * 16u + (std::size_t)type should be bounds checked. Honestly: Can't we just use the original code and suppress the warning from the msvc analyzer? Those casts add absolutely nothing in terms of correctness, readability or safety to the code.
What MSVC complains about is that computation is done in 32 bits first and then the result is widened to 64bits (a.k.a. std::size_t). Such a warning makes sense in circumstances where people expect the result of a computation might need more than 32 bits and thus assign the result to a 64 bit variable, but forget that the computation itself (and resulting overflow) still happens in 32 bits.

I agree with you.
What I worry about is that maybe it is not suitable for shutting up a compiler warning for a single line. But if people think it is ok, everything will be fine.


In addition, @nlohmann , what is this project's policy about these annoying compiler warnings?

nlohmann:
The "policy" of the library is to be warning-free where possible. However, this is getting more and more difficult, because some warnings are loaded with stylistic opinions, and it gets impossible to write code where all compilers remain silent.

In this specific case, it seems only MSVC complains, and though it would be nice to silence the warning, I am hesitant to achieve this "whatever it takes".

I put the `decode` function into [C++ Insights](https://cppinsights.io/lnk?code=I2luY2x1ZGUgPGNzdGRpbz4KI2luY2x1ZGUgPGNzdGRpbnQ+CiNpbmNsdWRlIDxhcnJheT4KI2luY2x1ZGUgPGNtYXRoPgoKICAgIHN0YXRpYyBjb25zdGV4cHIgc3RkOjp1aW50OF90IFVURjhfQUNDRVBUID0gMDsKICAgIHN0YXRpYyBjb25zdGV4cHIgc3RkOjp1aW50OF90IFVURjhfUkVKRUNUID0gMTsKCiAgICBzdGF0aWMgc3RkOjp1aW50OF90IGRlY29kZShzdGQ6OnVpbnQ4X3QmIHN0YXRlLCBzdGQ6OnVpbnQzMl90JiBjb2RlcCwgY29uc3Qgc3RkOjp1aW50OF90IGJ5dGUpIG5vZXhjZXB0CiAgICB7CiAgICAgICAgc3RhdGljIGNvbnN0IHN0ZDo6YXJyYXk8c3RkOjp1aW50OF90LCA0MDA+IHV0ZjhkID0KICAgICAgICB7CiAgICAgICAgICAgIHsKICAgICAgICAgICAgICAgIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIC8vIDAwLi4xRgogICAgICAgICAgICAgICAgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgMCwgLy8gMjAuLjNGCiAgICAgICAgICAgICAgICAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAwLCAvLyA0MC4uNUYKICAgICAgICAgICAgICAgIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIDAsIC8vIDYwLi43RgogICAgICAgICAgICAgICAgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgOSwgOSwgOSwgOSwgOSwgOSwgOSwgOSwgOSwgOSwgOSwgOSwgOSwgOSwgOSwgOSwgLy8gODAuLjlGCiAgICAgICAgICAgICAgICA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCA3LCAvLyBBMC4uQkYKICAgICAgICAgICAgICAgIDgsIDgsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIDIsIC8vIEMwLi5ERgogICAgICAgICAgICAgICAgMHhBLCAweDMsIDB4MywgMHgzLCAweDMsIDB4MywgMHgzLCAweDMsIDB4MywgMHgzLCAweDMsIDB4MywgMHgzLCAweDQsIDB4MywgMHgzLCAvLyBFMC4uRUYKICAgICAgICAgICAgICAgIDB4QiwgMHg2LCAweDYsIDB4NiwgMHg1LCAweDgsIDB4OCwgMHg4LCAweDgsIDB4OCwgMHg4LCAweDgsIDB4OCwgMHg4LCAweDgsIDB4OCwgLy8gRjAuLkZGCiAgICAgICAgICAgICAgICAweDAsIDB4MSwgMHgyLCAweDMsIDB4NSwgMHg4LCAweDcsIDB4MSwgMHgxLCAweDEsIDB4NCwgMHg2LCAweDEsIDB4MSwgMHgxLCAweDEsIC8vIHMwLi5zMAogICAgICAgICAgICAgICAgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMCwgMSwgMSwgMSwgMSwgMSwgMCwgMSwgMCwgMSwgMSwgMSwgMSwgMSwgMSwgLy8gczEuLnMyCiAgICAgICAgICAgICAgICAxLCAyLCAxLCAxLCAxLCAxLCAxLCAyLCAxLCAyLCAxLCAxLCAxLCAxLCAxLCAxLCAxLCAxLCAxLCAxLCAxLCAxLCAxLCAyLCAxLCAxLCAxLCAxLCAxLCAxLCAxLCAxLCAvLyBzMy4uczQKICAgICAgICAgICAgICAgIDEsIDIsIDEsIDEsIDEsIDEsIDEsIDEsIDEsIDIsIDEsIDEsIDEsIDEsIDEsIDEsIDEsIDEsIDEsIDEsIDEsIDEsIDEsIDMsIDEsIDMsIDEsIDEsIDEsIDEsIDEsIDEsIC8vIHM1Li5zNgogICAgICAgICAgICAgICAgMSwgMywgMSwgMSwgMSwgMSwgMSwgMywgMSwgMywgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMywgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSwgMSAvLyBzNy4uczgKICAgICAgICAgICAgfQogICAgICAgIH07CgogICAgICAgIGNvbnN0IHN0ZDo6dWludDhfdCB0eXBlID0gdXRmOGRbYnl0ZV07CgogICAgICAgIGNvZGVwID0gKHN0YXRlICE9IFVURjhfQUNDRVBUKQogICAgICAgICAgICAgICAgPyAoYnl0ZSAmIDB4M2Z1KSB8IChjb2RlcCA8PCA2dSkKICAgICAgICAgICAgICAgIDogKDB4RkZ1ID4+IHR5cGUpICYgKGJ5dGUpOwoKICAgICAgICBzdGF0ZSA9IHV0ZjhkWzI1NnUgKyBzdGF0ZSAqIDE2dSArIHR5cGVdOwogICAgICAgIHJldHVybiBzdGF0ZTsKICAgIH0KCgppbnQgbWFpbigpCnsKICBhdXRvIHRpbGVfaWQgPSA1OwogIGF1dG8geCA9IChsb2coc3RhdGljX2Nhc3Q8dWludDMyX3QgPih0aWxlX2lkKSkgLyBzdGQ6OmxvZygyKSkgLSAxNjsKfQo=&insightsOptions=&std=undefined&rev=1.0) which translates

```cpp
codep = (state != UTF8_ACCEPT)
        ? (byte & 0x3fu) | (codep << 6u)
        : (0xFFu >> type) & (byte);

state = utf8d[256u + state * 16u + type];
```

into

```cpp
codep = (static_cast<int>(state) != static_cast<int>(UTF8_ACCEPT))
      ? (static_cast<unsigned int>(byte) & 63U) | (codep << 6U)
      : (255U >> static_cast<int>(type)) & static_cast<unsigned int>((byte));
state = utf8d.operator[](static_cast<unsigned long>((256U + (static_cast<unsigned int>(state) * 16U)) + static_cast<unsigned int>(type)));
```

Maybe this can help to find a fix for the issue.

dota17:
> As byte is unsigned, the check whether it is >= 0 is redundant, right?

Right.
From #1911 , it said that `Warning C28020 The expression '0<=_Param_(1)&&_Param_(1)<=400-1' is not true at this call.`. So i added this. 
And from #859, it may be a false positive.

The newest version of this PR had move it out.

dota17:
There are some errors in CI. And i tried to fix it.
However, the errors in #2004, #2002 are same as this issue, especially #2004 just add a yml file and don't change any code.

dota17:
Ready to go.

nlohmann:
Thanks!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Fix a compiler warning in MSVC. #1969 #1911

* * *

PR #1974, opened 2020-03-06
Relative paths

taitov:
What do you think about this?

main.cpp:
```
#include "nlohmann/json/include/nlohmann/json.hpp"
int main() { return 0; }
```
run:
```
| command                              | before | after |
| g++ main.cpp -Inlohmann/json/include | pass   | pass  |
| g++ main.cpp                         | fail   | pass  |
```

:+1: or :-1:

### DIFF
--- a/include/nlohmann/adl_serializer.hpp
+++ b/include/nlohmann/adl_serializer.hpp
@@ -2,8 +2,8 @@
 
 #include <utility>
 
-#include <nlohmann/detail/conversions/from_json.hpp>
-#include <nlohmann/detail/conversions/to_json.hpp>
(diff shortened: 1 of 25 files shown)


nlohmann:
Can you please also run "make amalgamate" and check whether your approach also runs with the amalgamated single header?

taitov:
It doesn't work =(

main.cpp:
```
#include "single_include/nlohmann/json.hpp"
int main() { return 0; }
```
run:
```
cd nlohmann/json
rm single_include/nlohmann/json.hpp
make amalgamate
g++ main.cpp
... many errors ...
```

Please close the pull request if you think there is nothing more to discuss. Thanks!

PR #2011, opened 2020-03-25
.travis.yml: Switch to latest xcode 10/11 point release

t-b:


dota17:
Currently, people must wait for a long time due to there are a lot of ci tasks.

> Please add the images rather than replacing old ones. Please also update the README and add the respective macOS/Xcode versions.

If the newer version is compatible, i think it is better to hold one task for an env.

PR #2025, opened 2020-04-02
Fixes #1971 (memory leak in basic_json::push_back)

### LINKED ISSUE #1971: Possible memory leak on push_back
I'm experiencing a memory leak when using push_back to add strings to a json object

Steps to reproduce:
- Use a custom allocator to track memory allocations
- Create a json object as array: json myJson = json::array_t{};
- Add a string type object with push_back: myJson.push_back(String{"Hello leaking world"});
- Destroy the json instance.
- Experience the leak.

I'm using Visual Studio 2017 on Windows 10, compiling with standard Debug configuration in x64
My latest synch on this branch was on 28/02/2020 at 15:20 CET.

I've (probably) fixed the issue by removing (on the single header file) the line 19426: val.m_type = value_t::null; from the function void push_back(basic_json&& val).

From my debugging, the issue comes from the fact that the m_value.array->push_back(std::move(val)); is invoking the basic_json(const basic_json&) instead of the basic_json(basic_json&&), which creates a copy of the source value that should still be properly delete by the allocator. My fix shouldn't affect the final result since, if the move constructor is invoked in the push_back of 19422, the move constructor will take care of marking the type of the source basic_json as null.

Hope this will be helpful!

Cheers,

Davide.

ArtemSarmini:
Fixes leak in `basic_json::push_back(basic_json&&)`. Old code incorrectly assumed that `std::move` moves, which is not always the case. basic_json move constructor marks argument as null anyway, so that line of code was redundant.
I also added valgrind-oriented unit test to address such situation.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/test/src/unit-allocator.cpp
+++ b/test/src/unit-allocator.cpp
@@ -240,7 +240,7 @@ namespace
 template<class T>
 struct allocator_no_forward
 {
-    typedef std::remove_const_t<T> value_type;
+    typedef T value_type;
     template <typename U>
     struct rebind
     {
--- a/test/src/unit-allocator.cpp
+++ b/test/src/unit-allocator.cpp
@@ -240,8 +240,7 @@ namespace
 template<class T>
 struct allocator_no_forward : std::allocator<T>
 {
-    using std::allocator<T>::allocator;
-
+    allocator_no_forward() {}
     template <class U>
     allocator_no_forward(allocator_no_forward<U>) {}
 
@@ -253,6 +252,7 @@ struct allocator_no_forward : std::allocator<T>
     template <class... Args>
     void construct(T* p, const Args&... args)
     {
+        // force copy even if move is available
         ::new (static_cast<void*>(p)) T(args...);
     }
 };
--- a/test/src/unit-allocator.cpp
+++ b/test/src/unit-allocator.cpp
@@ -234,3 +234,51 @@ TEST_CASE("controlled bad_alloc")
         }
     }
 }
+
+namespace
+{
+template<class T>
+struct allocator_no_forward
+{
+    typedef std::remove_const_t<T> value_type;
+    template <typename U>
+    struct rebind
+    {
+        typedef allocator_no_forward<U> other;
+    };
+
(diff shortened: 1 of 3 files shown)


test/src/unit-allocator.cpp
```diff
@@ -234,3 +234,51 @@ TEST_CASE("controlled bad_alloc")
...
+    SECTION("my_allocator::construct doesn't forward")
+    {
+        using bad_alloc_json = nlohmann::basic_json<std::map,
+              std::vector,
+              std::string,
+              bool,
+              std::int64_t,
+              std::uint64_t,
+              double,
+              allocator_no_forward>;
```
dota17, line 278: wried code format

ArtemSarmini: I copied it from test above in the same file. Does this
```
using bad_alloc_json = nlohmann::basic_json<
    std::map, std::vector, std::string, bool, std::int64_t, std::uint64_t, double, allocator_no_forward>;
```
look better? Second line willl be 114 symbols long.

dota17:
```
using bad_alloc_json = nlohmann::basic_json<std::map,
                                            std::vector,
                                            std::string,
                                            bool,
                                            std::int64_t,
                                            std::uint64_t,
                                            double,
                                            allocator_no_forward>;
```
I think it will be better.

ArtemSarmini:
What do I do with this? https://travis-ci.org/github/nlohmann/json/jobs/672497870
> E: Unable to locate package g++-9
E: Couldn't find any package by regex 'g++-9'

nlohmann:
Thanks!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [x] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Fixed a memory leak in basic_json::push_back. #1971 #2025

* * *

ArtemSarmini:
Glad to help.

PR #2054, opened 2020-04-20
Fix bug in diff function

### LINKED ISSUE #1983: JSON patch diff for op=add formation is not as per standard (RFC 6902)
Hi,

When result_json = diff(source, target) is called with 

source = { "foo": [ "bar", "baz" ] }
dest = {"foo": [ "bar", "baz", "baa" ]}

NOTE: actual function is 

    static basic_json diff(const basic_json& source, const basic_json& target,
                           const std::string& path = "")

Obtained JSON is 
result_json = 
   [
     { "op": "add", "path": "/foo/1", "value": "qux" }
   ]

This is wrong as per RFC 6902 (refer section 4.1)
https://tools.ietf.org/html/rfc6902#page-4

Expected is 
result_json = 
   [
     { "op": "add", "path": "/foo/-", "value": "qux" }
   ]

Where is the problem here?

Or is it we need to interchange source and target to generate diff?
Waiting for your response.

Thanks and Regards,
Harish M

nlohmann:
Closes  #1983.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -7930,7 +7930,7 @@ class basic_json
                     result.push_back(
                     {
                         {"op", "add"},
-                        {"path", path + "/" + std::to_string(i)},
+                        {"path", path + "/-"},
                         {"value", target[i]}
                     });
                     ++i;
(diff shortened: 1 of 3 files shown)


nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [x] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Fix bug in diff function #1983 #2054
* * *

PR #2074, opened 2020-04-29
Add CMake fetchcontent documentation and tests

ArthurSonzogni:
Github issue:
https://github.com/nlohmann/json/issues/2073

nlohmann::json documents 2 way of depending on it using CMake
1) Copy-paste the project/source into your own project.
2) Install nlohman::json and then use find_package.

(1) pollutes your git repository, (2) requires everyone to install the
dependencies themselves.

Since 2018, CMake provide some kind of 'package manager' features using
[FetchContent](https://cmake.org/cmake/help/v3.17/module/FetchContent.html)
It gives the following:
~~~cmake
include(FetchContent)

FetchContent_Declare(json
  GIT_REPOSITORY https://github.com/nlohmann/json
  GIT_TAG v3.7.3)

FetchContent_GetProperties(json)
if(NOT json_POPULATED)
  FetchContent_Populate(json)
  add_subdirectory( ${json_SOURCE_DIR} ${json_BINARY_DIR} EXCLUDE_FROM_ALL)
endif()
~~~

Then declares the dependency in the target using it:
~~~cmake
target_link_library(my_project PRIVATE nlohmann_json::nlohmann_json
~~~

This patch updates the documentation and provides tests.

PR #2077, opened 2020-05-01
Fix README.md. Missing ```

ArthurSonzogni:
This patch follows the one from: https://github.com/nlohmann/json/pull/2074.

An error has been introduced by accepting the suggestions. See:
https://github.com/ArthurSonzogni/json/commit/4be4a038ccd17aab0a4cc85200a6a0b41e1cccf2

One suggestion was about removing ~~~, but it was meant to be replaced
by ``` in reality. This caused the README.md to be slightly broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ if(NOT json_POPULATED)
 endif()
 
 target_link_libraries(foo PRIVATE nlohmann_json::nlohmann_json)
+```
 
 **Note**: The repository https://github.com/nlohmann/json download size is huge.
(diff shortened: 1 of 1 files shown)


ArthurSonzogni:
Hi again,

It looks like there was a small issue in the previous patch. I blindly accepted all the suggestions, but I didn't realize one was removing ~~~ without replacing it with ```

Could you please take a look?

ArthurSonzogni:
Thanks!

nlohmann:
Thank you!

PR #1251, opened 2018-09-20
Tests/cmake external data

theodelrieu:
This PR solves #1185 and affiliates, by putting every data file (test and benchmarks) somewhere else (currently one of my repositories...), and fetching them with CMake `ExternalData` module.

The current repo size is 22MB, down from more than 150MB.

There are more than 1K files to download when running tests, which is a bit annoying. The CI will most likely be slower.

We could refactor the tests/cmake files and dedicate a specific CI job for tests requiring data files, and only trigger the 1K download when launching those specific tests.

Last but not least, we have to decide where to store those files.

Let me know what you think!

## Pull request checklist

Read the [Contribution Guidelines](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md) for detailed information.

- [x]  Changes are described in the pull request, or an [existing issue is referenced](https://github.com/nlohmann/json/issues).
- [ ]  The test suite [compiles and runs](https://github.com/nlohmann/json/blob/develop/README.md#execute-unit-tests) without error.
- [ ]  [Code coverage](https://coveralls.io/github/nlohmann/json) is 100%. Test cases can be added by editing the [test suite](https://github.com/nlohmann/json/tree/develop/test/src).
- [x]  The source code is amalgamated; that is, after making changes to the sources in the `include/nlohmann` directory, run `make amalgamate` to create the single-header file `single_include/nlohmann/json.hpp`. The whole process is described [here](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md#files-to-change).

PR #2081, opened 2020-05-03
Use external test data

### LINKED ISSUE #96: 17 MB / 90 MB repo size!?

### LINKED ISSUE #482: For a header-only library you have to clone 214MB
It would be nice if the tests and benchmark data could be separated from the include file. I just added this as a git submodule and discovered the clone is 214MB.

### LINKED ISSUE #556: Minimal repository (current size very large)
I want to use "json" as a submodule to my project, because:

1.  I want to be transparent is ownership, to give credits where they are due.
2.  To facilitate updating from source.

However, the current repository is extremely large, due to the tests and benchmarks. This hinders again the distribution of my code. Therefore it would be great if a minimal repository can be maintained.

Thanks!
Tom

### LINKED ISSUE #567: Need to improve ignores..
I love this project.. but why does it take 133.06 MiB?  Appears to be a short coming of your github ignore file.
```
ken@ken-VirtualBox:~/git-repos/septemmud$ git submodule add https://github.com/nlohmann/json.git
Cloning into 'json'...
remote: Counting objects: 31983, done.
remote: Compressing objects: 100% (52/52), done.
remote: Total 31983 (delta 21), reused 0 (delta 0), pack-reused 31929
Receiving objects: 100% (31983/31983), 133.06 MiB | 533.00 KiB/s, done.
Resolving deltas: 100% (26702/26702), done.
Checking connectivity... done.
ken@ken-VirtualBox:~/git-repos/septemmud$ 
```

### LINKED ISSUE #620: [Question] When using this as git submodule, will it clone the whole thing include test data and benchmark?
it seems the actual source is small but test and benchmark folder is huge (200MB ++).

Should I just copy the code from `json.hpp` or is there anyway to use this as submodule without cloning test folder and benchmark folder?

Thank you.

### LINKED ISSUE #732: Move test blobs to a submodule?
I am using the [Buckaroo](https://github.com/LoopPerfect/buckaroo) package for `nlohmann/json` and find it takes quite a long time to download. Looking at the repo, the test blobs seem to be the cause. Of course, these are not required when using the library. 

Perhaps the tests should be moved to a Git submodule, so that they can be fetched optionally?

### LINKED ISSUE #1066: Minimal branch?
**Feature Request**

Could we get a "minimal" or "release" branch which contains essentially `README` / `LICENSE` and `single_include/nlohmann/json.hpp` (and perhaps also `include/...`)?  This would allow a tiny shallow or single-branch checkout to track releases, rather than a (relatively slow) hundred-megabyte checkout or submodule.

### LINKED ISSUE #1184: CMakeLists.txt in release zips?
Hello, I've tried using [FetchContent](https://cmake.org/cmake/help/git-master/module/FetchContent.html) CMake module and this doesn't work as well as I expected. The problem is that after I download and unpack the zip, there's no CMakeLists.txt file, so I can't just get nlohmann_json target and link with it.

If you take a look at libraries like [fmt](https://github.com/fmtlib/fmt/releases), they include CMakeLists.txt in their releases, which makes fetching zips and then getting targets at configure step quite easy.

So, can we have CMake related files included in zips for future releases?

### LINKED ISSUE #1185: reduce repos size
When re-cloning this repo i keep noticing that it comes down as about 373MB. all this just to get a json parser/serializer. i would suggest outsourcing the test data (should it ever change?) and re-writing the history.

### LINKED ISSUE #1235: Move tests to a separate repo
The test directory is very large (currently >150MB), and checking out the repo requires checking out that entire test directory as well, even if you just want to build the code without running the tests.  Splitting the test directory into a separate repo and including it as a submodule or subtree would eliminate this issue and make it a lot more friendly for people to use this library from the git repo.

### LINKED ISSUE #1497: Repository is almost 450MB
The size of this repository is 436MB unpacked (although the download is 'only' ~150MB), which is insane for a tiny JSON parsing library. There are multiple cases where (IMO) the size can be trimmed down massively without actually impacting the repo.

The `.git` directory is 178MB, likely since you have a lot of changed binary files. You can easily slash the size of this directory by ~90% using `git repack` (which has drawbacks for repo history, but unless you really care about specific info about changes from years ago this is fine).

Why do you have a 23MB file of 1000000 signed integers in `test/data/regression/signed_ints.json` Do any of those ints besides the first few thousand really contribute any code coverage? I'd wager not. If the purpose of that many is to make sure the program can handle really large JSON files, why is there also another 23MB of unsigned integers? And 22MB of floats? Having lots of tests is good but this is seriously affecting users with slow internet speed who want to contribute to the project.

### LINKED ISSUE #1572: A "thinner" source code tar as part of release?
Would it be possible to have a lighter weight version of the source code tarball included with the release?  The benchmark and test directories are quite large, and while I want the structure and much of the content of your full release, it would be less burdensome if not all of the test and benchmark directories were included.

### LINKED ISSUE #1790: Unnecessary test data file in the release
The total repo exceeds 200MB in size. 
An ordinary user of this library doesn't need those data or test output data file. 
Please do not include those files in the release branch, some examples are:

test/data
test/reports
benchmark/data

Those files are needed by developers of this project.  

Without those files, the size of the repo can be reduced to 2.6MB.

nlohmann:
This PR removes the test data from this repository and downloads them from an external repository.

Closes #96
Closes #482
Closes #556
Closes #567
Closes #620
Closes #732
Closes #1066
Closes #1184
Closes #1185
Closes #1235
Closes #1497
Closes #1572
Closes #1790


cmake/download_test_data.cmake
```diff
@@ -0,0 +1,11 @@
+find_package(Git)
+
+# target to download test data
+add_custom_target(download_test_data
+    COMMAND test -d json_test_data || ${GIT_EXECUTABLE} clone https://github.com/nlohmann/json_test_data.git --quiet --depth 1
```
nickaein, line 5: Shouldn't we pin down the version of `json_test_data` we are referring to? Otherwise we have to worry about backward-compatibility of changes in `json_test_data` in future.

nickaein:
On a clean repo, `make check` command fails since `download_test_data` command doesn't get executed. 

```
isaac@ubuntu$ make check
make check -C test
make[1]: Entering directory '/home/isaac/nl-json/test'
[CXX]   src/unit-bson.o
src/unit-bson.cpp:37:25: fatal error: test_data.hpp: No such file or directory
compilation terminated.
Makefile:71: recipe for target 'src/unit-bson.o' failed
make[1]: *** [src/unit-bson.o] Error 1
make[1]: Leaving directory '/home/isaac/nl-json/test'
Makefile:64: recipe for target 'check' failed
make: *** [check] Error 2
isaac@ubuntu$
```

nlohmann:
> On a clean repo, `make check` command fails since `download_test_data` command doesn't get executed.

I am currently moving functionality from the plain `Makefile` to CMake. The header is generated by CMake, and the dependency to download the test data first is only expressed in CTest.

cmake/download_test_data.cmake
nlohmann: Good catch! I would use semantic versioning for the data repo, rightß

nickaein: Can versioning it as something like v1, v2, v3 do the job? There is no harm in using semantic versioning, but it _might_ be overkill in this case since we are not providing an API.

nlohmann: I totally agree. Will do.

nlohmann: I still decided for semantic versioning - you'll never know ;)

nlohmann:
Also related: #1251

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Moved test data in separate repository. #96 #482 #556 #567 #620 #732 #1066 #1184 #1185 #1235 #1497 #1572 #1790 #2081

* * *

tuananh:
Awesome! Thanks!

Mailaender:
This has the downside that tests will now fail in Linux distribution build environments that have no internet access.

nlohmann:
Is this a thing? I mean: having the library, having no internet connection, yet still want to execute the tests?

Mailaender:
I had to disable the tests for https://build.opensuse.org/request/show/856756 which is not nice but also not too serious. https://build.opensuse.org/package/rdiff/home:Mailaender:branches:devel:libraries:c_c++/nlohmann_json?linkrev=base&rev=2 is my try to disable the data download modules, but it seems they don't disable the tests that depend on it, but that may more be some kind of ctest problem.

Issue #2086, opened 2020-05-04
Value function return empty object even if it exist


xamix:
- Describe what you want to achieve.
I need to get a default empty object in case an item do not exist in the specified json.

- Describe what you tried.
Using the function `value`.

- Describe which system (OS, compiler) you are using.
GCC 9 on Ubuntu x64.

- Describe which version of the library you are using (release version, develop branch).
I use develop branch.

Here is a simple example to demonstrate my question:

```
#include <iostream>

#include <nlohmann/json.hpp>
using json = nlohmann::json;

int main()
{
    json j = {
        {"what", "i'm here"}
    };

    json k = j.value("what", json::value_t::object);

    std::cout << "j:" << j.dump(4) << std::endl;
    std::cout << "k:" << k.dump(4) << std::endl;
    return 0;
}
```

This output `k` as empty even if the item exist:

```
j:{
    "what": "i'm here"
}
k:""
```

However it works (`k` is filled with the good value `i'm here`) if I change the call of `value` with one of the following:

`json k = j.value<json>("what", json::value_t::object);`
or
`json k = j.value("what", json({}));`

Is this the expected behavior?

nlohmann:
This seems like a bug. A quick fix would be to pass `json::object()` instead of `json::value_t::object` as parameter.

xamix:
I confirm that it works with `json::object()`.

Also it seem to fail if I want a null object by using:

`json k = j.value("what", json::value_t::null);`

dota17:
`json::value_t::object` shold not be passed directly into value() unless it is converted to json type, like `value<json>("what", json::value_t::object)`. Obviously, this type of parameter is not handled correctly in the code, at least, it should remind the caller type error.

Or we should explicitly specify the parameter as the object type：
    template<class ValueType, typename std::enable_if<
                 std::is_convertible<~~basic_json_t~~ object_t, ValueType>::value, int>::type = 0>
    ValueType value(const typename object_t::key_type& key, const ValueType& default_value) const

nlohmann:
It should be ok to pass `json::value_t::object`, because `json(json::value_t::object)` is valid code to create an empty object. I have not understood why the code above does not work as expected, though.

nlohmann:
I had another look at this: The requirment for allowing code like

```cpp
json k = j.value("what", json::value_t::object);
```

is that the type of the second argument, `json::value_t`, can be converted to `basic_json`, and indeed it can be via member function

```cpp
[...]

PR #2104, opened 2020-05-13
Fix issue#2086: disallow json::value_t type parameter in value()

dota17:
This patch fix issue #2086 by adding a type-error to disallow json::value_t type parameter in value() to avoid it returns empty object even if the key and its mapped value exist.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -3759,7 +3759,8 @@ class basic_json
     @since version 1.0.0
     */
     template<class ValueType, typename std::enable_if<
-                 std::is_convertible<basic_json_t, ValueType>::value, int>::type = 0>
+                 std::is_convertible<basic_json_t, ValueType>::value
+                 and not std::is_same<value_t, ValueType>::value, int>::type = 0>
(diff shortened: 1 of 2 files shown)


include/nlohmann/json.hpp
```diff
@@ -3717,6 +3719,10 @@ class basic_json
                  std::is_convertible<basic_json_t, ValueType>::value, int>::type = 0>
     ValueType value(const typename object_t::key_type& key, const ValueType& default_value) const
     {
+        // check defalut_value type
+        if(std::is_same<json::value_t, ValueType>::value) {
```
nlohmann, line 3723: I do not like runtime error. I think it makes more sense to disallow the type via SFINAE.

dota17: nice suggestions😄

nlohmann:
Thanks!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Disallow json::value_t type parameter in value(). #2086 #2104

* * *

PR #2117, opened 2020-05-16
Add conversion from/to std::optional

### LINKED ISSUE #1749: [C++17] Allow std::optional to convert to nlohmann::json

nlohmann:
Closes #1749.


dota17:
I am working for checking these CI tasks.

Restart the travis CI tasks, and then they would be fine.
[This Appveyor CI task](https://ci.appveyor.com/project/nlohmann/json/builds/33010708/job/jnlg1wm6cnpewdfn) may be ok after we restart, it seems like a network error.

And about other failed Appveyor CI tasks, I find the difference is `CXX_FLAGS=/permissive- /std:c++latest`. Maybe it is the problem, and I would try to fix it.

nlohmann:
I restarted all jobs.

nlohmann:
The same AppVeyor jobs keep failing. Any ideas?

dota17:
**More Info :**
The `std::optional` code maybe not included. I had added logs in `unit-conversions.cpp` to check if the testcases of `std::optional` had been run:
```
    ....
    SECTION("traditional enum")
    {
        // check normal testcase had been run, result : print, run
        std::cout << "traditional enum" << std::endl; 
        ...
    }
#ifdef JSON_HAS_CPP_17
TEST_CASE("std::optional")
{
    SECTION("null")
    {
        // check the C++17 and testcase, reslut : not print, not run
        std::cout << "null" << std::endl; 
....
```
It seems that all jobs hadn't compile the `std::optional` code so that there are no compiler errors and and the testcases hadn't been run. If compiled, maybe they would had same errors.

nlohmann:
Yes, the `<optional>` tests are only executed in C++17 mode. Strangely, not all of them fail, but only a few of them.

dota17:
> Yes, the <optional> tests are only executed in C++17 mode. Strangely, not all of them fail, but only a few of them.

I find that the `<optional>` tests are only executed in `VS 2017 with CXX_FLAGS=/permissive- /std:c++latest`. I think they should be also executed in VS2019, but didn't. Thus, some success may be fake because they didn't run the tests.

What's more, in `VS 2017 with CXX_FLAGS=/permissive- /std:c++latest` jobs, they didn't call the method `from_json(const BasicJsonType& j, std::optional<T<& opt)`.

The actual behavior:
```
CHECK(std::optional<std::string>(j_string) == opt_string);  // call from_json(const BasicJsonType& j, ConstructibleStringType& s)
CHECK(std::optional<bool>(j_bool) == opt_bool); // call from_json(const BasicJsonType& j, typename BasicJsonType::boolean_t& b) 
CHECK(std::optional<int>(j_number) == opt_int); // call from_json(const BasicJsonType& j, ArithmeticType& val)
```
And the test `SECTION("null")` didn't call the expected method either.

1. We should ensure which envs/jobs would execute the `<optional>` tests.
2. It seems that the implement of `from_json(const BasicJsonType& j, std::optional<T<& opt)` doesn't work. We should fix it so that `CHECK(std::optional<int>(j_number) == opt_int);` calls the expected method. - I have no idea now.

nlohmann:
Since I don't use MSVC myself, I cannot debug this further. I will drop this feature from 3.8.0 to proceed. Feel free to dig in, but I will not pursue this issue in the moment.

kix-mcotton:
Just wanted to point out that I selectively added this to the 3.9.0 single header and was able to build with visual studio 16.6.4 with no issues.

nlohmann:
@kix-mcotton Thanks for reporting! I updated the branch to the latest develop version and will give it another try.

dota17:
glad to see some progress in this.

nlohmann:
Two tests at AppVeyor are still failing. :-/

karzhenkov:
I guess that some tests at Travis are successful due to a bug in GCC and Clang. Particularly, these compilers accept the following (while MSVC doesn't):
https://github.com/nlohmann/json/blob/fedf82c904af5560b47a922b69b23fead01afa23/test/src/unit-conversions.cpp#L1764

Consider [the simplified example](https://godbolt.org/z/saKKos):
```
#include <vector>

struct J
{
    template <typename T> operator T () const;
};

void test()
{
    J j;
    std::vector<int> v(j);
}
```
With `-std=c++14` both GCC and Clang correctly diagnose the ambiguity. Candidate functions are converting constructors:
```
vector<_Tp, _Alloc>::vector(vector<_Tp, _Alloc>&&);
vector<_Tp, _Alloc>::vector(const vector<_Tp, _Alloc>&);
vector<_Tp, _Alloc>::vector(const allocator_type&);
vector<_Tp, _Alloc>::vector(size_type, const allocator_type& = allocator_type());
vector<_Tp, _Alloc>::vector(initializer_list<_Tp>, const allocator_type& = allocator_type());
```
Note that specializations of `S::operator T() const` are not candidates for overload resolution.

When matching the first parameter against the initalizer of type `J` all the candidates require an implicit user-defined conversion,
and there is no best one among them. All applicable specializations of `S::operator T() const` provide exact match.

C++17 introduces a special case - when the initializer is of the same type as the object being initialized, no overload resolution is perfomed at all (and no conversion). We have different types, so this special case doesn't apply here; overload resolution should be performed as before.

However, with `-std=c++17` both GCC and Clang erroneously select the second candidate from the list above. This bug was [reported](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83445) for GCC 7 and seems to be fixed in GCC 8, but arises again in GCC 9 and later.

Exploiting a compiler bug is not as good choise. Perhaps support for `std::optional` should be deferred until compilers become more mature. Other possibility is to find suitable workarounds. Interestingly, the workaround is needed for MSVS, which exhibits a better conformance here.

karzhenkov:
The success at Travis has one more reason. Conversion tests for `std::optional` are not executed. There are three test cases in `unit-conversions.cpp` whereas [job log](https://travis-ci.org/github/nlohmann/json/jobs/715757748#L854) contains only two of them. Test case for `std::optional` requires C++17 but is excluded despite of `CXXFLAGS=-std=c++2a`.

This can be explained by the fact that the desired С++ standard is specified twice: in top-level `CMakeLists.txt` (compile feature `cxx_std_11`) and through aforementioned `CXXFLAGS` in `.travis.yml`. CMake probably doesn't analyze user flags, and compiler command line includes both options. The collision is resolved to C++11 (this is determined by order of options).

Problem can be fixed by using `cmake -DCMAKE_CXX_STANDARD=...` in Travis configuration instead of `CXXFLAGS` (somewhat like 7df758f41f97a43d1b562579bd260fee94187b9d from #2213).

nlohmann:
Wow, thanks for the detailed analysis! What do you propose to do?

karzhenkov:
I don't see any way to achieve desired conversion sequence for direct initialization when implicit conversions of `json` are enabled. We don't have control over constructors of "external" target classes such as `std::vector`. Conversion operators defined in `json` are included in the candidate set only for copy initialization which uses a different syntax, e. g. `std::vector<...> v = j` instead of `std::vector<...>(j)`.

The above only applies to "conforming" compilers, but each compiler vendor has their own approach to interpretation of standard. It will take time to reach a consensus on subtle topics (and on some trivial topics too). Anyway, it is dangerous to rely on the "new" overload resolution algorithm provided by GCC and Clang with `-std=c++17`. Language improvements should not change the meaning of the program. Meanwhile, the example discussed on [Bugzilla](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83445) compiles to executable that produces different output when compiled with `-std=c++14` or `-std=c++17`. I suspect significant difference between GCC 8.3 and GCC 8.4 (the latter is used by Travis, but is not avaliaible at [godbolt](https://godbolt.org/)).

Maybe it is reasonable to support `std::optional` only with `JSON_USE_IMPLICIT_CONVERSIONS` defined as `0` for a while.

nlohmann:
> Maybe it is reasonable to support std::optional only with JSON_USE_IMPLICIT_CONVERSIONS defined as 0 for a while.

Yes, I think this is a reasonable approach.

test/src/unit-conversions.cpp
```diff
@@ -1703,6 +1703,65 @@ TEST_CASE("JSON to enum mapping")
...
+#ifdef JSON_HAS_CPP_17
+TEST_CASE("std::optional")
+{
+    SECTION("null")
+    {
+        json j_null;
+        std::optional<std::string> opt_null;
+
+        CHECK(json(opt_null) == j_null);
+        CHECK(std::optional<std::string>(j_null) == std::nullopt);
```
karzhenkov, line 1715: Initialization of `optional<T>` from empty `json` should be considered as a `type_error`:
```suggestion
        CHECK_THROWS_AS(std::optional<std::string>(j_null), json::type_error&);
```
The reason is as follows. If `json` is implicitly convertible to `T`, then initialization of `optional<T>` from `json` almost necessarily goes through the "forwarding" constructor:
```
template <class U> optional(U&& value); // unrelated details omitted
```
This constructor provides the best parameter/argument match by value category, constness and a type. Its parameter binds directly to `json` in any form. The only "weakness" of the forwarding constructor is that it is a template. A conversion function that would be considered as a better candidate for overload resolution must be non-templated, but this is obviously unacceptable.

The forwarding constructor converts `json` to `T` (initializes the latter from the former) and stores the result, so the constructed `optional<T>` cannot be empty. Such a conversion from empty `json` to `T` is a `type error` (even if `T` itself is `optional<X>`, by induction).

The non-standard overload resolution performed by GCC and Clang when enabling C++17 does not change the outcome. The only way to achieve the "natural" behavior is to tune constructors of `std::optional<T>`, which is beyond the scope of the project.

test/src/unit-conversions.cpp
```diff
@@ -1703,6 +1703,65 @@ TEST_CASE("JSON to enum mapping")
...
+        CHECK(std::optional<int>(j_number) == opt_int);
+    }
+
+    SECTION("array")
+    {
+        json j_array = {1, 2, nullptr};
+        std::vector<std::optional<int>> opt_array = {{1, 2, std::nullopt}};
+
+        CHECK(json(opt_array) == j_array);
+        CHECK(std::vector<std::optional<int>>(j_array) == opt_array);
```
karzhenkov, line 1751: Direct initialization of `vector<...>` is not suitable here:
```suggestion
        CHECK(j_array.get<std::vector<std::optional<int>>>() == opt_array);
```
Such kind of initialization of `vector<...>` has not been tested even with simple item types and actually leads to a compile error for the reason considered [earlier](https://github.com/nlohmann/json/pull/2117#issuecomment-675366739). For example, the following will not compile unless using GCC or Clang with C++17 enabled (see  https://godbolt.org/z/oeooPG):
```
json j({1, 2, 3});
std::vector<int> v1 = j;         // Copy initialization: ok
auto v2 = std::vector<int>(j);   // Direct initialization: compile error 
```

test/src/unit-conversions.cpp
```diff
@@ -1703,6 +1703,65 @@ TEST_CASE("JSON to enum mapping")
...
+    }
+
+    SECTION("object")
+    {
+        json j_object = {{"one", 1}, {"two", 2}, {"zero", nullptr}};
+        std::map<std::string, std::optional<int>> opt_object {{"one", 1}, {"two", 2}, {"zero", std::nullopt}};
+
+        CHECK(json(opt_object) == j_object);
+        CHECK(std::map<std::string, std::optional<int>>(j_object) == opt_object);
```
karzhenkov, line 1760: Direct initialization of `map<...>` has the same trouble as with `vector<...>` above:
```suggestion
        CHECK(j_object.get<std::map<std::string, std::optional<int>>>() == opt_object);
```

.travis.yml
```diff
@@ -316,6 +316,7 @@ script:
 
   # make sure CXX is correctly set
   - if [[ "${COMPILER}" != "" ]]; then export CXX=${COMPILER}; fi
+  - if [[ "${CXXFLAGS}" != "" ]]; then export CXXFLAGS=${CXXFLAGS}; fi
```
karzhenkov, line 319: This line has no effect. It is probably intended to activate C++17 (solved by #2540).
```suggestion
```

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -31,6 +31,14 @@
     #define JSON_HAS_CPP_14
 #endif
 
+#ifdef JSON_HAS_CPP_17
+    #if __has_include(<optional>)
+        #include <optional>
+    #elif __has_include(<experimental/optional>)
+        #include <experimental/optional>
```
karzhenkov, line 38: Here we will have `std::experimental::optional` instead of `std::optional`. Perhaps it would be better to just `#include <optional>`.

karzhenkov (changes requested):

test/src/unit-conversions.cpp
karzhenkov: There is also the usual way to explicitly convert `json` to `optional<T>`. The following will work:
```
        CHECK(j_null.get<std::optional<std::string>>() == std::nullopt);
```

include/nlohmann/detail/macro_scope.hpp
nlohmann: I think now that C++17 support is a bit more stable, we can indeed drop the experimental headers.

test/src/unit-conversions.cpp
karzhenkov: Essentially, there is nothing new here. @dota17 proposed the same in #2213.
However, this approach (treat such a conversion as `type_error`) introduces some inconsistency:
```
std::optional<int> opt1 = json().get<std::optional<int>>(); // ok: std::nullopt
std::optional<int> opt2 = std::optional<int>(json());       // throws type_error
std::optional<int> opt3 = json();                           // throws type_error
```
Perhaps it makes sense to throw `type_error` in the first case too?
The "natural" conversion could be explicitly implemented by dedicated member function:
```
template <typename T> std::optional<T> get_optional() const;
```

karzhenkov: Moreover, as it turns out, this dedicated member function is the only meaningful thing `nlohmann::basic_json` can provide to support `std::optional`. Overloads of `get<std::optional<T>>` should probably be disabled to prevent confusion.

include/nlohmann/detail/macro_scope.hpp
nlohmann: Fixed in if [[ "${CXXFLAGS}" != "" ]]; then export CXXFLAGS=${CXXFLAGS}; fi.

.travis.yml
nlohmann: Fixed in 384442e8e5c7b684e2b1cc4acf0047d4d38e66d5

test/src/unit-conversions.cpp
nlohmann: I think the behavior is counter-intuitive. JSON's `null` is usually considered as "unset" value (in contrast to an empty value). Therefore, I would expect `null` to convert to `std::nullopt` of any optional type.

UndarkAido:
I don't know exactly how this is being implemented but my ideal implementation would be for a field not in the `from_json` to be interpreted as `std::nullopt` and for a `std::nullopt` to not appear in the `to_json` result.

(I'm trying to parse object from Discord's API and there's lots of optional fields. Right now I can't use the `NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE` because the fields just aren't there to parse.)

nlohmann:
I'm not sure how to proceed with this PR. Any ideas?

karzhenkov:
The primary issue we have is the conversion of `null` json values to `std::optional<T>`. To fix the counter-intuitive behavior observed [here](https://github.com/nlohmann/json/pull/2117#discussion_r550885358) (the cases that throw `type_error`) the "improvement" of `std::optional` is needed. Perhaps it is worth to consider a "better" version of `optional` as a part of this library (see [#2229](https://github.com/nlohmann/json/pull/2229#issuecomment-770432619)). Otherwise, we will have to adopt the existing counter-intuitive behavior, whether with or without this PR.

fredemmott:
I don't think this can be directly resolved without edge cases as 'may be null', and 'may be omitted' are distinct concepts in JSON, but people expect `std::optional<>` to work with both - but they have conflicting requirements.

I think these need to be separate types in C++ - perhaps a new `nlohmann::nullable<T>` type which acts essentially the same way as `std::optional<>` *except* for JSON encoding:
- a nullable field is mandatory, but may be null
- an optional field may be entirely missing
- an optional field may not be null, unless it's an `std::optional<nlohmann::nullable<T>>`

This would be similar to:
- `foo ?: T` (optional) vs `foo: T|null` (nullable) in TypeScript with `strictNullChecks` enabled
- `shape(?'foo' => T)` (optional) vs `shape('foo' => ?T)` (nullable) in Hack

fsandhei:
Is there any updates on this?

nlohmann:
Unfortunately not - any help welcome!

fsandhei:
If I may ask, what is the current blocker for this to go in to the library?

I see that the issues with overload resolutions taken up by @karzhenkov seem to still not be addressed by GCC/Clang.

From what I understand there is still also discussions about the case of initializing a `std::optional` from an empty JSON object.

Also the issues with implicit conversions hell with conversion operators / constructors seem avoidable (?) through `JSON_USE_IMPLICIT_CONVERSIONS = 0`

I would like to try to contribute to this to help, but I have to admit I'm unsure what I can help with.

nlohmann:
The branch first needs to be rebased to the current `develop` branch.

cschreib-ibex:
> I don't think this can be directly resolved without edge cases as 'may be null', and 'may be omitted' are distinct concepts in JSON, but people expect `std::optional<>` to work with both - and they have conflicting requirements.

My two cents: 

1. A variable of type `std::optional<T>` indicates a C++ variable that may or may not have a value. It is disjoint from any JSON concept. When used as a return type of a function, it means there may or may not be a return value for that function (again, no JSON here). The causes for the lack of return value are defined by the function, not by the return type. Therefore, I think "may be null" and "may be omitted" are cases that should be handled by different functions; they may both return `std::optional<T>`, but with different semantic.

2. If we consider that calling `j["some_field"].get<T>()` where `"some_field"` doesn't exist is an error today, for all `T`, then `j["some_field"].get<std::optional<T>>()` should also be an error. This preserves the invariant that, no matter the template parameter, it's always an error to call `get<T>()` on a non-existent value. So, `get<std::optional<T>>()` should only support "may be null".

3. We also want to support "may be omitted", therefore we can introduce another function `try_get<T>()` which returns `std::optional<T>`. This returns `nullopt` if the value is not present, and otherwise returns `get<T>()`. Then you can write `try_get<std::optional<T>>()` for a value that may not be present and may be null. Open question whether `try_get<T>()` should still throw in case of type mismatch, or return `nullopt` (in which case its semantic becomes "return `T` if possible, `nullopt` otherwise").

All choices are then covered:
```c++
j.get<std::string>();                    // must be a string
j.get<std::optional<std::string>>();     // must be a string or null
j.try_get<std::string>();                // may be a string, or may not exist
j.try_get<std::optional<std::string>>(); // may be a string or null, or may not exist
```

Granted, adding a `nullable` type would lift any ambiguity, especially with the 4th line above (`std::optional<std::optional<std::string>>` :-1:, `std::optional<nullable<std::string>>` :+1:).

nlohmann:
Closed in favor of #4036.

Issue #1989, opened 2020-03-17
Question about flatten and unflatten


dota17:
```
json j  = R"({
        "bad": {
            "0": "one",
            "1": "two",
            "2": "three"
        },
        "foo": "bar"
    })"_json;
json j_flatten = {
        {"/bad/0", "one"},
       {"/bad/1", "two"},
       {"/bad/2", "three"},
       {"/foo",   "bar"}
};
```

I had tested that `j.flatten() == j_flatten` is true and `j_flatten.unflatten() == j` is false.
From the source code, it might just look at the first pointer - if there is a `"0"` reference_token, it will start a new array. And the behavior will cause that under the following cases, we cannot use `unflatten()`:
```
    json j_flatten =
            {
                    {"/bad/0", "one"},
                    {"/bad/1", "two"},
                    {"/bad/t", "three"},
                    {"/foo",   "bar"}
            };
```
when calling `unflatten()`, `[json.exception.parse_error.109] parse error: array index 't' is not a number` will be raised.

dota17:
Related Issue : #1575 
But it seems that there is no exact solutions.

nlohmann:
Yes, there is no exact solution for this. The point you are describing (object keys which can be translated into integers) are ambiguous in this setting. The problem with your described approach is that it would require multi-pass parsing. one to collect all keys for all object/arrays, and then to create the final types. I am not sure whether this would fix the issue. But maybe I miss something.

dota17:
> object keys which can be translated into integers

```
json j_flatten =
            {
                    {"/bad/0", "one"},
                    {"/bad/1", "two"},
                    {"/bad/t", "three"},
                    {"/foo",   "bar"}
            };
```
In this case, integers can be object keys.
What i want to say is that when we get a `0`, it maybe not an array.
Of course, if all keys are integers, array can be our first choice.
But if not all integers, it should not raise a exception. Maybe we can try `object`.

dota17:
I had tried to fix this and got some troubles. 
* Trouble 1: How to translate `array` to `object` and free the memory of `array`?
* Trouble 2: The nullvalue probleam between `0` and `11`, this is the example:
```
json j_flatten =
            {
                    {"/bad/0", "one"},
                    {"/bad/11", "two"},
                    {"/bad/t", "three"},
                    {"/foo",   "bar"}
            };
```
Anyway, `fill up array with null values if given idx is outside range` is not a good choice for this case.

nlohmann:
[...]

Issue #1575, opened 2019-04-21
`unflatten` vs objects with number-ish keys


hoehrmann:
The documentation for `flatten` and `unflatten` should point out that JSON pointers as per RFC 6901 do not distinguish between arrays, and objects with number-ish strings as keys. There should also be a note about how `unflatten` decides whether to make an array or an object (e.g., it might just look at the first pointer, it might look at all of them, what happens for path elements like `1.0`, `010`, or `1E1`).

```c++
json x = json::object({
  {"0", json::object({{"1", "2"}})},
  {"9", json::object({{"3", "4"}})},
});
```

is actually

```json
{"0":{"1":"2"},"9":{"3":"4"}}
```

flattens to

```json
{"/0/1":"2","/9/3":"4"}
```

and that unflattens to

```json
[{"1":"2"},null,null,null,null,null,null,null,null,{"3":"4"}]
```

nlohmann:
Good point! This should go into the documentation.

levinandrew:
I think it would be helpful if there was a way to override the `unflatten()` behavior so that it didn't treat "0", "1", and "2" as list indices. In my application I actually want to use incrementing values from 0 as keys, and the automatic conversion to lists here makes `unflatten()` unusable for me
```
    // create JSON value
    json j_flattened =
    {
        {"/answer/everything", 42},
        {"/list/0", 1},
        {"/list/1", 0},
        {"/list/2", 2},
        {"/name", "Niels"},
    };

    // call unflatten()
    std::cout << std::setw(4) << j_flattened.unflatten() << '\n';
```
output: 
```
{
    "answer": {
        "everything": 42
    },
    "list": [
        1,
        0,
        2
    ],
    "name": "Niels"
}
```

nlohmann:
As the flatten and in particular the unflatten function are not really standardized, one could think about adding a parameter to not interpret numbers as strings.

PR #2123, opened 2020-05-18
fix issue#1989: only consider continuous numbers

dota17:
In `unflatten` we just consider creating `object`, which means when we meet continuous numbers, we also unflatten to an object to represent the actual array. And then we convert those object with continuous numbers key to array.

https://github.com/nlohmann/json/issues/1989#issuecomment-626424778

### DIFF
--- a/include/nlohmann/detail/json_pointer.hpp
+++ b/include/nlohmann/detail/json_pointer.hpp
@@ -395,31 +395,10 @@ class json_pointer
             switch (result->type())
             {
                 case detail::value_t::null:
-                {
-                    if (reference_token == "0")
-                    {
-                        // start a new array if reference token is 0
-                        result = &result->operator[](0);
-                    }
-                    else
-                    {
-                        // start a new object otherwise
-                        result = &result->operator[](reference_token);
-                    }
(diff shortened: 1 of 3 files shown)


dota17:
Please restart the job. It seems like a `timeout` error.

dota17:
Keep alive.
Is there any new review or idea about this?

nlohmann:
I am hesitant on this issue as it will break existing code in the sense that the unflattened values will differ.

dota17:
>  the unflattened values will differ.

The problem is that the different Json values can be flattened to the same result. Thus, we can not know which one is the origin one.
For the ambiguous cases, we should take a clear rule to decide which one is preferred.

> it will break existing code

Maybe `unflatten` is a sick funtion.
If we don't break, the origin function will still throw an exception for some scenarios that should be supported. Maybe we can clear the doc or comment to explain the abnormal situation. 
Or add an option to keep the origin code. https://github.com/nlohmann/json/issues/1575#issuecomment-516981344

Honestly, I don't know why the feature is included and which reference functions it is. The infos may help a lot.

PR #2125, opened 2020-05-19
Clean up implementation of binary type

nlohmann:
This PR polishes the integration of the binary type introduced in #1662 . 

Changes:

- Added example to README.
- Extended examples for type checking functions like `is_array`, etc.
- Added a swap function.
- Created separate header file for `byte_container_with_subtype`. `binary_t` is an alias to this and replaces the previous `internal_binary_t`. There is no alias for the `BinaryType` template parameter any more - if needed, it is `binary_t::container_type`.
- Added conversions between JSON and binary type.
- Changed JSON serialization of binary types to object with `bytes` array and `subtype` integer/null value, e.g. `{"bytes": [1,2,3], "subtype": 10}`. Serialization to JSON is only for debugging purposes and should always produce valid JSON.
- Added test cases to achieve 100% coverage.
- Added `get_binary` function for simple access to the binary value by (const) reference.


OmnipotentEntity:
Looks good to me.  Thanks so much for cleaning this up.

nlohmann:
> Looks good to me. Thanks so much for cleaning this up.

Thank you for implementing the feature in the first place!

Issue #1813, opened 2019-10-25
input_adapter not user extensible


abrownsword:
I need to implement a slight variation on one of the standard input adapters, but I have no way to implement the input_adapter_protocol and then connect it to the detail::input_adapter.  This is basically equivalent to my earlier issue about making output_adapter extensible (#1534 ) as well.

jzakrzewski:
Would be a great thing to have!

abrownsword:
Would like to keep this issue alive and see it addressed.  Input data very often comes from sources which are not std::istream or in-memory strings.

nlohmann:
This may be fixed in #1950.

FrancoisChabot:
Here's how I think this should look:

I suggest adding the following overload to the parse function(s):

```
template<typename IteratorType>
json parse(IteratorType begin, IteratorType end, bool strict);
```

Internally, this would check if `IteratorType::value_type` is `int8_t` , `uint8_t`, or 'wchar_t', and properly instantiate a matching input adapter.

A user-defined input would just be a matter of creating a user-defined iterator.

PR #2145, opened 2020-05-27
Fix Issue#1813: user defined input adapters

FrancoisChabot:
This is an attempt to address #1813 

This change adds generalized Iterator support for inputs. In effect, this is just 2 changes:

a) loosen the restriction on iterators having to be of continuous storage
b) harmonize the `parse()` API with the `from_*()` API.

All `parse()` and `from_*()` functions now have 3 (or 4, see below) overloads:

1. `parse(InputType)`: This will go into the specialized input adapters for `FILE`, `istream` or null-terminated string. As a fallback, it will  perform a std-using ADL-lookup on `begin(input)` and `end(input)` to invoke 2.

2. `parse(iterator, iterator)`: This will instantiate an input adapter for that specific iterator type. This will also handle all of the contiguous memory scenarios, (using `const char*` as the iterator type). The compiler has enough information to optimize this appropriately. If the `value_type` of the iterator is a multi-byte character type, wchar to char conversion will be performed.

3. `parse(span_input_adapter)`: This exists specifically to support the existing `parse({ptr, len})` syntax, but is marked as deprecated, as it's redundant with 2. apart from minor syntactic sugar.

User defined input is now a matter of implementing an arbitrary iterator type. It can also be done by having `begin(T)` and `end(T)` return an existing iterator type (like `const char*`)

From a very pedantic POV, this might be "technically" API breaking, because a lot of the SFINAE stuff that was in place to ensure that data is continuous in memory are now gone. Code that would have failed to compile will now result in functional, but not as optimized as it could be program. i.e. If someone tries to parse a `std::list<char>`, it'll work, and this may or may not be what was desired.

* * *

## Pull request checklist

- [X] [Code coverage](https://coveralls.io/github/nlohmann/json) is 100%. Test cases can be added by editing the [test suite](https://github.com/nlohmann/json/tree/develop/test/src).


include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -316,111 +323,92 @@ class wide_string_input_adapter
...
+           using iterator_type = IteratorType;
+           using char_type = typename std::iterator_traits<iterator_type>::value_type;
+           using base_adapter_type = iterator_input_adapter<iterator_type>;
+           using adapter_type = wide_string_input_adapter<base_adapter_type, char_type>;
+
+           static adapter_type create(IteratorType begin, IteratorType end)
 {
-    return input_buffer_adapter(reinterpret_cast<const char*>(b), l);
+    return adapter_type(base_adapter_type(std::move(begin), std::move(end)));
 }
```
dota17, line 353: Need to adjust brace style.

FrancoisChabot: Unfortunately, this was caused by astyle when I ran `make pretty` (that's not really a valid excuse, I should have checked. I'll try and rejigger the code a bit so that it doesn't confuse astyle anymore, but I'm not super optimistic.

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -316,111 +323,92 @@ class wide_string_input_adapter
...
-template<typename CharT,
-         typename std::enable_if<
-             std::is_pointer<CharT>::value and
-             std::is_integral<typename std::remove_pointer<CharT>::type>::value and
-             sizeof(typename std::remove_pointer<CharT>::type) == 1,
-             int>::type = 0>
-input_buffer_adapter input_adapter(CharT b, std::size_t l)
+template<typename IteratorType>
+struct iterator_input_adapter_factory<IteratorType,
+       typename std::enable_if<(sizeof(typename std::iterator_traits<IteratorType>::value_type)>1)>::type >
```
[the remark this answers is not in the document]
FrancoisChabot: Fixed

include/nlohmann/detail/input/input_adapters.hpp
FrancoisChabot: Fixed.

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -269,20 +280,20 @@ struct wide_string_input_helper<WideStringType, 2>
     }
 };
 
-template<typename WideStringType>
+// Wraps another input apdater to convert wide character types into individual bytes.
+template<typename BaseInputAdapter, typename WideCharType>
 class wide_string_input_adapter
 {
```
dota17, line 286: I am thinking if we could use more general name, like `wide_char_xxx` or `wide_byte_xxx`?
PS: never mind. Basically, I think all your changes are amazing!

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -247,9 +260,9 @@ struct wide_string_input_helper<WideStringType, 2>
             }
             else
             {
-                if (current_wchar < str.size())
+                if (!input.empty())
```
dota17, line 263: maybe forget to use `JSON_HEDLEY_UNLIKELY`?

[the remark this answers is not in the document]
FrancoisChabot: oops. sorry, not used to this at all :)

FrancoisChabot: good catch

nlohmann:
Thanks so much!!!

nlohmann:
The code looks great - I may add some more examples and tests in the next days.

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [x] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Input adapters are used-extendible. #1813 #2145

* * *

PR #2158, opened 2020-06-02
Fix PR#1006

dota17:
PR #1006  should made some change comment in the code.
and its test case need to change a bit: remove duplicate test items `list`, replaced by `object` item, the data comes from its original old commit.

### DIFF
--- a/include/nlohmann/detail/output/output_adapters.hpp
+++ b/include/nlohmann/detail/output/output_adapters.hpp
@@ -97,7 +97,8 @@ class output_string_adapter : public output_adapter_protocol<CharType>
   private:
     StringType& str;
 };
-
+/// dump to alternative string type, as defined in basic_json template
+/// see https://github.com/nlohmann/json/pull/1006
 template<typename CharType, typename StringType = std::basic_string<CharType>>
 class output_adapter
 {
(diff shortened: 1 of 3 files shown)


include/nlohmann/detail/output/output_adapters.hpp
```diff
@@ -97,7 +97,8 @@ class output_string_adapter : public output_adapter_protocol<CharType>
   private:
     StringType& str;
 };
-
+/// dump to alternative string type, as defined in basic_json template
```
nlohmann, line 100: Please remove this comment. It's technically correct, but does not help the user or developer.

dota17: Ok, fixed.

include/nlohmann/detail/output/output_adapters.hpp
```diff
@@ -97,7 +97,6 @@ class output_string_adapter : public output_adapter_protocol<CharType>
   private:
     StringType& str;
 };
-
```
nlohmann, line 100: Sorry to be pedantic, but can you please re-add the line?

dota17: Oh, sorry, will do now.

nlohmann:
Thanks!

PR #2160, opened 2020-06-03
[Oss-Fuzz] Adding parse_stl_fuzzer

tanuj208:
I went through the codebase and added a new fuzz target - parse_stl_fuzzer to improve the coverage of the existing fuzz targets.
By doing so, the coverage goes up by
Line coverage: 284 lines
Function coverage: 31 functions
Region coverage: 89 regions

[Coverage report](https://imgur.com/gallery/mhk6ckT)

Let me know if there is any incorrect usage of the API


nlohmann:
I may not understand the issue completely, but that change does not really increase the coverage of the parsers, but only adds some more constructors. In the end, any JSON input will be processed by the same parser - regardless whether it was read from a file, a string, a vector, etc.

How did you come up with the coverage numbers? Do you have insights into fuzzing? I never digged too deep into this.

nlohmann:
So far, all value from the fuzzers were to detect issues during parsing. I do not see value in extending the fuzzing to the input adapters, because they can be unit tested 100%. Though, technically, we would just wasting Google's resources here, I'd rather improve the coverage of the existing fuzz targets if it is not 100% yet.

tanuj208:
@nlohmann and @francoischabot thank you very much for your comments. I apologize for the wrong and confusing name of the fuzz target and lack of explanation. I have updated this pull request with more explanation about the functions that this fuzz target tests. I have also added how did I come up with the coverage stats.

I am an intern at Google and writing/improving fuzz target for this repo is a part of my internship project. I learned about fuzzing only a few days back.

I got the coverage stats for each of the functions in this fuzz target, and I found out that the improvement in coverage because of parsing or direct conversion from STL containers was very little, like you mentioned. The main improvement came because of loops in JSON map & JSON vectors, and because of methods like push_back and get.

I request you to please go through the pull request again and decide if it is of value.

nlohmann:
@tanuj208 That is cool to hear! I really appreciate the effort Google already put in checking this project with OSS-Fuzz.

I have a question: How exactly is the project integrated into OSS Fuzz? Do you just execute `make fuzzers` and then execute the generated binaries? I'm always afraid to break anything, hence I never touched these lines. It would be great to know more about this so I can document this properly for the future.

nlohmann:
I agree with @FrancoisChabot - however, exactly this check (parsing from a vector of bytes from OSS Fuzz) is already performed by `parse_afl_fuzzer`.

test/src/fuzzer-stl_operations.cpp
```diff
@@ -0,0 +1,140 @@
...
+    std::deque<uint8_t> deq(data, data + size);
+    std::list<uint8_t> lst(data, data + size);
+    std::forward_list<uint8_t> flist(data, data + size);
+    std::set<uint8_t> st(data, data + size);
+    std::unordered_set<uint8_t> uset(data, data + size);
+    std::multiset<uint8_t> multist(data, data + size);
+    std::unordered_multiset<uint8_t> umultiset(data, data + size);
+
+    // parsing from STL containers
+    json j_vector(vec);
```
nlohmann, line 43: This is a constructor call, not a parse call. Instead of interpreting the bytes in the containers to create a JSON value, the containers are interpreted as JSON array. As such, no real library code is tested, but just copy/move constructors from STL containers.

tanuj208:
@nlohmann and @FrancoisChabot, I understand what you were trying to explain and why the fuzz target I wrote cannot change code paths with different inputs. I will close this pull request for now. I am very thankful for all the quick replies.

@nlohmann regarding your question about OSS Fuzz, @oliverchang has more context, I will let him respond.

nlohmann:
If there is any way to improve the current fuzzing, e.g. with a grammar or so, please let us know!

oliverchang:
> @tanuj208 That is cool to hear! I really appreciate the effort Google already put in checking this project with OSS-Fuzz.
> 
> I have a question: How exactly is the project integrated into OSS Fuzz? Do you just execute `make fuzzers` and then execute the generated binaries? I'm always afraid to break anything, hence I never touched these lines. It would be great to know more about this so I can document this properly for the future.

Json is integrated into OSS-Fuzz [here](https://github.com/google/oss-fuzz/blob/master/projects/json/build.sh). We do essentially just execute `make fuzzers`.

PR #2164, opened 2020-06-04
Add clang support in JSON_HAS_CPP_17 macro

dota17:
Clang 3.4 and later implement all of the ISO C++2014 standard, 
and Clang 5.0 and later implement all of the ISO C++2017 standard,

### DIFF
--- a/include/nlohmann/detail/macro_scope.hpp
+++ b/include/nlohmann/detail/macro_scope.hpp
@@ -20,10 +20,15 @@
 #endif
 
 // C++ language standard detection
-#if (defined(__cplusplus) && __cplusplus >= 201703L) || (defined(_HAS_CXX17) && _HAS_CXX17 == 1) // fix for issue #464
+// fix for issue #464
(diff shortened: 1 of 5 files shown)


nlohmann:
I'm not sure whether this is correct. `JSON_HAS_CPP_17` is not to detect C++-17-compliant compilers, but to check if they actually compile with `-std=c++17`. If I understand you patch correctly, you assume any Clang 5+ will compile with C++17.

dota17:
I just wander when we defined `JSON_HAS_CPP_17`, we check it on GNU platform, and MSVC platform, why can't we check it on Clang platform.

jaredgrubb:
No,  this isn't required. `__cplusplus >= 201703L` is the correct way to check for C++17 and clang supports this.

nlohmann:
Even worse, the change breaks the build since the library uses C++17 code for Clang5+ though only C++11 is used. I think we can close the PR.

dota17:
Thank you all for your patience.

nlohmann:
No worries!

PR #2165, opened 2020-06-04
Add user-defined build option in CMake

dota17:
Add user-defined build option in CMake so that we can build the library with cmake like this:
> cmake  .. -DBUILD_CXX_STD=c++11

or
> cmake  .. -DBUILD_CXX_STD=c++14

or
> cmake  .. -DBUILD_CXX_STD=c++17

* * *

## Pull request checklist


nlohmann:
Why would we need this? The library only needs C++11, and if the client code compiles with a higher standard, that is fine. This flag would only make sense for our CI, right?

dota17:
@nlohmann  @FrancoisChabot  You are right!
The original reason I made this change is that when I debug something in my local, for example, I want to know  if `JSON_HAS_CPP_17` macro can really work, I have to complie the library with C++17, so I have to change `cxx_std_11` to `cxx_std_17` in CMakeLists.txt.
but now, as @FrancoisChabot explained, I can directly use `CMAKE_CXX_STANDARD` option, which I never know. So thank you all.
As this change may only make sense for CI, nor for the client, so I will close this.

PR #2176, opened 2020-06-06
C++20 support by removing swap specialization

gracicot:

In C++20, the rule for customization point has changed.

From [[namespace.std]/7](http://eel.is/c++draft/namespace.std#7):

> Other than in namespace `std` or in a namespace within namespace `std`, a program may provide an overload for any library function template designated as a customization point, provided that (a) the overload's declaration depends on at least one user-defined type and (b) the overload meets the standard library requirements for the customization point.[<sup>173</sup>](http://eel.is/c++draft/namespace.std#footnote-173)
[ *Note:* This permits a (qualified or unqualified) call to the customization point to invoke the most appropriate overload for the given arguments. — *end note* ]

This rule come from the paper [P0551: Thou Shalt Not Specialize std Function Templates!](wg21.link/p0551)

Swap is denoted as a customization point.

We can keep the specialization for older C++ version, but we should disable it for C++20 and above.

We also should provide a non-member swap function, ideally as a hidden friend.

This pull request implement those recommendations.

Please tell me if there's anything to fix or change in the patch.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/macro_scope.hpp
+++ b/include/nlohmann/detail/macro_scope.hpp
@@ -20,7 +20,11 @@
 #endif
 
 // C++ language standard detection
-#if (defined(__cplusplus) && __cplusplus >= 201703L) || (defined(_HAS_CXX17) && _HAS_CXX17 == 1) // fix for issue #464
+#if (defined(__cplusplus) && __cplusplus >= 202002L) || (defined(_MSVC_LANG) && _MSVC_LANG >= 202002L)
+    #define JSON_HAS_CPP_20
+    #define JSON_HAS_CPP_17
+    #define JSON_HAS_CPP_14
+#elif (defined(__cplusplus) && __cplusplus >= 201703L) || (defined(_HAS_CXX17) && _HAS_CXX17 == 1) // fix for issue #464
     #define JSON_HAS_CPP_17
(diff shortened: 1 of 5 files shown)


gracicot:
Overloading standard function inside the `std` namespace was never allowed. Specialization of functions inside the `std` namespace was allowed before C++20. This is a new C++20 restriction. The specialization was used if we typed `std::swap(json1, json2)`.

Without ADL, the patch will make such code pick the default `std::swap` implementation with C++20.

Although if everyone agrees, I will gladly remove the specialization entirely for all language version and promote the hidden friend instead.

jaredgrubb:
What would the deprecation say? Falling back to std::swap would probably work just fine right?

gracicot:
@jaredgrubb exactly. No code written today will suddenly stop working. It's just that the default swap implementation will be picked instead, and might be slightly less efficient since it uses the move constructor instead of member-wise swap calls.

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Remove swap specialization for C++20. #2176

* * *

Issue #2059, opened 2020-04-22
Library provides template arg for string_type but assumes std::string in some places


bcollins526:
I'm using json object with Electronic Arts' EASTL library, which provides high performance implementations of std containers. I use the following typedef:
```
    using stdjson = nlohmann::basic_json<std::map,
                                         std::vector,
                                         stdstring,
                                         bool,
                                         int64_t,
                                         uint64_t,
                                         double,
                                         jsonallocator,
                                         nlohmann::adl_serializer>;
```
where stdstring is eastl::string, which is their version of std::string. Then throughout my code, I just always use stdjson type. This works great, except for when json code assumes std::string. For example, this code:
```
   for (auto& cam : sjson["cameras"].items()) {
        stdstring cameraname(cam.key());
    }
```
fails to compile because of this code in iteration_proxy.hpp:
```
template<typename string_type>
void int_to_string( string_type& target, std::size_t value )
{
    target = std::to_string(value);
}
```
which assumes std::string is the string_type. I'm using Visual Studio 2019 on Windows 10 using the latest release of json.

ArtemSarmini:
It assumes that `string_type` is move-assignable or implicitly-move-constructible from `std::string` which is not true for `eastl::string`. To fix it we may add some sfinae to construct `string_type` either from `std::string` or from `std::to_string().data()` which I hope will return something meaningful for all `string_type`s we care of.
Another option is to introduce like `convertion_traits` in json template args which will handle such stuff. If one uses nonstandard string, one provides nonstandard traits, simple and neat. (It is also a possibility to consider opting in to use `std::from_chars`. ) Downside is that it is a very major change in api and requires quite lots of work: new code, new tests, new docs.
@nlohmann thoughts?

t-b:
> To fix it we may add some sfinae to construct string_type either from std::to_string(value) or from std::to_string(value).data() which I hope will return something meaningful for all string_types we care of.

That sounds like a good idea.

PR #2193, opened 2020-06-16
Fix consistency in function `int_to_string()`

dota17:
The function `int_to_string()` call `std::to_string()`, which will return std::string. For consistency, need to check `string_type` is `std::string`.

Refer: #2059

### DIFF
--- a/include/nlohmann/detail/iterators/iteration_proxy.hpp
+++ b/include/nlohmann/detail/iterators/iteration_proxy.hpp
@@ -12,7 +12,8 @@ namespace nlohmann
 {
 namespace detail
 {
-template<typename string_type>
+template<typename string_type, typename std::enable_if<
+             std::is_same<char, typename string_type::value_type>::value, int>::type = 0>
 void int_to_string( string_type& target, std::size_t value )
 {
     target = std::to_string(value);
--- a/include/nlohmann/detail/iterators/iteration_proxy.hpp
+++ b/include/nlohmann/detail/iterators/iteration_proxy.hpp
@@ -16,7 +16,9 @@ template<typename string_type, typename std::enable_if<
              std::is_same<char, typename string_type::value_type>::value, int>::type = 0>
 void int_to_string( string_type& target, std::size_t value )
 {
-    target = std::to_string(value);
+    // For ADL
+    using std::to_string;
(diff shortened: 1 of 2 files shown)


nlohmann:
What would be the previous behavior if `string_type::value_type` is not `char`? I understand now such types are excluded - but what would have happened earlier?

dota17:
> What would be the previous behavior if `string_type::value_type` is not `char`? I understand now such types are excluded - but what would have happened earlier?

If char_type is not char, there will be compiler error, the problem will arise before `int_to_string`, cause the library default string_type is std::string, default char_type is char. So now I might think I may not understand the issue correctly, and so far, I can't used the EASTL to reproduce the issue. The current solution is more of a temporary and partial solution, and more changes may be needed next.

nlohmann:
I am not sure whether this change is improving the situation. Instead, we may change the function to

```cpp
using std::to_string;
target = to_string(value);
```

to allow for ADL to choose the right function and the user-defined types to provide these functions.

dota17:
Thanks. This suggestion is very nice. 
And back to the original issue #2059, we may need more information from @bcollins526

bcollins526:
> 
> 
> Thanks. This suggestion is very nice.
> And back to the original issue #2059, we may need more information from @bcollins526

Sure just let me know what you need.

include/nlohmann/detail/iterators/iteration_proxy.hpp
```diff
@@ -12,10 +12,13 @@ namespace nlohmann
 {
 namespace detail
 {
-template<typename string_type>
+template<typename string_type, typename std::enable_if<
```
nlohmann, line 15: You can remove the `std::enable_if` part now. We now just assume that `to_string` will do the job of converting a `std::size_t` to `string_type`.

nlohmann:
Thanks!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Allow ADR inside iteration proxy. #2193

* * *

PR #2203, opened 2020-06-20
Use unsigned indizies for array index in json pointer

t-b:
The current code uses std::stoi to convert the input string to an int
array_index. This limits the maximum addressable array size to ~2GB on most
platforms.
    
But all callers immediately convert the result of array_index to
BasicJsonType::size_type.

So let's parse it as unsigned long long, which allows us to have as big arrays
as available memory. And also makes the call sites nicer to read.

### DIFF
--- a/include/nlohmann/detail/json_pointer.hpp
+++ b/include/nlohmann/detail/json_pointer.hpp
@@ -3,6 +3,7 @@
 #include <algorithm> // all_of
 #include <cassert> // assert
 #include <cctype> // isdigit
+#include <limits> // max
 #include <numeric> // accumulate
 #include <string> // string
 #include <utility> // move
@@ -328,9 +329,12 @@ class json_pointer
     @throw parse_error.106  if an array index begins with '0'
     @throw parse_error.109  if an array index begins not with a digit
     @throw out_of_range.404 if string @a s could not be converted to an integer
+    @throw out_of_range.410 if an array index exceeds size_type
     */
-    static int array_index(const std::string& s)
+    static typename BasicJsonType::size_type array_index(const std::string& s)
     {
+        using size_type = typename BasicJsonType::size_type;
+
         // error condition (cf. RFC 6901, Sect. 4)
         if (JSON_HEDLEY_UNLIKELY(s.size() > 1 and s[0] == '0'))
         {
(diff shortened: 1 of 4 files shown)


include/nlohmann/detail/json_pointer.hpp
```diff
@@ -344,10 +348,10 @@ class json_pointer
         }
 
         std::size_t processed_chars = 0;
-        int res = 0;
+        size_type res = 0;
         JSON_TRY
         {
-            res = std::stoi(s, &processed_chars);
+            res = std::stoull(s, &processed_chars);
```
nlohmann, line 354: Don't we need to cast the result of `std::stoull` to `size_type` here?

t-b: @nlohmann 

Yes and no. Yes casting will prevent a warning when `size_type` and `unsigned long long` don't coincide. But on the other hand if they don't coincide (32bit platform) size_type should be smaller than `unsigned long long` thus breaking it altogether.

Maybe I should do more something like the following?

```cpp

unsigned long long res = std::stoull(s, &processed_chars);

if(res >= static_cast<unsigned long long>(std::numeric_limits<size_type>::max()))
{
throw ...
}

return static_cast<size_type>(res);
````

Comment (nlohmann) [reply include/nlohmann/detail/json_pointer.hpp:354]: Yes, this looks good. I think throwing `json::out_of_range` would make sense. We don't have an exception id for this yet, so I would propose using 410, e.g. like

```cpp
JSON_THROW(out_of_range::create(410, "array index " + s + " exceeds size_type");
```

Can you make that change?

t-b: @nlohmann Done. Writing a test for it was a bit tricky though.

t-b: @nlohmann Done.

t-b:
@nlohmann The coverage decreased because that code only triggers when sizeof(size_type) < sizeof(unsigned long long) (aka 32bit) and it looks like coverage is calculated for 64-bit.

nlohmann:
I see. I tried to build a 32 bit binary, but the CI seems not to support this.

Could you add a comment before the `JSON_THROW` line that this code is only triggered in 32 bit mode, best with a link to this PR. Then add `//LCOV_EXCL_LINE` at the end of the `JSON_THROW` line.

t-b:
@nlohmann Done.

nlohmann:
Thanks a lot!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Use unsigned indices for array index in JSON Pointers. #2203

* * *

PR #2206, opened 2020-06-21
Simple ordered_json that works on all supported compilers

gatopeich:
Solved the Allocator issue that prevented using a container whose `value_type` is not the same as `std::map`'s.
The trick was in instantiating `AllocatorType` with `ObjectType::value_type`, instead of hard-coding to `std::pair<const Key, basic_json>`. This it stays the same for std::map but adapts to whatever other container we use in other specializations.

* * *

## Pull request checklist

## Please don't

[X] checked too 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -1527,7 +1527,7 @@ This library will not support comments in the future. If you wish to use comment
 
 ### Order of object keys
 
-By default, the library does not preserve the **insertion order of object elements**. This is standards-compliant, as the [JSON standard](https://tools.ietf.org/html/rfc8259.html) defines objects as "an unordered collection of zero or more name/value pairs". If you do want to preserve the insertion order, you can specialize the object type with containers like [`tsl::ordered_map`](https://github.com/Tessil/ordered-map) ([integration](https://github.com/nlohmann/json/issues/546#issuecomment-304447518)) or [`nlohmann::fifo_map`](https://github.com/nlohmann/fifo_map) ([integration](https://github.com/nlohmann/json/issues/485#issuecomment-333652309)).
+By default, the library does not preserve the **insertion order of object elements**. This is standards-compliant, as the [JSON standard](https://tools.ietf.org/html/rfc8259.html) defines objects as "an unordered collection of zero or more name/value pairs". If you do want to preserve the insertion order, you can try the new [`nlohmann::ordered_json`](https://github.com/nlohmann/json/issues/2179) specialization, or use a more sophisticated ordered map like [`tsl::ordered_map`](https://github.com/Tessil/ordered-map) ([integration](https://github.com/nlohmann/json/issues/546#issuecomment-304447518)) or [`nlohmann::fifo_map`](https://github.com/nlohmann/fifo_map) ([integration](https://github.com/nlohmann/json/issues/485#issuecomment-333652309)).
 
 ### Memory Release
 
--- a/README.md
+++ b/README.md
@@ -1527,7 +1527,9 @@ This library will not support comments in the future. If you wish to use comment
 
 ### Order of object keys
 
-By default, the library does not preserve the **insertion order of object elements**. This is standards-compliant, as the [JSON standard](https://tools.ietf.org/html/rfc8259.html) defines objects as "an unordered collection of zero or more name/value pairs". If you do want to preserve the insertion order, you can try the new [`nlohmann::ordered_json`](https://github.com/nlohmann/json/issues/2179) specialization, or use a more sophisticated ordered map like [`tsl::ordered_map`](https://github.com/Tessil/ordered-map) ([integration](https://github.com/nlohmann/json/issues/546#issuecomment-304447518)) or [`nlohmann::fifo_map`](https://github.com/nlohmann/fifo_map) ([integration](https://github.com/nlohmann/json/issues/485#issuecomment-333652309)).
+By default, the library does not preserve the **insertion order of object elements**. This is standards-compliant, as the [JSON standard](https://tools.ietf.org/html/rfc8259.html) defines objects as "an unordered collection of zero or more name/value pairs".
+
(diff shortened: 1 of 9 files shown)


gatopeich:
Almost there. Only missing default template arguments for our new ordered_map...

gatopeich:
Looks like GCC 4.9 on Ubuntu 14 cannot handle the partial template definition required for my current solution.
Not sure it is reasonable to try to support such obsolete platform?

nlohmann:
When I look at https://travis-ci.org/github/nlohmann/json/builds/700984944, there seem to be a lot of other compilers complaining. I restarted all failed builds, let's see.

gatopeich:
Using std::map allocator as a placeholder (acd748e) is the most old-compiler friendly way I could find for the recursive declaration of the Allocator. And it is still not working on some compilers.

Summarizing:
*  current implementation has `object_t::value_type` hardcoded to that of `std::map`: `std::pair<const Key, T>`
*

Let's try my second alternative: leave the `value_type` hardcoded

gatopeich:
Using std::map allocator as a placeholder (acd748e) is the most old-compiler friendly way I could find for the recursive declaration of the Allocator. And it is still not working on some compilers.

Summarizing:
* Current implementation has `object_t::value_type` hardcoded to that of `std::map`: `std::pair<const Key, T>`, cause `AllocatorType` is always specialized for that.
* Providing a non-hardcoded allocator seems impossible for certain compilers without potentially complex refactoring.

So I am trying the next alternative: Implement VecMap:erase that works with the same `value_type` as std:map...

nlohmann:
Any ideas how proceed here?

gatopeich:
I am quite frustrated about how it fails on GCC 4.9 (quite old anyway) and specially XCode 9.3 (not that old but I have no access to such build system).

For the time being we could restrict `ordered_json` to recent compilers?

_Fact is, if we cannot make this simple container work, the claim that users can roll their own seems ambitious... However the same applies: recent compiler requierement._

gatopeich:
Fixed for GCC 5.5, which was giving the same error as v4.9 on travis-ci.
After isolating the issue to one of constructor instantiation, I found a [cryptic reference to a retroactive change to the C++11 standard related to issues around constructor inheritance](https://en.cppreference.com/w/cpp/language/using_declaration#Defect_reports)...
So I declared explicit constructors instead, and GCC 5.5 seems happy now.

gatopeich:
Now it works on all environments.
Reviewing the comments and any unintended changes to white space etc...

gatopeich:
@nlohmann this is all yours now... 🙂

README.md
```diff
@@ -1527,7 +1527,7 @@ This library will not support comments in the future. If you wish to use comment
...
+By default, the library does not preserve the **insertion order of object elements**. This is standards-compliant, as the [JSON standard](https://tools.ietf.org/html/rfc8259.html) defines objects as "an unordered collection of zero or more name/value pairs". If you do want to preserve the insertion order, you can try the new [`nlohmann::ordered_json`](https://github.com/nlohmann/json/issues/2179) specialization, or use a more sophisticated ordered map like [`tsl::ordered_map`](https://github.com/Tessil/ordered-map) ([integration](https://github.com/nlohmann/json/issues/546#issuecomment-304447518)) or [`nlohmann::fifo_map`](https://github.com/nlohmann/fifo_map) ([integration](https://github.com/nlohmann/json/issues/485#issuecomment-333652309)).
```
nlohmann, line 1530:
```suggestion
By default, the library does not preserve the **insertion order of object elements**. This is standards-compliant, as the [JSON standard](https://tools.ietf.org/html/rfc8259.html) defines objects as "an unordered collection of zero or more name/value pairs".

If you do want to preserve the insertion order, you can try the type [`nlohmann::ordered_json`](https://github.com/nlohmann/json/issues/2179). Alternatively, you can use a more sophisticated ordered map like [`tsl::ordered_map`](https://github.com/Tessil/ordered-map) ([integration](https://github.com/nlohmann/json/issues/546#issuecomment-304447518)) or [`nlohmann::fifo_map`](https://github.com/nlohmann/fifo_map) ([integration](https://github.com/nlohmann/json/issues/485#issuecomment-333652309)).
```

include/nlohmann/json.hpp
```diff
@@ -496,8 +496,8 @@ class basic_json
     using object_t = ObjectType<StringType,
           basic_json,
           object_comparator_t,
-          AllocatorType<std::pair<const StringType,
-          basic_json>>>;
+          // Note: this forces object_t::value_type to match std::map's
```
nlohmann, line 499: I am not sure whether I understood this comment.

include/nlohmann/ordered_map.hpp
```diff
@@ -45,7 +54,14 @@ struct ordered_map : Container
         {
             if (it->first == key)
             {
-                Container::erase(it);
+                // Since we cannot move const Keys, re-construct them in place
+                for (auto next = it; ++next != this->end(); ++it)
+                {
+                    // *it = std::move(*next); // deleted
```
nlohmann, line 60: Do we need this commented code?

gatopeich: Not really, the comment above should be enough.

nlohmann: Can you remove it, please?

include/nlohmann/json.hpp
nlohmann: Please remove the comment and leave that part of the file as is.

include/nlohmann/ordered_map.hpp
gatopeich: Yes and I will try to address your comments today...

include/nlohmann/json.hpp
gatopeich: I added this comment to highlight the restriction imposed by an AllocatorType always specialized for `std::pair<const StringType, basic_json>`. I only managed to make `std::vector`  work through a tough workaround, but you may want to have a more flexible interface in the future.
To be flexible you could e.g. wrap `std::map` like this:
```
template <class Key, class T, class Comparator, template AllocatorTemplate>
class wrapped_map : public std::map<Key, T, Comparator, AllocatorTemplate<const Key, T>>
...
```

gatopeich: Anyway I removed the comment to move on with this :-)

include/nlohmann/ordered_map.hpp
nlohmann: The comment `// *it = std::move(*next); // deleted` is still there. Please remove.

gatopeich:
Ops! done now.

nlohmann:
Thanks a lot!!!

nlohmann:
I opened a new pull request #2258 to bring the feature to the main branch. Thanks so much for your patience and bringing this forward!

PR #2222, opened 2020-06-26
Enable CMake policy CMP0077

alexreinking:
Projects that import json via [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) or `add_subdirectory` pointed at a git submodule may want to set `JSON_BuildTests` to "NO". However, this doesn't work without creating an identical `option()` in the importing project. Enabling CMP0077 in CMake 3.13+ changes the behavior of `option()` to allow importing projects to set default values for the variables without touching the cache.

See the documentation for CMP0077 here: https://cmake.org/cmake/help/latest/policy/CMP0077.html

* * *

## Pull request checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,12 @@ include(ExternalProject)
 ##
 ## OPTIONS
 ##
+
+if (POLICY CMP0077)
+    # Allow CMake 3.13+ to override options when using FetchContent / add_subdirectory.
+    cmake_policy(SET CMP0077 NEW)
+endif ()
+
 option(JSON_BuildTests "Build the unit tests when BUILD_TESTING is enabled." ON)
 option(JSON_Install "Install CMake targets during install step." ON)
 option(JSON_MultipleHeaders "Use non-amalgamated version of the library." OFF)


t-b:
@alexreinking Can you clarify why that is needed?

I'm doing here

```
SET(JSON_BuildTests OFF CACHE INTERNAL "")
SET(JSON_Install OFF CACHE INTERNAL "")
ADD_SUBDIRECTORY(json)
```

and the JSON tests are not built and I also don't define these options. My cmake version is 3.15.2 on windows.

alexreinking:
The key is not touching the cache. The snippet you posted will no longer set JSON_BuildTests to OFF if I explicitly set it to "ON" in the cache. Additionally, `option()` will _not_ create a cache entry at all with CMP0077 enabled and the normal variable set.

> In CMake 3.13 and above the option() command prefers to do nothing when a normal variable of the given name already exists. It does not create or update a cache entry or remove the normal variable. The new behavior is consistent between the first and later runs in a build tree. This policy provides compatibility with projects that have not been updated to expect the new behavior.

Without the policy, the following does not work, but should, and is the default behavior of CMake going forward:

```
set(JSON_BuildTests OFF)
set(JSON_Install OFF)
add_subdirectory(json)
```

nlohmann:
Thanks!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Enable CMake policy CMP0077. #2222

* * *

t-b:
@alexreinking Thanks for the clarification. Makes sense!

PR #2229, opened 2020-06-28
Fix conversion from empty json to std::optional<>

karzhenkov:
Here is possible solution (or workaround) for issues in PR #2117, #2213.
The problem is caused by "wrong" user defined conversion selected by overload resolution. Compiler selects a converting constructor for `std::optional<std::string>`, while ` template <...> json::operator ValueType` is needed. This behavoir is correct because `json` can be implicitly converted to `std::string`. The constructor is unaware about `from_json` function.
"Non-template" version of conversion operator presented here "adjusts" the resolution process. However, the desired result is achieved for copy initailization only. Direct initialization still selects the converting constructor (this probably cannot be changed).

### DIFF
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -169,6 +169,14 @@
 #include <utility> // pair, declval
 #include <valarray> // valarray
 
+#ifdef JSON_HAS_CPP_17
+    #if __has_include(<optional>)
+        #include <optional>
+    #elif __has_include(<experimental/optional>)
+        #include <experimental/optional>
+    #endif
+#endif
+
 // #include <nlohmann/detail/exceptions.hpp>
 //     __ _____ _____ _____
 //  __|  |   __|     |   | |  JSON for Modern C++
@@ -4603,6 +4611,21 @@ inline void from_json(const BasicJsonType& j, typename std::nullptr_t& n)
     n = nullptr;
 }
 
+#ifdef JSON_HAS_CPP_17
+template<typename BasicJsonType, typename T>
+void from_json(const BasicJsonType& j, std::optional<T>& opt)
+{
+    if (j.is_null())
+    {
+        opt = std::nullopt;
+    }
+    else
+    {
+        opt = j.template get<T>();
+    }
+}
+#endif
+
 // overloads for basic_json template parameters
 template < typename BasicJsonType, typename ArithmeticType,
            enable_if_t < std::is_arithmetic<ArithmeticType>::value&&
@@ -5334,6 +5357,14 @@ class tuple_element<N, ::nlohmann::detail::iteration_proxy_value<IteratorType >>
 // #include <nlohmann/detail/value_t.hpp>
 
 
+#ifdef JSON_HAS_CPP_17
+    #if __has_include(<optional>)
+        #include <optional>
+    #elif __has_include(<experimental/optional>)
+        #include <experimental/optional>
+    #endif
+#endif
+
 NLOHMANN_JSON_NAMESPACE_BEGIN
 namespace detail
 {
@@ -5569,6 +5600,22 @@ struct external_constructor<value_t::object>
 /////////////
 // to_json //
 /////////////
+
+#ifdef JSON_HAS_CPP_17
+template<typename BasicJsonType, typename T,
+         enable_if_t<std::is_constructible<BasicJsonType, T>::value, int> = 0>
+void to_json(BasicJsonType& j, const std::optional<T>& opt)
+{
+    if (opt.has_value())
+    {
+        j = *opt;
+    }
+    else
+    {
+        j = nullptr;
+    }
+}
+#endif
 
 template<typename BasicJsonType, typename T,
          enable_if_t<std::is_same<T, typename BasicJsonType::boolean_t>::value, int> = 0>
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -4678,7 +4678,10 @@ class optional
         return std::move(base_value);
     }
 
-    constexpr optional() noexcept(noexcept(std::optional<noexcept_fix_t>())) = default;
+    constexpr optional() noexcept(noexcept(std::optional<noexcept_fix_t>()))
+        : base_value() // explicitly initialized to mitigate -Werror=effc++
+    {
+    }
 
     constexpr optional(std::nullopt_t /* unused */) noexcept
         : base_value(std::nullopt)
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -4771,7 +4771,6 @@ template <typename T> const std::optional<T>& cmp_val(const optional<T>& v)
 #define JSON_OPTIONAL_COMPARISON(OP, LHS, RHS) \
     template <typename A, typename B> \
     auto operator OP (const LHS& lhs, const RHS& rhs) \
-    noexcept(noexcept(JSON_OPTIONAL_COMPARISON_EXPR(OP))) \
     -> decltype(JSON_OPTIONAL_COMPARISON_EXPR(OP)) \
     { \
         return JSON_OPTIONAL_COMPARISON_EXPR(OP); \
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -4663,17 +4663,17 @@ class optional
 
   public:
 
-    base_type& base() &
+    base_type& base() & noexcept
     {
         return base_value;
     }
 
-    const base_type& base() const &
+    const base_type& base() const & noexcept
     {
         return base_value;
     }
 
-    base_type&& base() &&
+    base_type&& base() && noexcept
     {
         return std::move(base_value);
     }
@@ -4742,8 +4742,8 @@ class optional
     constexpr T* operator ->() noexcept { return base_value.operator ->(); }
     constexpr const T* operator ->() const noexcept { return base_value.operator ->(); }
 
-    operator base_type& () & { return base_value; }
-    operator base_type&& () && { return std::move(base_value); }
+    operator base_type& () & noexcept { return base_value; }
+    operator base_type&& () && noexcept { return std::move(base_value); }
 
     // *INDENT-ON*
 };
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -4599,15 +4599,15 @@ NLOHMANN_JSON_NAMESPACE_END
 #include <optional>
 #include <utility>
 
-namespace nlohmann
-{
+NLOHMANN_JSON_NAMESPACE_BEGIN
 
 template <typename T>
-class optional : public std::optional<T>
+class optional
 {
     // *INDENT-OFF*
 
     using base_type = std::optional<T>;
+    using value_type = T;
 
     template <typename U, typename = optional>
     struct has_conversion_operator : std::false_type { };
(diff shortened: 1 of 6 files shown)


karzhenkov:
Is the following really needed?
```
...\test\src\unit-concepts.cpp(124): ERROR: CHECK( std::is_standard_layout<json>::value ) is NOT correct!
```

nlohmann:
That assertion held true so far. If that changes now, this may indicate a breaking change.

karzhenkov:
Modified `json` may have two (or more) base classes of the same type, and thus not to be `StandardLayout`. This may happen if there are identical types in argument pack for `optional_conversion_helper`. Possible effects need to be explored.

karzhenkov:
There are no identical bases. Moreover, direct identical bases are not allowed.

dota17:
```
Direct initialization still selects the converting constructor (this probably cannot be changed).
```
As i said in #2213, this is why an exception is raised when calling `std::optional<std::string>(j_null)`. It is a problem of `std::optional` not this json library, i think.
IMO, we can add notes in documents to tell people that they should not use `std::optional<std::string>(j_null)` directly.

test/src/unit-conversions.cpp
```diff
@@ -1692,3 +1700,74 @@ TEST_CASE("JSON to enum mapping")
...
+#ifdef JSON_HAS_CPP_17
+TEST_CASE("std::optional")
+{
+    SECTION("null")
+    {
+        json j_null;
+        std::optional<std::string> opt_null;
+
+        CHECK(json(opt_null) == j_null);
+        CHECK_THROWS_WITH(std::optional<std::string>(j_null),
```
dota17, line 1713: Weird. In my local env, the exception would not raise. By from CI log, it will get the exception.

nlohmann:
> IMO, we can add notes in documents to tell people that they should not use `std::optional<std::string>(j_null)` directly if it still raise an exception.

I do not like this approach. I cannot believe this is an issue in `std::optional`. It feels like a bug in the library. If we cannot properly support a quite frequent type like `std::optional<std::string>`, then maybe we should not add support for `std::optional` to the library in the first place as this would be highly confusing.

bhardwajs:
> 
> 
> > IMO, we can add notes in documents to tell people that they should not use `std::optional<std::string>(j_null)` directly if it still raise an exception.
> 
> I do not like this approach. I cannot believe this is an issue in `std::optional`. It feels like a bug in the library. If we cannot properly support a quite frequent type like `std::optional<std::string>`, then maybe we should not add support for `std::optional` to the library in the first place as this would be highly confusing.

Are there specific STL implementations where the issue with `std::optional` is surfacing? gcc, clang, and MSVC don't always have the same implementations.

test/src/unit-conversions.cpp
karzhenkov: What is special in your local environment?

karzhenkov:
It seems the issue is caused by `std::optional` intented desing and language standard, not by implementation features.
Perhaps it would be better to disable implicit conversions of `json` to `std::string`, `bool`, etc. When doing so, the overload resolution will prefer conversion operator defined in `json` over converting constructor of `std::optional`. The conversion operator will call the proper `from_json` function. But, оf course, this is a breaking change.
On the other hand, it may be sufficient to 'disable' direct initialization by throwing an exception. A bit ugly, but maybe not so bad.

karzhenkov:
MSVC has bad `std::is_standard_layout`. Here is a [sample](https://godbolt.org/z/U6DJ3Z).
By the way, I see no reason why _StandardLayoutType_ concept may be useful when using `json`.

test/src/unit-conversions.cpp
dota17: ubuntu 18.04
gcc 7.4
cmake 3.10.2

dota17:
What is more, `std::optional<std::string>(j_null)` still raises an exception.
I added logs to check `SECTION("string")` and `bool`, `number`. Unfortunally, I found that they still didn't call the expected method - `from_json(basic_json, std::optional)`.

test/src/unit-conversions.cpp
karzhenkov: Somewhat has changed since GCC 7.4
https://godbolt.org/z/4MfL8i

test/src/unit-concepts.cpp
```diff
@@ -121,7 +121,27 @@ TEST_CASE("concepts")
...
+
+                struct D: C
+                {
+                    int v;
+                };
+
+                constexpr operator bool() const
+                {
+                    // should return true
+                    return std::is_standard_layout<D>::value;
```
karzhenkov, line 140: Microsoft doesn't consider their implemenation wrong. Their [story](https://developercommunity.visualstudio.com/content/problem/753436/msvs2017-incorrect-size-layout-and-first-data-memb.html) is about empty base optimization, backward binary compatibility and non-standard extenstion `__declspec(empty_bases)` which can be used to achieve standard behavior.

test/src/unit-conversions.cpp
karzhenkov: There is probably a bug in GCC 7. When used with "-std=c++17" it may perform wrong overload resolution for user-defined conversion. In previous example and also [here](https://godbolt.org/z/kWTWx9) (with no `std::optional` involved) it behaves as if the implicit copy constructor is selected, whereas there is a more suitable template constructor. Other major releases of GCC (including older versions) prefer template constructor.

nlohmann:
Same question as with #2117 - how to proceed here?

karzhenkov:
Here is a "better" `optional` that probably allows to achieve the desired behavior. The `nlohmann::optional` class is derived from its standard counterpart and defines some constructors. The goal of these definitions is to change the preferred conversion performed by converting constructor. If the object being converted has `operator optional<T>`, this operator will be called by corresponding constructor. Other constructors are defined to resemble `std::optional` in any other initialization scenario.

nlohmann:
@karzhenkov Can you please update to the latest develop branch?

karzhenkov:
Rebased, but there is a test failure. Codacy test is probably excessively strict. It requires single-argument constructors to be explicit, but even `std::optional` doesn't meet this. The implicit conversions that the standard implementation allows are also allowed for `nlohmann::optional`. In fact, one of constructors in question is conditionally explicit, which is also replicated. Is there a possibility to relax codacy requirements?

Here is also a [simple example](https://godbolt.org/z/6Gq9661cn) to consider.

nlohmann:
Don't worry about Codacity - there is also another test failing: `ci_single_binaries` fails, because header `optional.hpp` is not self-contained and seems not to include `<optional>`.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -36,6 +36,10 @@
     #define JSON_HAS_CPP_11
 #endif
 
+#ifdef JSON_HAS_CPP_17
```
nlohmann, line 39: Please do not move this include to this header. Rather, move it to all headers where `<optional>` is actually needed.

nlohmann (changes requested):

karzhenkov:
I hope the includes are arranged properly now. However, `ci_single_binaries` is cancelled. This is propably caused by wrong test environment, not by actual problem in the code. The test log contains the [following](https://github.com/nlohmann/json/pull/2229/checks?check_run_id=3454885602#step:5:18):
```
/bin/sh: 1: SCAN_BUILD_TOOL-NOTFOUND: not found
```

nlohmann:
Oh, I had to make some changes to the CI. See #2981. I will merge this soon. Sorry for this!

nlohmann:
Done. Please update your branch. This should fix the CI error.

gregmarr:
@jarni-ua You are correct that this doesn't address that particular case.  This just addresses using a `std::optional<T>` as the value half of that.  It accepts and produces this format:

    json::parse("{ \"name\" : \"foo\", \"value\": null }");

gregmarr:
@nlohmann Anything still holding this up?  There is a codacy warning about two constructors not being explicit in the `optional` wrapper class, but they're not explicit in `std::optional` either.

gregmarr:
@karzhenkov This PR is conflicting again.  Can you update?
@nlohmann Anything besides the merge conflict still outstanding here?

tests/src/unit-optional.cpp
```diff
@@ -0,0 +1,159 @@
+#include "doctest_compatibility.h"
+
+#include <nlohmann/json.hpp>
+
+#if (defined(__cplusplus) && __cplusplus >= 201703L) || (defined(_HAS_CXX17) && _HAS_CXX17 == 1) // fix for issue #464
+    #define JSON_HAS_CPP_17
+    #define JSON_HAS_CPP_14
+#elif (defined(__cplusplus) && __cplusplus >= 201402L) || (defined(_HAS_CXX14) && _HAS_CXX14 == 1)
+    #define JSON_HAS_CPP_14
+#endif
```
falbrechtskirchinger, line 10: This shouldn't be needed. `JSON_HAS_CPP_*` macros from the library are now available to the unit tests.

falbrechtskirchinger (commented):
I'd like to evaluate this branch in one of my projects to make sure there aren't any ugly side effects. Afterward, I'll come back for a more in-depth review.

tests/src/unit-conversions.cpp
```diff
@@ -51,15 +51,15 @@ using nlohmann::json;
 // NLOHMANN_JSON_SERIALIZE_ENUM uses a static std::pair
 DOCTEST_CLANG_SUPPRESS_WARNING_PUSH
 DOCTEST_CLANG_SUPPRESS_WARNING("-Wexit-time-destructors")
+DOCTEST_CLANG_SUPPRESS_WARNING("-Wunused-macros")
```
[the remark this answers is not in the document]
karzhenkov: I considered it, but I think it's hardly is better. I preferred a slightly more concise variant.

single_include/nlohmann/json.hpp
```diff
@@ -4344,9 +4344,9 @@ class optional : public std::optional<T>
         return *this;
     }
 
-    constexpr optional() noexcept = default;
+    constexpr optional() noexcept(noexcept(std::optional<T>())) = default;
 
-    constexpr optional(std::nullopt_t) noexcept
+    constexpr optional(std::nullopt_t _) noexcept
```
falbrechtskirchinger, line 4349: Try `optional(std::nullopt_t /*unused*/)` if clang-tidy, etc. complain.

karzhenkov:
It's not clear if `std::optional` has `noexcept` default constructor in GCC 8.1. Actually, it isn't declared `noexcept` there. This can be checked in [the example](https://godbolt.org/z/v5vE1bcbh) at godbolt. However, after the local variable definition in the example is uncommented, `std::optional` default constructor looks `noexcept`. I guess, there are both library and compiler bugs.

single_include/nlohmann/json.hpp
karzhenkov: My fix doesn't work. Eventually, I replaced the unused names with comments.

karzhenkov:
The remaining problems are related to three-way comparisons in C++20. I hope to fix them when I have time.

falbrechtskirchinger:
That failure is an ICE. Are you sure it's not just a GCC 13 issue? I'll check with GCC 12 (and I'm building a more current GCC 13).

karzhenkov:
The problem boils down to the following example, which complies in C++17 but not in C++20:
```c++
#include <optional>

struct opt : std::optional<int> {};

bool operator == (const opt&, const opt&);
bool operator >= (const opt&, const opt&);

int main()
{
    sizeof(opt() == opt());
    sizeof(opt() >= opt());
}
```
If any of the operator declarations is removed along with the corresponding check, the code compiles regardless of the language standard.

falbrechtskirchinger:
Digging a little bit deeper, the code triggers a recursion in `__partially_ordered_with` via `std::operator <=>`.
For `std::optional`, that recursion is broken via the constraint `__is_optional_v`.

Adding a specilization to `std` (for testing purposes only, of course!) makes the code compile without issue.
```cpp
namespace std {
template<>
inline constexpr bool __is_optional_v<opt> = true;
}
```

How do we work around this?

karzhenkov:
> How do we work around this?

It looks like a defect in the C++20 standard (or maybe in the standard library implementation). I think it is impossible to correctly define comparisons for a class derived from `std::optional` in C++20. My initial approach was to adjust the class constructors, but that doesn't seem viable now. Another option is to completely redefine the class, which also has disadvantages. All functionality needs to be duplicated and maintained; the issues of the interchangeability needs to be solved.

PR #2241, opened 2020-07-06
add travis task : compile and test under C++17

dota17:
From #2213, we could know that `JSON_HAS_CPP_17` is false in all travis jobs.
Thus, we should add a task to compile and test under CXX17.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -218,6 +218,17 @@ matrix:
         sources: ['ubuntu-toolchain-r-test']
         packages: ['g++-9', 'ninja-build']
 
+  - os: linux
+    compiler: gcc
+    env:
+      - COMPILER=g++-9
+      - CXXFLAGS=-std=c++17
+      - CMAKE_CXX_STANDARD=17
+    addons:
+      apt:
+        sources: ['ubuntu-toolchain-r-test']
+        packages: ['g++-9', 'ninja-build']
+
   # Linux / Clang
 
   - os: linux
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -323,7 +334,11 @@ script:
 
   # compile and execute unit tests
   - mkdir -p build && cd build
-  - cmake .. ${CMAKE_OPTIONS} -DJSON_MultipleHeaders=${MULTIPLE_HEADERS} -DJSON_BuildTests=On -GNinja && cmake --build . --config Release
+  - if [[ "${CMAKE_CXX_STANDARD}" != "" ]]; then
+    cmake .. ${CMAKE_OPTIONS} -DJSON_MultipleHeaders=${MULTIPLE_HEADERS} -DCMAKE_CXX_STANDARD=${CMAKE_CXX_STANDARD} -DJSON_BuildTests=On -GNinja && cmake --build . --config Release ;
```
nlohmann, line 338: I think it may be clearer to set `CMAKE_CXX_STANDARD` by default to `11` (if I understand the Travis YAML correctly, this needs to be done at the top of the file) and only override it to `17` where a different value is needed. This avoids this `if` here, and also makes it easier to add similar tests for `14` and `20`.

.travis.yml
```diff
@@ -218,6 +218,17 @@ matrix:
         sources: ['ubuntu-toolchain-r-test']
         packages: ['g++-9', 'ninja-build']
 
+  - os: linux
```
nlohmann, line 221: We already have a test case above that sets `CXXFLAGS=-std=c++2a`. Maybe this can be reused.

.travis.yml
```diff
@@ -218,6 +218,17 @@ matrix:
...
+    compiler: gcc
+    env:
+      - COMPILER=g++-9
```
nlohmann, line 224: Please also add similar tests for Clang.

.travis.yml
dota17: Unfortunately, It seems that `CXXFLAGS=-std=c++1z` didn't work under c++17.

.travis.yml
dota17: Fine, I will do it

nlohmann: Why did you close the PR?

dota17:
Reopening.
I've made a little investigation. 
In GCC,  the -std= command-line option have C++98, C++11, C++14, C++17, C++2a, so c++2a is different from c++17.
In Clang, the same option are the same, but use -std=c++1y in Clang 3.4 and earlier.

dota17:
At the same time, I think we should update Ubuntu distribution in .travis.yml, because `Precise` and `Trusty` are EOL by Canonial.

nlohmann:
> At the same time, I think we should update Ubuntu distribution in .travis.yml, because `Precise` and `Trusty` are EOL by Canonial.

As long as they are supported by Travis, I am fine with the working setup. I am afraid that the CI breaks if we are moving to a newer version.

.travis.yml
karzhenkov: `CXXFLAGS=-std...` is overriden by compile feature `cxx_std_11` in top-level `CMakeLists.txt` (and by `CMAKE_CXX_STANDARD`, if it is present; see [comment](https://github.com/nlohmann/json/pull/2117#issuecomment-676597609) in #2117)

PR #2242, opened 2020-07-06
Make assert configurable via JSON_ASSERT

### LINKED ISSUE #2239: Support of replacing assert with something else
The library uses `assert` in certain places to signal about fatal errors.
An example:
```
    void assert_invariant() const noexcept
    {
        assert(m_type != value_t::object or m_value.object != nullptr);
        assert(m_type != value_t::array or m_value.array != nullptr);
        assert(m_type != value_t::string or m_value.string != nullptr);
        assert(m_type != value_t::binary or m_value.binary != nullptr);
    }
```
Such kind of signal isn't friendly (when debugging, the code just crashes) if compared to `_ASSERT` (from `<crtdbg.h>`, which is available for Microsoft C++ compiler) or similar macros (like https://github.com/foonathan/debug_assert).
Please, provide a way to use another macro instead of `assert`.

It is not possible to globally redefine `assert` for the library, because it includes `<cassert>` directly.

- Compiler: Microsoft C++ compiler
- Operating system: Windows 7 64-bit

- [x] latest release version 3.7.3
- [ ] the `develop` branch

nlohmann:
This PR introduces a macro `JSON_ASSERT(x)` which defaults to `assert(x)`, but can be replaced by any other implementation.

Closes #2239.


nlohmann:
It seems there is a memory issue - Valgrind and ASAN are complaining about the new test case.

gregmarr:
I think any `assert` still needs to terminate the process somehow, or you're going to have issues with the algorithms not working properly because the preconditions are violated.  That means that you can't really test that an assert fires.

nlohmann:
> I think any `assert` still needs to terminate the process somehow, or you're going to have issues with the algorithms not working properly because the preconditions are violated. That means that you can't really test that an assert fires.

I think I chose a bad function to test - one where a non-aborting assertion left a non-void function without return value. I now chose a different function where first an assertion fires and even if it does not abort an exception is thrown.

nlohmann:
(The test fails if exceptions are switched off...)

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Macro `JSON_ASSERT(x)` can be used to replace `assert(x)` calls. #2239 #2242

* * *

PR #2244, opened 2020-07-06
Tag binary values in cbor if set

matthewbauer:
CBOR has tags, which work similarly to "subtype"s:

https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml

Unsure if this makes sense. Note that the subtype must just be one
byte wide.

* * *

## Pull request checklist

### DIFF
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -279,6 +279,12 @@ class binary_writer
 
             case value_t::binary:
             {
+                if (j.m_value.binary->has_subtype())
+                {
+                    write_number(static_cast<std::uint8_t>(0xd8));
+                    write_number(j.m_value.binary->subtype());
+                }
+
                 // step 1: write control byte and the binary array size
                 const auto N = j.m_value.binary->size();
                 if (N <= 0x17)
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -12641,6 +12641,12 @@ class binary_writer
 
             case value_t::binary:
             {
+                if (j.m_value.binary->has_subtype())
+                {
+                    write_number(static_cast<std::uint8_t>(0xd8));
+                    write_number(j.m_value.binary->subtype());
+                }
+
                 // step 1: write control byte and the binary array size
                 const auto N = j.m_value.binary->size();
                 if (N <= 0x17)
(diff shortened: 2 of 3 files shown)


include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -279,6 +279,12 @@ class binary_writer
 
             case value_t::binary:
             {
+                if (j.m_value.binary->has_subtype())
+                {
+                    write_number(static_cast<std::uint8_t>(0xd8));
```
nlohmann, line 284: What does `0xd8` mean?

matthewbauer: It's a single byte data tag.

> 0xd8..0xdb : more tagged items, 1/2/4/8 bytes and then a data item follow

from https://tools.ietf.org/html/rfc7049

if we want to support bigger than 1 byte, it would be 0xd9, 0xda, 0xdb instead.

matthewbauer:
Note this only works for encoding subtypes, not decoding. Decoding is hard because the tag can come for any data type - not just binary. So we would need to add more state to handle it properly.

Ericson2314:
For context, here is an issue from a Rust library about this exact issue https://github.com/pyfisch/cbor/issues/157 . Serde also wasn't originally made with CBOR tags in mind, so again the sticking point is derialization without polluting everything else with CBOR-specific concepts.

nlohmann:
As the library does not support tags in the moment (#1968), it would fail to parse its own output, or am I mistaken?

matthewbauer:
> As the library does not support tags in the moment (#1968), it would fail to parse its own output, or am I mistaken?

Yes - you wouldn't be able to parse a cbor with a subtype since it would have that tag.

Ericson2314:
Yes this change just attempts serialization for now, as a stop-gap which is good enough for our purposes.

nlohmann:
Hm. Then I think it would be helpful to also adjust the parser to skip tags, because otherwise roundtripping would not work any more. This does not need to be done in this PR, of course, but I'm still re-opening #1968.

Ericson2314:
Oh! We'll if you are fine doing it after is there anything more this PR needs?

nlohmann:
@Ericson2314 It would be great if you could have a look at #2273.

nlohmann:
Thanks!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Use binary subtype as tag in CBOR if set. #2244

* * *

Ericson2314:
Thank you!

PR #1900, opened 2020-01-11
Added flags to cmakelists in test project to treat all warnings as errors

jmoore91:
Refer to #1798

Changes on test/CMakeLists.txt to always use -Wall, -Werror on GCC/Clang and the /WX, /W4 flags on MSVC.

Fixed build issue for GCC
Fixed build issues for Clang

PR #2245, opened 2020-07-06
Build tests with warnings as errors enabled

t-b:
I'm trying to resurrect https://github.com/nlohmann/json/pull/1900.

- [x] Cleanup commits
- [x] `#if DOCTEST_CLANG >=` -> `#if DOCTEST_CLANG && DOCTEST_CLANG >=`
- [x] Fix https://github.com/nemequ/hedley/issues/40 upstream
- [x] Get it working

Close https://github.com/nlohmann/json/issues/1798.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -15,20 +15,6 @@ jobs:
     - name: test
       run: cd build ; ctest -j 10 -C Debug --exclude-regex "test-unicode" --output-on-failure
 
-  clang9:
-    runs-on: windows-latest
-
-    steps:
-      - uses: actions/checkout@v1
-      - name: install Clang
-        run: curl -fsSL -o LLVM9.exe https://releases.llvm.org/9.0.0/LLVM-9.0.0-win64.exe ; 7z x LLVM9.exe -y -o"C:/Program Files/LLVM"
-      - name: cmake
-        run: cmake -S . -B build -DCMAKE_CXX_COMPILER="C:/Program Files/LLVM/bin/clang++.exe" -G"MinGW Makefiles" -DCMAKE_BUILD_TYPE=Debug -DJSON_BuildTests=On
-      - name: build
-        run: cmake --build build --parallel 10
-      - name: test
-        run: cd build ; ctest -j 10 -C Debug --exclude-regex "test-unicode" --output-on-failure
-
   clang10:
     runs-on: windows-latest
 
--- a/test/src/unit-allocator.cpp
+++ b/test/src/unit-allocator.cpp
@@ -29,6 +29,10 @@ SOFTWARE.
 
 #include "doctest_compatibility.h"
 
+#if DOCTEST_CLANG && DOCTEST_CLANG >= DOCTEST_COMPILER(3, 6, 0)
+    DOCTEST_CLANG_SUPPRESS_WARNING("-Wkeyword-macro")
+#endif // clang 3.6.x
+
 #define JSON_TESTS_PRIVATE
 #include <nlohmann/json.hpp>
 using nlohmann::json;
@@ -95,6 +99,8 @@ struct my_allocator : std::allocator<T>
     {
         if (next_deallocate_fails)
         {
+            (void)p; // Ignored unused
+            (void)n; // ignored unused
             next_deallocate_fails = false;
             throw std::bad_alloc();
         }
--- a/test/src/unit-allocator.cpp
+++ b/test/src/unit-allocator.cpp
@@ -114,6 +114,7 @@ struct my_allocator : std::allocator<T>
     {
         if (next_destroy_fails)
         {
+            (void)p; // Ignored unused
             next_destroy_fails = false;
             throw std::bad_alloc();
         }
--- a/test/src/unit-items.cpp
+++ b/test/src/unit-items.cpp
@@ -39,6 +39,14 @@ using nlohmann::json;
     #define JSON_HAS_CPP_14
 #endif
 
+#if defined(DOCTEST_PLATFORM_WINDOWS) && defined(DOCTEST_CLANG) && DOCTEST_CLANG >= DOCTEST_COMPILER(10, 0, 0)
+    DOCTEST_CLANG_SUPPRESS_WARNING("-Wrange-loop-construct")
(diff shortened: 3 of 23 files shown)


nlohmann:
FYI: These are the flags I check regularly: https://github.com/nlohmann/json/blob/develop/Makefile#L93.

include/nlohmann/thirdparty/hedley/hedley.hpp
```diff
@@ -504,7 +504,8 @@
 #if \
     defined(__has_cpp_attribute) && \
     defined(__cplusplus) && \
-    (!defined(JSON_HEDLEY_SUNPRO_VERSION) || JSON_HEDLEY_SUNPRO_VERSION_CHECK(5,15,0))
+    (!defined(JSON_HEDLEY_SUNPRO_VERSION) || JSON_HEDLEY_SUNPRO_VERSION_CHECK(5,15,0)) && \
```
nlohmann, line 507: Why did you change the Hedley header? We copy this from https://github.com/nemequ/hedley, so the change is best made upstream.

test/src/unit-allocator.cpp
```diff
@@ -96,6 +100,8 @@ struct my_allocator : std::allocator<T>
     {
         if (next_deallocate_fails)
         {
+            (void)p; // Ignored unused
```
gregmarr, line 103: Why is this needed?  It's used in the `else` branch.

include/nlohmann/thirdparty/hedley/hedley.hpp
t-b: @nlohmann This is only for intermediate hacking to get it working. I've already opened https://github.com/nemequ/hedley/issues/40.

test/src/unit-allocator.cpp
t-b: @gregmarr MSVC 2015 does warn here. I presume it first uses the fact that `next_deallocate_fails == true` ditches the else part and then complains about `p` not being used.

gregmarr: That's very strange, because `next_deallocate_fails` is only ever `false`.  Ah well, fun with MSVC.

test/src/unit-items.cpp
```diff
@@ -39,6 +39,18 @@ using nlohmann::json;
...
+    DOCTEST_CLANG_SUPPRESS_WARNING("-Wrange-loop-construct")
+#endif
+#endif
+
+#ifdef DOCTEST_PLATFORM_MAC
+#if DOCTEST_CLANG && DOCTEST_CLANG >= DOCTEST_COMPILER(12, 0, 0)
+    DOCTEST_CLANG_SUPPRESS_WARNING("-Wrange-loop-analysis")
+#endif
+#endif
+
```
t-b, line 53: @nlohmann Any idea why the MacOSX part here does not work? https://travis-ci.org/github/nlohmann/json/jobs/707957810 still fails.

t-b: This is due to https://github.com/onqtam/doctest/issues/415, I have a workaround.

t-b:
@nlohmann Ready.

.github/workflows/windows.yml
```diff
@@ -15,20 +15,6 @@ jobs:
     - name: test
       run: cd build ; ctest -j 10 -C Debug --exclude-regex "test-unicode" --output-on-failure
 
-  clang9:
```
nlohmann, line 18: Why did you remove Clang 9?

test/src/unit-allocator.cpp
```diff
@@ -29,6 +29,10 @@ SOFTWARE.
 
 #include "doctest_compatibility.h"
 
+#if DOCTEST_CLANG && DOCTEST_CLANG >= DOCTEST_COMPILER(3, 6, 0)
+    DOCTEST_CLANG_SUPPRESS_WARNING("-Wkeyword-macro")
```
nlohmann, line 33: Do you still need this after `JSON_TESTS_PRIVATE` was introduced?

nlohmann (changes requested):

.github/workflows/windows.yml
t-b: The commit message explains why, see e2264601542bd29a88fc85570b95c4e897eb2781.

test/src/unit-allocator.cpp
t-b: I'll have a look if I find time.

MBalszun:
Is it possible to override this? 
E.g. when I want to compile the tests with a new compiler (with new warnings) on ci or my local computer.

t-b:
@MBalszun Not as of now. But a patch is welcome to add a cmake option JSON_WARNINGS_AS_ERRORS defaulting to ON.

.github/workflows/windows.yml
MBalszun: Could the problem be that windows-latest doesn't remain the same? Newer versions of Visual studio's standard c++ library also require newer versions of clang, so when testing llvm on windows, a fixed version of clang should probably be paired with a fixed version of the VS toolchain.

MBalszun:
So, what is the state here? Can this be merged? I see there is a merge conflict, but github doesn't show me what it is.

nlohmann:
I am currently overworking the targets for the CI. The goal is to move everything from the individual Travis/AppVeyor/GitHub Actions files as well as the project's Makefile into some CMake-based targets. This includes the warning flags for GCC and Clang I linked in https://github.com/nlohmann/json/pull/2245#issuecomment-654457381. Then I think we only need to add a similar target for MSVC.

I will open a PR to make the proposal visible. It is not done yet, but I think it will tackle most issues described here (apart of MSVC).

nlohmann:
See https://github.com/nlohmann/json/pull/2561.

MBalszun:
Thanks for the update. I wasn't aware that all the warning flags for gcc and clang are set in a separate make file.

That aside: You are aware tha travis.org is being shut down and that travis.com won't provide unlimited CI time for open source projects anymore right? Or have you negotiated a deal with them?

nlohmann:
I have no deal with travis. If they stop working the way they used to, I will rely on AppVeyor and GitHub Actions. In the meantime, I see whether I can get a self-hosted runner to work so that I can automate as many system-independent tests as possible.

PR #2264, opened 2020-07-13
Makefile: Always use SED variable

t-b:
We need to have a proper sed, even on MacOSX. So let's use the variable
introduced in 191aa0fd (:wrench: overworked maintaner targets,
2019-03-28) in more places.


nlohmann:
Thanks!

PR #2282, opened 2020-07-17
Add static assertion for missing binary function in SAX interface

nlohmann:
In #1662, we added function `bool binary(binary_t&)` to the SAX interface. If the user inherits from the `json_sax` class, then an error occurs complaining about instantiation of an abstract class, see #2235. This PR also adds a static assertion to detect the error in case of template usage.


nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Add static assertion for missing binary function in SAX interface. #2282

* * *

PR #2285, opened 2020-07-18
add inline specifier for detail::combine

T0b1-iOS:
fix compiles where multiple source files include the header and generate multiple definitions of `nlohmann::detail::combine` by adding an inline specifier.

I hope it's okay I skipped the unit test and stuff since I would figure this literally can't break anything.

Fixes #2281.

### DIFF
--- a/include/nlohmann/detail/hash.hpp
+++ b/include/nlohmann/detail/hash.hpp
@@ -9,7 +9,7 @@ namespace detail
 {
 
 // boost::hash_combine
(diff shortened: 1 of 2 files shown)


nlohmann:
Thanks! Can you please also run the amalgamation?

T0b1-iOS:
oh, yea. Sorry, I forgot that.

nlohmann:
I think we should add a regression test for this - at least something basic that consists of two translation units. But this would be out of scope of this PR.

nlohmann:
Thanks!

Issue #2280, opened 2020-07-17
NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE fails if the length of the argument is 10


mzyy94:
#### What is the issue you have?

A compilation error occurred while calling NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE and NLOHMANN_DEFINE_TYPE_INTRUSIVE with 9 or 10 members.
It' caused by wrong definition of NLOHMANN_JSON_PASTE10 and NLOHMAN_JSON_PASTE11 implemented by #2225 https://github.com/nlohmann/json/blob/0fe9f23254c98172c71a08ca5a1f75d4e6974811/include/nlohmann/detail/macro_scope.hpp#L141-L142

#### Please describe the steps to reproduce the issue.

1. Define a `struct` or `class` with 9 or 10 members.
2. Call NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE or NLOHMANN_DEFINE_TYPE_INTRUSIVE with them like shown in [this example](https://github.com/nlohmann/json#examples-1).
3. Compile the code.

#### Can you provide a small but working code example?

```cpp
#include "single_include/nlohmann/json.hpp"

namespace shopping {
struct customer {
    unsigned int postcode;
    std::string address1;
    std::string address2;
    std::string city;
    std::string country;
    unsigned int telephone;

    std::string name;
    unsigned int age;
    std::string sex;
};
NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(customer, postcode, address1, address2, city, country, telephone, name, age, sex);
} // namespace shopping

int main() {}
```

#### What is the expected behavior?

Compilation fails.

```
test.cpp:16:113: error: too many arguments provided to function-like macro invocation
NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(customer, postcode, address1, address2, city, country, telephone, name, age, sex);
                                                                                                                ^
./single_include/nlohmann/json.hpp:2166:9: note: macro 'NLOHMANN_JSON_PASTE8' defined here
#define NLOHMANN_JSON_PASTE8(func,  v1, v2, v3, v4, v5, v6, v7)              NLOHMANN_JSON_PASTE2(func, v1) NLOHMANN_JSON_PASTE7(func, v2, v3, v4, v5, v6, v7)
        ^
test.cpp:16:1: error: use of undeclared identifier 'NLOHMANN_JSON_PASTE8'
[...]

Issue #2267, opened 2020-07-14
Custom types : MACRO expansion bug


pfeatherstone:
In the following code:
```
#define NLOHMANN_JSON_EXPAND( x ) x
#define NLOHMANN_JSON_GET_MACRO(_1,_2,_3,_4,_5,_6, _7, _8, _9, _10, _11, NAME,...) NAME

#define NLOHMANN_JSON_PASTE(...) NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_GET_MACRO(__VA_ARGS__, NLOHMANN_JSON_PASTE11, \
        NLOHMANN_JSON_PASTE10, NLOHMANN_JSON_PASTE9, NLOHMANN_JSON_PASTE8, NLOHMANN_JSON_PASTE7, \
        NLOHMANN_JSON_PASTE6, NLOHMANN_JSON_PASTE5, NLOHMANN_JSON_PASTE4, NLOHMANN_JSON_PASTE3, \
        NLOHMANN_JSON_PASTE2, NLOHMANN_JSON_PASTE1)(__VA_ARGS__))
#define NLOHMANN_JSON_PASTE2(func,  v1)                                      func(v1)
#define NLOHMANN_JSON_PASTE3(func,  v1, v2)                                  NLOHMANN_JSON_PASTE2(func, v1) NLOHMANN_JSON_PASTE2(func, v2)
#define NLOHMANN_JSON_PASTE4(func,  v1, v2, v3)                              NLOHMANN_JSON_PASTE2(func, v1) NLOHMANN_JSON_PASTE3(func, v2, v3)
#define NLOHMANN_JSON_PASTE5(func,  v1, v2, v3, v4)                          NLOHMANN_JSON_PASTE2(func, v1) NLOHMANN_JSON_PASTE4(func, v2, v3, v4)
#define NLOHMANN_JSON_PASTE6(func,  v1, v2, v3, v4, v5)                      NLOHMANN_JSON_PASTE2(func, v1) NLOHMANN_JSON_PASTE5(func, v2, v3, v4, v5)
#define NLOHMANN_JSON_PASTE7(func,  v1, v2, v3, v4, v5, v6)                  NLOHMANN_JSON_PASTE2(func, v1) NLOHMANN_JSON_PASTE6(func, v2, v3, v4, v5, v6)
[...]

PR #2287, opened 2020-07-18
Serialisation macros: increase upper bound on number of member variables

pfeatherstone:
This pull request contains a bit of rough and ready code to generate the serialisation/deserialisation macros. The upper bound on number of member variables is increased to 64. You can increase the upper bound arbitrarily using the aforementioned code.


pfeatherstone:
This solves #2267

pfeatherstone:
@nlohmann There is one test that is failing in continuous-integration, specifically with configuration "Xcode: xcode10 C++ AMD64". All other tests are fine. It's failing on 'download_test_data'. Any ideas?

nlohmann:
I restarted the job.

pfeatherstone:
> I restarted the job.

Thanks. Ready for review

tools/NlohmannMacroBuilder/main.cpp
```diff
@@ -0,0 +1,85 @@
...
+        ss << "v" << i-1 << ") NLOHMANN_JSON_PASTE2(func, v1) NLOHMANN_JSON_PASTE" << i-1 << "(func, ";
+        for (int j = 2 ; j < i-1 ; j++)
+            ss << "v" << j << ", ";
+        ss << "v" << i-1 << ")" << endl;
+    }
+    
+    cout << ss.str() << endl;
+}
+
+struct example_struct
```
nlohmann, line 37: Can you please move this code into `test/src/unit-udt_macro.cpp`?

test/src/unit-udt_macro.cpp
```diff
@@ -128,3 +210,69 @@ TEST_CASE_TEMPLATE("Serialization/deserialization via NLOHMANN_DEFINE_TYPE_INTRU
...
+                   persons::person_with_public_alphabet)
+{
+    SECTION("alphabet")
+    {
+        {
+            T obj1;
+            nlohmann::json j = obj1; //via json object
+            T obj2;
+            j.get_to(obj2);
+            assert(obj1 == obj2);
```
nlohmann, line 225: Please use a doctest construct for this and the other `assert` usages.

pfeatherstone: So tried ```CHECK(obj1 == obj2)``` and didn't work. So replaced with ```bool ok = (obj1 == obj2); CHECK(ok);```. Don't know why the first version didn't work (I'm new to doctest) but the test is the same.

nlohmann: I think doctest needs a way to produce a string from an object in case something goes wrong. The best would be the following:

```cpp
CHECK((obj1 == obj2))
```

(note the double parentheses)

pfeatherstone:
Ok, the combination of valgrind with some of the older compilers doesn't like using std::tie and std::tuple to implement the ```operator==``` overload in the unit test. It complains with error ```Conditional jump or move depends on uninitialised value(s)```

pfeatherstone:
Rewrote ```operator==```. Doesn't use std::tie. Hopefully, valgrind and clang will no longer complain about uninitialised variables

pfeatherstone:
This is still not ready as I have to rebase the fork. This will require a battle with git which i don't have time for today.

nlohmann:
No need to rebase as long as there are no merge conflicts.

tools/NlohmannMacroBuilder/main.cpp
pfeatherstone: This is dealt with

pfeatherstone:
@nlohmann Seems to be ready

nlohmann:
Thanks!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [x] ⚡️ Improvement
- [ ] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Serialisation macros: increase upper bound on number of member variables. #2267 #2280 #2287

* * *

PR #2308, opened 2020-07-24
Fix bug in CBOR tag handling

nlohmann:
This PR fixes a bug introduced in #2273. The `skip_tags` parameter was not passed to all functions. This resulted in a parse error even if `json::cbor_tag_handler_t::ignore` was passed.


nlohmann:
(Release item is in #2273)

PR #2253, opened 2020-07-08
Add pkg-config file

ericonr:
Add pkg-config generation to the CMake build system, functionally the same as the pkg-config generated by Meson. This PR is picking up on the work from #1701 .

* * *

## Pull request checklist

- [*]  Changes are described in the pull request, or an [existing issue is referenced](https://github.com/nlohmann/json/issues).
- [*]  The test suite [compiles and runs](https://github.com/nlohmann/json/blob/develop/README.md#execute-unit-tests) without error.

PR #2318, opened 2020-07-27
cmake: install pkg-config file relative to current_binary_dir

eli-schwartz:
When the testsuite is enabled, the "cmake_add_subdirectory" test adds a second copy of the project into the build configuration, and the project files are installed twice.

This becomes super problematic when it tries to install a file from CMAKE_BINARY_DIR which is only available in CMAKE_CURRENT_BINARY_DIR and bombs out with the following error message:

```
[...]
-- Installing: <DESTDIR>/usr/lib/cmake/nlohmann_json/nlohmann_jsonTargets.cmake
CMake Error at test/cmake_add_subdirectory/nlohmann_json/cmake_install.cmake:73 (file):
  file INSTALL cannot find
  "......./nlohmann-json/builddir/test/cmake_add_subdirectory/nlohmann_json.pc":
  No such file or directory.
Call Stack (most recent call first):
  test/cmake_add_subdirectory/cmake_install.cmake:47 (include)
  test/cmake_install.cmake:49 (include)
  cmake_install.cmake:94 (include)
```

Fixes regression in #2253


nlohmann:
Thanks!

nlohmann:
* * *

### 🔖 Release item

#### Type

- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] ⚡️ Improvement
- [x] 🔨 Further Change
- [ ] 🔥 Deprecated function

#### Description

- Install pkg-config file to `CMAKE_CURRENT_BINARY_DIR` instead of `CMAKE_BINARY_DIR`. #2318

* * *

PR #2375, opened 2020-08-25
Properly select "Release" build for Travis

karzhenkov:
Ninja doesn't support multiple build configurations. The option `--config Release` has no effect when `cmake --build` is called. Essentially, all testing was performed for "Debug" configuration, which is the default. Maybe it doesn't really matter, but it's misleading. The proper way to select build configuration here is through `CMAKE_BUILD_TYPE`.


karzhenkov:
The coverage test requires the use of the "Debug" configuration. From the [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Gcov-and-Optimization.html#Gcov-and-Optimization):
> if you want to prove that every single line in your program was executed, you should not compile with optimization at the same time

nlohmann:
@karzhenkov Can you please update to the latest develop branch? I changed a lot of issues in the CI and moved some checks from Travis to GitHub Actions. Maybe the changes of this PR may need to be applied in `cmake/ci.cmake`.

karzhenkov:
I think there is nothing here that we can apply to `cmake\ci. cmake`.

.travis.yml
```diff
@@ -224,10 +224,13 @@ script:
   # append CXX_STANDARD to CMAKE_OPTIONS if required
   - CMAKE_OPTIONS+=${CXX_STANDARD:+ -DCMAKE_CXX_STANDARD=$CXX_STANDARD -DCMAKE_CXX_STANDARD_REQUIRED=ON}
 
+  # build configuration
+  - CMAKE_OPTIONS+=" -DCMAKE_BUILD_TYPE=Release"
```
nlohmann, line 228: But doesn't this mean we compile all tests in Release mode?

karzhenkov: The test configuration with `BUILD_TYPE=Debug` was previously used for coverage check only. This configuration is defined now in GitHub workflows. Others don't set build type explicitly and use the default which is `Release`. I'm not sure wheter the default should be `Release` or `Debug`. Is there a reason to prefer `Debug` here?

nlohmann: I see. I never thought about the "default" build type.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #2405, opened 2020-09-26
Fix move constructor of json_ref

### LINKED ISSUE #2387: Usage with -fno-elide-constructors causes dump() output to be array of `null`s

The following program...

```c++
#include "./json.hpp"

#include <iostream>

nlohmann::json makeobj() {
        return {
                {"type", "Foo"},
                {"location", {10, 20}}
        };
}

int main() {
        std::cout << makeobj().dump() << std::endl;
        return 0;
}
```

... compiled with the following flags ...

```console
$ c++ -Wall -Wextra -Werror -std=c++11 -Wshadow -pedantic-errors -fno-elide-constructors -Winit-self -Wold-style-cast -Woverloaded-virtual -Wuninitialized -Wmissing-declarations -fvisibility=hidde
n -Wunused -g3 -O0 -otest-json test-json.cc
```

... produces the following output:

```
[[null,null],[null,[null,null]]]
```

Omitting the `-fno-elide-constructors` flag, we get the following (correct) output:

```
{"location":[10,20],"type":"Foo"}
```

I've since removed the `-fno-elide-constructors` from the build script as it was leftover from another project and isn't necessary (or really a good idea) for this project. I just figured you might want to know about it.

I don't fully understand the nuances of the flag but I would imagine it's not a flag that'd be "compatible" with this library (understandably so). At the very least, some documentation in the `Notes` section in the readme might be a worthwhile PR as it was the first place I looked for an explanation of the `null` output. Caused a 15 minute headache tracking it down - not terrible, but just annoying enough to open this issue :)

Thanks for the great library!

```console
$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.1 LTS"

$ c++ --version
Ubuntu clang version 12.0.0-++20200829052612+bf21ce7b908-1~exp1~20200829153238.148
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/local/bin
```

Commit 1bcabd9e83ebc2d04b5767c7f51bf385e80398d6; single include header vendored into project.

karzhenkov:
The defaulted move constructor of `nlohmann::detail::json_ref` makes copy of internal pointer which points to a field inside the object being moved. It becomes a dangling pointer. A "manual" move implementation is needed.

Resolves #2387. Here is an [example at Godbolt](https://godbolt.org/z/9bKbTY).

### DIFF
--- a/include/nlohmann/detail/json_ref.hpp
+++ b/include/nlohmann/detail/json_ref.hpp
@@ -17,19 +17,14 @@ class json_ref
 
     json_ref(value_type&& value)
         : owned_value(std::move(value))
-        , value_ref(&owned_value)
-        , is_rvalue(true)
     {}
 
     json_ref(const value_type& value)
-        : value_ref(const_cast<value_type*>(&value))
-        , is_rvalue(false)
+        : value_ref(&value)
     {}
 
     json_ref(std::initializer_list<json_ref> init)
         : owned_value(init)
-        , value_ref(&owned_value)
-        , is_rvalue(true)
     {}
 
     template <
@@ -37,8 +32,6 @@ class json_ref
         enable_if_t<std::is_constructible<value_type, Args...>::value, int> = 0 >
     json_ref(Args && ... args)
         : owned_value(std::forward<Args>(args)...)
-        , value_ref(&owned_value)
-        , is_rvalue(true)
     {}
 
     // class should be movable only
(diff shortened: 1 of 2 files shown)


karzhenkov:
There is a also better solution: karzhenkov@42a9dc0bc38691b602a679d493065f732f09cf55.
No `const_cast` required; defaulted move constructor is pretty enough.

I'm ready to force-push if it's acceptable.

nlohmann:
@karzhenkov That would be great!

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

Qix-:
Great catch @karzhenkov, didn't expect #2387 to be a bug.

nlohmann:
Sorry everybody for waiting so long, and thanks for your patience!

PR #2444, opened 2020-10-23
Change argument name "subtype" in byte_container_with_subtype

linev:
Fix warnings on gcc 4.8.5: `declaration of ‘subtype’ shadows a member of 'this'`

### DIFF
--- a/include/nlohmann/byte_container_with_subtype.hpp
+++ b/include/nlohmann/byte_container_with_subtype.hpp
@@ -39,15 +39,15 @@ class byte_container_with_subtype : public BinaryType
         : container_type(std::move(b))
     {}
 
-    byte_container_with_subtype(const container_type& b, std::uint8_t _subtype) noexcept(noexcept(container_type(b)))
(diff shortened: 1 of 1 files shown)


include/nlohmann/byte_container_with_subtype.hpp
```diff
@@ -39,15 +39,15 @@ class byte_container_with_subtype : public BinaryType
         : container_type(std::move(b))
     {}
 
-    byte_container_with_subtype(const container_type& b, std::uint8_t subtype) noexcept(noexcept(container_type(b)))
+    byte_container_with_subtype(const container_type& b, std::uint8_t _subtype) noexcept(noexcept(container_type(b)))
```
nlohmann, line 42: Please do not use a leading underscore here. Rather use `subtype_`.

nlohmann (changes requested):

include/nlohmann/byte_container_with_subtype.hpp
linev: Ok, done

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

Issue #2528, opened 2020-12-13
Cannot assign from ordered_json vector<CustomStruct> to value in not ordered json


kotori2:
#### What is the issue you have?

I created a custom struct in unordered_json with vector, then assign it to an ordered json, it will throw error.

#### Please describe the steps to reproduce the issue.

Check the snippet below. 
Actually it works if both of them are ordered json or unordered json, or I removed the vector. 

#### Can you provide a small but working code example?

I defined my own `NLOHMANN_DEFINE_TYPE_INTRUSIVE` just to conveniently define them in unordered json. 
```
using nlohmann::json;
using ordered_json = nlohmann::ordered_json;

#define _NLOHMANN_DEFINE_TYPE_INTRUSIVE(Type, ...)  \
    friend void to_json(ordered_json& nlohmann_json_j, const Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__)) } \
    friend void from_json(const ordered_json& nlohmann_json_j, Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_FROM, __VA_ARGS__)) }

class releaseInfo {
public:
    uint64_t id = 0;
    std::string key;
    _NLOHMANN_DEFINE_TYPE_INTRUSIVE(releaseInfo, id, key);
};

int main() {
    std::vector<releaseInfo> test;
    json parent;
    parent["test"] = test;
    return 0;
}
```

#### What is the expected behavior?

No error happened. 

#### And what is the actual behavior instead?

```
/Users/kotori0/code/untitled/main.cpp:21:15: error: no viable overloaded '='
    a["test"] = test;
    ~~~~~~~~~ ^ ~~~~
/Users/kotori0/code/untitled/cmake-build-debug/_deps/json-src/single_include/nlohmann/json.hpp:18662:17: note: candidate function not viable: no known conversion from 'std::vector<releaseInfo>' to 'nlohmann::basic_json<std::map, std::vector, std::__1::basic_string<char>, bool, long long, unsigned long long, double, std::allocator, adl_serializer, std::__1::vector<unsigned char, std::__1::allocator<unsigned char> > >' for 1st argument
    basic_json& operator=(basic_json other) noexcept (
                ^
1 error generated.
```

#### Which compiler and operating system are you using?

[...]

PR #2532, opened 2020-12-14
Templated versions of NLOHMANN_DEFINE_TYPE_INTRUSIVE and NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE

YarikTH:
As I described in https://github.com/nlohmann/json/issues/2528#issuecomment-744100061
Ideally, NLOHMANN_DEFINE_TYPE_INTRUSIVE and NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE should work with any basic_json overload.
Like here:
https://godbolt.org/z/d16erK

In this pull request, I added my templated macroses as two new macroses with ugly names
NLOHMANN_DEFINE_TYPE_INTRUSIVE_2
NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_2
I'm not sure if their content should just replace NLOHMANN_DEFINE_TYPE_INTRUSIVE and NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE or they should be separated, but have a better name (suggestions are welcome!).

Also any way some tests should be written for them, but I don't know where. Maybe in unit-udt_macro.cpp? I have to include ordered_json to it in this case.

TODO:
- [X] Add tests for new macroses in unit-udt_macro.cpp
- [X] Choise better names for macroses
- [X] Constrain JsonType with enable_if? Don't need. Existing NLOHMANN_JSON_SERIALIZE_ENUM don't constrain JsonType. If it need, it should be done for all such cases in separate PR

* * *

## Pull request checklist


include/nlohmann/detail/json_ref.hpp
```diff
@@ -57,7 +57,7 @@ class json_ref
 
     value_type const* operator->() const
     {
-        return &**this;
+        return &** this;
```
YarikTH, line 60: result of "make amalgamate"

single_include/nlohmann/json.hpp
```diff
@@ -4515,15 +4533,15 @@ class byte_container_with_subtype : public BinaryType
         : container_type(std::move(b))
     {}
 
-    byte_container_with_subtype(const container_type& b, std::uint8_t subtype) noexcept(noexcept(container_type(b)))
+    byte_container_with_subtype(const container_type& b, std::uint8_t subtype_) noexcept(noexcept(container_type(b)))
```
YarikTH, line 4536: result of "make amalgamate"

include/nlohmann/detail/json_ref.hpp
nlohmann: I forgot to check in the amalgamated version. Please update your branch.

nlohmann:
Tests should to to `test/src/unit-udt_macro.cpp`.

nlohmann:
Thanks for the proposal, I like it! In fact, the introduction of `ordered_json` was quite rough and we forgot about it in many places.

I'm hesitant on replacing the existing macros, because this may break existing code, wouldn't it?

YarikTH:
> I'm hesitant on replacing the existing macros, because this may break existing code, wouldn't it?

Theoretically, it shouldn't I just added template parametrisation of json type, but who knows. There are only a few tests that check this macro. But problems may occur in some real-life usage that is hard to predict.

YarikTH:
I added TODO section in PR description

YarikTH:
> In fact, the introduction of `ordered_json` was quite rough and we forgot about it in many places.

Speaking about this. Actually, it's quite easy to find all the problematic places. Just search "nlohmann::json" with full words option.
So what I found:
```
/*!
@brief macro
@def NLOHMANN_DEFINE_TYPE_INTRUSIVE
@since version 3.9.0
*/
#define NLOHMANN_DEFINE_TYPE_INTRUSIVE(Type, ...)  \
    friend void to_json(nlohmann::json& nlohmann_json_j, const Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__)) } \
    friend void from_json(const nlohmann::json& nlohmann_json_j, Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_FROM, __VA_ARGS__)) }

/*!
@brief macro
@def NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE
@since version 3.9.0
*/
#define NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(Type, ...)  \
    inline void to_json(nlohmann::json& nlohmann_json_j, const Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__)) } \
    inline void from_json(const nlohmann::json& nlohmann_json_j, Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_FROM, __VA_ARGS__)) }
```

```
/*!
@brief SAX interface

This class describes the SAX interface used by @ref nlohmann::json::sax_parse.
Each function is called in different situations while the input is parsed. The
boolean return value informs the parser whether to continue processing the
input.
*/
struct json_sax
{
```

```
/// hash value for JSON objects
template<>
struct hash<nlohmann::json>
{
    /*!
    @brief return a hash value for a JSON object

    @since version 1.0.0
    */
    std::size_t operator()(const nlohmann::json& j) const
    {
        return nlohmann::detail::hash(j);
    }
};
```

```
/*!
@brief exchanges the values of two JSON objects

@since version 1.0.0
*/
template<>
inline void swap<nlohmann::json>(nlohmann::json& j1, nlohmann::json& j2) noexcept(
    is_nothrow_move_constructible<nlohmann::json>::value&&
    is_nothrow_move_assignable<nlohmann::json>::value
                              )
{
    j1.swap(j2);
}
```

```
/*!
@brief user-defined string literal for JSON values

This operator implements a user-defined string literal for JSON objects. It
can be used by adding `"_json"` to a string literal and returns a JSON object
if no parse error occurred.

@param[in] s  a string representation of a JSON object
@param[in] n  the length of string @a s
@return a JSON object

@since version 1.0.0
*/
JSON_HEDLEY_NON_NULL(1)
inline nlohmann::json operator "" _json(const char* s, std::size_t n)
{
    return nlohmann::json::parse(s, s + n);
}

/*!
@brief user-defined string literal for JSON pointer

This operator implements a user-defined string literal for JSON Pointers. It
can be used by adding `"_json_pointer"` to a string literal and returns a JSON pointer
object if no parse error occurred.

@param[in] s  a string representation of a JSON Pointer
@param[in] n  the length of string @a s
@return a JSON pointer object

@since version 2.0.0
*/
JSON_HEDLEY_NON_NULL(1)
inline nlohmann::json::json_pointer operator "" _json_pointer(const char* s, std::size_t n)
{
    return nlohmann::json::json_pointer(std::string(s, n));
}
```

It seems more like too strict glueing to nlohmann::json despite the fact that it's just default specialization, not only one.
But it's subject to another PR anyway. Or it isn't? I can convert PR to more general one, but I can't think of the proper name for it.

YarikTH:
Some possible improvements for NLOHMANN_DEFINE_TYPE_INTRUSIVE and NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE. I'm not sure should it be done in this PR or not.

```cpp
/*!
@brief macro to briefly define intrusive serialization of a given type to/from JSON
@note you can define your own specialized macroses like NLOHMANN_DEFINE_TYPE_INTRUSIVE
@def NLOHMANN_DEFINE_TYPE_INTRUSIVE_IMPL
@since version 3.9.2
*/
#define NLOHMANN_DEFINE_TYPE_INTRUSIVE_IMPL(BasicJsonType, Type, ...)  \
    friend void to_json(BasicJsonType& nlohmann_json_j, const Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__)) } \
    friend void from_json(const BasicJsonType& nlohmann_json_j, Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_FROM, __VA_ARGS__)) }

/*!
@brief macro to briefly define non-intrusive serialization of a given type to/from JSON
@note you can define your own specialized macroses like NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE
@def NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_IMPL
@since version 3.9.2
*/
#define NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_IMPL(BasicJsonType, Type, ...)  \
    inline void to_json(BasicJsonType& nlohmann_json_j, const Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__)) } \
    inline void from_json(const BasicJsonType& nlohmann_json_j, Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_FROM, __VA_ARGS__)) }

/*!
@brief macro to briefly define intrusive serialization of given type to/from nlohmann::json
@def NLOHMANN_DEFINE_TYPE_INTRUSIVE
@since version 3.9.0
@sa NLOHMANN_DEFINE_TYPE_INTRUSIVE_IMPL
*/
#define NLOHMANN_DEFINE_TYPE_INTRUSIVE(Type, ...) NLOHMANN_DEFINE_TYPE_INTRUSIVE_IMPL(nlohmann::json, Type, __VA_ARGS__)

/*!
@brief macro to briefly define non-intrusive serialization of given type to/from nlohmann::json
@def NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE
@since version 3.9.0
@sa NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_IMPL
*/
#define NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(Type, ...) NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_IMPL(nlohmann::json, Type, __VA_ARGS__)
```

nlohmann:
Thanks for checking - the other usages of `nlohmann::json` seem straightforward to fix.

YarikTH:
Tests are added. PR is ready to review.

test/thirdparty/doctest/doctest.h
```diff
@@ -2051,8 +2051,8 @@ int registerReporter(const char* name, int priority, bool isReporter) {
                                                        __LINE__, #expr, #__VA_ARGS__, message);    \
             try {                                                                                  \
                 DOCTEST_CAST_TO_VOID(expr)                                                         \
-            } catch(const doctest::detail::remove_const<                                           \
-                    doctest::detail::remove_reference<__VA_ARGS__>::type>::type&) {                \
+            } catch(const typename doctest::detail::remove_const<                                  \
+                    typename doctest::detail::remove_reference<__VA_ARGS__>::type>::type&) {       \
```
YarikTH, line 2055: Bugfix described in https://github.com/onqtam/doctest/issues/447 and fixed in v2.4.2

MHebes:
Are there any possible use cases where just making the original macros templated would break backwards compat?

I personally think it'd be a lot cleaner if the macros were changed rather than making new ones that do the same thing. Although, I'm sure you agree and are just erring on the side of caution 🙂 

Just not sure if the caution is needed necessarily.

YarikTH:
> Are there any possible use cases where just making the original macros templated would break backwards compat?
> 
> I personally think it'd be a lot cleaner if the macros were changed rather than making new ones that do the same thing. Although, I'm sure you agree and are just erring on the side of caution 
> 
> Just not sure if the caution is needed necessarily.

There are a lot of users with different compilers. Who knows what can go wrong. A separate macro at least minimizes the chance of problems. But of course, I can just replace the content of the existing macro with my templated version, if the owner says so. Unfortunately, 3 months passed and there is still zero feedback from nlohmann about the final version.

nlohmann:
@YarikTH Sorry for leaving this open for so long... Can you please update to the latest develop version. I think the only remaining todos are documentation.

PR #2545, opened 2020-12-25
Fix compilation of input_adapter(container) in edge cases

jasujm:
This fixes a compilation issue with the library if trying to parse JSON from a
custom container type that doesn't support detecting `begin()` and `end()`
functions via ADL.

`include/nlohmann/detail/input/input_adapters.hpp` defines convenience function
for creating input adapter for STL-like containers. It should support iterators
both via `std::begin()` and `std::end()`, as well as `begin()` and `end()` found
via ADL. However, the former doesn't actually work for iterable types not
defined in the `std` namespace, due to the way the return type deduction is
implemented:

    $ cat test.cc
    #include <iostream>
    #include <nlohmann/json.hpp>

    const char DATA[] = R"("Hello, world!")";

    struct MyIterable {
        using iterator = const char*;
        using const_iterator = iterator;
        const_iterator begin() const { return DATA; }
        const_iterator end() const { return DATA + sizeof(DATA); }
    };

    int main()
    {
        const auto c = MyIterable {};
        std::cout << nlohmann::json::parse(c) << "\n";
        return 0;
    }
    $ g++ --version
    g++ (Ubuntu 10.2.0-13ubuntu1) 10.2.0
    Copyright (C) 2020 Free Software Foundation, Inc.
    This is free software; see the source for copying conditions.  There is NO
    warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

    $ g++ -std=c++11 test.cc
    ...
    /usr/include/nlohmann/detail/input/input_adapters.hpp:375:6: note: candidate: ‘template<class ContainerType> decltype (nlohmann::detail::input_adapter(begin(container), end(container))) nlohmann::detail::input_adapter(const ContainerType&)’
      375 | auto input_adapter(const ContainerType& container) -> decltype(input_adapter(begin(container), end(container)))
          |      ^~~~~~~~~~~~~
    /usr/include/nlohmann/detail/input/input_adapters.hpp:375:6: note:   template argument deduction/substitution failed:
    /usr/include/nlohmann/detail/input/input_adapters.hpp: In substitution of ‘template<class ContainerType> decltype (nlohmann::detail::input_adapter(begin(container), end(container))) nlohmann::detail::input_adapter(const ContainerType&) [with ContainerType = MyIterable]’:
    ...

In the above example the trailing type deduction won't look for the `begin()`
and `end()` functions in the `std` namespace, although the function body would.

I don't know about a more elegant fix in C++11 except to spell out the return
type verbatim.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/input/input_adapters.hpp
+++ b/include/nlohmann/detail/input/input_adapters.hpp
@@ -372,7 +372,7 @@ typename iterator_input_adapter_factory<IteratorType>::adapter_type input_adapte
 
 // Convenience shorthand from container to iterator
 template<typename ContainerType>
-auto input_adapter(const ContainerType& container) -> decltype(input_adapter(begin(container), end(container)))
+typename iterator_input_adapter_factory<typename ContainerType::const_iterator>::adapter_type input_adapter(const ContainerType& container)
 {
     // Enable ADL
     using std::begin;
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -5179,7 +5179,7 @@ typename iterator_input_adapter_factory<IteratorType>::adapter_type input_adapte
 
 // Convenience shorthand from container to iterator
 template<typename ContainerType>
-auto input_adapter(const ContainerType& container) -> decltype(input_adapter(begin(container), end(container)))
+typename iterator_input_adapter_factory<typename ContainerType::const_iterator>::adapter_type input_adapter(const ContainerType& container)
 {
     // Enable ADL
     using std::begin;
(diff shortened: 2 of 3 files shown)


jasujm:
The PR imposes the requirement for the custom container to define `const_iterator`, which the previous code didn't. I'm sure this can be worked around with some more SFINAE if required.

jasujm:
I'll go back to the drawing board with this. Not all types that are eligible should be required to define `const_iterator`, even in the standard library. For example, the `span` type introduced in C++20 doesn't: https://en.cppreference.com/w/cpp/container/span

I'll try to make an improved PR that works on more kinds of container-like objects.

PR #2547, opened 2020-12-26
Add a CI test for the multi-header option

Mailaender:
For https://build.opensuse.org/package/show/home:Mailaender:branches:devel:libraries:c_c++/nlohmann_json some tests failed. This is just a check to verify this also happens in your integration environment.


Mailaender:
Okay, this is probably related to my `ctest --exclude-regex "cmake_fetch_content_.*|download_test_data"` then.

PR #2558, opened 2020-12-30
Strict size checks for conversions to fixed size containers

domsoll:
The current implementation accepts conversion of JSON arrays to fixed size containers like std::array, std::pair or std::tuple, as long as there are sufficient elements in the JSON array to fill the container.
The conversion only implicitly fails if there are to few elements via the out of range exception when trying to access further elements.

I suggest to strictly check for matching JSON array size when converting to fixed size containers instead.
In my opinion a mismatch in size should be considered a type error, as in the sense, that i.e. a std::array<int, 4> is a different type than a std::array<int, 5>.
Silently discarding extra JSON array elements might hide problems with mismatching serialized data and result in unexpected behavior.
If a fixed size container is extracted, it explicitly expresses that a certain number of elements is expected, if this expectation is not met by the JSON object, throwing a type error seems to be the safest way to handle the situation.

Though according to the test cases, the current behavior seems at least partially intentional, so I guess, this is up for debate, which approach is preferable.

Anyway, the commits in this pull request add strict size checks for the conversion to std::array, std::pair, std::tuple and c-style arrays.
There are also explicit JSON type checks for more expressive exceptions when the JSON type is not an array at all.
Test cases are adapted accordingly and some extra test cases are added to check for the new size restrictions.

* * *

## Pull request checklist

## Please don't


nlohmann:
The change will break existing code, so it cannot be introduced in the 3.x.x versions. I would propose to guard the test/throwing code with a preprocessor symbol so that the feature can be switched on if needed. What do you think?

domsoll:
Thank's for your input, you are right, an introduction in 3.x.x is problematic, I guess the preprocessor symbol is the best option then.
I've added the according preprocessor conditionals and a cmake option to define the symbol which is disabled by default.

The behavior without the strict container size option should now be unchanged, including the type of exceptions thrown in case of errors.
Code coverage is 100% and all test cases pass with and without the option.
I left the extra test cases for c-style arrays active without the new option enabled and extended them to check for the non-strict variant also, I guess that makes sense?

nlohmann:
@domsoll Can you please update to the latest `develop` version to resolve the conflicts?

nlohmann:
(Related: #3010)

domsoll:
The pull request is now updated to the latest `develop` version.

nlohmann:
Thanks! Can you please run the amalgamation (make amalgamate)?

domsoll:
I did update amalgamation before with make amalgamate.
Though without astyle/make pretty there is one closing brace placed differently (which the check complained about) and with astyle/make pretty I get a huge diff in the amalgamation and the source files.
For some reason my version of astyle seems to behave differently to what was used on the sources before.

Anyway, I reverted the brace placement manually again, so the check should hopefully pass now.

gregmarr:
@domsoll This PR has conflicts again.
@nlohmann Anything else outstanding here?

PR #2562, opened 2021-01-01
Better diagnostics

### LINKED ISSUE #932: Add Key name to Exception
**Bug Report**

(I'm interested in developing this feature, so pointers are appreciated)

- When working with an invalid json, the exception produced is not really helpfull.

I'v found out https://github.com/nlohmann/json/issues/160 - but that's a different issue than the one I'm having, and it will also be harder to fix because the values have no notion of key, the specific code I'm talking about on json.hpp follows this pattern:

```cpp
void from_json(const BasicJsonType& j, typename BasicJsonType::string_t& s)
{
    if (JSON_UNLIKELY(not j.is_string()))
    {
        JSON_THROW(type_error::create(302, "type must be string, but is " + std::string(j.type_name())));
    }
    s = *j.template get_ptr<const typename BasicJsonType::string_t*>();
}
```
but even if we have a small json to handle, the output is not really helpfull:

` std::exception: [json.exception.type_error.302] type must be string, but is number`

- Small code example:

```cpp
nlohmann::json j {{"firstElement", ""}};
j["firstElement"] = 0;

```
` std::exception: [json.exception.type_error.302] type for key firstElement must be string, but is number`

3.0.1

### LINKED ISSUE #1508: How to get position info or parser context with custom from_json() that may throw exceptions?
Sorry if the question is lame or was answered before (couldn't find anything similar in the issues history).

Is it possible to get a position info or, better, a typical `json::exception` with a context where the "parsing" failed, for any exceptions thrown inside my custom `from_json()` for third party classes that may throw errors on their `from_string()` builder functions?

nlohmann:
This PR improves the diagnostic information provided by the library in case of exceptions. The idea (originally described in https://github.com/nlohmann/json/issues/1508#issuecomment-683271573 and https://github.com/nlohmann/json/issues/932#issuecomment-363871828) is that every JSON value in an array or object contains a pointer to the enclosing container. From this parent relation, we can generate a JSON Pointer to the current value that allows to identify the erroneous value. As we need to store one additional pointer per JSON value, the diagnostic mode is guarded by a preprocessor macro `JSON_DIAGNOSTICS` that needs to be defined in order to enable the better diagnostics.

This PR is work in progress. There are a lot of open tasks, including

- [x] set the parent pointer in all situations
- [x] find a way to deal with iterators
- [x] add diagnostics to exceptions
- [x] add tests
- [x] add CMake option

Any help, input, review, or discussion is greatly appreciated!

Closes #1508. Closes #932. Closes #2552.


test/src/unit-json_patch.cpp
```diff
@@ -1268,7 +1280,7 @@ TEST_CASE("JSON patch")
             std::ifstream f(filename);
             json suite = json::parse(f);
 
-            for (const auto& test : suite)
+            for (const auto test : suite)
```
nlohmann, line 1283: The tests fail with a SIGSEGV if we process `suite` by a const reference. I have no idea why...

karzhenkov: There is probably an error in the move constructor of `basic_json`. When object is moved, its children still have old (no longer existing) parent. The following may help:
```
$ git diff include/nlohmann/json.hpp
diff --git a/include/nlohmann/json.hpp b/include/nlohmann/json.hpp
index 2fc497bae..f6e19eead 100644
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -2196,6 +2196,10 @@ class basic_json
         other.m_type = value_t::null;
         other.m_value = {};

+#if JSON_DIAGNOSTICS
+        for (auto& item : *this) { item.m_parent = this; }
+#endif
+
         assert_invariant();
     }
#endif
```
The copy constructor has more elaborated logic for the same purpose. By the way, the copy constructor initializes new object with no parent - i. e. with no danging pointer (if there is one). This is why processing `suite` by value works.

nlohmann: Thanks a lot! I missed that one. Fixed in 294fa343d54634856dfef4a645de5241bd2dcde1.

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -27,7 +27,7 @@ void from_json(const BasicJsonType& j, typename std::nullptr_t& n)
 {
     if (JSON_HEDLEY_UNLIKELY(!j.is_null()))
     {
-        JSON_THROW(type_error::create(302, "type must be null, but is " + std::string(j.type_name())));
+        JSON_THROW(type_error::create(302, j.diagnostics() + "type must be null, but is " + std::string(j.type_name())));
```
gregmarr, line 30: After seeing this several times, I'm wondering if `type_error::create` should instead take a second parameter of type `BasicJsonType` so that you can't forget to do the `j.diagnostics() + `.

include/nlohmann/json.hpp
```diff
@@ -1642,6 +1642,9 @@ class basic_json
             std::for_each(init.begin(), init.end(), [this](const detail::json_ref<basic_json>& element_ref)
             {
                 auto element = element_ref.moved_or_copied();
+#if JSON_DIAGNOSTICS
+                (*element.m_value.array)[1].m_parent = this;
```
gregmarr, line 1646: It's not obvious to me why this is `[1]` here.  Perhaps it deserves a comment, or perhaps like other places this could be `m_value.object->back().m_parent = this` after the following line?

Unrelated question, would this be less verbose using range-for instead of `std::for_each`?

gregmarr: Would it make sense to have a `set_parent()` function that is a no-op if `JSON_DIAGNOSTICS` is not defined, and save some `#if/#endif`?

We would still want it for the loop below, but it would save one-off functions.  On the other hand, I suppose that there could also be a function that takes an iterable and sets the parent pointer of every element.

    #if JSON_DIAGNOSTICS
    set_parent_pointer(auto const &iterable) {
        for (auto& element : iterable) {
            element.m_parent = this;
        }
    }
    #else
    set_parent_pointer(auto const &) {
    }
    #endif

gregmarr: I guess you'd need separate functions for array vs object.

include/nlohmann/json.hpp
```diff
@@ -2217,6 +2288,9 @@ class basic_json
     {
         assert_invariant();
         m_value.destroy(m_type);
+#if JSON_DIAGNOSTICS
+        m_parent = nullptr;
+#endif
```
gregmarr, line 2293: This should not be needed.

include/nlohmann/json.hpp
```diff
@@ -3318,17 +3446,23 @@ class basic_json
         {
             JSON_TRY
             {
+#if JSON_DIAGNOSTICS
+                reference result = m_value.array->at(idx);
+                result.m_parent = this;
```
gregmarr, line 3451: Could be interesting for debugging to have this check:

    JSON_ASSERT(result.m_parent == nullptr || result.m_parent == this);

gregmarr: If we had that `set_parent()` function it could go in there.

gregmarr: Could wrap this into a function as well:

    reference ensure_parent(reference result) {
        result.set_parent(this);
        return result
    }

    return ensure_parent(m_value.array->at(idx));

include/nlohmann/detail/conversions/from_json.hpp
gregmarr: Looks like there are many more uses than the ones here that add the diagnostics, so perhaps not.

nlohmann: I was thinking about this, too. I first wanted to get things "right" before cleaning up. You are definitely right!

nlohmann:
Thanks @gregmarr for the comments! I will look at them in detail later this week.

include/nlohmann/json.hpp
```diff
@@ -2146,6 +2185,9 @@ class basic_json
     basic_json(basic_json&& other) noexcept
         : m_type(std::move(other.m_type)),
           m_value(std::move(other.m_value))
+#if JSON_DIAGNOSTICS
+        , m_parent(other.m_parent)
```
karzhenkov, line 2189: This gives the user an opportunity to create an object with a dangling pointer. For example:
```c++
json j = std::move(json{0}[0]); // extract nested object from json rvalue
```
The dangling pointer leads to `SIGSEGV` when `diagnostics()` is called (see also https://godbolt.org/z/jKGsGc).
The object being initialized by move constructor doesn't actually reside inside of `*other.m_parent`. The latter still contains the `other` (until destruction) and has no relation to the new object, which should have no parent in its initial state.

include/nlohmann/json.hpp
```diff
@@ -2193,6 +2261,9 @@ class basic_json
         using std::swap;
         swap(m_type, other.m_type);
         swap(m_value, other.m_value);
+#if JSON_DIAGNOSTICS
+        m_parent = other.m_parent;
```
karzhenkov, line 2265: Here is essentially the same issue as with the move constructor above.

include/nlohmann/json.hpp
gregmarr: Man that's ugly code, but agreed that the constructors shouldn't be setting the parent to anything other than `nullptr`, which means it can just use the NSDMI in the definition.

include/nlohmann/json.hpp
gregmarr: Agreed.

include/nlohmann/json.hpp
```diff
@@ -3525,12 +3665,24 @@ class basic_json
                 m_value.array->insert(m_value.array->end(),
                                       idx - m_value.array->size() + 1,
                                       basic_json());
+#if JSON_DIAGNOSTICS
+                for (std::size_t i = idx + 1; i < m_value.array->size(); ++i)
```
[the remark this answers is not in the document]
gregmarr: Yes, need to save the old size from before the `insert` call.

include/nlohmann/json.hpp
```diff
@@ -3525,12 +3665,24 @@ class basic_json
...
+                {
+                    m_value.array->operator[](i).m_parent = this;
+                }
+#endif
             }
 
+#if JSON_DIAGNOSTICS
+            reference result = m_value.array->operator[](idx);
+            result.m_parent = this;
```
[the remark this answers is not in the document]
gregmarr: It can't, because the insert above only happens if the array needs to be expanded.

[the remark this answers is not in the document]
karzhenkov: The parent should be set only when the array is expanded. Existing elements already have the parent.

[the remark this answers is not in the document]
gregmarr: Ah, I see what you're saying now, that this assignment isn't necessary because it was previously done when the array was expanded to be large enough to handle this index.

include/nlohmann/json.hpp
nlohmann: Right - this was a leftover from debugging.

include/nlohmann/json.hpp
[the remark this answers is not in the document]
nlohmann: Good catch!

include/nlohmann/json.hpp
nlohmann: What do you propose as fix?

karzhenkov: 1. As @gregmarr said, the NSDMI in the class definition will be enough here. The above line needs just to be deleted. After the object is initialized, the parent must be `nullptr`.
2. Taking into account the previous, the assignment below should be performed after the element is moved into a container (as also proposed by @gregmarr). https://github.com/nlohmann/json/blob/294fa343d54634856dfef4a645de5241bd2dcde1/include/nlohmann/json.hpp#L1646
3. The `operator = ` should not modify the parent pointer.

include/nlohmann/json.hpp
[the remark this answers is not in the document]
nlohmann: Done.

include/nlohmann/json.hpp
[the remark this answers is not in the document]
nlohmann: In fact, only after fixing the move constructor, I could drop that part of the code :)

karzhenkov:
There is a missed point. The following assigment will be lost after the object is moved to the container:
https://github.com/nlohmann/json/blob/d4a91b7445e88d81730ca75a5905e514f58e9d05/include/nlohmann/json.hpp#L1646

nlohmann:
> There is a missed point. The following assigment will be lost after the object is moved to the container:
> 
> https://github.com/nlohmann/json/blob/d4a91b7445e88d81730ca75a5905e514f58e9d05/include/nlohmann/json.hpp#L1646

Unfortunately, I cannot create an example where the parent is lost.

Would you propose setting the parent after calling `emplace`?

karzhenkov:
I think it can be done something like this: 22bd1e1508913f1949b6a90a6b00474447b812e5

include/nlohmann/json.hpp
nlohmann: @gregmarr You're right - much better without `std::for_each`.

nlohmann:
> I think it can be done something like this: [karzhenkov/json@22bd1e1](https://github.com/karzhenkov/json/commit/22bd1e1508913f1949b6a90a6b00474447b812e5)

I thought so. Done.

include/nlohmann/detail/conversions/from_json.hpp
nlohmann: Done.

include/nlohmann/json.hpp
nlohmann: I added a `set_parent` function. Good cleanup!

include/nlohmann/json.hpp
nlohmann: I just need to find a place to add the assertion - then I think I'm done for the moment.

nlohmann: I added a check for the parent relation to the `assert_invariant` function. By adding it, I found quite some places I forgot to set the relation. Furthermore, there are two places where the invariant does not hold:

- in the destructor, and
- in the move constructor

I have not figured out why there is a problem - I guess because we do not destroy values in the order we created them, but added code to avoid a stack overflow. That optimization may yield invalid intermediate states. For this, I added a parameter `check_parents` to `assert_invariant` to not check the parent relation for these functions.

nlohmann:
@karzhenkov @gregmarr Thanks a lot for your review comments. I tried to address them all. It would be great if you could have another look.

include/nlohmann/json.hpp
```diff
@@ -1227,13 +1233,70 @@ class basic_json
...
+
+#if JSON_DIAGNOSTICS
+        JSON_ASSERT(!check_parents || !is_structured() || std::all_of(begin(), end(), [this](const basic_json & j)
+        {
+            return j.m_parent == this;
+        }));
+#endif
+    }
+
+    reference set_parent(reference j, bool recursive)
```
gregmarr, line 1261: I'm concerned that this form allows you to accidentally do `j.set_parent(j, false)` and set `j`'s `m_parent` to itself.  It's also very weird to see `j.set_parent(j...` in the code.  Could these be two separate functions, such as `set_parent(reference j)` and `set_parents()`?  This way it's clear that one is setting the parent of the passed in object, and the other is setting the parents within `this`.

include/nlohmann/detail/conversions/to_json.hpp
```diff
@@ -164,6 +167,9 @@ struct external_constructor<value_t::array>
         for (const bool x : arr)
         {
             j.m_value.array->push_back(x);
+#if JSON_DIAGNOSTICS
+            j.m_value.array->back().m_parent = &j;
```
gregmarr, line 171: `j.set_parent(...`

include/nlohmann/json.hpp
```diff
@@ -1227,13 +1233,70 @@ class basic_json
     invariant. Furthermore, it has to be called each time the type of a JSON
     value is changed, because the invariant expresses a relationship between
     @a m_type and @a m_value.
+
+    Furthermore, the parent relation is checked for arrays and objects: If
+    @a check_parents true and the value is an array or object, then the
+    container's elements must have the current value as parent.
+
+    @param[in] check_parents  whether the parent relation should be checked.
+               The value is true by default and should only be set to true
```
gregmarr, line 1242: ...only be set to `false`.

include/nlohmann/json.hpp
```diff
@@ -3522,15 +3592,26 @@ class basic_json
...
+                const auto previous_size = m_value.array->size();
+#endif
+
+                m_value.array->resize(idx + 1);
+
+#if JSON_DIAGNOSTICS
+                // set parent for values added above
+                for (auto i = previous_size; i <= idx; ++i)
+                {
+                    m_value.array->operator[](i).m_parent = this;
```
gregmarr, line 3606: I'm torn between leaving this as a direct access to `m_parent` because we're already inside `#if JSON_DIAGNOSTICS ` and using `set_parent(` for consistency.

include/nlohmann/json.hpp
```diff
@@ -5644,28 +5745,37 @@ class basic_json
...
+#if JSON_DIAGNOSTICS
+        iterator result = insert_iterator(pos, first.m_it.array_iterator, last.m_it.array_iterator);
+        for (typename iterator::difference_type i = 0; i < std::distance(first, last); ++i)
+        {
+            (result + i)->m_parent = this;
+        }
+        return result;
+#else
         return insert_iterator(pos, first.m_it.array_iterator, last.m_it.array_iterator);
+#endif
```
gregmarr, line 5778: As above

        iterator result = insert_iterator(pos, first.m_it.array_iterator, last.m_it.array_iterator);
    #if JSON_DIAGNOSTICS
        for (typename iterator::difference_type i = 0; i < std::distance(first, last); ++i)
        {
            set_parent(result + i);
        }
    #endif
        return result;

include/nlohmann/json.hpp
```diff
@@ -5697,17 +5807,27 @@ class basic_json
...
 
         // insert to array and return iterator
+#if JSON_DIAGNOSTICS
+        const auto size = ilist.size();
+        iterator result =  insert_iterator(pos, ilist.begin(), ilist.end());
+        for (std::size_t i = 0; i < size; ++i)
+        {
+            (result + static_cast<typename iterator::difference_type>(i))->m_parent = this;
+        }
+        return result;
```
gregmarr, line 5827: As above

        iterator result =  insert_iterator(pos, ilist.begin(), ilist.end());
    #if JSON_DIAGNOSTICS
        for (typename iterator::difference_type i = 0; i < static_cast<typename iterator::difference_type>(ilist.size()); ++i)
        {
            set_parent(result + i);
        }
    #endif
        return result;

include/nlohmann/json.hpp
```diff
@@ -5599,14 +5691,23 @@ class basic_json
...
+#if JSON_DIAGNOSTICS
+            iterator result = insert_iterator(pos, cnt, val);
+            for (size_type i = 0; i < cnt; ++i)
+            {
+                (result + static_cast<typename iterator::difference_type>(i))->m_parent = this;
+            }
+            return result;
+#else
             return insert_iterator(pos, cnt, val);
+#endif
```
gregmarr, line 5707: For the reasons above:

            iterator result = insert_iterator(pos, cnt, val);
    #if JSON_DIAGNOSTICS
            for (typename iterator::difference_type i = 0; i < static_cast<typename iterator::difference_type>(cnt); ++i)
            {
                set_parent(result + i);
            }
    #endif
            return result;

include/nlohmann/json.hpp
```diff
@@ -5548,14 +5634,20 @@ class basic_json
...
             }
 
             // insert to array and return iterator
+#if JSON_DIAGNOSTICS
+            iterator result = insert_iterator(pos, val);
+            result->m_parent = this;
+            return result;
+#else
             return insert_iterator(pos, val);
+#endif
```
gregmarr, line 5647: So that there is no possibility of mismatch between #if/#else should we do the following?

    iterator result = insert_iterator(pos, val);
    set_parent(result);
    return result;

include/nlohmann/json.hpp
gregmarr: This happens enough that it could be a function:

    void set_parents(iterator result, iterator::difference_type count)

include/nlohmann/json.hpp
gregmarr: This is the comment referred to as "As above" in the ones that actually appear above this one on the conversation view because I hadn't actually submitted this one when I submitted the other ones.

include/nlohmann/json.hpp
nlohmann: Good idea. Done in b9d3aa40670a57cedeb71e52c192a7678f0e23c2.

include/nlohmann/json.hpp
nlohmann: Good catch! Fixed in a77621687a826bfb020a3ceeda656f7b45479b44.

include/nlohmann/detail/conversions/to_json.hpp
nlohmann: Fixed in 0d1fb383b7f4572a154d7a26a1318e2c617b4de8.

include/nlohmann/json.hpp
nlohmann: Fixed in f8037660d06a0fbb6b1cde944d436032d04564ae.

include/nlohmann/json.hpp
nlohmann: Good idea, I took the first proposal. See f8037660d06a0fbb6b1cde944d436032d04564ae.

include/nlohmann/json.hpp
nlohmann: Fixed with f8037660d06a0fbb6b1cde944d436032d04564ae.

include/nlohmann/json.hpp
nlohmann: Fixed with f8037660d06a0fbb6b1cde944d436032d04564ae.

include/nlohmann/json.hpp
nlohmann: Fixed with f8037660d06a0fbb6b1cde944d436032d04564ae.

nlohmann:
Thanks @gregmarr for all your comments! `m_parent` is now only accessed in the `set_parent`/`set_parents` functions, and only one function still needs an `#if JSON_DIAGNOSTICS` block.

As remaining question for this PR I'm asking myself whether just adding the JSON Pointer to the exception is sufficient.

include/nlohmann/json.hpp
```diff
@@ -5886,6 +5983,8 @@ class basic_json
     {
         std::swap(m_type, other.m_type);
         std::swap(m_value, other.m_value);
+
+        set_parents();
```
gregmarr, line 5987: Need to call `other.set_parents();` too.

include/nlohmann/json.hpp
```diff
@@ -3522,15 +3603,26 @@ class basic_json
...
+
+                m_value.array->resize(idx + 1);
+
+#if JSON_DIAGNOSTICS
+                // set parent for values added above
+                for (auto i = previous_size; i <= idx; ++i)
+                {
+                    set_parent(m_value.array->operator[](i));
+                }
+#endif
```
gregmarr, line 3619: I think if you wanted this could be done as `set_parents(m_value.array->begin() + previous_size, idx + 1 - previous_size);`

gregmarr:
"As remaining question for this PR I'm asking myself whether just adding the JSON Pointer to the exception is sufficient."

nlohmann:
> "As remaining question for this PR I'm asking myself whether just adding the JSON Pointer to the exception is sufficient."
> 
> Do you mean putting a "JSON Pointer" in the exception, or creating a "JSON Pointer" and converting to string?

I just mean putting JSON Pointer strings like `/0` to the exceptions just like they are right now. I don't like the fact that the empty string is used to describe the top-level value. This is something to think about in this PR.

I'm also not sure whether the individual wording of the messages could be improved. I'm also thinking about overworking https://json.nlohmann.me/home/exceptions/ or create an individual page for every exception and add a URL to the `what()` message to avoid answering the same questions over and over again...

include/nlohmann/json.hpp
nlohmann: I thought about this, too, but thought such code would be too ugly. But since you had the same idea, I think I'll get rid of that loop.

include/nlohmann/json.hpp
nlohmann: Good catch!

nlohmann: Fixed in b0d8628c498d1a2be3391a1949886843746b9878.

include/nlohmann/json.hpp
nlohmann: Fixed in b0d8628c498d1a2be3391a1949886843746b9878.

include/nlohmann/json.hpp
```diff
@@ -3607,16 +3607,10 @@ class basic_json
...
                 m_value.array->resize(idx + 1);
 
-#if JSON_DIAGNOSTICS
                 // set parent for values added above
-                for (auto i = previous_size; i <= idx; ++i)
-                {
-                    set_parent(m_value.array->operator[](i));
-                }
-#endif
+                set_parents(begin() + previous_size, idx + 1 - previous_size);
```
gregmarr: Since you are now always using `previous_size` you need to remove the `#if/#endif` above.

gregmarr (changes requested):

include/nlohmann/json.hpp
gregmarr: ...or you can add them back in as you did in the next commit.  :)

include/nlohmann/detail/diagnostics_t.hpp
```diff
@@ -0,0 +1,86 @@
...
+            switch (current->m_parent->type())
+            {
+                case value_t::array:
+                {
+                    for (std::size_t i = 0; i < current->m_parent->m_value.array->size(); ++i)
+                    {
+                        if (current->m_parent->m_value.array->operator[](i) == *current)
+                        {
+                            tokens.emplace_back(std::to_string(i));
+                            continue;
```
gregmarr, line 42: `continue` is meaningless here, did you mean `break` to exit the loop?

include/nlohmann/detail/diagnostics_t.hpp
```diff
@@ -0,0 +1,86 @@
...
+                }
+
+                case value_t::object:
+                {
+                    for (const auto& element : *current->m_parent->m_value.object)
+                    {
+                        if (element.second == *current)
+                        {
+                            tokens.emplace_back(element.first.c_str());
+                            continue;
```
gregmarr, line 55: `continue` is meaningless here, did you mean `break` to exit the loop?

include/nlohmann/detail/diagnostics_t.hpp
```diff
@@ -0,0 +1,86 @@
...
+        {
+            return a + "/" + detail::escape(b);
+        }) + ") ";
+#else
+        return "";
+#endif
+    }
+
+  private:
+    const BasicJsonType* m_j = static_cast<const BasicJsonType*>(nullptr);
```
gregmarr, line 82: You shouldn't need the `static_cast` here.

include/nlohmann/detail/diagnostics_t.hpp
nlohmann: Good catch!

nlohmann: Fixed in d6ff059a901fc69163d9a42e15fcef87278fb23d.

include/nlohmann/detail/diagnostics_t.hpp
nlohmann: Fixed in d6ff059a901fc69163d9a42e15fcef87278fb23d.

include/nlohmann/detail/diagnostics_t.hpp
nlohmann: Fixed in d6ff059a901fc69163d9a42e15fcef87278fb23d.

doc/mkdocs/docs/features/macros.md
```diff
@@ -12,6 +12,14 @@ This macro overrides `#!cpp catch` calls inside the library. The argument is the
...
 See [Switch off exceptions](../home/exceptions.md#switch-off-exceptions) for an example.
 
+## `JSON_DIAGNOSTICS`
+
+This macro enables extended diagnostics for exception messages. Possible values are `1` to enable or `0` to disable (default).
+
+When enabled, exception messages contain a [JSON Pointer](json_pointer.md) to the JSON value that triggered the exception, see [Extended diagnostic messages](../home/exceptions.md#extended-diagnostic-messages) for an example.
+
+The diagnostics messages can also be controlled with the CMake option `JSON_Diagnostics` (`OFF` by default) which sets `JSON_DIAGNOSTICS` accordingly.
+
```
gregmarr, line 22: I think this should this mention that it increases the size of every object by one pointer, and makes some operations take a little bit longer, so that people know the cost of enabling it.  Is this pointer available for users to read, or is it completely private?  I don't remember seeing anything that would allow it.

gregmarr: Since you already write that in exceptions.md, perhaps just a reference to those docs.

nlohmann: Done in 7b7da08fb643c75dd18516ac8835ce2f2d5509a6.

nlohmann:
@gregmarr @karzhenkov I think this branch is ready now. Any further comment or observation?

karzhenkov (changes requested):

doc/examples/diagnostics_extended.link
```diff
@@ -0,0 +1 @@
+<a target="_blank" href="https://wandbox.org/permlink/pbz3ULoJ4maRnV8N"><b>online</b></a>
```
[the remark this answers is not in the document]
nlohmann: I also saw that. Maybe there is a caching issue on their side downloading the library? As that example is not part of Doxygen (and the Wandbox links are only showed there), I don't think that the link is an issue right now.

include/nlohmann/detail/diagnostics_t.hpp
```diff
@@ -0,0 +1,86 @@
...
+        std::vector<std::string> tokens;
+        for (const auto* current = m_j; current->m_parent != nullptr; current = current->m_parent)
+        {
+            switch (current->m_parent->type())
+            {
+                case value_t::array:
+                {
+                    for (std::size_t i = 0; i < current->m_parent->m_value.array->size(); ++i)
+                    {
+                        if (current->m_parent->m_value.array->operator[](i) == *current)
```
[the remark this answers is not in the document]
karzhenkov: The output is different: https://godbolt.org/z/Wb5Ejq. Did you `make amalgamate`? The single-header version needs also to be changed. Without the fix the wrong element index is reported.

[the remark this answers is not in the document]
nlohmann: I'll try again.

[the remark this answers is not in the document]
nlohmann: Thanks, good catch! Fixed in 380a613f2b5d32425021129cd1f371ddcfd54ddf.

include/nlohmann/detail/diagnostics_t.hpp
```diff
@@ -0,0 +1,86 @@
...
+                        }
+                    }
+                    break;
+                }
+
+                case value_t::object:
+                {
+                    for (const auto& element : *current->m_parent->m_value.object)
+                    {
+                        if (element.second == *current)
```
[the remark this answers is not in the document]
nlohmann: Fixed in 380a613f2b5d32425021129cd1f371ddcfd54ddf.

include/nlohmann/detail/json_pointer.hpp
```diff
@@ -926,7 +884,7 @@ class json_pointer
         {
             if (JSON_HEDLEY_UNLIKELY(!element.second.is_primitive()))
             {
-                JSON_THROW(detail::type_error::create(315, "values in object must be primitive"));
+                JSON_THROW(detail::type_error::create(315, "values in object must be primitive", diagnostics_t(element)));
```
karzhenkov, line 887:
```suggestion
                JSON_THROW(detail::type_error::create(315, "values in object must be primitive", diagnostics_t(element.second)));
```

karzhenkov:
Here are also some proposals to undo unnecessary changes and make the code a bit more concise: karzhenkov#1

nlohmann:
> Here are also some proposals to undo unnecessary changes and make the code a bit more concise: [karzhenkov#1](https://github.com/karzhenkov/json/pull/1)

I took over some of your ideas: 74cc0ab47098f0b25c0436564924c437c2254781.

karzhenkov:
It would also be nice to get rid of the empty json objects that are created in many places just to indicate "no diagnostics". This can be achived with an overload of `exception::diagnostics` taking `std::nullptr_t` and with some defaults (`BasicJsonType = std::nullptr_t`; `context = nullptr`) here:
https://github.com/nlohmann/json/blob/74cc0ab47098f0b25c0436564924c437c2254781/include/nlohmann/detail/exceptions.hpp#L187-L188
Other similar function templates need the same defaults.

As a possible point of additional extension, we can also consider other types of `context` that may be appropriate if no relevant json object is available.

gregmarr (commented):
Ugh, started a review yesterday and forgot to submit it.

include/nlohmann/detail/exceptions.hpp
```diff
@@ -258,9 +318,10 @@ caught.,type_error}
 class type_error : public exception
 {
   public:
-    static type_error create(int id_, const std::string& what_arg)
+    template<typename BasicJsonType>
+    static type_error create(int id_, const std::string& what_arg, const BasicJsonType& context)
     {
-        std::string w = exception::name("type_error", id_) + what_arg;
+        std::string w = exception::name("type_error", id_) + exception::diagnostics<BasicJsonType>(context) + what_arg;
```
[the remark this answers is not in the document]
nlohmann: Probably not. I'll check.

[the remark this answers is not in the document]
nlohmann: Fixed in 524eea58871f60968f3e7a3389942ae941b90c47.

include/nlohmann/detail/exceptions.hpp
```diff
@@ -305,9 +366,10 @@ caught.,out_of_range}
 class out_of_range : public exception
 {
   public:
-    static out_of_range create(int id_, const std::string& what_arg)
+    template<typename BasicJsonType>
+    static out_of_range create(int id_, const std::string& what_arg, const BasicJsonType& context)
     {
-        std::string w = exception::name("out_of_range", id_) + what_arg;
+        std::string w = exception::name("out_of_range", id_) + exception::diagnostics<BasicJsonType>(context) + what_arg;
```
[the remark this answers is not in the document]
nlohmann: Fixed in 524eea58871f60968f3e7a3389942ae941b90c47.

include/nlohmann/detail/exceptions.hpp
```diff
@@ -343,9 +405,10 @@ caught.,other_error}
 class other_error : public exception
 {
   public:
-    static other_error create(int id_, const std::string& what_arg)
+    template<typename BasicJsonType>
+    static other_error create(int id_, const std::string& what_arg, const BasicJsonType& context)
     {
-        std::string w = exception::name("other_error", id_) + what_arg;
+        std::string w = exception::name("other_error", id_) + exception::diagnostics<BasicJsonType>(context) + what_arg;
```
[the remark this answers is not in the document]
nlohmann: Fixed in 524eea58871f60968f3e7a3389942ae941b90c47.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -137,7 +137,7 @@ class binary_reader
             if (JSON_HEDLEY_UNLIKELY(current != std::char_traits<char_type>::eof()))
             {
                 return sax->parse_error(chars_read, get_token_string(),
-                                        parse_error::create(110, chars_read, exception_message(format, "expected end of input; last byte: 0x" + get_token_string(), "value")));
+                                        parse_error::create(110, chars_read, exception_message(format, "expected end of input; last byte: 0x" + get_token_string(), "value"), BasicJsonType()));
```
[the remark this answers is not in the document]
nlohmann: I tried earlier to work with a default value, but I got an error that the type could not be derived. I'll give it another check later.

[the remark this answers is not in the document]
gregmarr: Ah, it's a templated parameter, yeah that makes it harder to have a default.

gregmarr:
LGTM

PR #2637, opened 2021-02-11
Fix compilation error that can occur with custom StringType

sjanel:
Fix compilation errors occurring in some ```JSON_THROW``` when using custom ```StringType```.

For instance, here is the compilation error when using ```at()``` method with ```folly::fbstring```:
```
_deps/json-src/include/nlohmann/json.hpp:19903:68: error: cannot convert ‘folly::basic_fbstring<char>’ to ‘const string&’ {aka ‘const std::__cxx11::basic_string<char>&’}
19903 |                 JSON_THROW(out_of_range::create(403, "key '" + key + "' not found"));
      |                                                      ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
      |                                                                    |
      |                                                                    folly::basic_fbstring<char>
_deps/json-src/include/nlohmann/json.hpp:2065:41: note: in definition of macro ‘JSON_THROW’
 2065 |     #define JSON_THROW(exception) throw exception
      |                                         ^~~~~~~~~
```
I am a new user of this json library so there may be other areas which need to be updated as well.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -3499,7 +3499,7 @@ class basic_json
             JSON_CATCH (std::out_of_range&)
             {
                 // create better exception explanation
-                JSON_THROW(out_of_range::create(403, "key '" + key + "' not found", *this));
+                JSON_THROW(out_of_range::create(403, "key '" + std::string(key.begin(), key.end()) + "' not found", *this));
             }
         }
         else
@@ -3550,7 +3550,7 @@ class basic_json
             JSON_CATCH (std::out_of_range&)
             {
                 // create better exception explanation
-                JSON_THROW(out_of_range::create(403, "key '" + key + "' not found", *this));
+                JSON_THROW(out_of_range::create(403, "key '" + std::string(key.begin(), key.end()) + "' not found", *this));
             }
         }
         else
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -20282,7 +20282,7 @@ class basic_json
             JSON_CATCH (std::out_of_range&)
             {
                 // create better exception explanation
-                JSON_THROW(out_of_range::create(403, "key '" + key + "' not found", *this));
+                JSON_THROW(out_of_range::create(403, "key '" + std::string(key.begin(), key.end()) + "' not found", *this));
             }
         }
         else
@@ -20333,7 +20333,7 @@ class basic_json
             JSON_CATCH (std::out_of_range&)
             {
                 // create better exception explanation
-                JSON_THROW(out_of_range::create(403, "key '" + key + "' not found", *this));
+                JSON_THROW(out_of_range::create(403, "key '" + std::string(key.begin(), key.end()) + "' not found", *this));
             }
         }
         else


include/nlohmann/json.hpp
```diff
@@ -3499,7 +3499,7 @@ class basic_json
             JSON_CATCH (std::out_of_range&)
             {
                 // create better exception explanation
-                JSON_THROW(out_of_range::create(403, "key '" + key + "' not found", *this));
+                JSON_THROW(out_of_range::create(403, "key '" + std::string(key.begin(), key.end()) + "' not found", *this));
```
gregmarr, line 3502: This creates an unnecessary copy in the general case.  Does the custom string type not have the ability to form strings in this way, or is the issue in passing the formed string to the `create` function?

sjanel: Hum, normally throwing an exception should stay... exceptional so I don't think the additional copy makes any problem. 
The OOM of the cost of throwing an exception is huge compared to the creation of a new string.

gregmarr: Just trying to understand possible solutions.  Is the problem that `folly::basic_fbstring<char>` can't be combined with plain string literals?  Is it that the entire resulting string is a `folly::basic_fbstring<char>` and there is no implicit conversion from it to `std::string` as needed by the C++ exception types? If it's the latter, but there is an explicit conversion, then wrapping the entire thing in `std::string()` might provide the cleanest solution.

The library can't completely eliminate `std::string` usage as it's used in the standard exception types.

gregmarr: Looks like `fbstring` can be created from an `std::string`, or you can create a `std::string` using `toStdString`, but there's no conversion operator.  Could potentially do something like the `toStdString` function.  Might be better to make the strings in the `create` functions template types and do the conversions inside the `create` functions.

gregmarr: Looks like `fbstring` can be created from an `std::string`, or you can create a `std::string` using `toStdString`, but there's no conversion operator.  Could potentially do something like the `toStdString` function.  Might be better to make the strings in the `create` functions template types and do the conversions inside the `create` functions.

gregmarr: On the other hand, if this is the only location where we use a `StringType` to form an exception string, it may be worth it to just update this one location.

t-b: There is a expliciti conversion to std::string called `toStdString`, see https://github.com/facebook/folly/blob/2fd1e607d92f5af7118599dfbe4f1d7e0e8810ec/folly/FBString.h#L1098.

Stepping back a bit. Would it make sense to require custom string types to specialize something like `nlohmann::to_string`? In that way this is extensible and can get working with every string type and json (this library) does not have to think about it when using the std::string type.

sjanel: What about avoiding the use of + operator here and simply use the methods of the `StringType` like this:
```StringType s("key ");
s.append(key.begin(), key.end());
s.append(" not found");```
If the StringType has standard api it will always work.

t-b: Well not exactly pretty, but it does the job. Would you be able to provide a simple test case

nlohmann:
Can you please update to the latest develop branch? It contains a lot of changes to the CI.

sjanel:
Hello,

I don't reproduce the compiling issues with latest code base and `folly::fbstring`, closing the PR. Thanks!

PR #2729, opened 2021-04-21
Possibility to use without the dependency to file io and streams to use in intel sgx enclaves

### LINKED ISSUE #2728: Possibility to use without the dependency to file io and streams to use in intel sgx enclaves
I need to use your library in an intel sgx enclave. 
For security reasons it is not allowed to use file io there.

I created the pull request #2729 which includes some macros to be defined for using this great library without the file io.

theShmoo:
includes some macros to be defined for using without file io.


nlohmann (changes requested):
Please run `make amalgamate` to make sure the single header is properly generated. The CI currently fails because this.

nlohmann:
Sorry for the inactivity. I have one question left: you write in #2728 that the headers for I/O do not exist. But in your change you seem not to exclude any headers if `JSON_NO_IO` is defined.

theShmoo:
Ah that is true!
The headers exist in Intels SGX implementation of the standard library. But they are empty:

for example in `iostream`: https://github.com/intel/linux-sgx/blob/master/sdk/tlibcxx/include/iostream

Should I remove the includes so that it also usable for other applications?

nlohmann:
Yes, please. The use case may be still quite special, but at lease it would not assume include included headers are empty. I could imagine that that flag may be usable for some if they want to reduce compile times if they do not use I/O in their project.

theShmoo:
Does the build pipeline fail because of me or is this an error in the pipeline?

nlohmann:
This seems to be an issue in the pipeline. I need to investigate.

nlohmann:
Please update the branch - I just merged #2817 to fix the CI.

nlohmann (approved):
Looks good to me.

theShmoo:
Thank you very much for your support and your review!

nlohmann:
Thanks!

PR #2731, opened 2021-04-21
Fixes #2730

### LINKED ISSUE #2730: Possibility to use with custom c++ version to use in intel sgx enclaves
I need to use your library in an intel sgx enclave.
For that I am restricted to another standard library implementation (the one from intel)
You provide nice defines for the used c++ standard.
I use c++17 but the library of intel only has some c++11 features.
So I need to define the used standard version by my own.

I created a pull request that fixes this: #2731

theShmoo:
Added define JSON_VERSION_IS_PREDEFINED
for skipping automatic JSON_HAS_CPP_* detection

### DIFF
--- a/include/nlohmann/detail/macro_scope.hpp
+++ b/include/nlohmann/detail/macro_scope.hpp
@@ -20,8 +20,8 @@
 #endif
 
 // C++ language standard detection
-// if the user wants to manually specify the used c++ version this is skipped
-#ifndef JSON_VERSION_IS_PREDEFINED
+// if the user manually specified the used c++ version this is skipped
+#if !defined(JSON_HAS_CPP_20) && !defined(JSON_HAS_CPP_17) && !defined(JSON_HAS_CPP_14) && !defined(JSON_HAS_CPP_11)
     #if (defined(__cplusplus) && __cplusplus >= 202002L) || (defined(_MSVC_LANG) && _MSVC_LANG >= 202002L)
         #define JSON_HAS_CPP_20
         #define JSON_HAS_CPP_17
@@ -32,6 +32,8 @@
     #elif (defined(__cplusplus) && __cplusplus >= 201402L) || (defined(_HAS_CXX14) && _HAS_CXX14 == 1)
         #define JSON_HAS_CPP_14
     #endif
+    // the cpp 11 flag is always specified because it is the minimal required version
+    #define JSON_HAS_CPP_11
 #endif
 
 // disable documentation warnings on clang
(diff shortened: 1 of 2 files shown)


nlohmann (changes requested):
Please run `make amalgamate` to make sure the single header is properly generated. The CI currently fails because this.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -20,15 +20,18 @@
...
-    #define JSON_HAS_CPP_20
-    #define JSON_HAS_CPP_17
-    #define JSON_HAS_CPP_14
-#elif (defined(__cplusplus) && __cplusplus >= 201703L) || (defined(_HAS_CXX17) && _HAS_CXX17 == 1) // fix for issue #464
-    #define JSON_HAS_CPP_17
-    #define JSON_HAS_CPP_14
-#elif (defined(__cplusplus) && __cplusplus >= 201402L) || (defined(_HAS_CXX14) && _HAS_CXX14 == 1)
-    #define JSON_HAS_CPP_14
+// if the user wants to manually specify the used c++ version this is skipped
+#ifndef JSON_VERSION_IS_PREDEFINED
```
nlohmann, line 24: How about replacing this with

```cpp
#if !defined(JSON_HAS_CPP_20) && !defined(JSON_HAS_CPP_17) && !defined(JSON_HAS_CPP_14)
```

Wouldn't this have the same effect, but does not need an additional macro?

nlohmann (changes requested):

theShmoo:
Yeah this sounds great!

But I would have to define the new define JSON_HAS_CPP_11

Is this ok?

Niels Lohmann ***@***.***> schrieb am Sa., 24. Apr. 2021,
12:34:

nlohmann:
> Yeah this sounds great! But I would have to define the new define JSON_HAS_CPP_11 Is this ok?

Oh right - I forgot this. Then I guess the best would be the following:

- Adjust the `if` to also check for `JSON_HAS_CPP11`.
- Define `JSON_HAS_CPP11` if now other C++ standard has been detected. This way, the (otherwise unused) variable is at least always set. This allows to maybe make some compile-time decisions later on, but also to "force" any C++ standard different to how it would have been detected.

What do you think?

theShmoo:
Sounds great! Will do tomorrow!

Niels Lohmann ***@***.***> schrieb am Sa., 24. Apr. 2021,
14:09:

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

theShmoo:
Thank you very much for your support!

Issue #2802, opened 2021-06-02
Swift Package Manager support


richardtop:
[Swift Package Manager](https://swift.org/package-manager/) can be used to integrate together C++ libraries on Apple platforms. For example, with SPM it's possible to [add this project](https://developer.apple.com/documentation/swift_packages/adding_package_dependencies_to_your_app) as a dependency to an iOS app or library.

#### What is the issue you have?

I'd like to use Swift Package Manager to integrate this library into my project.

#### Which compiler and operating system are you using?

- Compiler: Apple Clang / GNU++17 / it doesn't matter
- Operating system: macOS, iOS

#### Which version of the library did you use?

- [x] latest release version 3.9.1
- [ ] other release - please state the version: ___
- [ ] the `develop` branch

PR #2807, opened 2021-06-03
Swift package manager support

richardtop:
Issue: https://github.com/nlohmann/json/issues/2802
Discussion: https://github.com/nlohmann/json/discussions/2806

This adds support for the Swift Package Manager on Apple platforms.

I had to add an empty file to make SPM build the target, as header-only targets are not currently supported in the Swift Packages.

Similar solution has been used in Apple's swift-numerics repository:
Discussion: https://forums.swift.org/t/header-only-library-using-swift-package-manager/42700
Repository: https://github.com/apple/swift-numerics/blob/main/Sources/_NumericsShims/_NumericsShims.c

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -41,7 +41,7 @@ let package = Package(
                 [
                     "_SwiftPackageManagerFile.cpp"
                 ],
-            publicHeadersPath: "./single_include/nlohmann/")
+            publicHeadersPath: "./single_include/")
     ],
     
     cxxLanguageStandard: .cxx11
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,48 @@
+// swift-tools-version:5.4
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "json",
+    platforms: [
+        .iOS(.v9), .macOS(.v10_10), .tvOS(.v9), .watchOS(.v2)
+    ],
+    products: [
+        .library(
+            name: "json",
+            targets: ["json"]),
+    ],
+    targets: [
+        .target(
+            name: "json",
+            dependencies: [],
+            path: ".",
+            exclude:
+                [
+                    "appveyor.yml",
+                    "benchmarks",
+                    "cmake",
+                    "ChangeLog.md",
+                    "doc",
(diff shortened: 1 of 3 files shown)


nlohmann:
Do all files need to be put into the repository root?

richardtop:
As for now, this is impossible, as the "Swift Package" is defined as:
"a git repository with semantically versioned tags, that contains Swift sources and a Package.swift manifest file at its root."
https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#creating-a-package
So, the `Package.swift` should be in the root. However, the empty file can be located anywhere.

richardtop:
Please, let me know if there is anything preventing this pull request from being merged, I'm happy to make changes to get this finalized.

nlohmann:
> However, the empty file can be located anywhere. Would you like me to relocate it somewhere?

is that file really needed?

richardtop:
Yes, unfortunately, there is no way around it. I've added the comment about why that file is needed, it's a known limitation of the Swift Package Manager. Even Apple's official repositories use this approach: https://github.com/apple/swift-numerics/blob/main/Sources/_NumericsShims/_NumericsShims.c

nlohmann:
I honestly do not like this. For Homebrew, I have a dedicated repository not to "spam" this repository: https://github.com/nlohmann/homebrew-json. Maybe this is a good approach also for the Swift package manager - especially if the file does not need to be updated with every release. What do you think?

richardtop:
While I also don't like it, I'm not sure if this is possible with the current state of the SPM, as it expects `Package.swift` to be in the root of the repository.
Since it's known limitation, I'm pretty sure it'll be resolved in next versions of the SPM, since it's developing really fast.
On the other hand, not supporting SPM would complicate usage of this library for Apple platforms developers, as SPM is becoming the industry standard when managing dependencies for Apple apps.

nlohmann:
I honestly do not like this solution. I am sorry that this is inconvenient for Apple users, but if every package manager dumps a few files to this repo, then it starts to get really ugly... Sorry.

richardtop:
No problem, I understand. Apple users always have to think different✌️

PR #2843, opened 2021-06-26
Add templated version of DEFINE_TYPE macros

kunaltyagi:
Picking up from #2532

- Add macros for custom types in DEFINE_TYPE_{NON_,}INTRUSIVE
- Add macros for template version of DEFINE_TYPE_{NON_,}INTRUSIVE
- Refactor and add tests for templated version of DEFINE_TYPE macros

`make amalgamate` was not found, so the change was performed manually

* * *

## Pull request checklist

### DIFF
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -18627,7 +18627,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         detail::parser_callback_t<basic_json>cb = nullptr,
         const bool allow_exceptions = true,
         const bool ignore_comments = false
-    )
+                                 )
     {
         return ::nlohmann::detail::parser<basic_json, InputAdapterType>(std::move(adapter),
                 std::move(cb), allow_exceptions, ignore_comments);
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -2684,6 +2684,40 @@ using is_detected_convertible =
     {                                                                             \
     }
 
+/*!
+@brief macro to briefly define intrusive serialization of a given type to/from any basic_json object
+@def NLOHMANN_DEFINE_TYPE_INTRUSIVE_T
+@since version 3.9.2
+*/
+#define NLOHMANN_DEFINE_TYPE_INTRUSIVE_T(Type, ...)                                             \
+    template<typename BasicJsonType>                                                            \
+    friend void to_json(BasicJsonType& nlohmann_json_j, const Type& nlohmann_json_t)            \
+    {                                                                                           \
+        NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__))                \
+    }                                                                                           \
+    template<typename BasicJsonType>                                                            \
+    friend void from_json(const BasicJsonType& nlohmann_json_j, Type& nlohmann_json_t)          \
+    {                                                                                           \
+        NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_FROM, __VA_ARGS__))              \
+    }                                                                                           \
+
+/*!
+@brief macro to briefly define non-intrusive serialization of a given type to/from any basic_json object
+@def NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_T
+@since version 3.9.2
+*/
+#define NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_T(Type, ...)                            \
+    template<typename BasicJsonType>                                               \
+    void to_json(BasicJsonType& nlohmann_json_j, const Type& nlohmann_json_t)      \
+    {                                                                              \
+        NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__))   \
+    }                                                                              \
+    template<typename BasicJsonType>                                               \
+    void from_json(const BasicJsonType& nlohmann_json_j, Type& nlohmann_json_t)    \
+    {                                                                              \
+        NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_FROM, __VA_ARGS__)) \
+    }
+
 #ifndef JSON_USE_IMPLICIT_CONVERSIONS
     #define JSON_USE_IMPLICIT_CONVERSIONS 1
 #endif
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -2607,9 +2607,15 @@ using is_detected_convertible =
 @def NLOHMANN_DEFINE_TYPE_INTRUSIVE_IMPL
 @since version 3.9.2
 */
-#define NLOHMANN_DEFINE_TYPE_INTRUSIVE_IMPL(BasicJsonType, Type, ...)  \
-    friend void to_json(BasicJsonType& nlohmann_json_j, const Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__)) } \
-    friend void from_json(const BasicJsonType& nlohmann_json_j, Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_FROM, __VA_ARGS__)) }
+#define NLOHMANN_DEFINE_TYPE_INTRUSIVE_IMPL(BasicJsonType, Type, ...)                  \
+    friend void to_json(BasicJsonType& nlohmann_json_j, const Type& nlohmann_json_t)   \
+    {                                                                                  \
+        NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__))       \
+    }                                                                                  \
+    friend void from_json(const BasicJsonType& nlohmann_json_j, Type& nlohmann_json_t) \
+    {                                                                                  \
+        NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_FROM, __VA_ARGS__))     \
+    }
 
 /*!
 @brief macro to briefly define non-intrusive serialization of a given type to/from JSON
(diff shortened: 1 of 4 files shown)


kunaltyagi:
In the macros added, it's impossible to follow `macro argument should be enclosed in parentheses [bugprone-macro-parentheses,-warnings-as-errors]` since the macro arguments are the class name and the visibility type. Thoughts?
```cpp
// sample code
#define start_class(A, type) class A { type:
// these wouldn't compile
#define start_class2(A, type) class (A) { type:
#define start_class3(A, type) class A { (type):
```

nlohmann (changes requested):
Flex check and fix the issue found in the CI:

- https://github.com/nlohmann/json/pull/2843/checks?check_run_id=2931115424 
- https://github.com/nlohmann/json/pull/2843/checks?check_run_id=2931115844

test/src/unit-udt_macro.cpp
```diff
@@ -36,236 +36,199 @@ using nlohmann::json;
...
-    person_with_private_data(std::string name_, int age_, json metadata_)
-        : name(std::move(name_))
-        , age(age_)
-        , metadata(std::move(metadata_))
-    {}
-
+#define START_CLASS_PERSON(ClassName, Visibility)                                  \
+    class ClassName                                                                \
+    {                                                                              \
+        Visibility : std::string name;                                             \
```
kunaltyagi, line 42:
```suggestion
        Visibility : std::string name{};                                           \
```

kunaltyagi:
Hopefully fixed the first. Regarding the second error, please find my analysis in the comment above yours (https://github.com/nlohmann/json/pull/2843#issuecomment-869672937)

nlohmann:
There is some issue with the CI - the job `ci_test_clang_sanitizer` seems to fail also on `develop`. I'll have a look. Sorry for the inconvenience.

nlohmann:
Can you please update to the latest develop branch? The CI issues should be fixed then. Sorry for the inconvenience.

gregmarr:
@kunaltyagi Can you update to the latest develop branch?

kunaltyagi:
Rebased, and also refactored the code to make the macros a bit less offensive

single_include/nlohmann/json.hpp
```diff
@@ -18563,7 +18627,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         detail::parser_callback_t<basic_json>cb = nullptr,
         const bool allow_exceptions = true,
         const bool ignore_comments = false
-                                 )
+    )
```
kunaltyagi, line 18630:
```suggestion
                                 )
```

kunaltyagi:
Almost all tests are passing. Current issue:
```
C:\projects\json\tests\src\unit-udt_macro.cpp(413): error C2220: warning treated as error - no 'object' file generated [C:\projects\json\tests\test-udt_macro_cpp11.vcxproj]
C:\projects\json\tests\src\unit-udt_macro.cpp(413): warning C4503: 'doctest::detail::instantiationHelper': decorated name length exceeded, name was truncated [C:\projects\json\tests\test-udt_macro_cpp11.vcxproj]
```

* Line 413 is the last line of the file
* I'm not able to experiment with windows because I have never developed on that platform
* C2220 can be ignore (essentially -Werror)
* C4503 explanation is found [here](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4503?view=msvc-170)

If someone can identify the source, I can make a workaround.

falbrechtskirchinger:
It seems safe to just add a suppression for 4503.

kunaltyagi:
All checks pass. PTAL

kunaltyagi:
Are there any more blockers? 😅

tests/src/unit-udt_macro.cpp
```diff
@@ -390,9 +367,9 @@ TEST_CASE_TEMPLATE("Serialization/deserialization of classes with 26 public/priv
 
         {
             T obj1;
-            nlohmann::json j1 = obj1; //via msgpack
-            std::vector<uint8_t> buf = nlohmann::json::to_msgpack(j1);
-            nlohmann::json j2 = nlohmann::json::from_msgpack(buf);
+            json_t j1 = obj1;  //via msgpack
```
falbrechtskirchinger, line 370: I don't know how much sense it makes to test the different formats here. To the extent that there are roundtrip differences, these are best tested elsewhere.
You're also missing BJData, and when the BON8 PR is merged, this is likely to be missed.

kunaltyagi: This is existing code, which I modified (parameterized on different types) to support different underlying implementations for json: ordered and unordered map

I'd prefer to tackle the missing BJData and BON8 in a separate PR. Thoughts?

gregmarr: Here's the PR where they were added: https://github.com/nlohmann/json/pull/2287

falbrechtskirchinger: I just find the choice of roundtripping bizarre. It's already tested extensively for each format and unless you're testing the corner cases (which the macros don't handle anyway), there's no reason to do so here. Let's just do *something* to pad the unit test! :man_shrugging:

falbrechtskirchinger: Still, why test different formats? Ignoring format-specific exceptions, a roundtrip doesn't change the `basic_json`.

```cpp
SECTION("alphabet")
{
    json_t j = obj1;
    T obj2;
    j.get_to(obj2);
    CHECK((obj1 == obj2));
}
```

Is this not sufficient?

gregmarr: Could be, don't know for sure.  @pfeatherstone did you have a particular reason for not just doing one format here?

nlohmann:
@kunaltyagi It would be great to have this feature in the library. Do you have time to work on the remaining issues or should someone take over?

kunaltyagi:
I think the sole unaddressed comment explicitly states that I don't need to take any action. What issues are remaining?

nlohmann:
Oh I'm sorry - this was meant for @pfeatherstone...

kunaltyagi:
Seems to be not available.

Btw, can we change the label on the PR?

falbrechtskirchinger:
Is there a reason we can't make `*_T` the default? It wouldn't break existing code and is more sensible anyway.

Instead, I think we should (in a separate PR, of course) explore refactoring and renaming some of the implementation macros (i.e., `NLOHMANN_DEFINE_TYPE_WITH_DEFAULT_IMPL`) to give users the tools for creating more elaborate macros themselves, without having to delve too deeply into Preprocessor madness.

kunaltyagi:
This PR was a pain to rebase to master. Is it possible to move it ahead?

kunaltyagi:
> Is there a reason we can't make *_T the default

Those are templated versions. They would behave a bit differently with overloading than the current versions. The existing code in the repo wouldn't be broken, but the user experience might be affected due to how templates participate in overload resolution (wrt implicit conversions), which is surprising when compared to non-overloaded versions

falbrechtskirchinger:
> > Is there a reason we can't make *_T the default
> 
> Those are templated versions. They would behave a bit differently with overloading than the current versions. The existing code in the repo wouldn't be broken, but the user experience might be affected due to how templates participate in overload resolution (wrt implicit conversions), which is surprising when compared to non-overloaded versions

Non-template functions are preferred over template functions during overload resolution – true. I did not consider that. Still, what are the practical, real-world implications in this particular context? When I have some time later, I'll try to construct a contrived example to trigger different behavior, but If I even manage to do so, I wouldn't expect to find this in the wild. Maybe [GitHub Code Search](https://cs.github.com/) can help.
Also, note that all internal conversions are template functions, and if we catered to these concerns, we would never be able to add another conversion or risk interfering with existing code.

A separate issue came to mind: `BasicJsonType` should be constrained with `nlohmann::detail::enable_if_t<typename nlohmann::detail::is_basic_json<BasicJsonType>::type, int> = 0`.

falbrechtskirchinger:
> This PR was a pain to rebase to master. Is it possible to move it ahead?

The open issues are:
* I believe there's no need for separate macros. The template versions should just replace the existing ones. IMHO, the impacts of such a change are far more benign than adding a new conversion (which was done in recent history, and pending PRs would do so again).
* Depending on whether new macros are added or the existing ones are replaced, the macro documentation needs to be updated.
* Should `BasicJsonType` be SFINAE-constrained on `to_json`/`from_json` or is that redundant because it's already constrained earlier in the call chain?
* The unit tests for other serialization formats should be removed before merge because they serve no purpose. Let's at least make some effort for higher quality unit tests. I'll die on this hill if I have to. :-)

These shouldn't be too difficult to resolve and I'm holding back work on substantial changes until after this PR is merged.

kunaltyagi:
> A separate issue came to mind: `BasicJsonType` should be constrained with `nlohmann::detail::enable_if_t<typename nlohmann::detail::is_basic_json<BasicJsonType>::type, int> = 0`

Sure. I'll check if that's needed. If it is, I'll add that to the PR.

> * The unit tests for other serialization formats should be removed before merge because they serve no purpose. Let's at least make some effort for higher quality unit tests. I'll die on this hill if I have to. :-)

Someone will have to get that merged in master outside the PR. I don't disagree on your stand, just that it's not nice to bundle it in this PR.

> * I believe there's no need for separate macros.

cc: @nlohmann 
Please make a decision so I can move ahead with the relevant changes. I'm not against adopting the PR, but it'd be better if we agree on a direction before I make the changes

falbrechtskirchinger:
> > * The unit tests for other serialization formats should be removed before merge because they serve no purpose. Let's at least make some effort for higher quality unit tests. I'll die on this hill if I have to. :-)
> 
> Someone will have to get that merged in master outside the PR. I don't disagree on your stand, just that it's not nice to bundle it in this PR.

Right, I forgot that that's not something you added.

nlohmann:
I agree fully with @falbrechtskirchinger.

falbrechtskirchinger:
@kunaltyagi Heads up. To satisfy the new `clang-tidy` const-correctness check, we must add `const` *everywhere*. I expect the PR to be merged to `develop` soon.
Rebasing shouldn't be too complicated. Keep your changes and add const where possible. Don't worry about the mixed use of east and west const. We're switching to `clang-format` soon and will reformat the code anyway.

Disservin:
Some months have passed whats the status on this? ;)

YarikTH:
How many programmers need to support templated BasicJsonType in NLOHMANN_DEFINE_TYPE_MEOW? This macro is cursed

nlohmann:
Closed in favor of #4597.

PR #2925, opened 2021-08-11
Guard GCC pragmas

### LINKED ISSUE #2924: warning C4068: unknown pragma 'GCC' on MSVC/cl

When compiling with MSVC, The following warnings are created:

```
warning C4068: unknown pragma 'GCC'
```

1. Compile any program including `json.hpp` with `cl`
2. Observe warnings

```
// main.cpp
#include "nlohmann/json.hpp"

int main() { return 0; }
```

```
$ cl /EHsc main.cpp
```

No warning should appear.

MSVC warning C4068 appears.

- Compiler: `cl /?`, `Microsoft (R) C/C++ Optimizing Compiler Version 19.27.29111 for x86` using the Developer Command Prompt for VS 2019.
- Operating system: Windows 10

- [ ] latest release version 3.9.1
- [x] the `develop` branch

- [x] yes

Workaround:

Adding

```
#pragma warning ( disable: 4068 )
```

to the top of either file disables the warning, or wrap all `#pragma GCC ...`s in `#if defined(__clang__)`.

nlohmann:
This PR adds preprocessor checks to only use GCC pragmas with compatible compilers.

Fixes #2924

### DIFF
--- a/include/nlohmann/detail/conversions/to_chars.hpp
+++ b/include/nlohmann/detail/conversions/to_chars.hpp
@@ -1066,8 +1066,10 @@ char* to_chars(char* first, const char* last, FloatType value)
         *first++ = '-';
     }
 
+#ifdef __GNUC__
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wfloat-equal"
+#endif
     if (value == 0) // +-0
     {
         *first++ = '0';
@@ -1076,7 +1078,9 @@ char* to_chars(char* first, const char* last, FloatType value)
         *first++ = '0';
         return first;
     }
+#ifdef __GNUC__
 #pragma GCC diagnostic pop
+#endif
 
     JSON_ASSERT(last - first >= std::numeric_limits<FloatType>::max_digits10);
 
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -1525,8 +1525,10 @@ class binary_writer
 
     void write_compact_float(const number_float_t n, detail::input_format_t format)
     {
+#ifdef __GNUC__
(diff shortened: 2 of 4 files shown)


MHebes:
Just curious—there are other `#pragma GCC `s which are guarded on the `__clang__` variable instead of `__GNUC__`. I think clang defines the latter, but gcc might not define the former. Would it make sense to change those to `__GNUC__` as well?

nlohmann:
> Just curious—there are other `#pragma GCC `s which are guarded on the `__clang__` variable instead of `__GNUC__`. I think clang defines the latter, but gcc might not define the former. Would it make sense to change those to `__GNUC__` as well?

The only other place is this:

```cpp
// disable documentation warnings on clang
#if defined(__clang__)
    #pragma GCC diagnostic push
    #pragma GCC diagnostic ignored "-Wdocumentation"
#endif
```

This pragma is meant for Clang, because only Clang has a `-Wdocumentation` warning. If we would change it to `__GNUC__`, then GCC would be confused about this unknown flag.

t-b:
@nlohmann What is missing to avoid things like this with the CI? warning-free tests on all platforms?

nlohmann:
> @nlohmann What is missing to avoid things like this with the CI? warning-free tests on all platforms?

I currently set no specific warning flag for MSVC. Is there something comparable to `-Werror` which is still pragmatic for MSVC?

t-b:
`/WX` for warnings as errors and `/W4` for maximum warnings. Maybe you can scavenge something from https://github.com/nlohmann/json/pull/2245/commits/0b8644c452d864dc88d801ce55b47207051857d7.

nlohmann:
I'll give it a try, see https://github.com/nlohmann/json/pull/2930 - feedback more than welcome!

PR #2930, opened 2021-08-13
Treat MSVC warnings as errors

nlohmann:
This PR is a follow up to #2924 and treats all MSVC warnings as errors.

### DIFF
--- a/test/src/unit-allocator.cpp
+++ b/test/src/unit-allocator.cpp
@@ -100,6 +100,9 @@ struct my_allocator : std::allocator<T>
         std::allocator<T>::deallocate(p, n);
     }
 
+    // the code below warns about p in MSVC 2015 - this could be a bug
+    DOCTEST_MSVC_SUPPRESS_WARNING_PUSH
+    DOCTEST_MSVC_SUPPRESS_WARNING(4100)
     void destroy(T* p)
     {
         if (next_destroy_fails)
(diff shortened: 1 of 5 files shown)


test/src/unit-allocator.cpp
```diff
@@ -100,6 +100,9 @@ struct my_allocator : std::allocator<T>
         std::allocator<T>::deallocate(p, n);
     }
 
+    // the code below warns about p in MSVC 2015 - this could be a bug
```
gregmarr, line 103: "C4100 can also be issued when code calls a destructor on a otherwise unreferenced parameter of primitive type. This is a limitation of the Microsoft C++ compiler."
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4100?view=msvc-160

test/src/unit-udt.cpp
```diff
@@ -809,6 +809,10 @@ TEST_CASE("an incomplete type does not trigger a compiler error in non-evaluated
     static_assert(!is_constructible_patched<json, incomplete>::value, "");
 }
 
+// the code below warns about t in MSVC 2015 - this could be a bug
+DOCTEST_MSVC_SUPPRESS_WARNING_PUSH
```
gregmarr, line 813: Same class of error as the other one, `sizeof(t)` is a compile-time constant based only on `T`, not on the value of `t`.

test/src/unit-allocator.cpp
nlohmann: Thanks, I fixed it.

test/src/unit-udt.cpp
nlohmann: Thanks, I also fixed it.

PR #2943, opened 2021-08-18
Updating the Homebrew Command

amirmasoudabdol:
So, apparently, `homebrew/core` has a `nlohmann-json` formula. This means that, even after tapping to `nlohmann/json`, the user has to specify what formula should be used. and running `brew install nlohmann-json` will use the `homebrew/core` instead of the formula located at `nlohmann/json`. To resolve this, the install command needs to be more explicit and point to your formula.

You can point to your formula in two fashions, `brew install nlohmann_json` or more explicitly, `brew install nlohmann/json/nlohmann_json`. This distinguish it from the existing formula `homebrew/core/nlohmann-json` which I believe is not being maintained by people in this repo.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/README.md
+++ b/README.md
@@ -1238,7 +1238,7 @@ endif()
 
 ### Package Managers
 
-:beer: If you are using OS X and [Homebrew](https://brew.sh), just type `brew tap nlohmann/json` and `brew install nlohmann/json/nlohmann_json` and you're set. If you want the bleeding edge rather than the latest release, use `brew install nlohmann/json/nlohmann_json --HEAD`.
+:beer: If you are using OS X and [Homebrew](https://brew.sh), just type `brew install nlohmann-json` and you're set. If you want the bleeding edge rather than the latest release, use `brew install nlohmann-json --HEAD`.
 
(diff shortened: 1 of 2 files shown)


amirmasoudabdol:
I think this needs more attention. I just realised from some test that installing `nlohmann_json` from your `nlohmann/json` does not setup the CMake correctly. This might be because I had the version from `homebrew/core` installed and now it's being confused. But, I've tried a clean install and had the same problem. 

As we discussed on Twitter, if you are willing to drop the `nlohmann/json` Formula, then we can make on the `homebrew/core` Formula and keep that updated. I try to update the homebrew formula for this update and reflect it on the README. However, we need to have this as part of the release checklist to make sure that it's been keeping up to date.

amirmasoudabdol:
The [PR at homebrew/core](https://github.com/Homebrew/homebrew-core/pull/83479) has been approved, and will be merged soon. So, I think this one can go after that one.

nlohmann (changes requested):
Can you please also adjust the documentation in https://github.com/nlohmann/json/blob/develop/doc/mkdocs/docs/integration/package_managers.md#homebrew ? Thanks!

amirmasoudabdol:
I think that should be done!

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #2950, opened 2021-08-18
🚨 fix gdb_pretty_printer failure on basic types

senyai:
`name` can be `None`, and `re.search` fails on `None`;
replaced `re` by `str.startswith` and `str.endswith`


nlohmann (approved):
Looks good to me.

nlohmann:
Thanks a lot!

PR #2956, opened 2021-08-19
Allow std::unordered_map as ObjectType

### LINKED ISSUE #2932: Support for unordered_map as object_t
Is `std::unordered_map` supposed to be supported by now? The documentation mentions this (https://json.nlohmann.me/api/basic_json/object_t/#template-parameters), but it still fails to compile.

I know there've been passt issues tagged with `wontfix`. Just asking because `std::map` actually seems to be a bottleneck in our use-case and the docs still mention `std::unordered_map`.

nlohmann:
This PR fixes the definition of `object_t` to allow to use `std::unordered_map` as `ObjectType`.

Fixes #2932.

### DIFF
--- a/include/nlohmann/detail/meta/type_traits.hpp
+++ b/include/nlohmann/detail/meta/type_traits.hpp
@@ -449,5 +449,23 @@ T conditional_static_cast(U value)
     return value;
 }
 
+// helper to check if a type has bucket_count function (and hence can tell a std::map and a std::unordered_map apart)
+template <typename T>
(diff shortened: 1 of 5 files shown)


edelmanjm:
Can I ask why this was closed?

gregmarr:
It wasn't possible to implement at the time.  @nlohmann I wonder if the introduction of the intermediate `data` struct holding the value type means that it is possible to do this now, since the `basic_json` type could be complete.

PR #3014, opened 2021-09-12
Define workflow order

nlohmann:
This PR adds some ordering to the Ubuntu workflow.


nlohmann:
Closing the PR - this would slow down the builds.

PR #3020, opened 2021-09-17
meta: fix is_compatible/constructible traits

theodelrieu:
(my previous PR was empty and got auto-closed, duh)

The previous version relied on the existence of an iterator member type.

As mentioned in comments, this is not the proper way to do it and causes issues with certain types (e.g. views from range-v3 which do not have such an alias).

Add a is_range trait that properly detects the return type of begin/end, and use it instead.

* * *

## Pull request checklist

## Please don't


theodelrieu:
I ran `make pretty` but it seems that it made a lot of changes 🤔

include/nlohmann/detail/meta/can_call_std.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include <nlohmann/detail/meta/detected.hpp>
+
+// source: https://stackoverflow.com/a/26745591
+
+#define CAN_CALL_STD_FUNC_IMPL(std_name)                                          \
```
gregmarr, line 7: This was replaced with `NLOHMANN_CAN_CALL_STD_FUNC_IMPL` right?

gregmarr:
>I ran make pretty but it seems that it made a lot of changes 🤔

Would it be worth doing that in a separate commit within the PR or even a separate PR?

theodelrieu:
> Would it be worth doing that in a separate commit within the PR or even a separate PR?

I'll fix it.

include/nlohmann/detail/meta/can_call_std.hpp
theodelrieu: Should be good

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -292,6 +293,41 @@
     inline void to_json(nlohmann::json& nlohmann_json_j, const Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__)) } \
     inline void from_json(const nlohmann::json& nlohmann_json_j, Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_FROM, __VA_ARGS__)) }
 
+
+// source: https://stackoverflow.com/a/26745591
```
nlohmann, line 297: Please do not use code from StackOverflow - its license (CC BY-SA 3.0) is incompatible to MIT.

nlohmann (changes requested):

include/nlohmann/detail/macro_scope.hpp
theodelrieu: The code is inspired from SO, but is completely different.

nlohmann: Then please still rephrase the comment - I had issues in the past that automatic license checking tools raised their concerns which blocked people from using the company until they could convince their lawyers otherwise...

theodelrieu: Is it fine if I drop it?

theodelrieu: reworded instead

nlohmann: Thanks!

PR #3028, opened 2021-09-24
Allow to add metadata to json nodes

barcode:
When processing json it is often useful to annotate nodes with some metadata. This PR adds an additional template parameter which allows to set the type of metadata. 
By default the parameter is set to void and no metadata is allowed. In this case the library behaves as it already did.
If a user explicitly adds metadata (last template parameter) each node has a data member of this type and access functions for this member.

This metadata can be used to tag nodes or store information such as the line and column number from an input file.

* * *

## Pull request checklist

- [x]  Changes are **described in the pull request**, or an [existing issue is referenced](https://github.com/nlohmann/json/issues).
- [x]  [Code coverage](https://coveralls.io/github/nlohmann/json) is 100%. Test cases can be added by editing the [test suite](https://github.com/nlohmann/json/tree/develop/test/src). **(should be. will update state aver coverall is done)**

### DIFF
--- a/include/nlohmann/detail/json_metadata.hpp
+++ b/include/nlohmann/detail/json_metadata.hpp
@@ -1,0 +1,30 @@
+#pragma once
+
+namespace nlohmann
+{
+namespace detail
+{
+template<class MetaDataType>
+class json_metadata
+{
+  public:
+    using metadata_t = MetaDataType;
+    metadata_t& metadata()
+    {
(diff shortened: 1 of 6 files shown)


barcode:
Closed in favour of https://github.com/nlohmann/json/pull/3110 which implements the more powerful feature of a custom base class and thus allows adding custom member functions as well.

PR #3041, opened 2021-09-28
add support for from_json/to_json member function

yt-sun:
add support for from_json/to_json member function. for a example:
```
using nlohmann::json;
class A {
private:
  int a;
public:
  void from_json(const json&){...}
  void to_json(json&) const {...}
}
void fun(){
  A a;
  json j;
  j=a;
  j.get_to(a)
}
```


gregmarr:
@yt-sun Please check the results of the CI and update to the latest develop branch.

PR #3071, opened 2021-10-10
Consolidate documentation

nlohmann:
This PR consolidates two streams of documentation: Doxygen-style comments in the source code and the newer MkDocs-based Markdown documentation in folder `doc/mkdocs`. Both documentations diverged over time, and since we reached the limits of what Doxygen can do, we decided to drop Doxygen and go all-in on MkDocs. As a result, we removed all Doxygen-style comments for the public API and updated/fixed the MkDocs accordingly. The result is a much cleaner documentation that documents all overloads of functions with the same name on the same page.

In particular:

- move images from `doc/images` to `doc/mkdocs/images`
- fix typos in README and comments
- remove Doxygen files and code comments
- add README how to build documentation locally
- remove Wandbox online code examples
- add more examples
- overworked documentation of public API (<https://json.nlohmann.me/api/basic_json/>)
- documented all deprecations and added hints how to replace deprecated calls
- added documentation linter to enforce standard section naming and ordering
- remove Doxygen documentation from public API and added links to online pages
- adjusted parameter namings to those of the documentation


nlohmann:
> Assuming I understand this right and the documentation is still in the distribution, just not in the header file, should the comment with the link to the latest documentation also contain the path to the documentation for this version? I still see utility in being able to see the documentation as it was written at the time that the version I'm using was released. Will the documentation site still contain the documentation for any functions that are removed with v4 comes out and several of the deprecated functions are removed?

Good point!

The motivation for this PR is the observation that the documentation in the code and that in the docs folder start to diverge. Furthermore, Doxygen is so inflexible that the current output is ugly and needs a lot of repetitions in the code. That said, I want to drop Doxygen.

As the `docs` folder is tagged together with the sources, it should be able to get the documentation for the future versions. What do you think? Is this sufficient?

Maybe (?) there is also a way to have URLs like https://json.nlohmann.me/v3.10.3/api/basic_json/accept/ or so that contain the documentation for a specific version.

include/nlohmann/json.hpp
```diff
@@ -4707,20 +4707,16 @@ std::string to_string(const NLOHMANN_BASIC_JSON_TPL& j)
...
 
-// specialization of std::swap, and std::hash
-namespace std
-{
-
 /// hash value for JSON objects
-template<>
-struct hash<nlohmann::json>
+template<typename BasicJsonType>
+struct std::hash
```
gregmarr, line 4712: @nlohmann This is now a full redefinition of the `std::hash` struct, which is not allowed.  You still need to keep it limited to the library's type.  It needs to stay as before but use `nlohmann::basic_json<>>` instead of `nlohmann::json`.  (I'm not sure the exact proper way to spell `nlohmann::basic_json<>` here, I would start with that and see if it works.  If not, you'll have to add more template parameters.)

gregmarr: I think you need to do it like this:

    namespace std
    {
       NLOHMANN_BASIC_JSON_TPL_DECLARATION
       struct hash<NLOHMANN_BASIC_JSON_TPL>
       {
           std::size_t operator()(const NLOHMANN_BASIC_JSON_TPL &j)

nlohmann: Thanks a lot! I actually saw your fix before I noticed the CI was red ;-)

I will check out the other functions as well - it seems just having `nlohmann::json` in the code base usually is a problem in the future, as it will not work for other specializations like `nlohmann::ordered_json`.

gregmarr: I didn't notice the CI was red.  :)  I wasn't sure if this was something that the CI would catch, or if it would just lead to a silent ODR violation later.

nlohmann: I recently read https://quuxplusone.github.io/blog/2021/10/27/dont-reopen-namespace-std/ and wanted to apply the ideas. With that, I notices we lacked the specialization for `nlohmann::ordered_json`. Interestingly though, GCC 4.8 rejects the approach of the blog post, so we're back to the original code, but with your fix to the specialization.

gregmarr: Yeah, looks like this was a GCC bug until 7.x.

akapocsi:
@nlohmann The consolidated docs looks great. 

Just wanted to point out that there are some stale links/references to the doxygen docs in the main README: https://github.com/nlohmann/json/blob/develop/README.md?plain=1#:~:text=doxygen

nlohmann:
Thanks @akapocsi! Fixed in eb2182414749825be086c825edb5229e5c28503d.

PR #3125, opened 2021-11-06
Fix spelling

axic:
I am trying to use the `single_include` version of this library in a project ([here](https://github.com/ethereum/solidity/pull/11967)), and to simplify matters I thought to just include the file (or perhaps via subtree).

This means most of the linter tools run on the project would check this file. This is how I found the following complaints from codespell. I certainly can exclude this file from that tool, but thought it may be useful here too.

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -42,7 +42,7 @@ enum class cbor_tag_handler_t
 
 @note from https://stackoverflow.com/a/1001328/266378
 */
-static inline bool little_endianess(int num = 1) noexcept
+static inline bool little_endianness(int num = 1) noexcept
 {
     return *reinterpret_cast<char*>(&num) == 1;
 }
@@ -2341,7 +2341,7 @@ class binary_reader
 
     @return whether conversion completed
 
-    @note This function needs to respect the system's endianess, because
+    @note This function needs to respect the system's endianness, because
           bytes in CBOR, MessagePack, and UBJSON are stored in network order
           (big endian) and therefore need reordering on little endian systems.
     */
@@ -2514,8 +2514,8 @@ class binary_reader
     /// the number of characters read
     std::size_t chars_read = 0;
 
-    /// whether we can assume little endianess
-    const bool is_little_endian = little_endianess();
+    /// whether we can assume little endianness
+    const bool is_little_endian = little_endianness();
 
     /// the SAX parser
     json_sax_t* sax = nullptr;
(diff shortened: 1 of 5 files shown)


include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -2514,7 +2514,7 @@ class binary_reader
     /// the number of characters read
     std::size_t chars_read = 0;
 
-    /// whether we can assume little endianess
+    /// whether we can assume little endianness
```
axic, line 2517: Now this of course is a controversial change, given there is a helper named `little_endianess`. It seems the [more common spelling is with two n-s](https://en.wikipedia.org/wiki/Endianness).

nlohmann:
The CI failed: You need to edit the individual files in the `include` directory and then execute `make amalgamate` to create the single header.

include/nlohmann/detail/input/binary_reader.hpp
nlohmann: I agree: it should be `endianness`. Feel free to also rename the helper.

nlohmann (changes requested):
Please edit the files in `include` and call `make amalgamate` to create the single header.

include/nlohmann/detail/input/binary_reader.hpp
axic: Do you think any user of the library could depend on this? If so this may be a breaking change.

axic:
> Please edit the files in include and call make amalgamate to create the single header.

I though I ran it, but apparently didn't. Fixed now.

include/nlohmann/detail/input/binary_reader.hpp
nlohmann: No, that function is in the `detail` namespace that nobody should depend on.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1625,7 +1625,7 @@ class binary_writer
     }
 
   private:
-    /// whether we can assume little endianess
+    /// whether we can assume little endianness
     const bool is_little_endian = little_endianess();
```
nlohmann, line 1629: Please rename this function and its calls to `little_endianness`.

nlohmann (changes requested):

include/nlohmann/detail/output/binary_writer.hpp
axic: Done!

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks a lot! I will merge once the CI is done.

PR #3145, opened 2021-11-17
Docs: Update `skip_comments` to `ignore_comments`

daniel-kun:
In the docs, the `ignore_comments` parameter to `nlohmann::json::parse()` is described as `skip_comments`, which does not match the actual implementation. While the demo code compiles and works, the parameter name is wrong.


nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #3110, opened 2021-10-30
Allow custom base class as node customization point

barcode:
This PR adds an additional template parameter which allows to set a custom base class for `nlohmann::json`. This class serves as an extension point and allows to add functionality to json node. Examples for such functionality might be metadata or additional member functions (e.g. visitors) or other application specific code.
By default the parameter is set to void and an empty base class is used. In this case the library behaves as it already did.

* * *

## Pull request checklist

Read the [Contribution Guidelines](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md) for detailed information.

- [x]  Changes are described in the pull request, or an [existing issue is referenced](https://github.com/nlohmann/json/issues).
- [x]  The test suite [compiles and runs](https://github.com/nlohmann/json/blob/develop/README.md#execute-unit-tests) without error.
- [x]  [Code coverage](https://coveralls.io/github/nlohmann/json) is 100%. Test cases can be added by editing the [test suite](https://github.com/nlohmann/json/tree/develop/test/src). (should be. will update state aver coverall is done)
- [x]  The source code is amalgamated; that is, after making changes to the sources in the `include/nlohmann` directory, run `make amalgamate` to create the single-header file `single_include/nlohmann/json.hpp`. The whole process is described [here](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md#files-to-change).

PR #3165, opened 2021-11-28
Additional information for sax parser via next token start stop

barcode:
This PR requires #3110.

This PR enables a sax parser to get additional information about line, column and byte position of a token from a parsed input (e.g. file). This can be used to give feedback to a user.
To use this feature a sax parser has to implement `void next_token_start(std::size_t pos)` and `void next_token_end(std::size_t pos)`. These functions are supported for `json` and binary formats (`cbor`, `bson`, `msgpack`, `ubjson`).
If the input is `json`, the sax parser can implement `template<class T1, class T2> void next_token_start(const nlohmann::detail::lexer<T1, T2>& lex)` and  `template<class T1, class T2> void next_token_end(const nlohmann::detail::lexer<T1, T2>& lex)` to get access to more detailed information (such as line and column).

For examples see the tests `unit-sax-parser-extended.cpp` and `unit-sax-parser-store-source-location.cpp`. The former is a general test for these functions, while the latter also serves as an example for a sax parser storing token start and end information in `json` nodes.

This PR is currently WIP and was only added at this time to 
1. document someone is working on this, 
2. enable people interested to give feedback or try it out and
3. run the CI to detect issues.

* * *

## Pull request checklist

**Will be filled out when tin PR is ready to merge.**


barcode:
I will add some documentation, but otherwise my work on this PR is finished.

gregmarr:
@barcode Are you able to add the documentation to finish up this PR?

raphael-grimm:
This PR is not dead, it was just sleeping very deeply for a very long time... I finally have some time to work on this again. 
(I am the same guy as barcode. Commits / Messages are now from my second account since it is on company time.)
I will update the code and write some documentation. After I am done, I will remove the draft status.

raphael-grimm:
I updated the code and added some documentation. 

Currently there is one thing I am unhappy about:
When parsing json data it would be nice to have more information than just the current byte (e.g., line + col).
Currently I achieve this by passing the lexer to the methods (in case they support this). Then the lexer can be used to get additional information about the current parse.
Primarily `lex.get_position()` would be of interest, but this way it is possible to add some additional introspection methods to the lexer and have them directly available in the SAX parser.

This comes with additional complexity in the user interface (compared with directly passing `nlohmann::detail::position_t` ). I don’t think this kind of future proofing can justify this additional complexity.
I think it would be better to directly pass `nlohmann::detail::position_t`. This makes the user interface and documentation easier to understand.
Also we should move `nlohmann::detail::position_t` to `nlohmann::position_t` to prevent leaking the `nlohmann::detail` namespace.

Before changing to `nlohmann::detail::position_t`, I wanted to ask what your (@gregmarr / @nlohmann)  opinion is on this matter. Is passing the lexer worth the additional complexity (maybe there is some issue where this would be useful) or do you see any issue with me moving `nlohmann::detail::position_t` to `nlohmann::position_t`?

barcode:
I switched to using `nlohmann::position_t` and added some more documentation.

barcode:
So either clang-tydy tells me:
https://github.com/nlohmann/json/actions/runs/3771999964/jobs/6412687531

    /__w/json/json/tests/src/unit-sax-parser-store-source-location.cpp:87:11: error: member initializer for 'object_element' is redundant [modernize-use-default-member-init,-warnings-as-errors]
            , object_element{nullptr}

or codacy complains about:
https://github.com/nlohmann/json/pull/3165/checks?check_run_id=10288084598

     tests/src/unit-sax-parser-store-source-location.cpp#L84
    Member variable 'sax_with_token_start_stop_metadata::object_element' is not initialized in the constructor.

Which one should i suppress?

/edit

I ended up suppressing clang-tidy, since i did not find a clean way of suppressing codacy.

barcode:

PR #3181, opened 2021-12-07
Separate number-to/from-string logic into their own classes

KeXu001:
Hello. This is a feature request:

**Changes**
The code that performs conversions between numbers and strings was moved into separate classes.

Originally, the code was inside the `lexer` and `serializer` classes. Now, the code lives inside new classes `numerizer` and `denumerizer` which are passed as template args to the former.

`numerizer` converts strings to numbers (strtof, strtoull, ...)
`denumerizer` converts numbers to strings (dump_integer, dump_float, ...)

**Reasoning**
- this enables customization of the number parsing/dumping behavior
- this enables custom numeric types to be used

To do this, users can replace or inherit from `numerizer` / `denumerizer`. After implementing the static functions to their liking, they pass their custom classes as template args

**Notes**
- In order to create an 'interface' for the classes, template arguments were added to `serializer` and `lexer`.
- Furthermore, two template arguments were added to `basic_json`: `LexerType` and `SerializerType`. (Should we create a config template class to avoid too many template args?)
- `number_buffer` and locale information is allocated on-demand rather than inside an object, which might impact performance

* * *

## Pull request checklist

### DIFF
--- a/include/nlohmann/detail/macro_scope.hpp
+++ b/include/nlohmann/detail/macro_scope.hpp
@@ -132,12 +132,13 @@
              class NumberUnsignedType, class NumberFloatType,              \
              template<typename> class AllocatorType,                       \
              template<typename, typename = void> class JSONSerializer,     \
-             class BinaryType>
+             class BinaryType,                                             \
+             template<typename, typename> class LexerType>
 
 #define NLOHMANN_BASIC_JSON_TPL                                            \
     basic_json<ObjectType, ArrayType, StringType, BooleanType,             \
     NumberIntegerType, NumberUnsignedType, NumberFloatType,                \
-    AllocatorType, JSONSerializer, BinaryType>
+    AllocatorType, JSONSerializer, BinaryType, LexerType>
 
 // Macros to simplify conversion from/to types
 
--- a/include/nlohmann/detail/macro_scope.hpp
+++ b/include/nlohmann/detail/macro_scope.hpp
@@ -133,12 +133,13 @@
              template<typename> class AllocatorType,                       \
              template<typename, typename = void> class JSONSerializer,     \
              class BinaryType,                                             \
-             template<typename, typename> class LexerType>
+             template<typename, typename> class LexerType,                 \
+             template<typename> class SerializerType>
 
 #define NLOHMANN_BASIC_JSON_TPL                                            \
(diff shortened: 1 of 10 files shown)


KeXu001:
**Additional note 1:**
While working on these changes, I was seeing an error from cppcheck:
```
Checking /__w/json/json/single_include/nlohmann/json.hpp ...
/__w/json/json/single_include/nlohmann/json.hpp:3471:54: error: syntax error [syntaxError]
         class BinaryType = std::vector<std::uint8_t>,
                                                     ^
```
I think this is a false alarm. As a temporary workaround I added an alias in json_fwd.hpp. Is there a better option?
```
using vector_of_uint8_t = std::vector<std::uint8_t>;
...
         class BinaryType = vector_of_uint8_t,
```

**Additional note 2:**
Should test cases be added along with this pull request?

KeXu001:
For an example of intended use please see [#4](https://github.com/KeXu001/json/pull/4) (default constructible custom floating point type).

Further changes would be necessary to support custom numeric types in general, including arbitrary precision/size integers and floating points.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -132,12 +132,14 @@
              class NumberUnsignedType, class NumberFloatType,              \
              template<typename> class AllocatorType,                       \
              template<typename, typename = void> class JSONSerializer,     \
-             class BinaryType>
+             class BinaryType,                                             \
+             template<typename, typename> class LexerType,                 \
```
nlohmann, line 136: I am not convinced that it is a good idea to template the type of the lexer and serializer. These are aspects of the parser/serializer, but not of the JSON type.

KeXu001: I think I understand, but let me know. Would it make more sense to template the class(es) directly (`JSONNumerizer`)? Or even put it all into `JSONSerializer`?

nlohmann: What I meant: The type of the lexer/serializer should not be a parameter of the JSON type. The JSON type should be defined by how values are stored. In the long term, a better library design could have the lexer/parser and serializer as free functions working on the public API of the JSON type. Then it would be easy to change these implementations without touching the JSON type in the first place.

Issue #3156, opened 2021-11-23
std::filesystem unavailable on macOS lower deployment targets


kevin--:
#### Steps to Reproduce
Building with C++17, and nlohmann v3.10.4, targeting macOS Deployment Target of 10.14

```
/Users/distiller/project/app/lib/nlohmann/single_include/nlohmann/json.hpp:4384:57: error: 'path' is unavailable: introduced in macOS 10.15
void from_json(const BasicJsonType& j, std::filesystem::path& p)
                                                        ^
```
etc,

Basically since C++ library features are implemented in macOS and not via a redistributable system, Xcode refuses to compile if it knows a feature will be unavailable. For example, `std::variant` does not work until 10.12

It is a product requirement for our minimum deployment target to be set at 10.14, but we are building with C++17 regardless and just need to avoid using features we can't rely on being present at runtime.

#### What is the expected behavior?

Use of `std::filesystem` is opt-in rather than automatically used if C++17 is available.

#### And what is the actual behavior instead?

I guess we would have to lower our compile standard to C++14, however since nlohmann is header only we can't lower our standard for the entire project.

#### Which compiler and operating system are you using?

MacOS 10.15.7
Xcode 12.4

Apple clang version 12.0.0 (clang-1200.0.32.29)
Target: x86_64-apple-darwin19.6.0
 
#### Which version of the library did you use?

- [x] latest release version 3.10.4
- [ ] other release - please state the version: ___
- [ ] the `develop` branch

#### If you experience a compilation error: can you [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests)?

- [x] yes
- [ ] no - please copy/paste the error message below

kevin--:
Can you point us to an older release that does not have any std::filesystem facilities? We can use that for now

nlohmann:
Duplicate of #3090. You can try 3.10.3 until the issue is fixed.

kevin--:
I would say that it is not an exact duplicate @nlohmann because if we change the macOS Deployment Target, the same toolchain can compile successfully (no use of experimental headers etc)

You may not be able to do a simple toolchain check to determine the behavior in this case.

nlohmann:
How would such a check look like?

kevin--:
in CMake, we use the variable `CMAKE_OSX_DEPLOYMENT_TARGET` ( see https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html )

[...]

Issue #3090, opened 2021-10-18
GCC 7.5.0 with --std=c++17: filesystem: No such file or directory


miikka75:
#### What is the issue you have?

Compilation fails using GCC 7.5.0 (default GCC compiler version in ubuntu 18.04) for a target which has --std=c++17 compiler option.

#### Please describe the steps to reproduce the issue.

GCC 7.5.0 declares: __cplusplus = 201703L, thus JSON_HAS_CPP_17 flag is set but in fact GCCv7 implements <experimental/filesystem>
https://stackoverflow.com/questions/45867379/why-does-gcc-not-seem-to-have-the-filesystem-standard-library

#### Can you provide a small but working code example?

See diff (to test project's CMakeLists.txt) at the end of bug report.

#### What is the expected behavior?

Compilation of json.hpp should work with GCCv7 with --std=c++17 compiler option.

#### And what is the actual behavior instead?

Compilation using GCCv7 outputs errors:
```
json-src/single_include/nlohmann/json.hpp:3954:14: fatal error: filesystem: No such file or directory
      #include <filesystem>

json-src/single_include/nlohmann/json.hpp:4384:45: error: ‘std::filesystem’ has not been declared
  void from_json(const BasicJsonType& j, std::filesystem::path& p)
```
#### Which compiler and operating system are you using?

- Compiler: GCC v7.5.0
- Operating system: Ubuntu 18.04

#### Which version of the library did you use?

- [x] latest release version 3.10.4
- [ ] other release - please state the version: ___
- [ ] the `develop` branch

#### If you experience a compilation error: can you [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests)?

- [x] yes
- [ ] no - please copy/paste the error message below

Compilation fails, if I make following change:
```diff
diff --git a/test/CMakeLists.txt b/test/CMakeLists.txt
index 447192c..c63ef61 100644
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -60,6 +60,7 @@ foreach(file ${files})
 
     add_executable(${testcase} $<TARGET_OBJECTS:doctest_main> ${file})
     target_compile_definitions(${testcase} PRIVATE DOCTEST_CONFIG_SUPER_FAST_ASSERTS)
+    set_property(TARGET ${testcase} PROPERTY CXX_STANDARD 17)
     target_compile_options(${testcase} PRIVATE
[...]

PR #3194, opened 2021-12-14
Allow disabling std::filesystem support

vadz:
This can be useful when the compiler claims C++17 support, but standard
library std::filesystem implementation is unavailable (MinGW) or can't
be used because the current macOS target is too low to allow it (#3156).

See #3090.

---

I think it could be useful to disable this by default when macOS min version target is lower than 10.15 and will add a commit doing this if there are no objections.


nlohmann:
The filesystem issue is fixed here: #3101 - it also implements macros `JSON_HAS_FILESYSTEM` and `JSON_HAS_EXPERIMENTAL_FILESYSTEM` to switch off filesystem support altogether. That PR currently has a failing PR, but it solves the issue completely. Hence, I will probably not merge this PR.

vadz:
Sorry for wasting your time, I've somehow failed to find #3101 which is, of course, much more complete and better than this version, thanks for pointing me to it!

gregmarr:
@vadz Thanks anyway for the contribution.

Issue #3215, opened 2021-12-26
Exception thrown during initialization causes a memory leak


namark:
Address sanitizer reports a memory leak after an exception is thrown in the initializer list of the json object and caught afterwards.

#### What is the issue you have?
Memory leak.

I did not diagnose the issue thoroughly, so I'm not sure about the details, but I consider it a bug because an equivalent standard data structure (std::map) does not leak memory in the same scenario. 

Compiler invocation and address sanitizer output:
```
g++ -g -fsanitize=address  -lasan  es.cpp   -o es

=================================================================
==3498808==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x7fc1a0a33ca1 in operator new(unsigned long) /build/gcc/src/gcc/libsanitizer/asan/asan_new_delete.cpp:99
    #1 0x55c6bbbb15e9 in __gnu_cxx::new_allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >::allocate(unsigned long, void const*) /usr/include/c++/11.1.0/ext/new_allocator.h:121
    #2 0x55c6bbbae461 in std::allocator_traits<std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >::allocate(std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >&, unsigned long) /usr/include/c++/11.1.0/bits/alloc_traits.h:460
    #3 0x55c6bbbb50db in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >* nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > >::create<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, char const (&) [5]>(char const (&) [5]) single_include/nlohmann/json.hpp:18478
[...]

PR #3222, opened 2021-12-28
Investigate memory leak

nlohmann:
This PR addresses #3215.


test/src/unit-regression2.cpp
```diff
@@ -835,6 +849,27 @@ TEST_CASE("regression tests 2")
...
         CHECK(j.dump() == "[1,4]");
     }
+
+    SECTION("issue #3215 - Exception thrown during initialization causes a memory leak")
+    {
+        std::string s;
+        try
+        {
+            // std::map<std::string, std::string> j // no leak
+            nlohmann::json j // leak
```
gregmarr, line 859: This looks like the error will be during the construction of the `initializer_list` and not during the `json` constructor.   Does it reproduce with `nlohmann::json::initializer_list_t j`?

nlohmann: I try: https://github.com/nlohmann/json/pull/3222/commits/66a2d18df6f70beabbd4924d12bf3d90c399635a

nlohmann: Yes, the leak remains: https://github.com/nlohmann/json/runs/4694680374?check_suite_focus=true

gregmarr: t looks to me like the compiler is constructing some of the initializer list data, then hitting the exception and not cleaning up what it already constructed.  I can't prove it at this point, but I don't see anything yet that says that the library is doing anything wrong.  I guess the next check would be `initializer_list<nlohmann::json>` to remove `json_ref` from the equation.

nlohmann: I'll try. Would be strange if this was a compiler bug.

gregmarr: Yes, strange, but it's a bit of an edge case.

gregmarr: It fails for 

    std::initializer_list<nlohmann::json>

but passes for

    std::initializer_list<std::pair<std::string, std::string>>

Not sure yet what that means.  

https://godbolt.org/z/hef9Yf8P5

gregmarr: Fails the same way for this too:

        nlohmann::json j{
            "smth", f_3215(true) // exception thrown here
        };

gregmarr:
This is a compiler bug in GCC and is fixed for GCC 12 (can be seen in trunk).  https://godbolt.org/z/6Tfshn39W

Issue #2676, opened 2021-03-16
Compiler warnings with NVCC 11.2


allisonvacanti:
#### What is the issue you have?

Instantiating the `json::basic_json` template and compiling with the CUDA NVCC compiler emits the warnings below.

#### Please describe the steps to reproduce the issue.

Compile the following with nvcc (observed on 11.2 with MSVC 2019 host):

```
#include <nlohmann/json.hpp>
int main()
{
  nlohmann::ordered_json json = {"Test"};
  json.dump();
}
```

The following warnings are emitted:

```
_deps/nlohmann_json-src/include\nlohmann/detail/hash.hpp(114): warning: missing return statement at end of non-void function "nlohmann::detail::hash(const BasicJsonType &) [with BasicJsonType=nlohmann::json]"
          detected during instantiation of "size_t nlohmann::detail::hash(const BasicJsonType &) [with BasicJsonType=nlohmann::json]" 
_deps/nlohmann_json-src/include\nlohmann/json.hpp(8709): here

_deps/nlohmann_json-src/include\nlohmann/detail/output/serializer.hpp(704): warning: pointless comparison of unsigned integer with zero
          detected during:
            instantiation of "void nlohmann::detail::serializer<BasicJsonType>::dump_integer(NumberType) [with BasicJsonType=nlohmann::basic_json<nlohmann::ordered_map, std::vector, std::string, __nv_bool, int64_t, uint64_t, double, std::allocator, nlohmann::adl_serializer, std::vector<uint8_t, std::allocator<uint8_t>>>, NumberType=uint8_t, <unnamed>=0]" 
(266): here
            instantiation of "void nlohmann::detail::serializer<BasicJsonType>::dump(const BasicJsonType &, __nv_bool, __nv_bool, unsigned int, unsigned int) [with BasicJsonType=nlohmann::basic_json<nlohmann::ordered_map, std::vector, std::string, __nv_bool, int64_t, uint64_t, double, std::allocator, nlohmann::adl_serializer, std::vector<uint8_t, std::allocator<uint8_t>>>]" 
_deps/nlohmann_json-src/include\nlohmann/json.hpp(2281): here
[...]

PR #3227, opened 2021-12-29
Add build step for NVCC and fix a warning

nlohmann:
This PR adds a CI step for NVCC and fixes a warning reported in #2676.


ax3l (approved):
Looks great to me, thanks a lot! :tada:

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -664,6 +664,19 @@ class serializer
         }
     }
 
+    // templates to avoid warnings about useless casts
+    template <typename NumberType, enable_if_t<std::is_signed<NumberType>::value, int> = 0>
+    bool is_negative_number(NumberType x)
```
lordnn, line 669:
```
template <typename NumberType>
enable_if_t<std::is_signed<NumberType>::value, bool> is_negative_number(NumberType x)
```

nlohmann: ?

lordnn: Try this. You use legal but verbose and very indirect way for `enable_if_t` utilize.

gregmarr: That is a normal way to do it for functions that are already templates.   https://en.cppreference.com/w/cpp/types/enable_if

lordnn: "legal but verbose", Ok
And how about this?
https://github.com/nlohmann/json/blob/700b95f4473cac1d900b807541daad0ca3d98f52/include/nlohmann/detail/conversions/from_json.hpp#L421-L423

gregmarr: I'm not sure why @theodelrieu used the longer form here compared to the rest of the file, perhaps there was a technical reason.  If not, it would make sense to use the value form as in the rest of the file. https://github.com/nlohmann/json/commit/c5e63fd68430469f3734b0468b1db2e9d4fc5674

PR #3228, opened 2021-12-29
Doctest 2.4.8

nlohmann:
This PR updates Doctest to the latest version 2.4.8.


nlohmann:
Test fail for MSVC 2015. See https://github.com/doctest/doctest/issues/573 for bug report.

nlohmann:
The bug was resolved in Doctest 2.4.8.

gregmarr:
That's a lot of clang tidy suppressing.  Is there a way to specify those just for the test directory?

nlohmann:
> That's a lot of clang tidy suppressing. Is there a way to specify those just for the test directory?

I have no idea. I am only aware of suppressing warnings in the code via `// NOLINT` comments.

I currently just want to make it pass the CI. I will create an issue upstream to inform the Doctest maintainers that their latest release triggers these warnings.

gregmarr:
I wonder if it would make sense to exclude the tests from clang tidy.  https://github.com/nlohmann/json/blob/1aca6cb949f4595994ba5e5188682416d3058688/cmake/ci.cmake#L635

nlohmann:
It would be great to only exclude file `doctest.h` from the results. I have not found a way to do this, and I also have not found a way to check the header without adding Clang-Tidy to the compilation process of the test suite.

nlohmann:
Blocked until https://github.com/doctest/doctest/issues/607 is addressed.

nlohmann:
We'll wait until Doctest 2.4.9 (see https://github.com/doctest/doctest/issues/607#issuecomment-1146950377).

PR #3229, opened 2021-12-29
Add build step for ICPC

nlohmann:
This PR adds a CI build step for the Intel C++ Compiler


nlohmann:
ICPC fails to build the test suite.

- https://github.com/nlohmann/json/runs/4665922491?check_suite_focus=true failed with

```
/__w/json/json/test/src/unit-regression2.cpp(810): error: more than one instance of constructor "std::filesystem::__cxx11::path::path" matches the argument list:
            function "std::filesystem::__cxx11::path::path(std::filesystem::__cxx11::path &&) noexcept"
            function "std::filesystem::__cxx11::path::path(std::filesystem::__cxx11::path::string_type &&, std::filesystem::__cxx11::path::format)"
            argument types are: (json)
          CHECK_THROWS_WITH_AS(nlohmann::detail::std_fs::path(json(1)), "[json.exception.type_error.302] type must be string, but is number", json::type_error);
          ^
/usr/include/c++/9/bits/fs_path.h(473): note: this candidate was rejected because mismatch in count of arguments
      path(basic_string_view<value_type> __str, _Type __type)
      ^
/usr/include/c++/9/bits/fs_path.h(213): note: this candidate was rejected because mismatch in count of arguments
        path(_InputIterator __first, _InputIterator __last, const locale& __loc,
        ^
/usr/include/c++/9/bits/fs_path.h(205): note: this candidate was rejected because mismatch in count of arguments
        path(_Source const& __source, const locale& __loc, format = auto_format)
        ^
/usr/include/c++/9/bits/fs_path.h(198): note: this candidate was rejected because mismatch in count of arguments
        path(_InputIterator __first, _InputIterator __last, format = auto_format)
        ^
/usr/include/c++/9/bits/fs_path.h(191): note: this candidate was rejected because at least one template argument could not be deduced
        path(_Source const& __source, format = auto_format)
        ^
```

- Switching off `std::filesystem` tests with `-DCMAKE_CXX_FLAGS="-DJSON_HAS_FILESYSTEM=0 -DJSON_HAS_EXPERIMENTAL_FILESYSTEM=0"` fails (https://github.com/nlohmann/json/runs/4666752017?check_suite_focus=true) with

```
In file included from /__w/json/json/test/src/unit-algorithms.cpp(32):
/__w/json/json/single_include/nlohmann/json.hpp(3821): error: this operator is not allowed in a constant expression
  #elif JSON_HAS_FILESYSTEM
        ^

In file included from /__w/json/json/test/src/unit-algorithms.cpp(32):
/__w/json/json/single_include/nlohmann/json.hpp(4254): error: this operator is not allowed in a constant expression
  #if JSON_HAS_FILESYSTEM || JSON_HAS_EXPERIMENTAL_FILESYSTEM
      ^

In file included from /__w/json/json/test/src/unit-algorithms.cpp(32):
/__w/json/json/single_include/nlohmann/json.hpp(4507): error: this operator is not allowed in a constant expression
  #elif JSON_HAS_FILESYSTEM
        ^

In file included from /__w/json/json/test/src/unit-algorithms.cpp(32):
/__w/json/json/single_include/nlohmann/json.hpp(4887): error: this operator is not allowed in a constant expression
  #if JSON_HAS_FILESYSTEM || JSON_HAS_EXPERIMENTAL_FILESYSTEM
      ^
```

falbrechtskirchinger:
Have a look at #3377. I had to disable the `CHECK_THROWS_WITH_AS` in #3379.

nlohmann:
> Have a look at #3377. I had to disable the `CHECK_THROWS_WITH_AS` in #3379.

Thanks for pointing this out. I am now confused as well. If I would just comment out the test, I'm afraid that I would ignore something not working...

falbrechtskirchinger:
Adding `$<$<CXX_COMPILER_ID:Intel>:-diag-disable=2196>` or `$<$<CXX_COMPILER_ID:Intel>:-wd2196>` (deprecated) should get rid of
```
warning #2196: routine is both "inline" and "noinline"
```
if that is still an issue with current doctest.

.github/workflows/ubuntu.yml
```diff
@@ -112,3 +112,15 @@ jobs:
...
+
+  ci_icpc:
+    runs-on: ubuntu-latest
+    container: ghcr.io/nlohmann/json-ci:v2.2.0
+    steps:
+      - uses: actions/checkout@v2
+      - name: cmake
+        run: cmake -S . -B build -DJSON_CI=On
+      - name: build
+        run: >
```
falbrechtskirchinger, line 124: I'd suggest using the literal style block scalar `|` over the flow style one `>` for multiline commands:
```yaml
        run: |
          . /opt/intel/oneapi/setvars.sh
          cmake --build build --target ci_icpc
```

tests/CMakeLists.txt
```diff
@@ -49,6 +49,9 @@ target_compile_options(test_main PUBLIC
     # https://github.com/nlohmann/json/issues/1114
     $<$<CXX_COMPILER_ID:MSVC>:/bigobj> $<$<BOOL:${MINGW}>:-Wa,-mbig-obj>
 
+    # https://github.com/nlohmann/json/pull/3229
+    $<$<CXX_COMPILER_ID:Intel>:-diag-disable=2196>
```
falbrechtskirchinger, line 53: This may need a version check. It's `-wd2196` for older versions but I'm not familiar with ICPC, so I don't know since when the new flag is available. :man_shrugging:

cmake/ci.cmake
```diff
@@ -877,6 +877,22 @@ add_custom_target(ci_cuda_example
...
 
+###############################################################################
+# Intel C++ Compiler
+###############################################################################
+
+add_custom_target(ci_icpc
+    COMMAND ${CMAKE_COMMAND}
+        -DCMAKE_BUILD_TYPE=Debug -GNinja
+        -DCMAKE_C_COMPILER=icc -DCMAKE_CXX_COMPILER=icpc
+        -DCMAKE_CXX_FLAGS="-DJSON_HAS_FILESYSTEM=0 -DJSON_HAS_EXPERIMENTAL_FILESYSTEM=0"
```
falbrechtskirchinger, line 888: Appears as `-DJSON_HAS_FILESYSTEM=0\ -DJSON_HAS_EXPERIMENTAL_FILESYSTEM=0` on the commandline and results in:
```
/__w/json/json/single_include/nlohmann/json.hpp(4116): error: this operator is not allowed in a constant expression
  #elif JSON_HAS_FILESYSTEM
```

Remove? Don't know why `<filesystem>` detection was an issue on ICPC, but might have been fixed by now?

tests/CMakeLists.txt
falbrechtskirchinger: Also 1786 to suppress deprecation warnings. We're not enabling warnings as errors (should we?), so it'd be just for cleaner output.

falbrechtskirchinger:
Intel needs to be excluded here as well. :-(

```cpp
#if JSON_HAS_FILESYSTEM || JSON_HAS_EXPERIMENTAL_FILESYSTEM
    // JSON_HAS_CPP_17 (do not remove; see note at top of file)
    SECTION("issue #3070 - Version 3.10.3 breaks backward-compatibility with 3.10.2 ")
    {
        nlohmann::detail::std_fs::path text_path("/tmp/text.txt");
        json j(text_path);

        CHECK(j_path == text_path);

#if !defined(_MSC_VER) && !(defined(__GNUC__) && __GNUC__ == 8 && __GNUC_MINOR__ < 4)
        // works everywhere but on MSVC and GCC <8.4
        CHECK_THROWS_WITH_AS(nlohmann::detail::std_fs::path(json(1)), "[json.exception.type_error.302] type must be string, but is number", json::type_error);
#endif
    }
#endif
```

Maybe this condition should just be changed to:
```cpp
#if defined(__clang__) || JSON_HEDLEY_GNUC_VERSION JSON_HEDLEY_GNUC_VERSION_CHECK(8, 4, 0)
        // only known to work on Clang and GCC >=8.4
        CHECK_THROWS_WITH_AS(nlohmann::detail::std_fs::path(json(1)), "[json.exception.type_error.302] type must be string, but is number", json::type_error);
#endif
```

tests/CMakeLists.txt
falbrechtskirchinger: I couldn't find a version number in the documentation.

falbrechtskirchinger:
Another issue:

```cpp
    SECTION("issue #1647 - compile error when deserializing enum if both non-default from_json and non-member operator== exists for other type")
    {
#if !(defined(JSON_HAS_CPP_20) && defined(__INTEL_COMPILER))
        {
            json j;
            NonDefaultFromJsonStruct x(j);
            NonDefaultFromJsonStruct y;
            CHECK(x == y);
        }
#endif

        auto val = nlohmann::json("one").get<for_1647>();
        CHECK(val == for_1647::one);
        json j = val;
    }
```

Test fails to compile on ICPC when targeting C++20.

```
/__w/json/json/tests/src/unit-regression2.cpp(517): error: too many initializer values
              NonDefaultFromJsonStruct x(j);
```

These lines compile fine:
```
    auto x2 = j.get<NonDefaultFromJsonStruct>();
    auto x3 = j.operator NonDefaultFromJsonStruct();
```

falbrechtskirchinger:
@nlohmann I'm still waiting for CI, so I don't know if anything else is needed. In any event, there are now a quite a few changes. You can either reference my branch [here](https://github.com/falbrechtskirchinger/json/tree/icpc) or I can open a PR, whichever you prefer.

nlohmann:
> @nlohmann I'm still waiting for CI, so I don't know if anything else is needed. In any event, there are now a quite a few changes. You can either reference my branch [here](https://github.com/falbrechtskirchinger/json/tree/icpc) or I can open a PR, whichever you prefer.

Please open a PR. You're already much deeper in this topic.

falbrechtskirchinger:
Okay. Of course CI just failed so there's at least one more issue to solve...

Issue #1529, opened 2019-03-20
How can I use std::string_view as the json_key to "operator []" ?


jencoldeng:
//as follow, this will fail as the key_type pof map is std::string_view
std::map<string_view, bool> mv{{"111",true}, {"222", false}, {"333", true}};
nlohmann::json j4{mv};

//another: fail again, the std::string_view is the json_key
nlohmann::json j5;
std::string_view key = "my_key"sv;
 j5[key] = 100;

nlohmann:
The library does not support string views there. Object keys are `std::string`, so even if you could pass a string view, it would be first converted to a `std::string` for the lookup.

jencoldeng:
but I can use string_view in map<std::string, bool>, like this:

std::map<std::string, bool, std::less<>> mp{{"111",true}, {"222", false}, {"333", true}}; //use std::less<>
std::string_view k = "111"sv;
const auto& it = mp.find(k);
cout<<it->second<<endl;


as your json_object is a map type, why can't you do it similarly ?

jencoldeng:
Maybe some part of your code I have not understand, : )

nlohmann:
Maybe under the hood, `std::map` is ready to do the job, but the interfaces of the library mess it up. I need to check. Unfortunately, the library targets C++11, so I cannot do too much for `std::string_view` being a C++17 feature...

jencoldeng:
What a pity  : )

nlohmann:
Transparent comparators seem to be a C++14 extension. :/

nlohmann:
Any idea how to proceed?

jencoldeng:
en....  May be this code is ok:

`
        //by jencol
    template<typename keyT>
    reference operator[](keyT&& key)
    {
        // implicitly convert null to object
        if (is_null())
        {
            m_type = value_t::object;
            m_value = value_t::object;
            assert_invariant();
        }

        // at only works for objects
        if (JSON_LIKELY(is_object()))
        {
                        auto it = m_value.object->find(key);
                        if(it == m_value.object->end())
                        {
                                typename std::remove_reference<decltype(*(m_value.object))>::type::key_type _key{key};
                                return (*(m_value.object))[_key];
                        }
                        else
                        {
                                return it->second;
                        }
        }

        JSON_THROW(type_error::create(305, "cannot use operator[] with a string argument with " + std::string(type_name())));
    }
`

jencoldeng:
See my code, I add a `operator[]`:

`
[...]

PR #3237, opened 2022-01-02
String view2

nlohmann:
This PR tries to get a green CI for the original string_view PR.


nlohmann:
See https://github.com/nlohmann/json/issues/1529#issuecomment-1003733316

falbrechtskirchinger:
I might have a solution: 19c26cc
Similar overloads may be needed for other member functions like `at(json_pointer)`.

falbrechtskirchinger:
Wait … this is not the expected behavior at all, is it? Why is it in the regression test?
My "solution" is based on this comment from the unit test:
```
// mixed type - the JSON Pointer is implicitly converted into a string "/root"
```
I've thoroughly confused myself now. Help.

PR #3244, opened 2022-01-03
Remove stringstream

### LINKED ISSUE #3239: Replace use of standard IO from error handling
In
https://github.com/nlohmann/json/blame/develop/include/nlohmann/detail/output/serializer.hpp#L598-L599
and
https://github.com/nlohmann/json/blame/develop/include/nlohmann/detail/output/serializer.hpp#L504-L505
std::stringstream is used.

Use fmt instead. Instead of:

```cpp
std::stringstream ss;
ss << std::uppercase << std::setfill('0') << std::setw(2) << std::hex << (byte | 0);
JSON_THROW(type_error::create(316, "invalid UTF-8 byte at index " + std::to_string(i) + ": 0x" + ss.str(), BasicJsonType()));
```

use:

```cpp
JSON_THROW(type_error::create(316, fmt::format("invalid UTF-8 byte at index {}: {:#04x}", i, byte), BasicJsonType()));
```

nlohmann:
This PR replaces string streams by a direct implementation.

Fixes #3239.

### DIFF
--- a/include/nlohmann/detail/output/serializer.hpp
+++ b/include/nlohmann/detail/output/serializer.hpp
@@ -10,7 +10,6 @@
 #include <limits> // numeric_limits
 #include <string> // string, char_traits
 #include <iomanip> // setfill, setw
-#include <sstream> // stringstream
 #include <type_traits> // is_same
 #include <utility> // move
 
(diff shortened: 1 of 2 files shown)


include/nlohmann/detail/output/serializer.hpp
```diff
@@ -664,6 +659,20 @@ class serializer
         }
     }
 
+    /*!
+     * @brief convert a byte to a uppercase hex representation
+     * @param[in] byte byte to represent
+     * @return representation ("00".."FF")
+     */
+    inline std::string hex_bytes(std::uint8_t byte)
```
gregmarr, line 667: Minor, especially since this is already merged: this is inside the class definition, so `inline` is implicit.  Also, this doesn't access any data, and so can be `static`.

nlohmann: Thanks for noting. Fixed in 6d8d043adde8c95589f6e65e1fe228cb1f0d8419.

Issue #2907, opened 2021-08-01
pkgconfig integration wrongly rendered if tests are run


dvzrv:
When running the tests for the project after building it (with `-DCMAKE_INSTALL_PREFIX=/usr`), the pkgconfig file is rendered with `/usr/local/include` instead of `/usr/include`.

#### What is the issue you have?

When building and running the tests, the `CMAKE_INSTALL_PREFIX` is not honored when creating the pkgconfig integration.

#### Please describe the steps to reproduce the issue.

```
  cmake -DCMAKE_INSTALL_PREFIX='/usr' \
        -DCMAKE_INSTALL_LIBDIR='/usr/lib' \
        -DCMAKE_BUILD_TYPE='None' \
        -DBUILD_TESTING=ON \
        -DJSON_BuildTests=ON \
        -DJSON_MultipleHeaders=ON \
        -Wno-dev \
        -B "build" \
        -S .
  make VERBOSE=1 -C "build"
  make -k test -C "build"
  make DESTDIR="${pkgdir}" install -C build
```

The pkgconfig integration will contain `/usr/local/include`, but should include `/usr/include`.

When building without tests, `CMAKE_INSTALL_PREFIX` is honored and the pkgconfig file contains `/usr/include`:

```
  cmake -DCMAKE_INSTALL_PREFIX='/usr' \
        -DCMAKE_INSTALL_LIBDIR='/usr/lib' \
        -DCMAKE_BUILD_TYPE='None' \
        -DJSON_MultipleHeaders=ON \
        -Wno-dev \
        -B "build" \
        -S .
  make VERBOSE=1 -C "build"
  make DESTDIR="${pkgdir}" install -C build
```

#### Can you provide a small but working code example?

n/a

#### What is the expected behavior?

`CMAKE_INSTALL_PREFIX` is always honored. Tests never influence the installed files.

#### And what is the actual behavior instead?

`CMAKE_INSTALL_PREFIX` is not honored if tests are run. Running tests influences the installed files.

#### Which compiler and operating system are you using?

- Compiler: gcc 11.1.0
- Operating system: Arch Linux

#### Which version of the library did you use?

- [x] latest release version 3.9.1
- [ ] other release - please state the version: ___
- [ ] the `develop` branch

#### If you experience a compilation error: can you [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests)?

- [ ] yes
- [ ] no - please copy/paste the error message below

nlohmann:
[...]

PR #3248, opened 2022-01-05
pkgconfig: Fix path relocation for mingw

Biswa96:
Just separate the path to prefix and includedir variables.

* * *

## Pull request checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,7 @@ endif()
 CONFIGURE_FILE(
     "${CMAKE_CURRENT_SOURCE_DIR}/cmake/pkg-config.pc.in"
     "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}.pc"
+    @ONLY
 )
 
 ##
--- a/cmake/pkg-config.pc.in
+++ b/cmake/pkg-config.pc.in
@@ -1,4 +1,7 @@
-Name: ${PROJECT_NAME}
+prefix=@CMAKE_INSTALL_PREFIX@
+includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
+
+Name: @PROJECT_NAME@
 Description: JSON for Modern C++
-Version: ${PROJECT_VERSION}
-Cflags: -I${CMAKE_INSTALL_FULL_INCLUDEDIR}
+Version: @PROJECT_VERSION@
+Cflags: -I${includedir}


Biswa96:
@doronbehar Is this allowed according to the rules?

nlohmann:
Thanks for your contribution! Is this also related to #2907?

Biswa96:
I am not sure if this change is related to that issue. But I would not like to ruin my Arch. I can wait for their comments @dvzrv @eli-schwartz

This change is for msys2/mingw environment. For example, with the prefix variable, pkgconfig automatically changes/relocates `/mingw64/include` path to `C:\msys64\mingw64\include` (assuming C:\msys64 is the install directory of msys2).

nlohmann:
As I have no experience with pkgconfig, but see several issues on this: is there a way I can test this in the CI? We are using mingw in GitHub Actions.

Biswa96:
I am not sure what to test. But with this change, the output of a command will be like this:

```
$ pkg-config -cflags nlohmann_json
-IF:/msys64/ucrt64/include
```

Without this change, output is nothing.

nlohmann:
Do I need to install anything before this?

eli-schwartz:
This change is generally speaking correct. Although the `prefix` variable is not actually being used, because you set the `includedir` variable to cmake's "full" includedir anyway.

When configured via `-DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_INCLUDEDIR=include` it's possible (and preferred) to say in the pkg-config file:

```
prefix=/usr
includedir=${prefix}/include
```

But this will be wrong in the case of `-DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_INCLUDEDIR=/usr/include`:
```
prefix=/usr
includedir=${prefix}/usr/include
```

Cmake doesn't have a way of correctly handling this. Meson has `join_paths('${prefix}', get_option('includedir')` which correctly handles edge cases in path handling and won't add `${prefix}` if the includedir is already an absolute path. (But also, Meson's pkg-config.generate() function automatically generates a .pc file for you that handles this.)

This PR is to my knowledge the best way to handle the issue in cmake, and perfect is the enemy of the good, so... It looks good to me.

eli-schwartz:
It is NOT related to the other issue. The other issue is a problem where running the testsuite will (somehow) cause the nlohmann_json project options (specifically, the prefix and includedir) to be overwritten by the testsuite's project options.

Biswa96:
> Although the prefix variable is not actually being used

At first, I thought about this 

```
includedir=@CMAKE_INSTALL_PREFIX@/@CMAKE_INSTALL_INCLUDEDIR@
```

or this

```
includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
````

But in the previous commit, it was changed to `CMAKE_INSTALL_FULL_INCLUDEDIR` by @doronbehar. So, I assume full includedir is the _rule_.

**Comment by Biswa96** [comment]
> Although the prefix variable is not actually being used

At first, I thought about this 

```
includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
````

But in the previous commit, it was changed to `CMAKE_INSTALL_FULL_INCLUDEDIR` by @doronbehar. So, I assume full includedir is the _rule_.

eli-schwartz:
Right, and it links to https://github.com/jtojnar/cmake-snips#assuming-cmake_install_dir-is-relative-path which is really just another way of examining what I just mentioned.

In fact, right below that is https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files which offers a cmake module reimplementing meson's built-in `join_paths()` function. This can solve the issue fully, "the right way".

Biswa96:
OK, understood.

eli-schwartz:
Hmm? Can't you just update this PR to switch to the other method? :) Why did you close it?

Biswa96:
I am not sure if that method is perfect in every case.

eli-schwartz:
It should be, but either that or this PR in its current state would be useful improvements over the status quo.

PR #3375, opened 2022-03-04
[NFC] Provide exception overloads without context

falbrechtskirchinger:
Currently the static `create()` member functions of exceptions require a `const BasicJsonType &context` parameter. This PR adds overloads without a `context` parameter and replaces uses of `JSON_THROW(exception_type::create(..., BasicJsonType()))` and similar with the `context`-less version.

This removes one obstacle to freeing `json_pointer` from its class template parameter in the future.

### DIFF
--- a/include/nlohmann/detail/exceptions.hpp
+++ b/include/nlohmann/detail/exceptions.hpp
@@ -124,21 +124,26 @@ class parse_error : public exception
     @param[in] what_arg  the explanatory string
     @return parse_error object
     */
+    static parse_error create(int id_, const position_t& pos, const std::string& what_arg)
+    {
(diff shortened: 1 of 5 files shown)


gregmarr:
I'm not wild about the duplication, and this makes it easier to forget to see if there is an appropriate context parameter available.  What about changing it from a reference to a pointer so that the callers that don't have one can just pass `nullptr`?

falbrechtskirchinger:
Fair enough. I don't like the duplication either. This is a low-priority item and I'll get to it eventually.

falbrechtskirchinger:
This has been folded into #3415 with changes requested.

Issue #3377, opened 2022-03-05
Regression test for #3070 is not being run and fails when enabled


falbrechtskirchinger:
Working on some unit test improvements, I discovered at least two unit tests that are not being run. This issue is about
```cpp
#if JSON_HAS_FILESYSTEM || JSON_HAS_EXPERIMENTAL_FILESYSTEM
    SECTION("issue #3070 - Version 3.10.3 breaks backward-compatibility with 3.10.2 ")
    {
        static_assert(false); // Added to confirm this test is never run
        nlohmann::detail::std_fs::path text_path("/tmp/text.txt");
        json j(text_path);

        const auto j_path = j.get<nlohmann::detail::std_fs::path>();
        CHECK(j_path == text_path);

        CHECK_THROWS_WITH_AS(nlohmann::detail::std_fs::path(json(1)), "[json.exception.type_error.302] type must be string, but is number", json::type_error);
    }
#endif
```
added in 6d3115924c.

`nlohmann::detail::std_fs::path(json(1))` fails to compile, because the constructor call is ambiguous.

@theodelrieu @nlohmann Was/is this expected to work?

(The fact, that this test isn't being run will be fixed by an upcoming PR.)

falbrechtskirchinger:
Clarification: If I remember correctly the ambiguity arises from `std::fs::path` accepting both `std::string` and `std::fs::path` (copy ctor, obviously) and `basic_json<>` can implicitly convert to both. I don't see a way to resolve this ambiguity.

@nlohmann Thoughts?

nlohmann:
Thanks for clarifying! I'll think about this. Good to know that it never worked - this allows us to make changes without breaking existing implementations ;-)

nlohmann:
I am confused - locally, the test is executed and succeeds... (Apple clang version 13.0.0)

nlohmann:
So it seems the test is indeed not executed on Windows - maybe the filesystem detection is broken and all compilers are skipped. The test is executed on macOS and Linux for some compilers. However, not for all C++17 enabled compilers. 🤔

falbrechtskirchinger:
While I didn't dig into why it is now running in my builds (#3380). I consolidated the test macros into a separate header and copied them over from `include/nlohmann/detail/macro_scope.hpp`. The unit test embedded macros were either defined incorrectly or missing.

I'm also working on printing out build information with each test to better understand which test are compiled using which flags and standard version, etc.

nlohmann:
[...]

PR #3380, opened 2022-03-06
Improve unit testing (Part 1)

falbrechtskirchinger:
This PR refactors the CMake code for creating unit tests (see commit message cac4e9fb8324d5ddd6035d15e17efef50b5751e5).

It also removes duplicate macro definitions from unit tests and reuses the ones provided by `json.hpp` (see commit message 02138cfb72808f59fe838a7d78b97622edf7cfd3). To that end the macro `JSON_TEST_KEEP_MACROS` is introduced which keeps some macros from being undefined in `macro_unscope.hpp`.

A new CMake cache variable `JSON_TestStandards` is added to replace `-DCMAKE_CXX_STANDARD=<version> -DCMAKE_CXX_STANDARD_REQUIRED=ON`.

CMake 3.13 has been added to the `ci_cmake_flags` test.

Various compilation issues have been fixed. (These could be moved into a separate PR, if desired.)

As described in #3384, some tests had to be disabled. These failures are not _caused_ by this PR but _uncovered_, since it enables some previously untested configurations.

Documentation may need updating.


falbrechtskirchinger:
Reminder to myself to document that overrides need to be declared before adding the to-be-overridden tests.

falbrechtskirchinger:
Coverage decrease is due to this disabled check: #3377

cmake/testcase.cmake
```diff
@@ -0,0 +1,245 @@
...
+#     [COMPILE_DEFINITIONS <args>...]
+#     [COMPILE_FEATURES <args>...]
+#     [COMPILE_OPTIONS <args>...]
+#     [LINK_LIBRARIES <args>...]
+#     [LINK_OPTIONS <args>...])
+#
+# Override build settings of <testcases> for each standard in CXX_STANDARDS.
+# <testcases> may be a list (e.g., "test-foo;test-bar") to apply the
+# overrides to multiple tests.
+# If an empty string is used, overrides apply to all tests.
```
falbrechtskirchinger, line 104: Whoops. Forgot that "all" is translated into "". Should update doc.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -97,6 +97,20 @@
...
+    #endif
+#endif
+
+#ifdef JSON_HAS_THREE_WAY_COMPARISON
+    #undef JSON_HAS_THREE_WAY_COMPARISON
+    #define JSON_HAS_THREE_WAY_COMPARISON 1
+#else
+    #define JSON_HAS_THREE_WAY_COMPARISON 0
+#endif
+
```
falbrechtskirchinger, line 113: Missing in `macro_unscope.h`.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -97,6 +97,20 @@
...
+#ifndef JSON_HAS_THREE_WAY_COMPARISON
+    #if defined(__cpp_lib_three_way_comparison) && __cpp_lib_three_way_comparison >= 201907L \
+        && defined(__cpp_impl_three_way_comparison)&& __cpp_impl_three_way_comparison >= 201907L
+        #define JSON_HAS_THREE_WAY_COMPARISON
+    #endif
+#endif
+
+#ifdef JSON_HAS_THREE_WAY_COMPARISON
+    #undef JSON_HAS_THREE_WAY_COMPARISON
+    #define JSON_HAS_THREE_WAY_COMPARISON 1
```
gregmarr, line 109: Why not just do the `#define JSON_HAS_THREE_WAY_COMPARISON 1` above?

falbrechtskirchinger: To cover the case when it has been externally defined for whatever reason.
Following the pattern established here:
```
// C++ language standard detection
// if the user manually specified the used c++ version this is skipped
#if !defined(JSON_HAS_CPP_20) && !defined(JSON_HAS_CPP_17) && !defined(JSON_HAS_CPP_14) && !defined(JSON_HAS_CPP_11)
```

I plan on changing all boolean macros to be defined as 0 or 1 eventually (useful for an extension to part 2 of this PR).

gregmarr: In those other cases, if it has been externally defined, we just leave it as-is.  We don't blindly change it to `1`.  This will break if someone has `#define JSON_HAS_THREE_WAY_COMPARISON 0`.

falbrechtskirchinger: You're right.
My end goal is to be able to `static_cast<bool>(ANY_BOOLEAN_MACRO)`, expose that through `basic_json<>::meta()`, and output it as part of the build information (see part 2 of this PR).
I was trying to avoid 
```
#ifdef BOOLEAN_MACRO
meta["BOOLEAN_MACRO"] = true;
#else
meta["BOOLEAN_MACRO"] = false;
#endif
```
and just write
```
meta["BOOLEAN_MACRO"] = static_cast<bool>(BOOLEAN_MACRO);
```

gregmarr: Makes sense, so you just need this:

    #ifndef JSON_HAS_THREE_WAY_COMPARISON
        #if defined(__cpp_lib_three_way_comparison) && __cpp_lib_three_way_comparison >= 201907L \
            && defined(__cpp_impl_three_way_comparison)&& __cpp_impl_three_way_comparison >= 201907L
            #define JSON_HAS_THREE_WAY_COMPARISON 1
        #else
            #define JSON_HAS_THREE_WAY_COMPARISON 0
        #endif
    #endif

falbrechtskirchinger: Done.

falbrechtskirchinger:
I cannot explain the test failure (segfault) seen [here](https://github.com/nlohmann/json/runs/5526651560?check_suite_focus=true) but suspect it might disappear on re-run.
Maybe #1001 isn't completely fixed? I've seen similar sporadic failures but don't recall where.

cmake/ci.cmake
```diff
@@ -434,21 +434,21 @@ add_custom_target(ci_test_clang
...
         COMMENT "Compile and test with GCC for C++${CXX_STANDARD}"
     )
 
     add_custom_target(ci_test_clang_cxx${CXX_STANDARD}
-        COMMAND CXX=${CLANG_TOOL} ${CMAKE_COMMAND}
+        COMMAND CXX=${CLANG_TOOL} CXXFLAGS="${CLANG_CXXFLAGS}" ${CMAKE_COMMAND}
             -DCMAKE_BUILD_TYPE=Debug -GNinja
-            -DCMAKE_CXX_STANDARD=${CXX_STANDARD} -DCMAKE_CXX_STANDARD_REQUIRED=ON
-            -DJSON_BuildTests=ON
+            -DJSON_BuildTests=ON -DJSON_FastTests=ON
```
nlohmann, line 450: Why are `DJSON_FastTests` set to `ON` here?

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -464,7 +464,8 @@ void from_json(const BasicJsonType& j, std_fs::path& p)
 {
     if (JSON_HEDLEY_UNLIKELY(!j.is_string()))
     {
-        JSON_THROW(type_error::create(302, "type must be string, but is " + std::string(j.type_name()), j));
+        // Not tested because of #3377 (related #3070)
```
nlohmann, line 467: I assume the test will be deleted with #3377, right?

include/nlohmann/detail/meta/cpp_future.hpp
```diff
@@ -147,8 +147,12 @@ struct static_const
     static constexpr T value{};
 };
 
-template<typename T>
-constexpr T static_const<T>::value; // NOLINT(readability-redundant-declaration)
+#ifndef JSON_HAS_CPP_17
```
nlohmann, line 150: Why was this guard added?

nlohmann (commented):
Some first comments - I am not done yet reviewing `cmake/test.cmake` and `test/CMakeLists.txt`.

cmake/ci.cmake
falbrechtskirchinger: While an unintentional copy&paste mistake, I left it in for parity with ci_test_gcc_cxx above. Of course we could go the other direction but I believe the coverage provided by other tests (e.g., ci_test_gcc & ci_test_clang) should be adequate, no?

include/nlohmann/detail/meta/cpp_future.hpp
falbrechtskirchinger: See d7954b29fcf1c0da74033435bdb82d24b6259cc1.

> Redelcaration of inline constexpr static data members in namespace scope
> was deprecated in C++17. Fixes -Werror=deprecated compilation failures.

cmake/ci.cmake
nlohmann: No, it's fine. Thanks for the quick response.

include/nlohmann/detail/conversions/from_json.hpp
nlohmann: I was just triggered by the coverage exclusion - we should be able to have a test for every part of the library - especially for "visible" effects like exceptions. I would like to keep it as is here and address it in a later issue.

falbrechtskirchinger: Alright. We have issue #3377 to keep track of it.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #3385, opened 2022-03-09
Improve unit testing (Part 2)

falbrechtskirchinger:
This is part 2 of the unit test improvements and therefore depends on #3380. It contains all commits from part 1 and needs to be rebased.

It is a fairly hefty addition but has proven valuable in debugging C++20/gcc-12 issues already and has helped in fixing some of the issues uncovered in part 1.

What does it do?
* Prints build information at the start of each unit test.
* Adds a target `print_buildinfo` which prints test non-specific build information per (used) C++ standard.
* (Supposedly) all CI steps had `cmake --build <build_dir> --target print_buildinfo` prepended to the test command.

In a first attempt to implement this functionality, I relied upon CMake `configure_file()` and `file(GENERATE)`. The end result was usable but suffered from some inaccuracies, as the order of compile/link options extracted from target properties does not mirror the order in the final command line, nor does it always contain all desirable flags (most importantly `-std=`).

For that reason, this solution relies upon the CMake file API [1] which requires parsing JSON files and therefore needs significant amounts of C++ and another library [2] to platform-independently iterate a directory to find the initial JSON file containing the CMake reply.

To avoid parsing `json.hpp` twice each unit test includes `buildinfo.cpp` directly at the end. This is also enforced in CMake to avoid new unit tests forgetting it.

I intend to expand upon this but will hold of for now, as I expect this PR to be somewhat controversial due to size/reward considerations and the ongoing maintenance burden.

[1] https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html
[2] https://github.com/cxong/tinydir/
     Not quite the very first search result but close and there might be better alternatives.


falbrechtskirchinger:
Note to self: Add ABI string from #3590 to `meta()`.

nlohmann:
This starts to become a very large PR. Please consider breaking it into smaller pieces, such as the whole fuzzer part.

falbrechtskirchinger:
> This starts to become a very large PR. Please consider breaking it into smaller pieces, such as the whole fuzzer part.

Yes. I'm just using this for CI. Locally I already have a separate branch for the fuzzing stuff.

PR #3405, opened 2022-03-30
Refactor unit tests to use more convenient doctest assertion macros (Part 2)

kkarbowiak:
Also, removed some now redundant code and adjusted the expected error messages. The new messages reflect what would be communicated if the old assertion macros were reordered (first `CHECK_THROWS_WITH` and then `CHECK_THROWS_AS`).

### DIFF
--- a/test/src/unit-regression1.cpp
+++ b/test/src/unit-regression1.cpp
@@ -771,7 +771,7 @@ TEST_CASE("regression tests 1")
             ss << "   ";
             json j;
             CHECK_THROWS_WITH_AS(ss >> j,
-                              "[json.exception.parse_error.101] parse error at line 1, column 4: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal", json::parse_error&);
+                                 "[json.exception.parse_error.101] parse error at line 1, column 4: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal", json::parse_error&);
         }
 
         SECTION("one value")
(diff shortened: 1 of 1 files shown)


test/src/unit-regression1.cpp
```diff
@@ -1413,24 +1411,14 @@ TEST_CASE("regression tests 1")
         auto p1 = R"([{"op": "move",
                        "from": "/one/two/three",
                        "path": "/a/b/c"}])"_json;
-        CHECK_THROWS_AS(model.patch(p1), json::out_of_range&);
+        CHECK_THROWS_WITH_AS(model.patch(p1),
+                             "[json.exception.out_of_range.403] key 'a' not found", json::out_of_range&);
 
-        auto p2 = R"([{"op": "move",
```
nlohmann, line 1418: Why did you remove the test for `move` here?

kkarbowiak: Not really removed. The test was executed twice: once to assert an exception is thrown, and second time to assert the exception's message. The new assertion macro does both checks at once, so I removed the redundant code.

gregmarr: I had the exact same question, and then I realized that my confusion was mostly due to the way that the diff was aligned.  :)

kkarbowiak: Yes, the diff aligning is a bit unfortunate in this case. The situation is more clear if you inspect the changes by commits, though.

nlohmann (approved):
Looks good to me.

kkarbowiak:
Thank you for taking the time to review and accept the changes.

PR #3416, opened 2022-04-04
CI: windows-2016 has been deprecated; remove jobs

falbrechtskirchinger:
As previously mentioned windows-2016 is now deprecated. Remove msvc2017* jobs from the workflow.
GH offers no replacement for jobs requiring "Visual Studio 15 2017".

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -29,41 +29,6 @@ jobs:
       - name: test
         run: cd build ; ctest -j 10 -C Debug --output-on-failure
 
-  msvc2017:
-    runs-on: windows-2016
-    strategy:
-      matrix:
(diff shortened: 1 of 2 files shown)


nlohmann (changes requested):
Can you please also update the README section and remove the lines for MSVC 2017?

falbrechtskirchinger:
Done.

The compiler list in README needs a thorough update at some point. GCC 11 (experimental) doesn't seem right, for example.

nlohmann:
Thanks!

nlohmann:
> Done.
> 
> The compiler list in README needs a thorough update at some point. GCC 11 (experimental) doesn't seem right, for example.

Fixed in #3420.

PR #3421, opened 2022-04-05
Restore disabled check for #3070 (except on MSVC)

### LINKED ISSUE #3377: Regression test for #3070 is not being run and fails when enabled

falbrechtskirchinger:
Restore the previously disabled check for regression #3070 on all compilers but MSVC.

To summarize the issue:
Given `namespace fs = std::filesystem`.
On MSVC attempting to construct an `fs::path` from `json` results in an ambiguous overload resolution because `fs::path` can be constructed from both a `std::string` as well as another `fs::path`. To the best of my knowledge there is no way to fix an ambiguous overload situation involving a type we do not control and with `json` implicitly converting to both `std::string` and `fs::path`.

Re-enabling the check where it compiles and keeping it disabled for MSVC is the best we can do.

Closes #3377 and #3382.

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -464,8 +464,7 @@ void from_json(const BasicJsonType& j, std_fs::path& p)
 {
     if (JSON_HEDLEY_UNLIKELY(!j.is_string()))
     {
-        // Not tested because of #3377 (related #3070)
-        JSON_THROW(type_error::create(302, "type must be string, but is " + std::string(j.type_name()), j)); // LCOV_EXCL_LINE
+        JSON_THROW(type_error::create(302, "type must be string, but is " + std::string(j.type_name()), j));
     }
     p = *j.template get_ptr<const typename BasicJsonType::string_t*>();
 }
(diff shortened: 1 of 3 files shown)


falbrechtskirchinger:
Looking at #3229 we may want to disable it for ICPC as well. Should allow the PR to pass CI and be merged.

nlohmann (approved):
Looks good to me.

PR #2685, opened 2021-03-24
string_view support for object access

nlohmann:
This PR adds support for `std::string_view` for object access (`operator[]`, `at`, `value`, `erase`).

Todo:

- [x] Add tests.
- [x] Update documentation.
- [x] Add examples.
- [x] Check if other functions (e.g., `contains`) also need adjustment.

Closes #1529.

PR #3423, opened 2022-04-05
Implement support for string_view (attempt no. 3)

### LINKED ISSUE #1529: How can I use std::string_view as the json_key to "operator []" ?

falbrechtskirchinger:
Based heavily on the previous 2 attempts of @nlohmann with some cleanups of mine.

Closes #1529.
Closes #2685.
Closes #3237.

~Note: Depends on #3415 and needs to be rebased.~ Done.

<hr/>

**Add `key_compare` member to `ordered_map`**

**Replace `==` with `key_compare` in `ordered_map`**

**Expose the actual comparison function used by `object_t`**

`nlohmann::ordered_map` uses a different comparison function than the one provided via template parameter.
* Introduce a type trait to detect if `object_t` has a `key_compare` member.
* Rename `object_comparator_t` to `default_object_comparator_t`.
* Add `object_comparator_t` to be conditionally defined as `object_t::key_compare`, if available, or `default_object_comparator_t` otherwise.
* Update the documentation accordingly.

**Add type traits to check if a type is usable as object key**

Add type trait to check:
* if a type is a specialization of a template.
* if a type is a `json_pointer`.
* if a type is a `basic_json::{const_,}iterator`.
* if two types are comparable using a given comparison functor.
* if a type is comparable to `basic_json::object_t::key_type`.
* if a type is usable as object key.

Based on previous work by @nlohmann.

**Rework `basic_json` element access to accept more key types**

Rework `basic_json` element access member functions and operators to accept any type that meets the requirements defined by type trait `detail::is_usable_as_key_type`.

Member functions and operators:
* `at()`
* `operator[]`
* `value()`
* `erase()`
* `find()`
* `count()`
* `contains()`

Update documentation to reflect these changes.

Add unit tests to excercise the new functions using `std::string_view`.

Based on previous work by @nlohmann.


falbrechtskirchinger:
I'm trying to update the list of linked issues. Trying a comment with the keywords since editing the PR descriptions has no effect.
Closes #1529.
Closes #2685.
Closes #3237.

falbrechtskirchinger:
I'd like to make `key_compare` a member of `ordered_map` and update the constructors to look more like `std;;map`.

Also, there are a few occurrences of `KeyT` that should be renamed to `KeyType` to be consistent with the other uses and the documentation.

Most of the code was written in early March before the `json_pointer` refactor, so I'll definitely want to take a closer look at it myself soon. Otherwise I think this is ready for review.

doc/mkdocs/docs/api/basic_json/at.md
```diff
@@ -172,5 +181,5 @@ Strong exception safety: if an exception occurs, the original value stays intact
 ## Version history
 
 1. Added in version 1.0.0.
-2. Added in version 1.0.0.
+2. Added in version 1.0.0. Added `KeyType` template in version 3.10.6.
```
nlohmann, line 184:
```suggestion
2. Added in version 1.0.0. Added `KeyType` template in version 3.11.0.
```

doc/mkdocs/docs/api/basic_json/contains.md
```diff
@@ -90,5 +92,5 @@ Logarithmic in the size of the JSON object.
 
 ## Version history
 
-1. Added in version 3.6.0.
+1. Added in version 3.6.0. Extended template `KeyType` to support any comparable type in version 3.10.6.
```
nlohmann, line 95:
```suggestion
1. Added in version 3.6.0. Extended template `KeyType` to support any comparable type in version 3.11.0.
```

doc/mkdocs/docs/api/basic_json/count.md
```diff
@@ -53,3 +55,4 @@ This method always returns `0` when executed on a JSON type that is not an objec
 ## Version history
 
 - Added in version 1.0.0.
+- Extended template `KeyType` to support any comparable type in version 3.10.6.
```
nlohmann, line 58:
```suggestion
- Extended template `KeyType` to support any comparable type in version 3.11.0.
```

doc/mkdocs/docs/api/basic_json/default_object_comparator_t.md
```diff
@@ -0,0 +1,18 @@
...
+
+```cpp
+using default_object_comparator_t = std::less<StringType>;  // until C++14
+
+using default_object_comparator_t = std::less<>;            // since C++14
+```
+
+The default comparator used by [`object_t`](object_t.md).
+
+Since C++14 a transparent comparator is used which prevents unnecessary string construction.
```
nlohmann, line 11:
```suggestion
Since C++14, a transparent comparator is used which prevents unnecessary string construction.
```

doc/mkdocs/docs/api/basic_json/default_object_comparator_t.md
```diff
@@ -0,0 +1,18 @@
...
+
+The actual comparator used depends on [`object_t`](object_t.md) and can be obtained via
+[`object_comparator_t`](object_comparator_t.md).
+
+## Version history
+
+- Added in version 3.10.6.
```
nlohmann, line 18:
```suggestion
- Added in version 3.11.0.
```

doc/mkdocs/docs/api/basic_json/erase.md
```diff
@@ -174,3 +182,4 @@ Strong exception safety: if an exception occurs, the original value stays intact
 
 - Added in version 1.0.0.
 - Added support for binary types in version 3.8.0.
+- Added `KeyType` template in version 3.10.6.
```
nlohmann, line 185:
```suggestion
- Added `KeyType` template in version 3.11.0.
```

doc/mkdocs/docs/api/basic_json/find.md
```diff
@@ -61,3 +63,4 @@ This method always returns `end()` when executed on a JSON type that is not an o
 ## Version history
 
 - Added in version 1.0.0.
+- Extended template `KeyType` to support any comparable type in version 3.10.6.
```
nlohmann, line 66:
```suggestion
- Extended template `KeyType` to support any comparable type in version 3.11.0.
```

doc/mkdocs/docs/api/basic_json/object_comparator_t.md
```diff
@@ -1,16 +1,14 @@
...
-The comparator used in [`object_t`](object_t.md).
-
-When C++14 is detected, a transparent com parator is used which, when combined with perfect forwarding on find() and
-count() calls, prevents unnecessary string construction.
+The comparator used by [`object_t`](object_t.md). Defined as `#!cpp typename object_t::key_compare`, if available, and [`default_object_comparator_t`](default_object_comparator_t.md) otherwise.
 
 ## Version history
 
-- Unknown.
+- Changed to use `#!cpp typename object_t::key_compare` or `default_object_comparator_t` in version 3.10.6.
```
nlohmann, line 14:
```suggestion
- Changed to use `#!cpp typename object_t::key_compare` or `default_object_comparator_t` in version 3.11.0.
```

doc/mkdocs/docs/api/basic_json/operator[].md
```diff
@@ -192,5 +192,6 @@ Strong exception safety: if an exception occurs, the original value stays intact
 ## Version history
 
 1. Added in version 1.0.0.
-2. Added in version 1.0.0. Overloads for `T* key` added in version 1.1.0.
+2. Added in version 1.0.0. Overloads for `T* key` added in version 1.1.0. Template `T* key` replaced by template `KeyType`
+   in version 3.10.6 which now also supports `std::string_view`.
```
Comment (nlohmann) [inline doc/mkdocs/docs/api/basic_json/operator[].md:196]:
```suggestion
   in version 3.11.0 which now also supports `std::string_view`.
```

doc/mkdocs/docs/api/basic_json/value.md
```diff
@@ -125,5 +129,5 @@ changes to any JSON value.
 
 ## Version history
 
-1. Added in version 1.0.0.
+1. Added in version 1.0.0. Added `KeyType` template parameter in 3.10.6.
```
nlohmann, line 132:
```suggestion
1. Added in version 1.0.0. Added `KeyType` template parameter in 3.11.0.
```

doc/mkdocs/docs/api/ordered_map.md
```diff
@@ -68,3 +74,4 @@ A minimal map-like container that preserves insertion order for use within [`nlo
 ## Version history
 
 - Added in version 3.9.0 to implement [`nlohmann::ordered_json`](ordered_json.md).
+- Added **key_compare** member in version 3.10.6.
```
nlohmann, line 77:
```suggestion
- Added **key_compare** member in version 3.11.0.
```

include/nlohmann/detail/meta/type_traits.hpp
```diff
@@ -54,11 +54,6 @@ struct is_basic_json_context :
     || std::is_same<BasicJsonContext, std::nullptr_t>::value >
 {};
 
-template<typename> struct is_json_pointer : std::false_type {};
```
nlohmann, line 57: Is this no longer needed?

include/nlohmann/json.hpp
```diff
@@ -2433,15 +2442,24 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+        {
+            JSON_THROW(type_error::create(307, detail::concat("cannot use erase() with ", type_name()), this));
+        }
+
+        const auto it = m_value.object->find(key);
+        if (it != m_value.object->end())
         {
-            return m_value.object->erase(key);
+            m_value.object->erase(it);
+            return 1;
```
nlohmann, line 2459: This may be really nitpicky. The documentation of the return value states

> Number of elements removed. If ObjectType is the default std::map type, the return value will always be 0 (key was not found) or 1 (key was found).

I am not sure whether there is a relevant specialization of the ObjectType (`std::multimap`?) where any other values make sense. If not, then the wording of the documentation should be adjusted to not make it sound as if the type has any influence there.

doc/mkdocs/docs/api/basic_json/find.md
nlohmann, line 66: Should also the change from `KeyT&&` to `const KeyT&` be mentioned?

nlohmann (changes requested):

include/nlohmann/json.hpp
gregmarr: Now that this is guaranteed to always be 0 or 1, it makes sense to update the documentation.

doc/mkdocs/docs/api/basic_json/default_object_comparator_t.md
falbrechtskirchinger: Actually, that is grammatically correct. Since + verb and noun would be followed by a comma, since + noun phrase doesn't need one.

While comma placement is a definite weakness of mine, be it in my native language or English, this one I did get right. :-)

On a different technical note I just realized that we're never checking for the presence of `is_transparent` in our comparator. I'll check if and where that is needed.

doc/mkdocs/docs/api/basic_json/find.md
falbrechtskirchinger: I don't understand the reasoning behind `template<typename K> std::map::erase(K &&)` accepting an rvalue reference. Anything binds to a const lvalue reference. That lack of understanding is making me doubt myself at the moment.

We can mention it but it shouldn't have an effect on existing code, right?

include/nlohmann/json.hpp
falbrechtskirchinger: `std::multimap` isn't usable because it lacks `operator[]`. If it had `operator[]` (and possibly other missing functions) would it work? I.e., are there checks for duplicate keys in the serialization/deserialization code paths?

I'm leaning towards the function returning what the `erase()` call returns.

include/nlohmann/detail/meta/type_traits.hpp
falbrechtskirchinger: It's still needed but I had a different implementation in this branch before I added this one with the `json_pointer` refactor. The implementation in this branch introduces a generic `is_specialization_of` trait. Using this trait `is_json_pointer` becomes a simple alias template declaration and other traits can later be rewritten/shortened as well.

include/nlohmann/json.hpp
nlohmann: Yes, this is what the previous code did:

```cpp
return m_value.object->erase(key);
```

include/nlohmann/detail/meta/type_traits.hpp
```diff
@@ -454,6 +468,51 @@ struct is_constructible_tuple : std::false_type {};
 template<typename T1, typename... Args>
 struct is_constructible_tuple<T1, std::tuple<Args...>> : conjunction<is_constructible<T1, Args>...> {};
 
+template<typename BasicJsonType, typename T>
+struct is_json_iterator : std::false_type {};
```
falbrechtskirchinger, line 472: Should this be called `is_json_iterator_of` to better reflect that the trait takes both a `BasicJsonType` and an iterator?

doc/mkdocs/docs/api/basic_json/find.md
nlohmann: I don't think it makes a difference. But maybe adding it to the version history would be good.

include/nlohmann/detail/meta/type_traits.hpp
nlohmann: Good idea. I'm not sure about the argument order then. `is_json_iterator_of<json, it>` or `is_json_iterator_of<it, json>`?

include/nlohmann/detail/meta/type_traits.hpp
```diff
@@ -160,6 +155,25 @@ struct has_to_json < BasicJsonType, T, enable_if_t < !is_basic_json<T>::value >>
         T>::value;
 };
 
+template<typename T, typename = void>
+struct has_key_compare : std::false_type
+{};
+
+template<typename T>
+struct has_key_compare<T, void_t<typename T::key_compare>> : std::true_type {};
```
falbrechtskirchinger, line 163: Could be done in even fewer lines using `is_detected`.

include/nlohmann/detail/meta/type_traits.hpp
nlohmann: Thanks for clarifying.

doc/mkdocs/docs/api/basic_json/default_object_comparator_t.md
nlohmann: Thanks - it's before 9 am and I already learned something :-)

include/nlohmann/json.hpp
falbrechtskirchinger: I don't know if you changed that or I did but, thinking about it, I believe the previous behavior is preferable.

include/nlohmann/detail/meta/type_traits.hpp
falbrechtskirchinger: "of" referrers to the `BasicJsonType` so it makes more sense as first argument to me.

doc/mkdocs/docs/api/basic_json/default_object_comparator_t.md
falbrechtskirchinger: The current implementation checks via `is_comparable` whether to SFINAE-enable various functions. This ignores possible conversions and temporaries. It would be better to add `is_transparent` as an additional requirement. It then forces the user to explicitly convert if desired.

falbrechtskirchinger:
Downgrading back to draft status. I made substantial changes and want to see if they pass CI. Will update the doc next and re-request review when it's ready.

falbrechtskirchinger:
@nlohmann I'm trying something new and wanted to get your feedback before I continue.

I've added the `mkdocs-macros-plugin` and have defined re-used phrases in a separate file that is included where needed. I'm testing these changes in `at.md` and `contains.md`.

Example:
```
3. See 2. {{ snippets.keytype_is_transparent }}
```
translates to:
```
3. See 2. This overload is only available if `KeyType` is comparable with `#!cpp typename object_t::key_type` and `#!cpp typename object_comparator_t::is_transparent` denotes a type.
```

Essentially, if we want to tweak the wording we can do so in one place now.

Shall I continue or do want me to roll that back?

nlohmann:
> @nlohmann I'm trying something new and wanted to get your feedback before I continue.
> 
> I've added the `mkdocs-macros-plugin` and have defined re-used phrases in a separate file that is included where needed. I'm testing these changes in `at.md` and `contains.md`.
> 
> Example:
> 
> ```
> 3. See 2. {{ snippets.keytype_is_transparent }}
> ```
> 
> translates to:
> 
> ```
> 3. See 2. This overload is only available if `KeyType` is comparable with `#!cpp typename object_t::key_type` and `#!cpp typename object_comparator_t::is_transparent` denotes a type.
> ```
> 
> Essentially, if we want to tweak the wording we can do so in one place now.
> 
> Shall I continue or do want me to roll that back?

I'd rather have it explicit, because this makes it easier for others to edit the Markdown files. I am aware of the DRY violation etc., but I'm afraid this will add too much complexity.

doc/mkdocs/docs/api/basic_json/default_object_comparator_t.md
nlohmann: I guess that was part of all attempts so far... Let's hope you're more successful!

falbrechtskirchinger: > I guess that was part of all attempts so far... Let's hope you're more successful!

Tests succeed. Just need to finish the documentation and I'm procrastinating. ;-)

(Same goes for C++20 stuff … even more documentation. Yikes.)

nlohmann: Thank you so much for your huge effort and your persistence!

Ericson2314:
Would be very nice if this could happen!

doc/mkdocs/docs/api/basic_json/at.md
```diff
@@ -134,7 +152,7 @@ Strong exception safety: if an exception occurs, the original value stays intact
     --8<-- "examples/at__object_t_key_type_const.output"
     ```
 
-??? example "Example (3) access specified element via JSON Pointer"
+??? example "Example (4) access specified element via JSON Pointer"
```
falbrechtskirchinger, line 155: Speaking of examples earlier … this is where I'd really need to add some. If not in this PR then soon after.

doc/mkdocs/docs/api/basic_json/contains.md
```diff
@@ -1,22 +1,31 @@
 # <small>nlohmann::basic_json::</small>contains
 
+{% import 'snippets.md' as snippets %}
+
```
falbrechtskirchinger, line 4: Left over from an experiment. Remove.

doc/mkdocs/docs/api/basic_json/erase.md
```diff
@@ -56,7 +70,8 @@ void erase(const size_type idx);
    is returned.
 3. Number of elements removed. If `ObjectType` is the default `std::map` type, the return value will always be `0`
    (`key` was not found) or `1` (`key` was found).
-4. /
+4. See 3.
+5. /
```
falbrechtskirchinger, line 74: Especially seeing it out of context I never quite know what the slash is supposed to mean.

It's used all over and could be clearer.

doc/mkdocs/docs/api/basic_json/find.md
```diff
@@ -1,20 +1,28 @@
...
+// (2)
+template<typename KeyType>
+iterator find(KeyType&& key);
+template<typename KeyType>
+const_iterator find(KeyType&& key) const;
 ```
 
-Finds an element in a JSON object with key equivalent to `key`. If the element is not found or the JSON value is not an
-object, `end()` is returned.
+1. Finds an element in a JSON object with key equivalent to `key`. If the element is not found or the
```
falbrechtskirchinger, line 15: "... with **a** key ..."

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -106,6 +106,16 @@
...
 
+// see https://devblogs.microsoft.com/cppblog/msvc-cpp20-and-the-std-cpp20-switch/
+// #if defined(_MSC_FULL_VER) && _MSC_FULL_VER >= 192829913
+//     #define JSON_NO_UNIQUE_ADDRESS [[msvc::no_unique_address]]
+// #el
+#if JSON_HEDLEY_HAS_ATTRIBUTE(no_unique_address)
+    #define JSON_NO_UNIQUE_ADDRESS [[no_unique_address]]
+#else
+    #define JSON_NO_UNIQUE_ADDRESS
+#endif
```
falbrechtskirchinger, line 117: Still don't know if it's worth adding on MSVC. Long-term we may want to add a macro for ABI stability. Then, if users opt-out of ABI stability, we could enable the optimization on MSVC. But that would only be worth the effort if it could be used elsewhere.

include/nlohmann/ordered_map.hpp
```diff
@@ -225,6 +230,14 @@ template <class Key, class T, class IgnoredLess = std::less<Key>,
             insert(*it);
         }
     }
+
+private:
+    JSON_HEDLEY_DIAGNOSTIC_PUSH
+#ifdef _MSC_VER
+#pragma warning(disable:4848)
```
falbrechtskirchinger, line 237: Only needed if we enable `[[msvc::no_unique_address]]`. (Currently commented out.)

doc/mkdocs/docs/api/basic_json/at.md
```diff
@@ -10,13 +10,28 @@ reference at(const typename object_t::key_type& key);
...
+
+// (4)
 reference at(const json_pointer& ptr);
 const_reference at(const json_pointer& ptr) const;
 ```
 
 1. Returns a reference to the array element at specified location `idx`, with bounds checking.
 2. Returns a reference to the object element at with specified key `key`, with bounds checking.
-3. Returns a reference to the element at with specified JSON pointer `ptr`, with bounds checking.
+3. See 2. This overload is only available if `KeyType` is comparable with `#!cpp typename object_t::key_type` and
```
falbrechtskirchinger, line 25: Fix typo in line above.

doc/mkdocs/docs/api/basic_json/at.md
```diff
@@ -10,13 +10,28 @@ reference at(const typename object_t::key_type& key);
...
+   `#!cpp typename object_comparator_t::is_transparent` denotes a type.
+4. Returns a reference to the element at with specified JSON pointer `ptr`, with bounds checking.
```
falbrechtskirchinger, line 27: "element at with"

That mistake seems to have been copied a few times. Fix.

include/nlohmann/json.hpp
falbrechtskirchinger: Turns out there was a reason for this code. The required `std::map::erase()` overload is a recent addition to the standard and so we need to use the iterator based `erase()` which doesn't return the number of erased elements.

`erase_internal()` handles this via SFINAE.

doc/mkdocs/docs/api/basic_json/value.md
```diff
@@ -125,5 +140,6 @@ changes to any JSON value.
 
 ## Version history
 
-1. Added in version 1.0.0.
-2. Added in version 2.0.2.
+1. Added in version 1.0.0. Changed parameter `default_value` type from `const ValueType&` to `ValueType&&`.
```
nlohmann, line 143: It seems a version number is missing here.

doc/mkdocs/docs/api/basic_json/erase.md
nlohmann: I could replace it with "(none)" just like cppreference is doing.

doc/mkdocs/docs/api/basic_json/default_object_comparator_t.md
nlohmann, line 11: Maybe add half a sentence where it is used?

include/nlohmann/detail/macro_scope.hpp
nlohmann: Regarding ABI stability, I am torn. On the one hand, it's nice to have and all. On the other hand, it's so hard to check and brings so many limitations that it could start to become a blocker for a header-only library...

doc/mkdocs/docs/api/basic_json/default_object_comparator_t.md
falbrechtskirchinger: "Since C++14 a transparent comparator is used which prevents unnecessary string construction when looking up a key in an object."

doc/mkdocs/docs/api/basic_json/erase.md
falbrechtskirchinger:
```
api/basic_json/basic_json.md:1. /
api/basic_json/basic_json.md:3. /
api/basic_json/basic_json.md:4. /
api/basic_json/basic_json.md:6. /
api/basic_json/basic_json.md:8. /
api/basic_json/insert.md:5. /
api/basic_json/to_bson.md:2. /
api/basic_json/to_cbor.md:2. /
api/basic_json/to_msgpack.md:2. /
api/basic_json/to_ubjson.md:2. /
api/basic_json/erase.md:5. /
api/basic_json/erase.md:2. /
api/basic_json/erase.md:5. /
```

Would you like me to replace all occurrences?

nlohmann: Yes, please.

include/nlohmann/detail/macro_scope.hpp
falbrechtskirchinger: I'm leaning towards removing `[[msvc::no_unique_address]]` but keeping `[[no_unique_address]]`. It's a fairly minor optimization.

nlohmann: What is the issue with `[[msvc::no_unique_address]]`?

falbrechtskirchinger: It's enabled regardless of C++ standard version and might catch users off guard. In terms of ABI stability which would otherwise not be affected.

nlohmann: Then I agree. I think it's one homework to decide to which extend ABI stability is something to strive for.

falbrechtskirchinger:
I've discovered some last minute issues yesterday that were bad enough to make me question the overall soundness of this PR. After some careful tweaks I am now cautiously optimistic that everything has been resolved (pending green CI and 100% coverage).
A few more overloads were necessary for the API to match the documented behavior.

I plan on reviewing everything once again side-by-side with the documentation to make sure they match up.

doc/mkdocs/docs/api/basic_json/operator[].md
```diff
@@ -42,7 +48,8 @@ const_reference operator[](const json_pointer& ptr) const;
 
 1. reference to the element at index `idx`
 2. reference to the element at key `key`
-3. reference to the element pointed to by `ptr`
+3. see 2
```
Comment (nlohmann) [inline doc/mkdocs/docs/api/basic_json/operator[].md:51]: Nitpicky, but I'd rather repeat `Logarithmic in the size of the container.` here. The level of indirection for that short sentence confuses me.

doc/mkdocs/docs/api/basic_json/operator[].md
```diff
@@ -70,7 +77,8 @@ Strong exception safety: if an exception occurs, the original value stays intact
 
 1. Constant if `idx` is in the range of the array. Otherwise, linear in `idx - size()`.
 2. Logarithmic in the size of the container.
-3. Constant
+3. See 2.
```
Comment (nlohmann) [inline doc/mkdocs/docs/api/basic_json/operator[].md:80]: Nitpicky, but I'd rather repeat `Logarithmic in the size of the container.` here. The level of indirection for that short sentence confuses me.

doc/mkdocs/docs/api/basic_json/erase.md
```diff
@@ -103,14 +119,16 @@ Strong exception safety: if an exception occurs, the original value stays intact
        - strings and binary: linear in the length of the member
        - other types: constant
 3. `log(size()) + count(key)`
-4. Linear in distance between `idx` and the end of the container.
+4. See 3.
```
nlohmann, line 122: Nitpicky, but I'd rather repeat `Logarithmic in the size of the container.` here. The level of indirection for that short sentence confuses me.

doc/mkdocs/docs/api/basic_json/contains.md
```diff
@@ -39,7 +47,8 @@ Strong exception safety: if an exception occurs, the original value stays intact
 ## Exceptions
 
 1. The function does not throw exceptions.
-2. The function can throw the following exceptions:
+2. See 1.
```
nlohmann, line 50: Nitpicky, but I'd rather repeat `Logarithmic in the size of the container.` here. The level of indirection for that short sentence confuses me.

doc/mkdocs/docs/api/basic_json/at.md
```diff
@@ -70,7 +87,8 @@ Strong exception safety: if an exception occurs, the original value stays intact
 
 1. Constant
 2. Logarithmic in the size of the container.
-3. Constant
+3. See 2.
```
nlohmann, line 90: Nitpicky, but I'd rather repeat `Logarithmic in the size of the container.` here. The level of indirection for that short sentence confuses me.

doc/mkdocs/docs/api/basic_json/at.md
```diff
@@ -25,15 +40,16 @@ const_reference at(const json_pointer& ptr) const;
...
 `ptr` (in)
 :   JSON pointer to the desired element
-    
+
 ## Return value
 
 1. reference to the element at index `idx`
 2. reference to the element at key `key`
-3. reference to the element pointed to by `ptr`
+3. see 2
```
nlohmann, line 51: Nitpicky, but I'd rather repeat `Logarithmic in the size of the container.` here. The level of indirection for that short sentence confuses me.

doc/mkdocs/docs/api/basic_json/operator[].md
Comment (falbrechtskirchinger) [reply doc/mkdocs/docs/api/basic_json/operator[].md:80]: While we're on the subject of complexities: Shouldn't any access via `json_pointer` be at least linear? `O(N)` with `N` being the number of reference tokens. And that ignores the complexity of each access per reference token.

doc/mkdocs/docs/api/basic_json/operator[].md
Comment (gregmarr) [reply doc/mkdocs/docs/api/basic_json/operator[].md:51]: What about

2. & 3. reference to the element at key `key`

Comment (falbrechtskirchinger) [reply doc/mkdocs/docs/api/basic_json/operator[].md:51]: He copy & pasted the message. I understand what he means.

doc/mkdocs/docs/api/basic_json/operator[].md
Comment (nlohmann) [reply doc/mkdocs/docs/api/basic_json/operator[].md:80]: Constant makes no sense here. It's logarithmic in the size of the container (because of the way reference tokens are interpreted for objects; a JSON pointer that only traverses through arrays like `/0/1/2/3` would be constant).

"O(N) with N being the number of reference tokens" is true, but the length of a JSON Pointer should be negligible compared to the size of a container.

doc/mkdocs/docs/api/basic_json/operator[].md
Comment (gregmarr) [reply doc/mkdocs/docs/api/basic_json/operator[].md:51]: I understood it too, I'm suggesting

    2. & 3. reference to the element at key `key`

rather than

    2. reference to the element at key `key`
    3. reference to the element at key `key`

Comment (nlohmann) [reply doc/mkdocs/docs/api/basic_json/operator[].md:51]: I wish this would possible, but that does not work in Markdown lists.

If it's in the middle of a list, then the numbering breaks. If it's at the end of the list, then numbering is fine, but indentation looks odd.

Comment (gregmarr) [reply doc/mkdocs/docs/api/basic_json/operator[].md:51]: Ah, okay.  That's too bad.

Comment (falbrechtskirchinger) [reply doc/mkdocs/docs/api/basic_json/operator[].md:51]: I have come across a "better lists" markdown extension. Maybe it is possible? I may look into that later if I remember.

doc/mkdocs/docs/api/basic_json/operator[].md
Comment (nlohmann) [reply doc/mkdocs/docs/api/basic_json/operator[].md:80]: The documentation apparently needs more eyes on it...

doc/mkdocs/docs/api/basic_json/operator[].md
Comment (falbrechtskirchinger) [reply doc/mkdocs/docs/api/basic_json/operator[].md:51]: Except that the numbering is of course done by HTML `<ol>` tags. So probably not...

doc/mkdocs/docs/api/basic_json/operator[].md
Comment (falbrechtskirchinger) [reply doc/mkdocs/docs/api/basic_json/operator[].md:80]: @nlohmann Because I just noticed that I'm missing an exception description for an overload: A more advanced feature of `check_structure.py` could be to check if sections like "Return value", "Exceptions", "Complexity" use numbered lists, that the number of entries is the same.

(This is also an example of something I might put into `Roadmap.md`.)

Comment (nlohmann) [reply doc/mkdocs/docs/api/basic_json/operator[].md]: Good point. Though I still think issues or discussions are a better place to collect such thoughts...

Comment (nlohmann) [reply doc/mkdocs/docs/api/basic_json/operator[].md]: The issue templates basically allow anything. Past experiences however made me request a lot of standardized information, because people tend to be lazy and create issues where a lot of back and forth is needed to understand what it's about.

Comment (nlohmann) [reply doc/mkdocs/docs/api/basic_json/operator[].md]: See #3455. I think Discussion/Ideas are the format we need.

falbrechtskirchinger:
On it.

falbrechtskirchinger:
I've gone through the documentation and the code side-by-side to make sure that all documented functions are in fact there and correct. I've also added a comment to a particularly unrecognizable overload:
```cpp
    // this is the value(const typename object_t::key_type&) overload
    template < class KeyType, class ValueType, detail::enable_if_t <
                   std::is_same<KeyType, typename object_t::key_type>::value
                   && detail::is_getable<basic_json_t, ValueType>::value
                   && !std::is_same<value_t, ValueType>::value, int > = 0 >
    typename std::decay<ValueType>::type value(const KeyType& key, ValueType && default_value) const
    {
```

I believe this thing is now done and ready for final reviews.

My suggestion would be to merge #3336 first, though. I'll have an easier time resolving merge conflicts, if there are any.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks a lot! This is really a milestone!!!

PR #3431, opened 2022-04-08
Document which version introduced the macros

nlohmann:
Closes #3429.

### DIFF
--- a/doc/mkdocs/docs/features/macros.md
+++ b/doc/mkdocs/docs/features/macros.md
@@ -14,6 +14,8 @@ This macro controls which code is executed for runtime assertions of the librari
     #define JSON_ASSERT(x) assert(x)
     ```
 
+The macro was introduced in version 3.9.0.
+
 ## `JSON_CATCH_USER(exception)`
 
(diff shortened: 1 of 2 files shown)


doc/mkdocs/docs/features/macros.md
```diff
@@ -63,6 +67,8 @@ The diagnostics messages can also be controlled with the CMake option `JSON_Diag
     #define JSON_DIAGNOSTICS 0
     ```
 
+The macro was introduced in version 3.10.0.
+
 ## `JSON_HAS_CPP_11`, `JSON_HAS_CPP_14`, `JSON_HAS_CPP_17`, `JSON_HAS_CPP_20`
```
falbrechtskirchinger, line 72: These macros aren't user-visible. They're `#undef`'ed in `macro_unscope.hpp`. Which raises the broader question: should internal macros be documented and specifically labeled as such? Also, there's `JSON_TEST_KEEP_MACROS` which does mean we can "leak" some internal macros.

Food for thought for a future documentation update…

PR #3451, opened 2022-04-22
Disable regression test for #3070 on GCC <8.4

falbrechtskirchinger:
Disables part of the regression test for #3070 on GCC <8.4 (already disabled for MSVC).
Discussed [here](https://github.com/nlohmann/json/discussions/3299#discussioncomment-2614557).

Once again, the ambiguous overload resolution isn't something we can fix (easily, if at all).

### DIFF
--- a/test/src/unit-regression2.cpp
+++ b/test/src/unit-regression2.cpp
@@ -48,6 +48,7 @@ using ordered_json = nlohmann::ordered_json;
 #endif
 
 #if JSON_HAS_EXPERIMENTAL_FILESYSTEM
+// JSON_HAS_CPP_17 (magic keyword; do not remove)
 #include <experimental/filesystem>
 namespace nlohmann::detail
 {
@@ -60,7 +61,6 @@ namespace nlohmann::detail
 namespace std_fs = std::filesystem;
 } // namespace nlohmann::detail
 #endif
-
 
 #ifdef JSON_HAS_CPP_20
     #include <span>
@@ -793,8 +793,8 @@ TEST_CASE("regression tests 2")
         const auto j_path = j.get<nlohmann::detail::std_fs::path>();
         CHECK(j_path == text_path);
 
-#ifndef _MSC_VER
-        // works everywhere but on MSVC
+#if !defined(_MSC_VER) && !(defined(__GNUC__) && __GNUC__ == 8 && __GNUC_MINOR__ < 4)
+        // works everywhere but on MSVC and GCC <8.4
         CHECK_THROWS_WITH_AS(nlohmann::detail::std_fs::path(json(1)), "[json.exception.type_error.302] type must be string, but is number", json::type_error);
 #endif
     }


test/src/unit-regression2.cpp
```diff
@@ -48,6 +48,7 @@ using ordered_json = nlohmann::ordered_json;
 #endif
 
 #if JSON_HAS_EXPERIMENTAL_FILESYSTEM
+// JSON_HAS_CPP_17 (magic keyword; do not remove)
```
falbrechtskirchinger, line 51: This is unrelated but I want to make sure these test are built with the correct standard version and the CMake code only checks for the presence of `JSON_HAS_CPP_*` macros. `JSON_HAS_CPP_17` is mentioned right above but if this file is ever refactored, that might change.

test/src/unit-regression2.cpp
```diff
@@ -793,8 +793,8 @@ TEST_CASE("regression tests 2")
         const auto j_path = j.get<nlohmann::detail::std_fs::path>();
         CHECK(j_path == text_path);
 
-#ifndef _MSC_VER
-        // works everywhere but on MSVC
+#if !defined(_MSC_VER) && !(defined(__GNUC__) && __GNUC__ == 8 && __GNUC_MINOR__ < 4)
+        // works everywhere but on MSVC and GCC <8.4
```
falbrechtskirchinger, line 797: `<filesystem>` is only available since GCC 8 so this check should be enough. I tested 8.1, 8.2, and 8.3 on Compiler explorer.

test/src/unit-regression2.cpp
nlohmann: Please add a comment that explains this "magic", e.g.

> cmake/test.cmake expects C++-17 tests to contain JSON_HAS_CPP_17 keyword

or so.

nlohmann: Alternatively, make the requirement explicit and add

```cpp
#ifdef JSON_HAS_CPP_17
```

before

```cpp
#if JSON_HAS_EXPERIMENTAL_FILESYSTEM
```

falbrechtskirchinger: I don't want to clutter the body of the tests with repeated explanations. Ideally every usage of a macro linked to a standard version (e.g., `JSON_HAS_RANGES`, `JSON_HAS_THREE_WAY_COMPARISON`, and probably more with each new standard) should also have that keyword close by, to avoid a repeat of tests not being built for the right standard.

I suggest a detailed note at the top of each test and a short referral to that note at each point of use. This CMake requirement is not documented or explained anywhere else right now.

```
// cmake/test.cmake selects the C++ standard versions with which to build the
// unit tests based on the presence of JSON_HAS_CPP_<VERSION> macros.
// When using macros that are only defined for particular versions of the standard
// (e.g., JSON_HAS_FILESYSTEM for C++17 and up), please mention the corresponding
// version macro in a comment close by, like this:
// JSON_HAS_CPP_<VERSION> (do not remove; see note at top of file)
```

I'd make these changes in one of the other PRs (C++20 fixes probably, since I'm using similar language there) and update all unit tests in one go. Does that work for you?

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

LeSpocky:
Currently building on Debian GNU/Linux 10 (buster) here, and hence stumbled over this. Will this be backported to the 3.10.x branch?

nlohmann:
No, we will not backport changes. This will be fixed in the 3.11.0 release.

PR #3452, opened 2022-04-23
Optimize copy / assignment and size of json iterator

sjanel:
Hello,

I stambled accross this clang-tidy warning in my project and investigated further, `it` being a `json::const_iterator`:
```
the parameter 'it' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
```

It looks like `json::iterator` is actually non trivially copyable (and quite big), which is unexpected for an iterator and the reason why clang-tidy is complaining. In templated code it is common practice to pass iterators by value and not by reference. After investigation and reading of the code I realized that it was made like that at the beginning as a workaround for a MSVC bug (see https://github.com/nlohmann/json/issues/1608). It is sad to impact all other compilation modes and compilers just for a MSVC bug.

I redesigned `internal_iterator_t` with two distinct implementations, one optimized in the case `primitive_iterator_t`, `object_t::iterator` and `array_t::iterator` are all trivially copyable (standard case, mainly for non-MSVC compilers), one for the general case (mainly used by MSVC, and other compilers with exotic vector and/or map iterator implementations).

This is the fix that I propose, in the hope that since then, the MSVC issue is not a concern anymore. After this PR, all operations on iterators are expected to be faster and iterators can be passed by value without any performance impact like standard operators.

### DIFF
--- a/docs/mkdocs/requirements.txt
+++ b/docs/mkdocs/requirements.txt
@@ -12,7 +12,7 @@ httplib2==0.20.4
 idna==3.3
 importlib-metadata==4.12.0
 Jinja2==3.1.2
-joblib==1.1.0
+joblib==1.2.0
 jsmin==3.0.1
 livereload==2.6.3
 lunr==0.6.2


nlohmann:
Thanks! It seems @falbrechtskirchinger is working in #3446 on similar issues - can you please check whether his adjustments go in the same direction?

falbrechtskirchinger:
> * Remove default value initialization, which does not seem to be used, such that union is valid
> * Made `iter_impl` copy construction default, such that it is trivially copyable (did the same for copy assignment)

Some of this is probably redundant due to my pending PR. Will look at the code in a second.

> This is the fix that I propose, in the hope that since then, the MSVC issue is not a concern anymore. After this PR, all operations on iterators are expected to be faster and iterators can be passed by value without any performance impact like standard operators.

MSVC still can't handle it:
"error C2338: union cannot be safely used with non trivial types"

But I agree that pessimizing this for everyone else is bad practice.

> PS: I don't know why, but formatter slightly changes output of the file here and there... is it coming from my clang-format version (I use v14) ? Please help me setting my environment with the expected formatter settings

This project uses Artistic Style to format code which is executed by `make amalgamate` or `make pretty`. You should disable `clang-format` when working on this code.

sjanel:
> Thanks! It seems @falbrechtskirchinger is working in #3446 on similar issues - can you please check whether his adjustments go in the same direction?

Thanks for your comment. I checked the PR and it's not exactly the same thing, it's more about preparing for C++20 support with ranges. This PR is mainly an optimization of the `json::iterator` type.

sjanel:
> > * Remove default value initialization, which does not seem to be used, such that union is valid
> > * Made `iter_impl` copy construction default, such that it is trivially copyable (did the same for copy assignment)
> 
> Some of this is probably redundant due to my pending PR. Will look at the code in a second. Edit: Actually I hardly touched `iter_impl` my fixes target proxy iteration.
> 
> > This is the fix that I propose, in the hope that since then, the MSVC issue is not a concern anymore. After this PR, all operations on iterators are expected to be faster and iterators can be passed by value without any performance impact like standard operators.
> 
> MSVC still can't handle it: "error C2338: union cannot be safely used with non trivial types"
> 
> But I agree that pessimizing this for everyone else is bad practice.
> 
> > PS: I don't know why, but formatter slightly changes output of the file here and there... is it coming from my clang-format version (I use v14) ? Please help me setting my environment with the expected formatter settings
> 
> This project uses Artistic Style to format code which is executed by `make amalgamate` or `make pretty`. You should disable `clang-format` when working on this code.

Thanks for taking a look at my PR! I am currently working on a cleaner `iter_impl` redesign to handle general case and (hopefully) fix MSVC compilation error. It complains because it does not like union with non trivial types (even if they are trivially copyable).

include/nlohmann/detail/iterators/internal_iterator.hpp
```diff
@@ -1,25 +1,189 @@
...
+    // take the maximum alignment of all three iterators
+    static constexpr std::size_t kMaxIteratorAlignment = std::alignment_of<ObjectIterator>::value < std::alignment_of<ArrayIterator>::value ? (std::alignment_of<ArrayIterator>::value < std::alignment_of<primitive_iterator_t>::value
+            ? std::alignment_of<primitive_iterator_t>::value
+            : std::alignment_of<ArrayIterator>::value)
+            : (std::alignment_of<ObjectIterator>::value < std::alignment_of<primitive_iterator_t>::value
+               ? std::alignment_of<primitive_iterator_t>::value
+               : std::alignment_of<ObjectIterator>::value);
+
+  protected:
+    typename std::aligned_storage<kMaxIteratorSize, kMaxIteratorAlignment>::type m_data;
```
falbrechtskirchinger, line 67: Please use `alignas(kMaxIteratorAlignment) std::uint8_t m_data[kMaxIteratorSize];`

`std::aligned_storage` has been deprecated.

sjanel: PR is ready for review (Microsoft issues fixed ;)), although I still have different formatter output with `make amalgamate` (my `astyle` version is `Artistic Style Version 3.1`). Is it an issue, and if so, why do I have all these differences ? (see `include/nlohmann/json.hpp` for instance)

include/nlohmann/json.hpp
```diff
@@ -51,19 +51,19 @@ SOFTWARE.
...
-#include <utility> // declval, forward, move, pair, swap
-#include <vector> // vector
+    #include <iosfwd>  // istream, ostream
+#endif                 // JSON_NO_IO
+#include <iterator>    // random_access_iterator_tag
+#include <memory>      // unique_ptr
+#include <numeric>     // accumulate
+#include <string>      // string, stoi, to_string
+#include <utility>     // declval, forward, move, pair, swap
+#include <vector>      // vector
```
falbrechtskirchinger, line 66: About the formatting: I assume you reformatted with `clang-format` and then used `astyle`?
`astyle` does not produce consistent output for equal source input. It keeps some of the input formatting, especially some white space as you can see here.

I think you'll have to find a way to undo that formatting. Otherwise reviewing and resolving merge conflicts isn't going to be much fun.

include/nlohmann/json.hpp
```diff
@@ -3053,12 +3046,12 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
-        auto insert_pos = std::distance(m_value.array->begin(), pos.m_it.array_iterator);
-        m_value.array->insert(pos.m_it.array_iterator, std::forward<Args>(args)...);
-        result.m_it.array_iterator = m_value.array->begin() + insert_pos;
+        auto insert_pos = std::distance(m_value.array->begin(), pos.m_it.array_iterator());
+        m_value.array->insert(pos.m_it.array_iterator(), std::forward<Args>(args)...);
+        result.m_it.set_array_iterator(m_value.array->begin() + insert_pos);
 
         // This could have been written as:
-        // result.m_it.array_iterator = m_value.array->insert(pos.m_it.array_iterator, cnt, val);
+        // result.m_it.set_array_iterator(m_value.array->insert(pos.m_it.array_iterator(), cnt, val));
```
falbrechtskirchinger, line 3054: I've come across a related Codacy issue today ... both things could be solved with a helper. Doesn't have to be here though.

include/nlohmann/detail/iterators/internal_iterator.hpp
```diff
@@ -1,25 +1,279 @@
...
+                   // NOLINTNEXTLINE
+                   : alignof(object_iterator_t))) std::uint8_t m_t[sizeof(object_iterator_t) < sizeof(array_iterator_t) ? (sizeof(array_iterator_t) < sizeof(primitive_iterator_t)
+                                                     ? sizeof(primitive_iterator_t)
+                                                     : sizeof(array_iterator_t))
+                                                     : (sizeof(object_iterator_t) < sizeof(primitive_iterator_t)
+                                                        ? sizeof(primitive_iterator_t)
+                                                        : sizeof(object_iterator_t))];
+#endif
+    } m_data {};
+};
```
falbrechtskirchinger, line 73: Horrible, indeed. I think I have a better solution. Will get back to you when I have more time.

falbrechtskirchinger (changes requested):
Some initial thoughts.
The formatting of `json.hpp` is very distracting. I scrolled through the file very quickly and probably missed most of your changes.

include/nlohmann/json.hpp
sjanel: "astyle does not produce consistent output for equal source input." Now I understood the issue, yes I have clang-format on save by default in my IDE. I will fix all the formatting issues manually, sorry for the inconvenience.

include/nlohmann/detail/iterators/internal_iterator.hpp
falbrechtskirchinger: This should work:

https://godbolt.org/z/MGhhEsd3G

```cpp
namespace detail
{

template<std::size_t... Is>
struct max;

template<>
struct max<> {
    using type = std::integral_constant<std::size_t, 0>;
};

template<std::size_t I, std::size_t... Rest>
struct max<I, Rest...> {
    using rest_type = typename max<Rest...>::type;
    using type = typename std::conditional<
        (I > rest_type::value),
        std::integral_constant<std::size_t, I>,
        rest_type>::type;
};

template<typename... Ts>
using max_alignof = typename max<alignof(Ts)...>::type;

template<typename... Ts>
using max_sizeof = typename max<sizeof(Ts)...>::type;

}

constexpr std::size_t data_align =
    detail::max_alignof<int, double, char>::value;
constexpr std::size_t data_size =
    detail::max_sizeof<int, double, char>::value;

alignas(data_align) std::uint8_t data[data_size];
```

Maybe add a new file `meta/algorithm.hpp` or similar.

nlohmann:
Please update to the latest develop branch (we renamed some folders in #3462).

sjanel:
> Please update to the latest develop branch (we renamed some folders in #3462).

Ok, done

include/nlohmann/detail/iterators/internal_iterator.hpp
```diff
@@ -1,25 +1,372 @@
...
+#include <nlohmann/detail/meta/memory.hpp>
+#include <nlohmann/detail/meta/type_traits.hpp>
+#include <nlohmann/detail/value_t.hpp>
 
 namespace nlohmann
 {
 namespace detail
 {
+
+template<class ObjectIterator, class ArrayIterator>
```
nlohmann, line 19: Please add some documentation on the idea behind the class.

include/nlohmann/detail/iterators/internal_iterator.hpp
```diff
@@ -1,25 +1,372 @@
...
+                return iterator_type::primitive;
+            }
+        }
+    }
+
+    struct internal_data_t
+    {
+        internal_data_t() noexcept = default;
+
+#ifdef JSON_HAS_CPP_14
```
nlohmann, line 119: Please document the idea behind this piece of code.

include/nlohmann/detail/iterators/internal_iterator.hpp
```diff
@@ -1,25 +1,372 @@
...
+        this->array_iterator() = std::forward<ArrIt>(it);
+    }
 };
+
+/*!
+@brief an iterator value
+*/
+template<class ObjectIterator, class ArrayIterator>
+using internal_iterator = internal_iterator_impl < ObjectIterator, ArrayIterator,
+#if defined(__clang__) || !defined(JSON_HEDLEY_GNUC_VERSION) || JSON_HEDLEY_GNUC_VERSION_CHECK(5, 0, 0)
```
falbrechtskirchinger, line 362: This condition isn't needed anymore, right? It's handled in your type trait implementation.

include/nlohmann/detail/meta/type_traits.hpp
```diff
@@ -570,5 +570,100 @@ T conditional_static_cast(U value)
     return value;
 }
 
+#if defined(__clang__) || !defined(JSON_HEDLEY_GNUC_VERSION) || JSON_HEDLEY_GNUC_VERSION_CHECK(5, 0, 0)
```
falbrechtskirchinger, line 573: This can be simplified to:
```cpp
#if !defined(JSON_HEDLEY_GNUC_VERSION) || JSON_HEDLEY_GNUC_VERSION_CHECK(5, 0, 0)
```

falbrechtskirchinger (changes requested):
I'd still prefer something like my proposed `max_alignof` and `max_sizeof` traits over nested tertiary operators and separate code paths for C++11 and C++14 and up.

include/nlohmann/detail/iterators/internal_iterator.hpp
falbrechtskirchinger: And a note about the `std::aligned_storage` deprecation in C++23.

nlohmann (changes requested):
Please address @falbrechtskirchinger 's comments.

PR #3478, opened 2022-05-08
Document fuzzer usage

nlohmann:
This PR adds documentation how to use LibFuzzer or afl++.

### DIFF
--- a/tests/fuzzing.md
+++ b/tests/fuzzing.md
@@ -1,12 +1,12 @@
 # Fuzz testing
 
 Each parser of the library (JSON, BJData, BSON, CBOR, MessagePack, and UBJSON) can be fuzz tested. Currently,
-[LibFuzzer](https://llvm.org/docs/LibFuzzer.html) and [afl++](https://github.com/AFLplusplus/AFLplusplus) are supported.
+[libFuzzer](https://llvm.org/docs/LibFuzzer.html) and [afl++](https://github.com/AFLplusplus/AFLplusplus) are supported.
 
 ## Corpus creation
 
 For most effective fuzzing, a [corpus](https://llvm.org/docs/LibFuzzer.html#corpus) should be provided. A corpus is a
-directory with some simple  input files that cover several features of the parser and is hence a good starting point
+directory with some simple input files that cover several features of the parser and is hence a good starting point
 for mutations.
 
 ```shell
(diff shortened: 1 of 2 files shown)


tests/fuzzing.md
```diff
@@ -0,0 +1,79 @@
...
+for mutations.
+
+```shell
+TEST_DATA_VERSION=3.1.0
+wget https://github.com/nlohmann/json_test_data/archive/refs/tags/v$TEST_DATA_VERSION.zip
+unzip v$TEST_DATA_VERSION.zip
+rm v$TEST_DATA_VERSION.zip
+for FORMAT in json bjdata bson cbor msgpack ubjson
+do
+  rm -fr mkdir corpus_$FORMAT
```
falbrechtskirchinger, line 19: Remove "mkdir".

tests/fuzzing.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+The generated corpus can be used with both LibFuzzer and afl++.
+
+## LibFuzzer
+
+To use LibFuzzer, you need to pass `-fsanitize=fuzzer` as `FUZZER_ENGINE`:
+
+```shell
+make fuzzers FUZZER_ENGINE="-fsanitize=fuzzer"
```
falbrechtskirchinger, line 33: If CXX isn't set in the current shell or doesn't point to `clang++` built with `libfuzzer`-enabled runtime, it has to be set here.

tests/fuzzing.md
```diff
@@ -0,0 +1,79 @@
...
+The fuzzer should be able to run indefinitely without crashing. In case of a crash, the tested input is written to the
+directory `out`.
+
+## OSS-Fuzz
+
+The library is further fuzz-tested 24/7 by Google's [OSS-Fuzz project](https://github.com/google/oss-fuzz). It uses
+the same `fuzzers` target as above and also relies on the `FUZZER_ENGINE` variable. See the used
+[build script](https://github.com/google/oss-fuzz/blob/master/projects/json/build.sh) for more information.
+
+In case the build at OSS-Fuzz fails, and issue will be created automatically.
```
falbrechtskirchinger, line 79: "... fails, an issue ..."

falbrechtskirchinger (changes requested):
In addition to my review comments, you should also mention the working directory for the commands.

tests/fuzzing.md
nlohmann: Done: f92e6d0979190347796329a66b76fd9b16ad22a5

tests/fuzzing.md
nlohmann: Done: f92e6d0979190347796329a66b76fd9b16ad22a5

tests/fuzzing.md
nlohmann: Done: f92e6d0979190347796329a66b76fd9b16ad22a5

nlohmann:
@falbrechtskirchinger Can I merge this? Or should we update the AFL wrapper with the one you used with AFL++?

falbrechtskirchinger:
> @falbrechtskirchinger Can I merge this? Or should we update the AFL wrapper with the one you used with AFL++?

Let me give it another read, then you can merge it. We can update the AFL wrapper some other time.

I'd like to understand how the make targets are used by OSS-fuzz (haven't looked at `build.sh` which I've seen linked in comments) and maybe move the fuzz testing into the CMake code since the more advanced AFL++ use requires building two fuzzer binaries and then running the fuzzer requires a different command as well. All of that would be easier to do in CMake (we're also already downloading the test date which could simplify and automate corpus generation).

falbrechtskirchinger (approved):
Looks good to me.

nlohmann:
Alright. I would like to avoid requiring OSS-Fuzz to use CMake. But the targets in the root `Makefile` can be removed/replaced.

PR #3487, opened 2022-05-11
Make the library signature more difficult to remove

falbrechtskirchinger:
Use compiler-specific mechanisms to prevent the removal of the identifying hash value and version number.

I've just stumbled over this while perusing the AFL++ source code and remembered this awkward, abused, unreachable exception.


falbrechtskirchinger:
In the end, I couldn't quite make it work without redefinition issues.

Issue #3490, opened 2022-05-12
UBSAN error while parsing UBJSON


nlohmann:
### Description

OSS-Fuzz reports a runtime error: load of value 16, which is not a valid value for type 'bool' when fuzzing with UBSAN.

- https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=47403
- https://oss-fuzz.com/testcase-detail/5120699125071872

### Reproduction steps

- Compile with UBSAN
- Parse input [clusterfuzz-testcase-minimized-parse_bjdata_fuzzer-5120699125071872.zip](https://github.com/nlohmann/json/files/8675069/clusterfuzz-testcase-minimized-parse_bjdata_fuzzer-5120699125071872.zip)

### Expected vs. actual results

Expected: parse error or valid JSON value returned

Actual: UBSAN runtime error

### Error messages

```Shell
[Environment] UBSAN_OPTIONS=print_stacktrace=1:silence_unsigned_overflow=1
+----------------------------------------Release Build Stacktrace----------------------------------------+
Command: /mnt/scratch0/clusterfuzz/resources/platform/linux/unshare -c -n /mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds_json_3dc7f07cae4ab217c21b70d40f93a3acccc6b431/revisions/parse_bjdata_fuzzer -rss_limit_mb=2560 -timeout=60 -runs=100 /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/crash-fb1efa0b5e5f3aa16f32b3d68449242cabca995e
Time ran: 0.09805822372436523
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3403573080
INFO: Loaded 1 modules   (5891 inline 8-bit counters): 5891 [0x5944d0, 0x595bd3),
INFO: Loaded 1 PC tables (5891 PCs): 5891 [0x530a68,0x547a98),
/mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds_json_3dc7f07cae4ab217c21b70d40f93a3acccc6b431/revisions/parse_bjdata_fuzzer: Running 1 inputs 100 time(s) each.
Running: /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/crash-fb1efa0b5e5f3aa16f32b3d68449242cabca995e
[...]

PR #3498, opened 2022-05-17
Add assertion for invariant in SAX-DOM parser

nlohmann:
Add assertions to catch misbehavior detected with #3490.


nlohmann:
I would merge this once #3500 is merged and OSS-Fuzz confirms the related issues as closed.

(If I would merge it now, OSS-Fuzz would likely just find the same issues as triggering the assertions.)

Issue #3492, opened 2022-05-12
ASAN error while parsing BJData (Heap-buffer-overflow READ 1)


nlohmann:
### Description

OSS-Fuzz reports a buffer overflow when fuzzing with ASAN

- https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=47391
- https://oss-fuzz.com/testcase-detail/5698044521218048

### Reproduction steps

- Compile with ASAN
- Parse input
[clusterfuzz-testcase-minimized-parse_bjdata_fuzzer-5698044521218048.zip](https://github.com/nlohmann/json/files/8675124/clusterfuzz-testcase-minimized-parse_bjdata_fuzzer-5698044521218048.zip)
 

### Expected vs. actual results

Expected: parse error or valid JSON value returned

Actual: ASAN runtime error

### Error messages

```Shell
+----------------------------------------Release Build Stacktrace----------------------------------------+
Command: /mnt/scratch0/clusterfuzz/resources/platform/linux/unshare -c -n /mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds_json_26b1464c0c18fac23c49bf26ed996090f90e682a/revisions/parse_bjdata_fuzzer -rss_limit_mb=2560 -timeout=60 -runs=100 /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/crash-9c5801955f5d26b496b0a3a159e11fc8baa926bf
Time ran: 0.3587803840637207
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 937550358
INFO: Loaded 1 modules   (6059 inline 8-bit counters): 6059 [0x694960, 0x69610b),
INFO: Loaded 1 PC tables (6059 PCs): 6059 [0x624770,0x63c220),
/mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds_json_26b1464c0c18fac23c49bf26ed996090f90e682a/revisions/parse_bjdata_fuzzer: Running 1 inputs 100 time(s) each.
Running: /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/crash-9c5801955f5d26b496b0a3a159e11fc8baa926bf
=================================================================
==101886==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000000140 at pc 0x000000576031 bp 0x7ffcb2e4ef80 sp 0x7ffcb2e4ef78
[...]

Issue #3491, opened 2022-05-12
UBSAN error while parsing BJData (Null-dereference)


nlohmann:
### Description

OSS-Fuzz reports a runtime error: member access within null pointer of type 'std::__tree_end_node<std::__tree_node_base<void *> *>' when fuzzing with UBSAN.

- https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=47406
- https://oss-fuzz.com/testcase-detail/6432243825901568

### Reproduction steps

- Compile with UBSAN
- Parse input
[clusterfuzz-testcase-minimized-parse_bjdata_fuzzer-6432243825901568.zip](https://github.com/nlohmann/json/files/8675101/clusterfuzz-testcase-minimized-parse_bjdata_fuzzer-6432243825901568.zip)
 

### Expected vs. actual results

Expected: parse error or valid JSON value returned

Actual: UBSAN runtime error

### Error messages

```Shell
[Environment] UBSAN_OPTIONS=print_stacktrace=1:silence_unsigned_overflow=1
+----------------------------------------Release Build Stacktrace----------------------------------------+
Command: /mnt/scratch0/clusterfuzz/resources/platform/linux/unshare -c -n /mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds_json_3dc7f07cae4ab217c21b70d40f93a3acccc6b431/revisions/parse_bjdata_fuzzer -rss_limit_mb=2560 -timeout=60 -runs=100 /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/6cd12eba683055cf14980d433e5156c264b292204f76005b9c062f0af87a5aa0
Time ran: 0.025225162506103516
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 529278620
INFO: Loaded 1 modules   (5891 inline 8-bit counters): 5891 [0x5944d0, 0x595bd3),
INFO: Loaded 1 PC tables (5891 PCs): 5891 [0x530a68,0x547a98),
/mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds_json_3dc7f07cae4ab217c21b70d40f93a3acccc6b431/revisions/parse_bjdata_fuzzer: Running 1 inputs 100 time(s) each.
Running: /mnt/scratch0/clusterfuzz/bot/inputs/fuzzer-testcases/6cd12eba683055cf14980d433e5156c264b292204f76005b9c062f0af87a5aa0
[...]

PR #3500, opened 2022-05-17
Discard optimized containers with negative counts in UBJSON/BJData (#3491,#3492,#3490)

fangq:
This may partially address the fuzzer errors reported in #3491, #3492 and #3490.

Specifically, the above failed fuzzer data appears to be triggered when an ND-array optimized header contains 0 in any of the dimensional vector elements. This causes partially written objects and array constructs.

Another issue this patch fixes is related to the implementation-spec mismatch discussed in https://github.com/nlohmann/json/issues/3492#issuecomment-1129318362

a negative count should result in an invalid UBJSON and BJData input.


tests/src/unit-bjdata.cpp
```diff
@@ -2518,25 +2520,11 @@ TEST_CASE("BJData")
...
-                else
-                {
-                    CHECK_THROWS_WITH_AS(_ = json::from_bjdata(v1), "[json.exception.out_of_range.408] excessive array size: 4294967281", json::out_of_range&);
-                    CHECK_THROWS_WITH_AS(_ = json::from_bjdata(v2), "[json.exception.out_of_range.408] excessive array size: 4294967282", json::out_of_range&);
-                    CHECK_THROWS_WITH_AS(_ = json::from_bjdata(v3), "[json.exception.out_of_range.408] excessive array size: 4294967272", json::out_of_range&);
-                    CHECK_THROWS_WITH_AS(_ = json::from_bjdata(v4), "[json.exception.out_of_range.408] excessive array size: 4294967287", json::out_of_range&);
-                }
-                CHECK_THROWS_WITH_AS(_ = json::from_bjdata(v5), "[json.exception.parse_error.110] parse error at byte 11: syntax error while parsing BJData number: unexpected end of input", json::parse_error&);
-
+                CHECK(json::from_bjdata(v1, true, false).is_discarded());
```
nlohmann, line 2523: Since all these inputs would be throwing unless you set `allow_exceptions=false`, it would be better to check these exceptions explicitly via `CHECK_THROWS_WITH_AS`.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -2120,6 +2120,15 @@ class binary_reader
                 }
                 if (!dim.empty())  // if ndarray, convert to an object in JData annotated array format
                 {
+                    for (auto i : dim)
```
nlohmann, line 2123: Please add a comment why this loop is needed.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -2016,7 +2016,7 @@ class binary_reader
             case 'i':
             {
                 std::int8_t number{};
-                if (JSON_HEDLEY_UNLIKELY(!get_number(input_format, number)))
+                if (JSON_HEDLEY_UNLIKELY(!get_number(input_format, number) || number < 0))
```
nlohmann, line 2019: In case of a negative size value - wouldn't it make more sense to create an exception right here with a proper diagnostics like "size values must be positive" or so?

nlohmann (changes requested):
I checked the fix with the inputs of issues #3491, #3492, and #3490, and none encountered the assertions I added in #3498. I only have some smaller remarks.

include/nlohmann/detail/input/binary_reader.hpp
fangq: done

nlohmann:
I let the fuzzer run on the now-merged branch and see no issues after letting it run for half an hour. 👍

nlohmann:
@fangq 

I think I found another issue:

In the binary reader in function `get_ubsize_size_value`, we have

```cpp
string_t key = "_ArraySize_";
if (JSON_HEDLEY_UNLIKELY(!sax->start_object(3) || !sax->key(key) || !sax->start_array(dim.size())))
{
    return false;
}
```

In an example input, I get the following SAX events:

```xml
<object size="3">
    <key key="_ArraySize_" />
    <array size="2">
        <number_integer val="2148925440" />
        <number_integer val="6701356247676222464" />
    </array>
```

However, the object that is started with `sax->start_object(3)` is never ended with a closing `end_object()` call. This yields issues down the road in the SAX-DOM parser, because, we when we call the `return sax->end_array();` line at the end of the code block, we are still inside the unclosed object we started before.

Here is the respective input: [crash.bjdata.zip](https://github.com/nlohmann/json/files/8720313/crash.bjdata.zip)

I will re-run the fuzzer with the enabled assertions from #3498 and see if I can find smaller examples.

nlohmann:
Here is a 14 byte example:

```
00000000: 5b5b 235b 5500 5b55 ff69 5d5d 5d5d       [[#[U.[U.i]]]]
```

which produces these SAX events:

```xml
<array>
    <object size="3">
        <key key="_ArraySize_" />
        <array size="2">
            <number_integer val="255" />
            <number_integer val="93" />
        </array>
        <array size="0">
        </array>
    </array>
```

The more I look at it: why is there a hard-coded `3` for the object size?

[minimized.bjdata.zip](https://github.com/nlohmann/json/files/8720388/minimized.bjdata.zip)

fangq:
@nlohmann, let me know if the patch if #3502 fixes the errors.

PR #3502, opened 2022-05-18
prevent ndarray dimension vector from recusive array, nlohmann/json#3500

fangq:
I believe the issue @nlohmann found in https://github.com/nlohmann/json/pull/3500#issuecomment-1130464847 is because the ndarray dimension vector parsing is too relax - permitting embedded arrays (which can result in unpredictable results).

I am adding an exception here to avoid such case. let me know if you still see errors.

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -1938,7 +1938,7 @@ class binary_reader
     {
         std::pair<std::size_t, char_int_type> size_and_type;
         size_t dimlen = 0;
-        bool is_ndarray = false;
+        bool is_ndarray = true;
 
         if (JSON_HEDLEY_UNLIKELY(!get_ubjson_size_type(size_and_type)))
         {
@@ -1994,7 +1994,6 @@ class binary_reader
     */
     bool get_ubjson_size_value(std::size_t& result, bool& is_ndarray, char_int_type prefix = 0)
     {
-        is_ndarray = false;
         if (prefix == 0)
         {
             prefix = get_ignore_noop();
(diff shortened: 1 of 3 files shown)


nlohmann:
This fix does not address the issue I described: you seem to open an object with fixed size of 3 which is never closed, yielding a problem in the SAX parser. I fuzzed this branch (together with the assertions from #3498) and got the following failing input:
[crash2.bjdata.zip](https://github.com/nlohmann/json/files/8723316/crash2.bjdata.zip)

This generates the following SAX events:

```xml
<array>
    <array>
        <object>
            <key key="" />
            <array>
                <number_integer val="27739" />
                <boolean val="true" />
                <object size="3">                   <!-- open an object with size 3 -->
                    <key key="_ArraySize_" />
                    <array size="2">
                        <number_integer val="6689" />
                        <number_integer val="0" />
                    </array>                        <!-- the first object entry -->
                    <array size="0">                <!-- ERROR: a value without key -->
                    </array>
                    <boolean val="true" />          <!-- ERROR: a value without key -->
                    <boolean val="true" />          <!-- ERROR: a value without key -->
                    <boolean val="true" />          <!-- ERROR: a value without key -->
                    <boolean val="true" />          <!-- ERROR: a value without key -->
                </array>                            <!-- ERROR: closing array (not object) -->
                <key key="" />
                <number_unsigned val="30069" />
                <parse_error id="71" token="<end of file>" />
```

nlohmann:
I moved the open issue to #3503. This PR fixes another bug in the BJData reader, so it should be merged.

nlohmann (approved):
Looks good to me.

Issue #3503, opened 2022-05-20
Binary reader for BJData creates incorrect SAX events


nlohmann:
The BJData reader seem to open an object with fixed size of 3 which is never closed, yielding a problem in the SAX parser. I fuzzed this branch (together with the assertions from #3498) and got the following failing input: [crash2.bjdata.zip](https://github.com/nlohmann/json/files/8723316/crash2.bjdata.zip)

This generates the following SAX events:

```xml
<array>
    <array>
        <object>
            <key key="��" />
            <array>
                <number_integer val="27739" />
                <boolean val="true" />
                <object size="3">                   <!-- open an object with size 3 -->
                    <key key="_ArraySize_" />
                    <array size="2">
                        <number_integer val="6689" />
                        <number_integer val="0" />
                    </array>                        <!-- the first object entry -->
                    <array size="0">                <!-- ERROR: a value without key -->
                    </array>
                    <boolean val="true" />          <!-- ERROR: a value without key -->
                    <boolean val="true" />          <!-- ERROR: a value without key -->
                    <boolean val="true" />          <!-- ERROR: a value without key -->
                    <boolean val="true" />          <!-- ERROR: a value without key -->
                </array>                            <!-- ERROR: closing array (not object) -->
                <key key="" />
                <number_unsigned val="30069" />
                <parse_error id="71" token="<end of file>" />
``` 

Again, for line

```cpp
if (JSON_HEDLEY_UNLIKELY(!sax->start_object(3) || !sax->key(key) || !sax->start_array(dim.size())))
```

we need to check

- Why is the size `3` hard-coded?
- Where to add the required `sax->end_object()` call?

_Originally posted by @nlohmann in https://github.com/nlohmann/json/issues/3502#issuecomment-1131176795_

nlohmann:
I just ran the fuzzer again and wanted to add another (shorter) example: [minimized-from-48f6d80fc879d78470bda5e99e4a1213d4d963a1.zip](https://github.com/nlohmann/json/files/8738181/minimized-from-48f6d80fc879d78470bda5e99e4a1213d4d963a1.zip)

```xml
<object>
    <key key="N" />
    <array>
        <object size="3">
            <key key="_ArraySize_" />
            <array size="2">
                <number_integer val="9216" />
                <number_integer val="792633534417207296" />
            </array>
            <array size="0">. <!-- expected a key here -->
            </array>
        </array>              <!-- expected a end_object() here -->
```

fangq:
[...]

PR #3505, opened 2022-05-22
Prevent ndarray size vector from recursive use, fix nlohmann/json#3503

fangq:
It seems that the broken ND array objects in #3503 were results of embedding arrays inside an ND-array size. In #3500, I added a flag to prevent recursive embedding but missed the `get_ubjson_size_type` function. now this is added. The test data throw errors as expected.

let me know if the fuzzer test still find something else after this patch.

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -2249,10 +2249,7 @@ class binary_reader
                     return sax->parse_error(chars_read, get_token_string(), parse_error::create(112, chars_read,
                                             exception_message(input_format, "ndarray can not be recursive", "size"), nullptr));
                 }
-                else
-                {
-                    result.second |= (1 << 8); // use bit 8 to indicate ndarray, all UBJSON and BJData markers should be ASCII letters
-                }
+                result.second |= (1 << 8); // use bit 8 to indicate ndarray, all UBJSON and BJData markers should be ASCII letters
             }
             return is_error;
         }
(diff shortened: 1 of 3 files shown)


nlohmann:
I'll check.

nlohmann:
No problems found so far.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -2240,6 +2244,11 @@ class binary_reader
             bool is_error = get_ubjson_size_value(result.first, is_ndarray);
             if (input_format == input_format_t::bjdata && is_ndarray)
             {
+                if (inside_ndarray)
+                {
+                    return sax->parse_error(chars_read, get_token_string(), parse_error::create(112, chars_read,
+                                            exception_message(input_format, "ndarray can not be recursive", "size"), nullptr));
```
nlohmann, line 2250: Please add a test that triggers this exception.

nlohmann (changes requested):

include/nlohmann/detail/input/binary_reader.hpp
fangq: done

nlohmann: The coverage is still not complete:

image

(see https://coveralls.io/builds/49342015/source?filename=include%2Fnlohmann%2Fdetail%2Finput%2Fbinary_reader.hpp#L2577)

fangq: my bad, did not check the report carefully. now it is added.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks a lot! I will merge this once AppVeyor is done.

PR #3517, opened 2022-06-03
Add assertion to converting constructor

### LINKED ISSUE #3425: Conversion from alt_json to json fails to compile
### Description

This is just a placeholder to have an issue to refer to in PR #3415. Will fill in the details after more testing at a later date.

### Reproduction steps

To do

To do

### Minimal code example

_No response_

### Error messages

_No response_

GCC 12.0.1

### Library version

trunk

### Validation

falbrechtskirchinger:
The converting `basic_json` constructor can inadvertently change the value type of its parameter. Assert that both `basic_json` values are of the same value type after conversion.

This is not a fix for #3425, but instead of surprising the user with strings converting to arrays, etc., those conversions will at least fail.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -916,6 +916,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
             default:            // LCOV_EXCL_LINE
                 JSON_ASSERT(false); // NOLINT(cert-dcl03-c,hicpp-static-assert,misc-static-assert) LCOV_EXCL_LINE
         }
+        JSON_ASSERT(m_type == val.type());
         set_parents();
         assert_invariant();
     }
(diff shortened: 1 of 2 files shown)


nlohmann (approved):
Looks good to me.

include/nlohmann/json.hpp
```diff
@@ -916,6 +916,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
             default:            // LCOV_EXCL_LINE
                 JSON_ASSERT(false); // NOLINT(cert-dcl03-c,hicpp-static-assert,misc-static-assert) LCOV_EXCL_LINE
         }
+        JSON_ASSERT(m_type == val.type());
```
gregmarr, line 919: In which case is it possible for this to not be true?

falbrechtskirchinger: See #3425.

gregmarr: Sorry that I wasn't clear, I meant which case in this switch statement, if you know.

falbrechtskirchinger: I've observed it in the `value_t::string` case. I can't rule it out for others. `to_json` doesn't enforce the destination value type and my attempt at injecting that information into the conversion process is failing right now.

nlohmann:
Please change the wording about fixing: GitHub does not understand the NOT and would close #3425 when merging.

include/nlohmann/json.hpp
gregmarr: Ah, because `other_string_t` is a different type than `string_t`.  Thanks.

falbrechtskirchinger:
@nlohmann Done.

Issue #2649, opened 2021-02-18
String type change breaks C++ type matching


sirzooro:
Usually when one of template parameters is changed, reference to new type cannot be implicitly converted to reference to old type. However for some reason this is not true for nlohmann::json. I tried to change std::string to my custom string class, and it turned out that references to both types are compatible for compiler. Here is example code which demonstrates this problem. It uses std::wstring instead of std::string:

```cpp
#include <json.hpp>

template <
    template <typename U, typename V, typename... Args> class ObjectType = std::map,
    template <typename U, typename... Args> class ArrayType              = std::vector,
    class StringType = std::wstring, class BooleanType = bool,
    class NumberIntegerType = std::int64_t, class NumberUnsignedType = std::uint64_t,
    class NumberFloatType = double, template <typename U> class AllocatorType = std::allocator,
    template <typename T, typename SFINAE = void> class JSONSerializer = nlohmann::adl_serializer>
using MyBasicJson = nlohmann::basic_json<
    ObjectType, ArrayType, StringType, BooleanType, NumberIntegerType, NumberUnsignedType,
    NumberFloatType, AllocatorType, JSONSerializer>;

using MyJson = MyBasicJson<>;

void load(const nlohmann::json& json);
//void load(const MyJson& json);

void test(MyJson* json)
{
    load(*json);
}
```
During compilation I should get error that parameter passed to load() has incorrect type. However this code compiles cleanly. I tried to use gcc 10.2, clang 10.0, both compiles this code without any complain. I use nlohmann::json 3.9.1. I compiled it on CentOS and Ubuntu using following command:

`g++ -c -o dupa.o test.cpp -O3 -Wall -Wextra -Werror -std=c++11 -I.`

I played a bit with this code trying to create minimum example. I found that when removed include and copied forward declarations from json.hpp to my file, g++ reported error as expected. So it looks that something is wrong in other part of json.hpp.

Initially I thought that this may be some gcc error, but clang also does not complain, so this is unlikely.

#### Which compiler and operating system are you using?

gcc 10.2 on CentOS7 (installed from RedHat SCL packages), clang 10.0 on Ubuntu Ubuntu 20.04.2 LTS

#### Which version of the library did you use?

- [x] latest release version 3.9.1
- [ ] other release - please state the version: ___
- [ ] the `develop` branch

gregmarr:
Discussed recently at https://github.com/nlohmann/json/discussions/2628#discussioncomment-350903

[...]

PR #3518, opened 2022-06-04
Fix basic_json converting constructor yielding wrong value type and mark constructor JSON_EXPLICIT

### LINKED ISSUE #3425: Conversion from alt_json to json fails to compile

falbrechtskirchinger:
This is an attempt to fix #3425.
The desired value type is communicated to the serializer by wrapping the source value in one of `string_type_wrapper`, `object_type_wrapper`, or `array_type_wrapper`.
It started fairly simple, but the ability to supply a custom serializer means this solution will only work with `nlohmann::adl_serializer` (and requires an indirection to compile with a custom serializer). The wrapper types could be made part of the public API to allow users of custom serializers to handle these conversions as well.
Additionally, the PR addresses aspects of #2649 (as discussed [here](https://github.com/nlohmann/json/discussions/2628#discussioncomment-350903)), by making the converting constructor explicit depending on `JSON_USE_IMPLICIT_CONVERSIONS`.

The converting `basic_json` constructor is kind of broken and with the current serializer API, I couldn't think of a better way to convey the destination type to the serializer. A well-designed solution should be an item for 4.0.

In the meantime, I'm rather uncertain as to how to proceed. At least this PR improves the situation for `adl_serializer` users, which is probably the vast majority of users anyway.

@nlohmann @gregmarr I'd appreciate some general feedback on the idea. The code isn't ready for a thorough review.

It also currently fails to build on MSVC due to the serializer's SFINAE template parameter, I'll look into that depending on feedback.

### DIFF
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -227,6 +227,16 @@ template<>
 struct external_constructor<value_t::object>
 {
     template<typename BasicJsonType>
+    static void construct(BasicJsonType& j)
+    {
+        j.m_value.destroy(j.m_type);
+        j.m_type = value_t::object;
+        j.m_value.object = j.template create<typename BasicJsonType::object_t>();
+        j.set_parents();
+        j.assert_invariant();
+    }
+
+    template<typename BasicJsonType>
     static void construct(BasicJsonType& j, const typename BasicJsonType::object_t& obj)
     {
         j.m_value.destroy(j.m_type);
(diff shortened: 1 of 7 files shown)


gregmarr:
@nlohmann Do we have a timeframe for a major version where we can address some of these issues that are being help up by backwards compatibility requirements?  Is it soon enough that we should just fix this properly instead of trying to fit it in within the current API?

shaischaudhry (approved):

PR #3520, opened 2022-06-06
Small documentation fixes

nlohmann:
- add a contributor
- update macOS/iOS versions shipped with the library
- fix documentation as reported in #3511

(CC @falbrechtskirchinger who had the last item on his list)


falbrechtskirchinger (approved):
Looks good to me.

PR #3528, opened 2022-06-10
Use new CI image

nlohmann:
- use CI image 2.4.0
- add new flags for GCC 13

### DIFF
--- a/README.md
+++ b/README.md
@@ -1110,8 +1110,8 @@ The following compilers are currently used in continuous integration at [AppVeyo
 | Clang 11.0.0 (11.0.0-2~ubuntu20.04.1)                                                                  | Ubuntu 20.04.3 LTS | GitHub Actions |
 | Clang 12.0.0 (12.0.0-3ubuntu1~20.04.3)                                                                 | Ubuntu 20.04.3 LTS | GitHub Actions |
 | Clang 13.0.1 (13.0.1-++20211015123032+cf15ccdeb6d5-1exp120211015003613.5)                              | Ubuntu 20.04.3 LTS | GitHub Actions |
-| Clang 14.0.0 (14.0.0-++20211221052852+55c71c9eac9b-1exp120211221172954.95)                             | Ubuntu 20.04.3 LTS | GitHub Actions |
-| Clang 15.0.0 (15.0.0-++20220403052648+896770c9a92e-1~exp1~20220403172744.209)                          | Ubuntu 20.04.3 LTS | GitHub Actions |
+| Clang 14.0.5-++20220603124341+2f0a69c32a4c-1~exp1~20220603124352.149                                   | Ubuntu 20.04.3 LTS | GitHub Actions |
+| Clang 15.0.0 (15.0.0-++20220530052901+b7d2b160c3ba-1~exp1~20220530172952.268)                          | Ubuntu 20.04.3 LTS | GitHub Actions |
 | GCC 4.8.5 (Ubuntu 4.8.5-4ubuntu2)                                                                      | Ubuntu 20.04.3 LTS | GitHub Actions |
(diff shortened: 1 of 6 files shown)


cmake/ci.cmake
```diff
@@ -390,6 +398,8 @@ set(GCC_CXXFLAGS
     -Wunused-result
     -Wunused-value
     -Wunused-variable
+    -Wuse-after-free
+    -Wuse-after-free=3
```
gregmarr, line 402: need both?

gregmarr: Okay, that this was generated makes me feel better.  :D

nlohmann: See https://github.com/nlohmann/gcc_flags

cmake/ci.cmake
```diff
@@ -343,6 +349,7 @@ set(GCC_CXXFLAGS
     -Wstrict-aliasing
     -Wstrict-aliasing=3
     -Wstrict-null-sentinel
+    -Wno-strict-overflow
```
falbrechtskirchinger, line 352: We just removed this one. When you fix the duplication issue, you should also add a blacklist.

nlohmann:
@falbrechtskirchinger PTAL - 32 bit support is now also enabled.

README.md
```diff
@@ -1123,7 +1123,8 @@ The following compilers are currently used in continuous integration at [AppVeyo
 | GCC 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04)                                                               | Ubuntu 20.04.3 LTS | GitHub Actions |
 | GCC 10.2.0 (Ubuntu 10.2.0-5ubuntu1~20.04)                                                              | Ubuntu 20.04.3 LTS | GitHub Actions |
 | GCC 11.1.0                                                                                             | Ubuntu (aarch64)   | Drone CI       |
-| GCC 12.0.0 20211219 (experimental)                                                                     | Ubuntu 20.04.3 LTS | GitHub Actions |
```
gregmarr, line 1126: Why drop GCC 12?

falbrechtskirchinger: This is always the latest snapshot build and GCC 12 hasn't made it into the r-toolchain PPA yet (https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test).

nlohmann: Exactly. It would be about time to have multiple images to install gcc 12 just from a recent Ubuntu, yet then I would lose the older Clangs…

falbrechtskirchinger:
Welp. I hoped you'd merge one of my 32bit PRs into this one first before I'd do a final review.

The addition of `readability-simplify-boolean-expr` triggers a bunch of errors on my system with a recent Clang 15 build. Specifically in `HEDLEY_(UN)LIKELY`.

nlohmann:
Oh, sorry. But `readability-simplify-boolean-expr` is excluded. In this PR.

falbrechtskirchinger:
> Oh, sorry. But `readability-simplify-boolean-expr` is excluded. In this PR.

Oh, good. I just quickly glanced at it and misinterpreted the plus sign from the diff. I'll see if there's anything else.

nlohmann:
I wished Clang-Tidy was more flexible and one could define a list of sources that you care about and treat warnings outside these files as out of scope. If I had a stupid Boolean tautology checked in "my" code, I would act on it. But not if it's the consequence of piling up calls from different libraries...

falbrechtskirchinger:
> I wished Clang-Tidy was more flexible and one could define a list of sources that you care about and treat warnings outside these files as out of scope. If I had a stupid Boolean tautology checked in "my" code, I would act on it. But not if it's the consequence of piling up calls from different libraries...

Agreed.

Everything else looks mostly good. The `bool()` cast could probably have been removed entirely. `std::conditional` doesn't do that, I'm sure.

README.md
gregmarr: Makes sense, thanks.

PR #3530, opened 2022-06-11
CI: Enable 32bit unit test (simplified)

falbrechtskirchinger:
Replaces #3529.

### DIFF
--- a/cmake/ci.cmake
+++ b/cmake/ci.cmake
@@ -416,7 +416,7 @@ set(GCC_CXXFLAGS
 add_custom_target(ci_test_gcc
     COMMAND CXX=${GCC_TOOL} CXXFLAGS="${GCC_CXXFLAGS}" ${CMAKE_COMMAND}
         -DCMAKE_BUILD_TYPE=Debug -GNinja
-        -DJSON_BuildTests=ON -DJSON_MultipleHeaders=ON
+        -DJSON_BuildTests=ON
         -S${PROJECT_SOURCE_DIR} -B${PROJECT_BINARY_DIR}/build_gcc
     COMMAND ${CMAKE_COMMAND} --build ${PROJECT_BINARY_DIR}/build_gcc
     COMMAND cd ${PROJECT_BINARY_DIR}/build_gcc && ${CMAKE_CTEST_COMMAND} --parallel ${N} --output-on-failure
@@ -426,7 +426,7 @@ add_custom_target(ci_test_gcc
 add_custom_target(ci_test_clang
     COMMAND CXX=${CLANG_TOOL} CXXFLAGS="${CLANG_CXXFLAGS}" ${CMAKE_COMMAND}
         -DCMAKE_BUILD_TYPE=Debug -GNinja
-        -DJSON_BuildTests=ON -DJSON_MultipleHeaders=ON
+        -DJSON_BuildTests=ON
         -S${PROJECT_SOURCE_DIR} -B${PROJECT_BINARY_DIR}/build_clang
     COMMAND ${CMAKE_COMMAND} --build ${PROJECT_BINARY_DIR}/build_clang
     COMMAND cd ${PROJECT_BINARY_DIR}/build_clang && ${CMAKE_CTEST_COMMAND} --parallel ${N} --output-on-failure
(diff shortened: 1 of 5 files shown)


cmake/ci.cmake
```diff
@@ -577,6 +584,13 @@ add_custom_target(ci_test_amalgamation
     COMMAND ${ASTYLE_TOOL} ${ASTYLE_FLAGS} ${INDENT_FILES}
     COMMAND cd ${PROJECT_SOURCE_DIR} && for FILE in `find . -name '*.orig'`\; do false \; done
 
+    COMMAND CXX=${GCC_TOOL} CXXFLAGS="${GCC_CXXFLAGS}" ${CMAKE_COMMAND}
```
nlohmann, line 587: Why did you add this to `ci_test_amalgamation`?

tests/CMakeLists.txt
```diff
@@ -124,34 +116,42 @@ message(STATUS "${msg}")
...
-    target_include_directories(test_main32 ${test_main_INCLUDE_DIRECTORIES})
-    target_link_libraries(test_main32 ${test_main_LINK_LIBRARIES})
-    target_link_options(test_main32 PUBLIC -m32)
-
-    json_test_add_test_for("src/unit-32bit.cpp" MAIN test_main32
-        CXX_STANDARDS ${test_cxx_standards} ${test_force})
+if(NOT "${json_32bit_test}" STREQUAL ONLY)
+    # test legacy comparison of discarded values
+    json_test_set_test_options(test-comparison_legacy
+        COMPILE_DEFINITIONS JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON=1
```
nlohmann, line 147: Why does this depend on `json_32bit_test`?

cmake/ci.cmake
falbrechtskirchinger: It seemed related. Amalgamation should be the only reason why the single-header could possibly behave differently than the multi-header version.

tests/CMakeLists.txt
falbrechtskirchinger: It depends on `JSON_32bitTest=ONLY`. No other tests should be built. Technically, that also applies to the CMake tests below.

cmake/ci.cmake
nlohmann: But the test only aims at the structure. The goal is not to detect different behavior, but is to make sure changes are applied to both the `include` and the `single_include` folder.

tests/CMakeLists.txt
nlohmann: This feels over-engineered to me. Why do we need the flag in the first place?

Wouldn't it be sufficient to check, inside the coverage target, to add the 32 bit unit test with `-m32` flag?

cmake/ci.cmake
falbrechtskirchinger: I guess I just wanted to avoid adding another job to the workflow.

Look forward to an issue (and PR) on `json-ci` soon. Preview:
> The `json-ci` image is about 14GB in size. The time spent initializing containers is 170 seconds (average over ~350 jobs) or almost 3 minutes. That amounts to a ridiculous 2.5 hours per workflow run.

nlohmann: Oh I so agree! We need to cut off the easy targets from the harder ones. I guess 2-3 Debian images could be sufficient to cover 90% of the compilers. Then we'd only need 1 image for the bleeding edge stuff and the specific tools.

tests/CMakeLists.txt
falbrechtskirchinger: It's not possible to "just" build a single unit test with `-m32` because that also needs the shared code (test_main) built with `-m32`. That's what I wanted to avoid in this "simplified" version. Instead, we *only* build the 32bit test. But I do have another idea. Just add a `return()` to the `CMakeLists.txt` and skip everything else in the file. I'll check if that's possible.

cmake/ci.cmake
falbrechtskirchinger: Do you want a test for the single-header at all?

nlohmann: Good point. So far, we tested it by chance somewhere, because we did not switch on the multiple-header version in all of the tests. Now that we have the multiple-header version as default, a dedicated test for the amalgamated version should be in order.

PR #3531, opened 2022-06-12
Fix iterator_input_adapter consuming more than it should

falbrechtskirchinger:
As identified by @imwhocodes, `iterator_input_adapter` consumes an additional character from the input during parsing. This PR based on #3442 by @imwhocodes fixes this. I'm finishing his work by incorporating requested changes and fixing CI issues.

At the moment, it's missing a unit test.

Replaces #3442.

### DIFF
--- a/include/nlohmann/detail/input/input_adapters.hpp
+++ b/include/nlohmann/detail/input/input_adapters.hpp
@@ -123,6 +123,7 @@ class input_stream_adapter
 };
 #endif  // JSON_NO_IO
 
+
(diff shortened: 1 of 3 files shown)


falbrechtskirchinger:
I've tried to write a unit test and the "fixed" `iterator_input_adapter` isn't consuming enough. Either my test is flawed or there was never a problem to begin with. :confused:

falbrechtskirchinger:
I've now added the unit test which fails to trigger the alleged issue.

PR #3532, opened 2022-06-12
CI: Enable 32bit unit test (3)

falbrechtskirchinger:
Third time's the charm.

Some of the decision-making logic for whether the 32bit test should be built has been moved into `test.cmake` and the code for constructing unit tests is now slightly less awkward. When only the 32bit unit test is requested, processing of the `CMakeLists.txt` is stopped early.

A single-header build has been added to CI.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -43,6 +43,16 @@ jobs:
       - name: build
         run: cmake --build build --target ${{ matrix.target }}
 
+  ci_test_single_header:
+    runs-on: ubuntu-latest
+    container: ghcr.io/nlohmann/json-ci:v2.4.0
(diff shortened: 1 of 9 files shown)


falbrechtskirchinger:
@fangq Can you help me out with a test case for this exception (possibly on 32bit only)? 
https://github.com/falbrechtskirchinger/json/blob/9ba90f5b70fa9c05978e5ad6c8222f7bf6aaf04f/include/nlohmann/detail/input/binary_reader.hpp#L2083-L2087

I only need a `std::vector` with the byte sequence. Thanks!

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks a lot!

PR #3534, opened 2022-06-13
Add to_json() for std::vector<bool>::reference

falbrechtskirchinger:
Avoids surprises as seen in #3533.

### DIFF
--- a/tests/src/unit-constructor1.cpp
+++ b/tests/src/unit-constructor1.cpp
@@ -474,6 +474,13 @@ TEST_CASE("constructors")
             json j(false);
             CHECK(j.type() == json::value_t::boolean);
         }
+
+        SECTION("from std::vector<bool>::refrence")
+        {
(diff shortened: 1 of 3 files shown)


falbrechtskirchinger:
Changed to `static_cast<>()`. Should be good now.

nlohmann (approved):
Looks good to me.

tests/src/unit-constructor1.cpp
```diff
@@ -474,6 +474,13 @@ TEST_CASE("constructors")
             json j(false);
             CHECK(j.type() == json::value_t::boolean);
         }
+
+        SECTION("from std::vector<bool>::refrence")
```
mdrost, line 478: typo in "refrence" :P

falbrechtskirchinger: Thanks, I'll add it to #3643.

PR #3544, opened 2022-06-18
Use `std::iterator_traits` to extract `iterator_category`

Mike-Leo-Smith:
In third-party STL implementations, `array_t::iterator` might be a pointer (e.g., `vector` in [EASTL](https://github.com/electronicarts/EASTL)) rather than a class, in which case directly using `array_t::iterator::iterator_category` is invalid. This commit fixes it with `std::iterator_traits`, which handles pointers correctly.


nlohmann (changes requested):
Please run `make amalgamate` to also add the changes to the single-header version, `single_include/json.hpp`.

Mike-Leo-Smith:
> Please run `make amalgamate` to also add the changes to the single-header version, `single_include/json.hpp`.

Thanks for the instructions. Should be good now.

falbrechtskirchinger (approved):
Looks good to me.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #3545, opened 2022-06-18
Doctest 2.4.9

nlohmann:
Update to Doctest 2.4.9

(Replaces #3228)

### DIFF
--- a/cmake/ci.cmake
+++ b/cmake/ci.cmake
@@ -928,6 +928,7 @@ add_custom_target(ci_icpc
         -DCMAKE_BUILD_TYPE=Debug -GNinja
         -DCMAKE_C_COMPILER=icc -DCMAKE_CXX_COMPILER=icpc
         -DJSON_BuildTests=ON -DJSON_FastTests=ON
+        -DCMAKE_CXX_FLAGS="-D_MSC_VER=0 -D_MSC_FULL_VER=0"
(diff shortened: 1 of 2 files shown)


nlohmann:
There is an issue with IPCC; reported upstream: https://github.com/doctest/doctest/issues/664

cmake/ci.cmake
```diff
@@ -928,6 +928,7 @@ add_custom_target(ci_icpc
         -DCMAKE_BUILD_TYPE=Debug -GNinja
         -DCMAKE_C_COMPILER=icc -DCMAKE_CXX_COMPILER=icpc
         -DJSON_BuildTests=ON -DJSON_FastTests=ON
+        -DCMAKE_CXX_FLAGS="-D_MSC_VER=0 -D_MSC_FULL_VER=0"
```
falbrechtskirchinger, line 931:
```suggestion
        -DCMAKE_CXX_FLAGS="-D_MSC_VER=0;-D_MSC_FULL_VER=0"
```

nlohmann: Thanks for the hint. But since this was a only a hack in the first place, I'd undo it and wait for a fix from upstream.

falbrechtskirchinger: Just a reminder to use semicolons when dealing with CMake lists in the future. :-)

nlohmann: This was not the first one, and I fear it was also not the last one :-)

shaischaudhry (approved):

PR #3546, opened 2022-06-18
Use REUSE framework

nlohmann:
See https://reuse.software

- Added license headers to relevant source files.
- Added CI step to check REUSE compliance.


nlohmann:
> Is it expected that the single include now has all the individual license blocks in it?

I guess yes. It’s not nice, but accurate.

gregmarr:
Those files in `LICENSES/` just look like templates, not specific licenses for specific files.

falbrechtskirchinger:
This seems to be the way to handle this today. Keep the specific copyright information with SPDX identifier in the source and bundle the license "templates".

gregmarr:
After digging further, it looks like the licenses are also in the source files in those directories, so no separate license file is needed.

nlohmann:
The idea of REUSE is to make the project a bit more machine-readable. You can call `pipx run reuse spdx` to get a bill of materials that contains, for each file, the license and copyright.

I removed all license files from the project, because REUSE only requires one copy in the `LICENSES` folder for every used license - all things that would be filled in there are exactly the copyrights we have in the files. I kept the root `LICENSE.MIT` to help GitHub understand that the main part of the project is MIT-licensed. If I have to guess, GitHub will eventually also overwork their license support.

include/nlohmann/thirdparty/hedley/hedley.hpp
```diff
@@ -1,15 +1,16 @@
...
 
+//     __ _____ _____ _____
+//  __|  |   __|     |   | |  JSON for Modern C++
+// |  |  |__   |  |  | | | |  version 3.10.5
+// |_____|_____|_____|_|___|  https://github.com/nlohmann/json
+//
+// SPDX-FileCopyrightText: 2013-2022 Niels Lohmann <https://nlohmann.me>
+// SPDX-FileCopyrightText: 2016-2021 Evan Nemerson <evan@nemerson.com>
+// SPDX-License-Identifier: MIT
+
```
falbrechtskirchinger, line 11: This file seems to be under CC0-1.0 and (C) Evan Nemerson. The only change is the rename from `HEDLEY_*` to `JSON_HEDLEY_*`?

include/nlohmann/thirdparty/hedley/hedley_undef.hpp
```diff
@@ -1,3 +1,11 @@
+//     __ _____ _____ _____
+//  __|  |   __|     |   | |  JSON for Modern C++
+// |  |  |__   |  |  | | | |  version 3.10.5
+// |_____|_____|_____|_|___|  https://github.com/nlohmann/json
+//
+// SPDX-FileCopyrightText: 2013-2022 Niels Lohmann <https://nlohmann.me>
+// SPDX-License-Identifier: MIT
+
```
falbrechtskirchinger, line 8: Same as `hedley.hpp`.

.github/workflows/ubuntu.yml
```diff
@@ -134,3 +134,14 @@ jobs:
         run: |
           . /opt/intel/oneapi/setvars.sh
           cmake --build build --target ci_icpc
+
+
```
falbrechtskirchinger, line 138: Nitpick: extraneous new line.
```suggestion

```

include/nlohmann/thirdparty/hedley/hedley.hpp
nlohmann: Yes, the only change is the renaming. The result is licensed as MIT to avoid confusion.

falbrechtskirchinger: IANAL, but …

The re-licensing seems fine. Are you sure the copyright assignments are proper, though?

Evan has disclaimed his copyright, which in the US, I believe, he can do. Other jurisdictions (i.e., Germany) do not recognize this. So, in this case, CC0 applies and Evan is the copyright holder under german law.

Under US copyright law, adding a prefix to some names isn't considered transformative, i.e., doesn't "add something new for a different purpose or of a different character". I imagine german copyright law equally requires something more for a work to be copyrightable.

Copyright in an open-source project seems like such a nightmare, but for this file at least, it seems pretty clear.

I'll stop now. :-)

PR #3548, opened 2022-06-20
Add unit test to make sure iterator_input_adapter advances iterators correctly

falbrechtskirchinger:
When parsing a string containing two JSON values using iterators, after parsing, `iterator_input_adapter` should have advanced to the first character of the second value.

The current API doesn't return the `first` iterator after parsing. To achieve the same effect a `proxy_iterator` is added that advances the iterator passed by reference, even if the proxy is moved.

### DIFF
--- a/tests/src/unit-deserialization.cpp
+++ b/tests/src/unit-deserialization.cpp
@@ -33,6 +33,7 @@ SOFTWARE.
 using nlohmann::json;
 
 #include <iostream>
+#include <iterator>
 #include <sstream>
 #include <valarray>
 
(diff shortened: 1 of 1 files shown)


falbrechtskirchinger:
@nlohmann @gregmarr This PR is a proactive step to make sure the iterator position after parsing reflects expectations [1]. I hope the parse API can be changed or amended in the future to return the iterator to where parsing stopped. This is in clear conflict with #3442, which aims to change this to point to the last processed character.

[1] To be fair, those are my expectations but I believe they're reasonable.

falbrechtskirchinger:
@nlohmann Ping. Do you need me to do anything here?

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks for reminding me.

PR #3550, opened 2022-06-21
Use DOCTEST_* compiler macros and suppress pragmas warning

falbrechtskirchinger:
Use `DOCTEST_*` macros in place of predefined compiler macros for compiler detection and version checks.

The suppression of warning `-Wrange-loop-construct` in `unit-items.cpp` causes GCC<11 to warn about pragmas.

### DIFF
--- a/tests/src/unit-items.cpp
+++ b/tests/src/unit-items.cpp
@@ -34,7 +34,9 @@ using nlohmann::json;
 
 // This test suite uses range for loops where values are copied. This is inefficient in usual code, but required to achieve 100% coverage.
 DOCTEST_GCC_SUPPRESS_WARNING_PUSH
-DOCTEST_GCC_SUPPRESS_WARNING("-Wrange-loop-construct")
+#if DOCTEST_GCC >= DOCTEST_COMPILER(11, 0, 0)
+    DOCTEST_GCC_SUPPRESS_WARNING("-Wrange-loop-construct")
+#endif
 DOCTEST_CLANG_SUPPRESS_WARNING_PUSH
 DOCTEST_CLANG_SUPPRESS_WARNING("-Wrange-loop-construct")
 
--- a/tests/src/unit-regression2.cpp
+++ b/tests/src/unit-regression2.cpp
@@ -803,7 +803,7 @@ TEST_CASE("regression tests 2")
         const auto j_path = j.get<nlohmann::detail::std_fs::path>();
         CHECK(j_path == text_path);
 
-#if defined(__clang__) || ((defined(__GNUC__) && !defined(__INTEL_COMPILER)) && (__GNUC__ > 8 || (__GNUC__ == 8 && __GNUC_MINOR__ >= 4)))
+#if DOCTEST_CLANG || DOCTEST_GCC >= DOCTEST_COMPILER(8, 4, 0)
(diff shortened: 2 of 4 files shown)


falbrechtskirchinger:
For reference:
Here's a g++10 run before this PR: https://github.com/nlohmann/json/runs/6973062108?check_suite_focus=true#step:5:92
With the patch, it builds cleanly: https://github.com/nlohmann/json/runs/6978632144?check_suite_focus=true#step:5:89

tests/CMakeLists.txt
```diff
@@ -86,6 +86,10 @@ elseif (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     #json_test_set_test_options(test-disabled_exceptions COMPILE_DEFINITIONS _HAS_EXCEPTIONS=0 COMPILE_OPTIONS /EH)
 endif()
 
+# Suppress GCC<11 warning about unknown pragmas
+json_test_set_test_options(test-items COMPILE_OPTIONS
```
nlohmann, line 90: Just to get it right: `-Wno-pragma` cannot be set by a pragma inside the sources?

falbrechtskirchinger: It could be done in the source but this option seemed less clumsy. Let me try to do it in the source and see what it takes to compare.

falbrechtskirchinger: We don't use `HEDLEY_*` because those aren't exported. But because of the doctest 2.4.9 issue I finally understand how doctest's compiler version macros work. We really should replace `#if defined(__GNUC__)` etc. with those.

It's a lot cleaner. I'll rename this PR and do it now all of this if that's OK with you?

nlohmann: Yes, that would be great. Sorry for the confusion - I know we're not using Hedley - with Hedley-like I meant exactly that kind of abstraction.

tests/src/unit-items.cpp
```diff
@@ -34,7 +34,9 @@ using nlohmann::json;
 
 // This test suite uses range for loops where values are copied. This is inefficient in usual code, but required to achieve 100% coverage.
 DOCTEST_GCC_SUPPRESS_WARNING_PUSH
+#if DOCTEST_GCC && DOCTEST_GCC >= DOCTEST_COMPILER(11, 0, 0)
```
falbrechtskirchinger, line 37: `DOCTEST_GCC && ` seems redundant.

tests/src/unit-regression2.cpp
```diff
@@ -803,7 +803,7 @@ TEST_CASE("regression tests 2")
         const auto j_path = j.get<nlohmann::detail::std_fs::path>();
         CHECK(j_path == text_path);
 
-#if defined(__clang__) || ((defined(__GNUC__) && !defined(__INTEL_COMPILER)) && (__GNUC__ > 8 || (__GNUC__ == 8 && __GNUC_MINOR__ >= 4)))
+#if DOCTEST_CLANG || DOCTEST_GCC >= DOCTEST_COMPILER(8, 4, 0)
```
falbrechtskirchinger, line 806: `!defined(__INTEL_COMPILER)` is included in the `DOCTEST_GCC` definition in case anyone's wondering.

nlohmann:
Thanks!

PR #3557, opened 2022-06-27
Make certain usage patterns more prominent in the README

jez:
We use this library extensively where I work. After pairing with many
teammates, I've learned that one common stumbling point when using this
library is how to first get a JSON value in the first place (because
once they have a variable of type `json` in hand, the API is intuitive
and mostly does what you expect).

With that in mind, I've added two subsections to the top of the Examples
section: how to read JSON from a file, and how to hardcode a JSON
literal.

I understand that these are already documented elsewhere in the Examples
section, and so these new subsections are technically redundant. But to
defend this choice:

- Redundancy in docs is actually good, because not everyone consumes
  docs in the same way or in the same order.

- Having these things called out explicitly in isolation encourages
  people to get something working faster. In later sections, these
  examples are presented alongside many other options. Having to choose
  between alternatives gets in the way of quickly prototyping something.

If you have further suggestions for how to improve or modify these docs
I'm open to them. The changes included here would really make a huge
improvement in the productivity of newer members of my team using this
library.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,6 @@
   - [Read JSON from a file](#read-json-from-a-file)
   - [Creating `json` objects from JSON literals](#creating-json-objects-from-json-literals)
   - [Reading and writing individual object keys](#reading-and-writing-individual-object-keys)
-  - [JSON as first-class data type](#json-as-first-class-data-type)
   - [Serialization / Deserialization](#serialization--deserialization)
   - [STL-like access](#stl-like-access)
   - [Conversion from STL containers](#conversion-from-stl-containers)
--- a/README.md
+++ b/README.md
@@ -162,7 +162,9 @@ json ex3 = {
 
 ### JSON as first-class data type
 
-Assume you want to create this JSON object by individually setting each key-value pair:
+Here are some examples to give you an idea how to use the class.
+
+Assume you want to create the JSON object
 
 ```json
 {
--- a/README.md
+++ b/README.md
@@ -23,6 +23,9 @@
 - [Sponsors](#sponsors)
 - [Support](#support) ([documentation](https://json.nlohmann.me), [FAQ](https://json.nlohmann.me/home/faq/), [discussions](https://github.com/nlohmann/json/discussions), [API](https://json.nlohmann.me/api/basic_json/), [bug issues](https://github.com/nlohmann/json/issues))
 - [Examples](#examples)
+  - [Read JSON from a file](#read-json-from-a-file)
+  - [Creating `json` objects from JSON literals](#creating-json-objects-from-json-literals)
+  - [Reading and writing individual object keys](#reading-and-writing-individual-object-keys)
   - [JSON as first-class data type](#json-as-first-class-data-type)
   - [Serialization / Deserialization](#serialization--deserialization)
   - [STL-like access](#stl-like-access)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,6 +23,9 @@
 - [Sponsors](#sponsors)
 - [Support](#support) ([documentation](https://json.nlohmann.me), [FAQ](https://json.nlohmann.me/home/faq/), [discussions](https://github.com/nlohmann/json/discussions), [API](https://json.nlohmann.me/api/basic_json/), [bug issues](https://github.com/nlohmann/json/issues))
 - [Examples](#examples)
+  - [Read JSON from a file](#read-json-from-a-file)
+  - [Creating `json` objects from JSON literals](#creating-json-objects-from-json-literals)
+  - [Reading and writing individual object keys](#reading-and-writing-individual-object-keys)
   - [JSON as first-class data type](#json-as-first-class-data-type)
```
nlohmann, line 29: This section has disappeared.

nlohmann (changes requested):
Looks goo

README.md
jez: good catch, thank you!

nlohmann: I think it was good having a section titled "JSON as first-class data type" - so instead of removing it altogether I think it should remain.

README.md
```diff
@@ -97,13 +99,70 @@ There is also a [**docset**](https://github.com/Kapeli/Dash-User-Contributions/t
...
+)"_json;
+
+// Using initializer lists
+json ex3 = {
+  {"happy", true},
+  {"pi", 3.141},
+};
+```
+
+### Reading and writing individual object keys
```
jez, line 163:
```suggestion
### JSON as first-class data type
```

README.md
```diff
@@ -23,7 +23,9 @@
 - [Sponsors](#sponsors)
 - [Support](#support) ([documentation](https://json.nlohmann.me), [FAQ](https://json.nlohmann.me/home/faq/), [discussions](https://github.com/nlohmann/json/discussions), [API](https://json.nlohmann.me/api/basic_json/), [bug issues](https://github.com/nlohmann/json/issues))
 - [Examples](#examples)
-  - [JSON as first-class data type](#json-as-first-class-data-type)
+  - [Read JSON from a file](#read-json-from-a-file)
+  - [Creating `json` objects from JSON literals](#creating-json-objects-from-json-literals)
+  - [Reading and writing individual object keys](#reading-and-writing-individual-object-keys)
```
jez, line 28:
```suggestion
  - [JSON as first-class data type](#json-as-first-class-data-type)
```

README.md
```diff
@@ -97,13 +99,70 @@ There is also a [**docset**](https://github.com/Kapeli/Dash-User-Contributions/t
...
+json ex3 = {
+  {"happy", true},
+  {"pi", 3.141},
+};
+```
+
+### JSON as first-class data type
 
-Assume you want to create the JSON object
+Assume you want to create this JSON object by individually setting each key-value pair:
```
jez, line 165:
```suggestion
Here are some examples to give you an idea how to use the class.

Assume you want to create the JSON object
```

README.md
jez: I have put the old section back. If you have a moment, I'd love another review.

nlohmann (approved):
Looks good to me.

PR #3581, opened 2022-07-15
add patch_inplace function

wolfv:
I am trying to `patch` a >100Mb JSON file and it turns out that the copying / destruction of JSON objects takes quite long.

Not allocating a new JSON for each patch helps a lot in terms of speed (>100x improvement).

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/docs/examples/patch_inplace.cpp
+++ b/docs/examples/patch_inplace.cpp
@@ -1,0 +1,34 @@
+#include <iostream>
+#include <iomanip>
+#include <nlohmann/json.hpp>
+
+using json = nlohmann::json;
+
+int main()
+{
+    // the original document
+    json doc = R"(
+        {
+          "baz": "qux",
+          "foo": "bar"
+        }
+    )"_json;
+
+    // the patch
+    json patch = R"(
+        [
+          { "op": "replace", "path": "/baz", "value": "boo" },
+          { "op": "add", "path": "/hello", "value": ["world"] },
+          { "op": "remove", "path": "/foo"}
+        ]
+    )"_json;
+
+    // output original document
+    std::cout << "Before\n" << std::setw(4) << doc << std::endl;
+
+    // apply the patch
+    doc.patch_inplace(patch);
+
+    // output patched document
+    std::cout << "After\n" << std::setw(4) << doc << std::endl;
+}
--- a/docs/examples/patch_inplace.output
+++ b/docs/examples/patch_inplace.output
@@ -1,0 +1,13 @@
+Before
+{
+    "baz": "qux",
+    "foo": "bar"
+}
+
+After
+{
+    "baz": "boo",
+    "hello": [
+        "world"
+    ]
+}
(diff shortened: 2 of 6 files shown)


falbrechtskirchinger:
A couple of things:
* Please run `make amalgamate` (see [Contribution Guidelines](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md#files-to-change)).
* I think `patch_inplace` would be the better name, so it appears next to `patch` when sorted.
* The documentation needs to be updated. I suggest you move `docs/mkdocs/docs/api/basic_json/patch.md` to `patch_inplace.md` and add a new `patch.md`, explaining `patch()` in terms of `patch_inplace()` and linking to `patch_inplace.md` for details. Example files need to be duplicated and adjusted.

(Separately, we should address the quite frequent failure of MinGW in CI. Can we rehost the MinGW package without violating the license?)

wolfv:
@falbrechtskirchinger thanks for the notes. What do you think about keeping docs in a single document? Otherwise there would be quite some duplication.

This really drastically improves the performance for our use case as copying and destroying a 100Mb json file is _really_ slow (on the tune of ~1 second per patch, vs a couple nanoseconds without the copy).

I wonder if the destruction could be optimized in #3583

falbrechtskirchinger:
> @falbrechtskirchinger thanks for the notes. What do you think about keeping docs in a single document? Otherwise there would be quite some duplication.

@wolfv I don't think there's precedent for similarly named functions being documented in one file and while I agree with you about the duplication, @nlohmann will probably want a dedicated `patch_inplace.md` file. Let's wait and see.

You're at least missing an entry in `docs/mkdocs/mkdocs.yml`.
```yaml
- 'patch_inplace': api/basic_json/patch.md
```

Also, can you please prepare the examples in `docs/examples` and include them once a decision has been made as to where `patch_inplace()` should be documented?
Basically, copy `patch.cpp` -> `patch_inplace.cpp` and `patch.output` -> `patch_inplace.output` and modify `patch_inplace.cpp`:

```cpp
    // output original document
    std::cout << std::setw(4) << doc << "\n\n";

    // apply the patch inplace
    doc.patch_inplace(patch);

    // output patched document
    std::court << std::setw(4) << doc << std::endl;
```

Did you run `make amalgamate`? You need to commit the generated `single_include/nlohmann/json.hpp` file as well.

nlohmann:
> @wolfv I don't think there's precedent for similarly named functions being documented in one file and while I agree with you about the duplication, @nlohmann will probably want a dedicated `patch_inplace.md` file. Let's wait and see.

Yes, please do so.

nlohmann (changes requested):

wolfv:
Thanks for the review! I applied your suggestions.

nlohmann (approved):
Looks good to me.

falbrechtskirchinger (approved):
Looks good to me.

nlohmann:
Note to me:

- [ ] Add new documentation to mkdocs.

wolfv:
@nlohmann thanks for approving! Is the remaining task something I should do?

nlohmann:
Thanks for your work!

> @nlohmann thanks for approving! Is the remaining task something I should do?

It's basically checking that the new Markdown file is added to docs/mkdocs/docs/api/basic_json/index.md, docs/mkdocs/mkdocs.yml, and docs/docset/docSet.sql. I'll take care of it.

Issue #3584, opened 2022-07-17
ERROR: ThreadSanitizer: SEGV on unknown address


abaelhe:
### Description

cd ~/Downloads/json-develop/build/tests/Debug
for x in ./test-* ; do ./$x 1> /dev/null  || echo $x ; done

./test-cbor_cpp11
==36891==ERROR: ThreadSanitizer: requested allocation size 0xffffffffffffffff exceeds maximum supported size of 0x10000000000
    #0 operator new(unsigned long) <null> (libclang_rt.tsan_osx_dynamic.dylib:x86_64h+0x7596b)
    #1 void* std::__1::__libcpp_operator_new<unsigned long>(unsigned long) new:235 (test-cbor_cpp11:x86_64+0x10004abc4)

==36891==HINT: if you don't care about these errors you may set allocator_may_return_null=1
SUMMARY: ThreadSanitizer: allocation-size-too-big (libclang_rt.tsan_osx_dynamic.dylib:x86_64h+0x7596b) in operator new(unsigned long)
Abort trap: 6
./test-cbor_cpp11

./test-testsuites_cpp11

ThreadSanitizer:DEADLYSIGNAL
==36941==ERROR: ThreadSanitizer: SEGV on unknown address 0x60000005d000 (pc 0x7ff80f562adb bp 0x7ff7b5762f10 sp 0x7ff7b5762f00 T6880876)
==36941==The signal is caused by a READ memory access.
    #0 flockfile <null>:197228464 (libsystem_c.dylib:x86_64+0x12ada)
    #1 fgetc <null>:197228464 (libsystem_c.dylib:x86_64+0x3382e)
    #2 nlohmann::detail::file_input_adapter::get_character() input_adapters.hpp:56 (test-testsuites_cpp11:x86_64+0x100044ca7)
    #3 nlohmann::detail::lexer<nlohmann::basic_json<std::__1::map, std::__1::vector, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, bool, long long, unsigned long long, double, std::__1::allocator, nlohmann::adl_serializer, std::__1::vector<unsigned char, std::__1::allocator<unsigned char> > >, nlohmann::detail::file_input_adapter>::get() lexer.hpp:1341 (test-testsuites_cpp11:x86_64+0x100044b0b)
[...]

PR #3593, opened 2022-07-20
Add assertion if nullptr is passed to parse function

nlohmann:
Addresses #3584 by adding an assertion and documenting the preconditions.

### DIFF
--- a/docs/mkdocs/docs/api/basic_json/accept.md
+++ b/docs/mkdocs/docs/api/basic_json/accept.md
@@ -29,7 +29,7 @@ Unlike the [`parse`](parse.md) function, this function neither throws an excepti
 :   A compatible input, for instance:
     
     - an `std::istream` object
-    - a `FILE` pointer
+    - a `FILE` pointer (must not be null)
     - a C-style array of characters
     - a pointer to a null-terminated string of single byte characters
     - a `std::string`
@@ -64,10 +64,6 @@ Whether the input is valid JSON.
 
 Strong guarantee: if an exception is thrown, there are no changes in the JSON value.
 
-## Exceptions
-
-Throws [`parse_error.116`](../../home/exceptions.md#jsonexceptionparse_error116) if passed `#!cpp FILE` pointer is `#!cpp nullptr`.
-
 ## Complexity
 
 Linear in the length of the input. The parser is a predictive LL(1) parser.
@@ -75,6 +71,11 @@ Linear in the length of the input. The parser is a predictive LL(1) parser.
 ## Notes
 
 (1) A UTF-8 byte order mark is silently ignored.
+
+!!! danger "Runtime assertion"
+
+    The precondition that a passed `#!cpp FILE` pointer must not be null is enforced with a
+    [runtime assertion](../../features/assertions.md).
 
 ## Examples
 
--- a/docs/mkdocs/docs/api/basic_json/accept.md
+++ b/docs/mkdocs/docs/api/basic_json/accept.md
@@ -64,6 +64,10 @@ Whether the input is valid JSON.
 
 Strong guarantee: if an exception is thrown, there are no changes in the JSON value.
 
+## Exceptions
+
+Throws [`parse_error.116`](../../home/exceptions.md#jsonexceptionparse_error116) if passed `#!cpp FILE` pointer is `#!cpp nullptr`.
+
 ## Complexity
 
 Linear in the length of the input. The parser is a predictive LL(1) parser.
--- a/docs/mkdocs/docs/api/basic_json/parse.md
+++ b/docs/mkdocs/docs/api/basic_json/parse.md
@@ -28,7 +28,7 @@ static basic_json parse(IteratorType first, IteratorType last,
 :   A compatible input, for instance:
     
     - an `std::istream` object
-    - a `FILE` pointer (will throw [json.exception.parse_error.116](../../home/exceptions.md#jsonexceptionparse_error116) if passed pointer is `#!cpp nullptr`)
+    - a `FILE` pointer
     - a C-style array of characters
     - a pointer to a null-terminated string of single byte characters
     - a `std::string`
@@ -70,6 +70,14 @@ Deserialized JSON value; in case of a parse error and `allow_exceptions` set to 
 ## Exception safety
 
 Strong guarantee: if an exception is thrown, there are no changes in the JSON value.
+
+## Exceptions
+
+- Throws [`parse_error.101`](../../home/exceptions.md#jsonexceptionparse_error101) in case of an unexpected token.
+- Throws [`parse_error.102`](../../home/exceptions.md#jsonexceptionparse_error102) if to_unicode fails or surrogate
+  error.
+- Throws [`parse_error.103`](../../home/exceptions.md#jsonexceptionparse_error103) if to_unicode fails.
+- Throws [`parse_error.116`](../../home/exceptions.md#jsonexceptionparse_error116) if passed `#!cpp FILE` pointer is `#!cpp nullptr`.
 
 ## Complexity
 
--- a/docs/mkdocs/docs/api/basic_json/parse.md
+++ b/docs/mkdocs/docs/api/basic_json/parse.md
@@ -28,7 +28,7 @@ static basic_json parse(IteratorType first, IteratorType last,
 :   A compatible input, for instance:
     
     - an `std::istream` object
-    - a `FILE` pointer
+    - a `FILE` pointer (must not be null)
     - a C-style array of characters
     - a pointer to a null-terminated string of single byte characters
     - a `std::string`
@@ -77,7 +77,6 @@ Strong guarantee: if an exception is thrown, there are no changes in the JSON va
 - Throws [`parse_error.102`](../../home/exceptions.md#jsonexceptionparse_error102) if to_unicode fails or surrogate
   error.
 - Throws [`parse_error.103`](../../home/exceptions.md#jsonexceptionparse_error103) if to_unicode fails.
-- Throws [`parse_error.116`](../../home/exceptions.md#jsonexceptionparse_error116) if passed `#!cpp FILE` pointer is `#!cpp nullptr`.
 
 ## Complexity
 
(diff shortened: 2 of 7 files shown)


docs/mkdocs/docs/api/basic_json/parse.md
```diff
@@ -28,7 +28,7 @@ static basic_json parse(IteratorType first, IteratorType last,
 :   A compatible input, for instance:
     
     - an `std::istream` object
-    - a `FILE` pointer
+    - a `FILE` pointer (will throw [json.exception.parse_error.116](../../home/exceptions.md#jsonexceptionparse_error116) if passed pointer is `#!cpp nullptr`)
```
falbrechtskirchinger, line 31: Should this go into an "Exception" section?

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -48,9 +50,14 @@ class file_input_adapter
...
 
     JSON_HEDLEY_NON_NULL(2)
-    explicit file_input_adapter(std::FILE* f) noexcept
+    explicit file_input_adapter(std::FILE* f)
         : m_file(f)
-    {}
+    {
+        if (m_file == nullptr)
+        {
+            JSON_THROW(parse_error::create(116, 0, detail::concat("input file is invalid: ", reinterpret_cast<const char*>(std::strerror(errno))), nullptr));
```
falbrechtskirchinger, line 58: I would not rely on `errno` here. We don't know if another function that sets `errno` was called after `fopen`.

docs/mkdocs/docs/api/basic_json/parse.md
nlohmann: Done.

include/nlohmann/detail/input/input_adapters.hpp
nlohmann: Right.

tests/src/unit-deserialization.cpp
```diff
@@ -332,6 +332,14 @@ TEST_CASE("deserialization")
         {
             CHECK_THROWS_WITH_AS("[\"foo\",1,2,3,false,{\"one\":1}"_json, "[json.exception.parse_error.101] parse error at line 1, column 29: syntax error while parsing array - unexpected end of input; expected ']'", json::parse_error&);
         }
+
+        SECTION("FILE*")
+        {
+            std::FILE* f = std::fopen("nonexisting_file", "r"); // NOLINT(cppcoreguidelines-owning-memory)
```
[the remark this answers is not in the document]
nlohmann: Now that we don't use `errno` any longer, you're right.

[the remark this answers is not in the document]
gregmarr: I didn't even see the `errno` version until I read some more of the email thread.  :)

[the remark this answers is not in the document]
nlohmann: Now UBSAN complains, because the function is marked `NONNULL`: https://github.com/nlohmann/json/runs/7433214543?check_suite_focus=true#step:5:235

I'll revert.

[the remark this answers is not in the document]
gregmarr: I suppose that's a good thing that it's marked that way.  :D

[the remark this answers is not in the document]
nlohmann: Yes - it's been a while since I saw a tool complain about the violation of such an annotation.

Nhattan9961 (approved):

docs/mkdocs/docs/api/basic_json/accept.md
```diff
@@ -64,6 +64,10 @@ Whether the input is valid JSON.
 
 Strong guarantee: if an exception is thrown, there are no changes in the JSON value.
 
+## Exceptions
+
+Throws [`parse_error.116`](../../home/exceptions.md#jsonexceptionparse_error116) if passed `#!cpp FILE` pointer is `#!cpp nullptr`.
```
falbrechtskirchinger, line 69: Nitpick: `FILE` or `FILE*` (`FILE*` is used below).

docs/mkdocs/docs/home/exceptions.md
```diff
@@ -354,6 +354,16 @@ A UBJSON high-precision number could not be parsed.
     [json.exception.parse_error.115] parse error at byte 5: syntax error while parsing UBJSON high-precision number: invalid number text: 1A
     ```
 
+### json.exception.parse_error.116
+
+A `#!cpp FILE*` pointer passed to the [parse](../api/basic_json/parse.md) function is `#!cpp nullptr`; that is, a previous call to `#!cpp std::fopen` failed.
```
falbrechtskirchinger, line 359: Nitpick: `FILE` or `FILE*` (`FILE` is used above).

tests/src/unit-deserialization.cpp
```diff
@@ -166,6 +166,21 @@ struct SaxEventLoggerExitAfterStartArray : public SaxEventLogger
...
+// Passing a NULL pointer to the input adapter violates its NON_NULL attribute which is detected by UBSAN.
+// To still test whether exceptions are thrown, we need to exclude these tests from UBSAN which can only
+// be done with a function attribute. See
+// https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#disabling-instrumentation-with-attribute-no-sanitize-undefined
+#if defined(__clang__)
+    __attribute__((no_sanitize("undefined")))
+#endif
+void test_file_exception()
+{
+    std::FILE* f = std::fopen("nonexisting_file", "r"); // NOLINT(cppcoreguidelines-owning-memory)
```
falbrechtskirchinger, line 178: Can you use `std::FILE* f = nullptr;` (again) now that UBSan is disabled?

tests/src/unit-deserialization.cpp
```diff
@@ -170,6 +170,11 @@ struct SaxEventLoggerExitAfterStartArray : public SaxEventLogger
...
 // be done with a function attribute. See
 // https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#disabling-instrumentation-with-attribute-no-sanitize-undefined
+#if defined(__clang__)
+    __attribute__((no_sanitize("undefined")))
+#endif
+void test_file_exception();
+
 #if defined(__clang__)
     __attribute__((no_sanitize("undefined")))
 #endif
```
falbrechtskirchinger, line 180: I don't think you need to repeat the attribute. Not 100% sure though.

nlohmann: See discussion below.

tests/src/unit-deserialization.cpp
nlohmann: See discussion below.

nlohmann:
> I agree that keeping exceptions out of constructors is a good idea. I'd still prefer a way to differentiate parse errors from invalid input adapters. This goes beyond the scope of this PR but is what I've been asking for from the start.

Yes, this is also a pain point for me, because there are issues again and again where people seem to not understand that "unexpected end of file" at position 0 means that a file could not be opened in the first place.

> To that end, an alternative might be to add `operator bool()` to `input_adapter`, check that somewhere in or near the `parse`/`sax_parse`/`accept` functions, and throw an exception from there.
> 
> For `file_input_adapter` the operator would simply return `m_file != nullptr`, for `input_stream_adapter` (which ought to be renamed to `stream_input_adapter` before being added to the public API) `is->good()`, for `iterator_input_adapter` `current != end`, etc.

Sounds good.

But for this PR, should we go for the behavior that a null pointer yields EOF?

gregmarr:
I guess we assert in the constructor for easy diagnosis in the cases where assert is enabled, and add the nullptr check in `get_character()` with the `unlikely` attribute.

falbrechtskirchinger:
And we can remove the `nullptr` check from `get_character()` once my idea (or equivalent) is implemented.

docs/mkdocs/docs/features/assertions.md
```diff
@@ -102,3 +102,30 @@ behavior and yields a runtime assertion.
     ```
     Assertion failed: (m_object != nullptr), function operator++, file iter_impl.hpp, line 368.
     ```
+
+### Reading from a null `FILE` pointer
+
+Reading from a null `#!cpp FILE` pointer is undefined behavior and yields a runtime assertion. This can happen when
+calling `#!cpp std::fopen` on a nonexisting file.
```
falbrechtskirchinger, line 109: I think "nonexistent" is more commonly used.
```suggestion
calling `#!cpp std::fopen` on a nonexistent file.
```

docs/mkdocs/docs/features/assertions.md
```diff
@@ -102,3 +102,30 @@ behavior and yields a runtime assertion.
...
+    The following code will trigger an assertion at runtime:
+
+    ```cpp
+    #include <nlohmann/json.hpp>
+    
+    using json = nlohmann::json;
+    
+    int main()
+    {
+      std::FILE* f = std::fopen("nonexisting_file.json", "r");
```
falbrechtskirchinger, line 122:
```suggestion
      std::FILE* f = std::fopen("nonexistent_file.json", "r");
```

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -61,7 +64,11 @@ class file_input_adapter
 
     std::char_traits<char>::int_type get_character() noexcept
     {
-        return std::fgetc(m_file);
+        if (JSON_HEDLEY_LIKELY(m_file != nullptr))
+        {
+            return std::fgetc(m_file);
+        }
+        return std::char_traits<char>::eof(); // LCOV_EXCL_LINE
```
gregmarr, line 71: Are we not able to write tests that assert?

nlohmann: I could try to set the `FILE` pointer to null after creating it.

falbrechtskirchinger: Wierd. :-)

gregmarr: It would have to be a build without ubsan and with assertions disabled.

nlohmann: I removed the `if`. The assertion, function annotation, and documentation state that the passed pointer must not be null. I don't see the point of playing what-if and spending more effort to simulate a situation where a user that misuses the API is still saved from the consequences.

gregmarr: Okay.

gregmarr (approved):

PR #3609, opened 2022-07-25
Use swap() by ADL

falbrechtskirchinger:
Bring `std::swap` into scope and call `swap()` to allow for argument-dependent lookup when using non-`std` containers.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -3452,7 +3452,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         }
         else
         {
-            JSON_THROW(type_error::create(310, detail::concat("cannot use swap() with ", type_name()), this));
+            JSON_THROW(type_error::create(310, detail::concat("cannot use swap(array_t&) with ", type_name()), this));
         }
     }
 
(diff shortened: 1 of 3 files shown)


include/nlohmann/json.hpp
```diff
@@ -3447,7 +3447,8 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         // swap only works for arrays
         if (JSON_HEDLEY_LIKELY(is_array()))
         {
-            std::swap(*(m_value.array), other);
+            using std::swap;
+            swap(*(m_value.array), other);
```
gregmarr, line 3451: While you're here, should the error strings be updated to provide the parameter type too?

    "cannot use swap(array_t&) with "

gregmarr (approved):
LGTM with a minor suggestion on related error messages, could definitely be a separate PR.

include/nlohmann/json.hpp
falbrechtskirchinger: Good idea. Done.

falbrechtskirchinger: I need to change the unit tests as well. Duh!

PR #3614, opened 2022-07-27
Fix Unicode test timeout (for real this time!)

### LINKED ISSUE #3579: unicode4 test consistently fails on RISC-V hardware
### Description

When compiling ``nlohmann_json`` on NixOS, on a RISC-V 64 bit system, StarFive VisionFive, the ``unicode4`` test times out and fails the build.
This does not happen on regular x86_64 systems, so I believe a fix may be needed for either the test, or the thing being tested, regarding RISC-V compatibility.

### Reproduction steps

Attempt to build ``nlohmann_json`` on a RISC-V system, either hardware or through a virtual machine like QEMU, on NixOS.
If you happen to have a StarFive VisionFive board, I suggest using this PR to reproduce: https://github.com/NixOS/nixpkgs/pull/168826

``unicode4`` test should not be timing out.

### Minimal code example

_No response_

### Error messages
https://gist.github.com/Madouura/edf2ac1b06bf6e63507b9dc9bb338064

GCC11 IIRC, NixOS

### Library version

3.10.5

### Validation

falbrechtskirchinger:
The previous attempt didn't quite do the job because `TIMEOUT_AFTER_MATCH` doesn't work as expected. This sets the timeout of `test-unicode4` to twice the default (i.e., 3000s).

Fixes #3579.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -86,10 +86,8 @@ elseif (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     #json_test_set_test_options(test-disabled_exceptions COMPILE_DEFINITIONS _HAS_EXCEPTIONS=0 COMPILE_OPTIONS /EH)
 endif()
 
-# set timeouts for Unicode tests
-json_test_set_test_options("test-unicode2;test-unicode3;test-unicode4;test-unicode5"
-    TEST_PROPERTIES "TIMEOUT_AFTER_MATCH;1500$<SEMICOLON>UTF-8 strings checked"
-)
+# raise timeout of expensive Unicode test
+json_test_set_test_options(test-unicode4 TEST_PROPERTIES TIMEOUT 3000)
 
 #############################################################################
 # add unit tests


tests/CMakeLists.txt
```diff
@@ -86,10 +86,8 @@ elseif (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     #json_test_set_test_options(test-disabled_exceptions COMPILE_DEFINITIONS _HAS_EXCEPTIONS=0 COMPILE_OPTIONS /EH)
 endif()
 
-# set timeouts for Unicode tests
-json_test_set_test_options("test-unicode2;test-unicode3;test-unicode4;test-unicode5"
-    TEST_PROPERTIES "TIMEOUT_AFTER_MATCH;1500$<SEMICOLON>UTF-8 strings checked"
-)
+# raise timeout of expensive Unicode test
+json_test_set_test_options(test-unicode4 TEST_PROPERTIES TIMEOUT 3000)
```
nlohmann, line 90: What about the other unicode tests? This seems to only tackle test-unicode4.

falbrechtskirchinger: They have an order of magnitude shorter runtime. I only included the others because they all generate output.

nlohmann (approved):
Looks good to me.

Issue #1697, opened 2019-08-01
CMake version config file is architecture-dependent


uhoreg:
CMake's `write_basic_package_version_file` by default creates a file that checks that the word size is the same as the word size used when the file was created.  That means that if the file was created on a 64-bit machine, then CMake will refuse to use it on a 32-bit machine (or vice versa).  This is a problem for the Debian package, since the package is architecture-independent, and so it is only built once on one architecture, and that package used when building other packages on other architectures.  See, for example, https://buildd.debian.org/status/fetch.php?pkg=nheko&arch=i386&ver=0.6.4-1&stamp=1564680125&file=log in which it tries to build a package on a 32-bit architecture, using the JSON library package built on a 64-bit architecture.

In CMake 3.14, they added a `ARCH_INDEPENDENT` flag to `write_basic_package_version_file` to drop this check.  However, if you  use that, it would require bumping the minimum version of CMake to 3.14.  This also doesn't help us in Debian, since we still have CMake 3.13.  For Debian, I've worked around this for now by creating my own template for `nlohmann_jsonConfigVersion.cmake`:

```
set(PACKAGE_VERSION "@PROJECT_VERSION@")

if(PACKAGE_VERSION VERSION_LESS PACKAGE_FIND_VERSION)
  set(PACKAGE_VERSION_COMPATIBLE FALSE)
else()

  if(PACKAGE_FIND_VERSION_MAJOR STREQUAL "@PROJECT_VERSION_MAJOR@")
    set(PACKAGE_VERSION_COMPATIBLE TRUE)
  else()
    set(PACKAGE_VERSION_COMPATIBLE FALSE)
  endif()

  if(PACKAGE_FIND_VERSION STREQUAL PACKAGE_VERSION)
      set(PACKAGE_VERSION_EXACT TRUE)
  endif()
endif()
```

nlohmann:
Would you mind opening a PR?

uhoreg:
Sure, I can do that.  Would you prefer bumping the CMake version requirement to 3.14 so that the ARCH_INDEPENDENT flag can be used, or using the template that I adapted for Debian?

nlohmann:
Definitely not bumping the required CMake version.

uhoreg:
Hi, I'm still alive, and will write a PR, but was on vacation for most of August.

uhoreg:
Hi stale bot, I made a PR to fix this issue.

nlohmann:
Fixed by merging #1746

PR #3619, opened 2022-07-28
build: install .pc and .cmake files to share/

Tachi107:
As nlohmann_json is a header-only library, its pkg-config and cmake config files should be installed to `share/` (GNUInstallDirs' DATADIR), as `share/` is the canonical directory where architecture-independent files should be stored, while `lib/` is for architecture-dependent stuff (see the [FHS][]).

Apart from being technically correct, installing to `share/` can help with cross-compiling, for example in Debian-based environments. If you're interested, I'd suggest reading this [thread][].

Related to #1697

[FHS]: https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrshareArchitectureindependentData
[thread]: https://github.com/marzer/tomlplusplus/pull/165#issuecomment-1193146210

* * *

### Pull request checklist


Tachi107:
@uhoreg, as you're the maintainer of the [nlohmann-json3](https://tracker.debian.org/pkg/nlohmann-json3) Debian package, would you like to share some comment?

uhoreg:
> @uhoreg, as you're the maintainer of the [nlohmann-json3](https://tracker.debian.org/pkg/nlohmann-json3) Debian package, would you like to share some comment?

I don't know much about cross-compiling, so I can't say much about how it would affect that, but it does at least seem plausible that putting it in `share` would work better for cross-compiling.

In general, `share` is for architecture-independent files, and `lib` is for architecture-dependent files.  Since this library is header-only, `share` seems like the logical place for those files.  (Even with libraries that are architecture-dependent, some of them place their `.pc` files in `share`, since the contents of the `.pc` file are the same on every architecture.)

falbrechtskirchinger (approved):
Looks good to me.

nlohmann (approved):
Looks good to me.

Tachi107:
Thanks for merging :)

nlohmann:
Thanks for opening the PR!

PR #3621, opened 2022-07-28
Disable exceptions on ICPC

falbrechtskirchinger:
We currently don't disable exceptions on ICPC for `test-disabled_exceptions`. Let's see what happens if we do.

Also simplifies the CMake code a lot.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -78,13 +78,18 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
 endif()
 
 # disable exceptions for test-disabled_exceptions
-json_test_set_test_options(test-disabled_exceptions COMPILE_DEFINITIONS JSON_NOEXCEPTION)
-if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    json_test_set_test_options(test-disabled_exceptions COMPILE_OPTIONS -fno-exceptions)
-elseif (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
-    # disabled due to https://github.com/nlohmann/json/discussions/2824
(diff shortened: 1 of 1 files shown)


falbrechtskirchinger:
> CMake Error at cmake/test.cmake:136 (target_link_libraries):
>   Error evaluating generator expression:
> 
>     $<CXX_COMPILER_ID:AppleClang,Clang,GNU,Intel,IntelLLVM>
> 
>   $<CXX_COMPILER_ID> expression requires one or zero parameters.

Apparently a CMake 3.15 feature. :-(

> $<CXX_COMPILER_ID:compiler_ids>
> where compiler_ids is a comma-separated list. 1 if CMake's compiler id of the CXX compiler matches any one of the entries in compiler_ids, otherwise 0.

nlohmann:
Doesn’t this run in the CI container? There we have control over the CMake version.

falbrechtskirchinger:
The CMake flag test uses 3.13. That one fails.

nlohmann:
> The CMake flag test uses 3.13. That one fails.
> 
> 3.13 was released around Nov 2018. Bumping to 3.15 might be a bit too soon for some distros. 🤷‍♂️

Yes, the 3.13 is for externals to execute the unit tests. Not sure if the specific "Intel without exceptions" should be seen as a test someone other than a maintainer should execute. I guess this goes in a similar direction as #3623 - we need to better separate maintainer targets from the rest.

falbrechtskirchinger:
Unfortunately, I can't separate them. Not if I want to replace the `if()` construct with a single generator expression.

Basically, I can make it work, but I can't also clean up the CMake code. Oh well …

nlohmann:
> Unfortunately, I can't separate them. Not if I want to replace the `if()` construct with a single generator expression.
> 
> Basically, I can make it work, but I can't also clean up the CMake code. Oh well …

Please don't invest more work to this as necessary...

falbrechtskirchinger:
I won't. I'll revert and just add "Intel" and "IntelLLVM" to the `if()`.

nlohmann (approved):
Looks good to me.

PR #3622, opened 2022-07-28
Adjust JSON Pointer examples

nlohmann:
Fix documentation issue after #3600.

Also add test to check the examples' outputs to the CI.


falbrechtskirchinger (approved):
Looks good to me.

nlohmann:
File `patch_inplace.output` has an issue with a missing newline. Let's see if the CI detects this.

falbrechtskirchinger:
There's one more issue with `multimap` I found while regenerating the outputs on my machine. I don't remember which example ...

falbrechtskirchinger:
Please replace `make` with `$(MAKE)` in `docs/Makefile`. Should get rid of those "jobserver unavailable" warnings.

falbrechtskirchinger:
Example `basic_json__CompatibleType.cpp` differs in line 5 (on my machine).

> {"one":true,"three":false,"two":true}

>{"one":true,"three":true,"two":true}

The others are `patch_inplace.cpp` (as you mentioned), `max_size.cpp`, and `std_hash.cpp`.

nlohmann:
> Example `basic_json__CompatibleType.cpp` differs in line 5 (on my machine).
> 
> > {"one":true,"three":false,"two":true}
> 
> > {"one":true,"three":true,"two":true}

That is odd. I cannot find a hint in the documentation of `std::unordered_multimap` that the constructor's behavior is platform-dependent. Maybe I'll use `emplace` to avoid this.

> The others are `patch_inplace.cpp` (as you mentioned), `max_size.cpp`, and `std_hash.cpp`.

Oh, I totally forgot about those. I will also add notes to the API pages using these examples that the output is platform-dependent.

falbrechtskirchinger:
> > Example `basic_json__CompatibleType.cpp` differs in line 5 (on my machine).
> > > {"one":true,"three":false,"two":true}
> > 
> > 
> > > {"one":true,"three":true,"two":true}
> 
> That is odd. I cannot find a hint in the documentation of `std::unordered_multimap` that the constructor's behavior is platform-dependent. Maybe I'll use `emplace` to avoid this.

Isn't that about which of multiple objects matching the same key is returned? I'd assume the standard doesn't require anything specific.

nlohmann:
You're right - I forgot that (of course) the container stores multiple values and only the conversion to JSON drops all but one. Alright, I'll add it to the list of platform-dependent examples and then we should be fine.

PR #3626, opened 2022-07-29
Update CI

### LINKED ISSUE #3612: macOS 10.15 Actions runner image deprecation
The macOS 10.15 Actions runner image is being removed.

https://github.com/actions/virtual-environments/issues/5583

falbrechtskirchinger:
* Bump macos-10.5 runner to macos-11 and update Xcode versions
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md#xcode
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-12-Readme.md#xcode
https://gist.github.com/yamaya/2924292#file-xcode-clang-vers

* Fix CodeQL workflow warning:
  > 2 issues were detected with this workflow: git checkout HEAD^2 is no longer necessary. Please remove this step as Code Scanning recommends analyzing the merge commit for best results. Please make sure that every branch in on.pull_request is also in on.push so that Code Scanning can compare pull requests against the state of the base branch.

* Only start AppVeyor builds if a commit touches these paths:
  ```
  cmake/
  include/
  tests/
  CMakeLists.txt
  ```

Fixes #3612.

### DIFF
--- a/.github/external_ci/appveyor.yml
+++ b/.github/external_ci/appveyor.yml
@@ -1,5 +1,19 @@
 version: '{build}'
 
+# only build PRs and commits to develop branch
+# (see https://help.appveyor.com/discussions/questions/55079-two-builds-per-commit-to-pull-request)
+branches:
+  only:
+    - develop
+
+only_commits:
+  files:
+    - .github/external_ci/appveyor.yml
+    - cmake/
+    - include/
+    - tests/
+    - CMakeLists.txt
+
 environment:
   matrix:
     - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
@@ -75,9 +89,3 @@ test_script:
   # occasional timeouts on AppVeyor.
   # More info: https://github.com/nlohmann/json/pull/1570
   - if "%configuration%"=="Debug" ctest --exclude-regex "test-unicode" -C "%configuration%" --parallel 2 --output-on-failure
-
-# only build PRs and commits to develop branch
-# (see https://help.appveyor.com/discussions/questions/55079-two-builds-per-commit-to-pull-request)
-branches:
-  only:
-    - develop
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -23,38 +23,15 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v3
-      with:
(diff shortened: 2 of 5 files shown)


.github/external_ci/appveyor.yml
```diff
@@ -1,5 +1,18 @@
...
+branches:
+  only:
+    - develop
+
+only_commits:
+  files:
+    - cmake/
+    - include/
+    - tests/
+    - CMakeLists.txt
```
nlohmann, line 14: What about `.github/external_ci/appveyor.yml`?

falbrechtskirchinger: Done.

.github/external_ci/appveyor.yml
```diff
@@ -1,5 +1,19 @@
 version: '{build}'
 
+# only build PRs and commits to develop branch
+# (see https://help.appveyor.com/discussions/questions/55079-two-builds-per-commit-to-pull-request)
+branches:
+  only:
+    - develop
+
+only_commits:
```
nlohmann, line 9: I gave this another thought: I do not think that we should be "too clever" about when to trigger a CI job. There may always be edge cases where a file outside this list can break a build. I get that we don't want to fire up dozens of Docker images for typo fixes, but at the same time, I would want to state with confidence "this does not break anyone's build".

nlohmann (changes requested):

.github/external_ci/appveyor.yml
falbrechtskirchinger: Note that this only affects AppVeyor, though. The vast majority of tests are still run.

nlohmann: I see. Let's hope this doesn't bite us in the future.

nlohmann (approved):
Looks good to me.

falbrechtskirchinger:
One more note for you:

> xcode: ['11.7', '12.4', '12.5.1', '13.0']

`13.0` turns into `13` and CI fails. Seems like good practice to use strings in cases like this in the future.

nlohmann:
> One more note for you:
> 
> > xcode: ['11.7', '12.4', '12.5.1', '13.0']
> 
> `13.0` turns into `13` and CI fails. Seems like good practice to use strings in cases like this in the future.

And that answers my unasked question why you added quotes around all the version numbers :-)

PR #3628, opened 2022-07-29
Fix patch::add creating nonexistent parents

### LINKED ISSUE #3134: Applying JSON patch creates parent object
Running the following function

```cpp
void test_prob() {
    auto doc = R"(
        {
            "a": {}
        }
    )"_json;

    auto patch = R"(
        [
            {
                "op": "add",
                "path": "/a/b/c/d",
                "value": 1
            }
        ]
    )"_json;

    auto doc1 = doc.patch(patch);
    auto d = doc1.dump(2);
    printf("%s\n", d.c_str());
}
```

results in creating `b` and `c` parent objects:

```js
{
  "a": {
    "b": {
      "c": {
        "d": 1
      }
    }
  }
}
```

while according to [RFC6902](https://datatracker.ietf.org/doc/html/rfc6902) it is an error:

"Because this operation is designed to add to existing objects and
   arrays, its target location will often not exist.  Although the
   pointer's error handling algorithm will thus be invoked, this
   specification defines the error handling behavior for "add" pointers
   to ignore that error and add the value as specified.

   **However, the object itself or an array containing it does need to
   exist, and it remains an error for that not to be the case.** "

BTW, great library, thank you very much!

falbrechtskirchinger:
Replacement for #3199. (Original author MIA.)

Fixes #3134.
Closes #3199.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -4683,7 +4683,8 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
             // get reference to parent of JSON pointer ptr
             const auto last_path = ptr.back();
             ptr.pop_back();
-            basic_json& parent = result[ptr];
+            // parent must exist when performing patch add per RFC6902 specs
+            basic_json& parent = result.at(ptr);
 
             switch (parent.m_type)
             {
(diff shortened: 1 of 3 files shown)


gregmarr (approved):

nlohmann (approved):
Looks good to me.

PR #3637, opened 2022-07-31
Minor BJData fixes

falbrechtskirchinger:
These are the previously discussed minor fixes to the BJData `binary_reader` code.

* Replace the `std::vector`/`std::map` lookup tables with `std::array`s.
   This is meant to be an optimization (mostly avoiding unnecessary, repeated heap allocations). Whether `std::lower_bound()` is actually faster than `std::find`/`std::find_if` in this scenario is unknown.

* Replace `string_t::npos` with `binary_reader::npos`.

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -20,7 +20,6 @@
 #include <string> // char_traits, string
 #include <utility> // make_pair, move
 #include <vector> // vector
-#include <map> // map
 
 #include <nlohmann/detail/exceptions.hpp>
 #include <nlohmann/detail/input/input_adapters.hpp>
@@ -2240,10 +2239,9 @@ class binary_reader
 
         if (current == '$')
         {
-            std::vector<char_int_type> bjdx = {'[', '{', 'S', 'H', 'T', 'F', 'N', 'Z'}; // excluded markers in bjdata optimized type
-
             result.second = get();  // must not ignore 'N', because 'N' maybe the type
-            if (JSON_HEDLEY_UNLIKELY( input_format == input_format_t::bjdata && std::find(bjdx.begin(), bjdx.end(), result.second) != bjdx.end() ))
+            if (input_format == input_format_t::bjdata
+                    && JSON_HEDLEY_UNLIKELY(std::binary_search(bjd_optimized_type_markers.begin(), bjd_optimized_type_markers.end(), result.second)))
             {
                 auto last_token = get_token_string();
                 return sax->parse_error(chars_read, last_token, parse_error::create(112, chars_read,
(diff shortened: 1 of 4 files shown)


include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -2492,23 +2494,23 @@ class binary_reader
...
+            if (JSON_HEDLEY_UNLIKELY(it == bjd_types_map.end() || it->first != size_and_type.second))
             {
                 auto last_token = get_token_string();
                 return sax->parse_error(chars_read, last_token, parse_error::create(112, chars_read,
                                         exception_message(input_format, "invalid byte: 0x" + last_token, "type"), nullptr));
             }
 
-            if (JSON_HEDLEY_UNLIKELY(!sax->key(key) || !sax->string(bjdtype[size_and_type.second]) ))
+            string_t type = it->second;
+            if (!sax->key(key) || !sax->string(type))
```
nlohmann, line 2513: Why did you remove the `JSON_UNLIKELY` here? So far, we always added it for the error cases.

falbrechtskirchinger: Probably a mistake introduced while resolving merge conflicts.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -2492,23 +2494,23 @@ class binary_reader
         // if bit-8 of size_and_type.second is set to 1, encode bjdata ndarray as an object in JData annotated array format (https://github.com/NeuroJSON/jdata):
         // {"_ArrayType_" : "typeid", "_ArraySize_" : [n1, n2, ...], "_ArrayData_" : [v1, v2, ...]}
 
-        if (input_format == input_format_t::bjdata && size_and_type.first != string_t::npos && (size_and_type.second & (1 << 8)) != 0)
+        if (input_format == input_format_t::bjdata && size_and_type.first != npos && (size_and_type.second & (1 << 8)) != 0)
         {
-            std::map<char_int_type, string_t> bjdtype = {{'U', "uint8"},  {'i', "int8"},  {'u', "uint16"}, {'I', "int16"},
```
nlohmann, line 2497: I think the `<map>` header can now also be removed from `binary_reader.hpp`.

include/nlohmann/detail/meta/cpp_future.hpp
```diff
@@ -152,15 +153,19 @@ template<> struct priority_tag<0> {};
...
 #ifndef JSON_HAS_CPP_17
-
     template<typename T>
-    constexpr T static_const<T>::value; // NOLINT(readability-redundant-declaration)
-
+    constexpr T static_const<T>::value;
 #endif
 
+template<typename T, typename... Args>
+inline constexpr std::array<T, sizeof...(Args)> make_array(Args&& ... args)
```
gregmarr, line 165: Should this be more like `to_array` from C++20?

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -2967,7 +2970,44 @@ class binary_reader
...
+    bjd_type{'d', "single"},               \
+    bjd_type{'i', "int8"},                 \
+    bjd_type{'l', "int32"},                \
+    bjd_type{'m', "uint32"},               \
+    bjd_type{'u', "uint16"})
+
+  JSON_PRIVATE_UNLESS_TESTED:
+    // lookup tables
+    // NOLINTNEXTLINE(cppcoreguidelines-non-private-member-variables-in-classes)
+    const decltype(JSON_BINARY_READER_MAKE_BJD_OPTIMIZED_TYPE_MARKERS_) bjd_optimized_type_markers =
```
gregmarr, line 2995: Can this be `static const` and then you can use `auto` and not have to have the `decltype` and macro to repeat?

falbrechtskirchinger: I don't think so … that'll just end in clang-tidy (or is it a Clang warning flag?) complaining.

gregmarr: What would it be complaining about?

include/nlohmann/detail/meta/cpp_future.hpp
falbrechtskirchinger: Maybe. This was initially based on Library TS but isn't needed anymore in newer C++ versions. Let me look at `to_array`...

include/nlohmann/detail/input/binary_reader.hpp
falbrechtskirchinger: Exit-time destructors, same as last time we had this conversation. :-)
Don't remember which PR...

include/nlohmann/detail/meta/cpp_future.hpp
falbrechtskirchinger: No, it's definitely `make_array` from the Library Fundamentals TS. I put it in `cpp_future.hpp` because C++17 solved this problem by introducing deduction guides.

include/nlohmann/detail/input/binary_reader.hpp
gregmarr: Ah, thanks, didn't remember that.

include/nlohmann/detail/meta/cpp_future.hpp
falbrechtskirchinger: Yes? That's what I said?

> based on Library [Fundamentals] TS but isn't needed anymore in newer C++ versions

> I put it in cpp_future.hpp because C++17 solved this problem by introducing deduction guides.

gregmarr: I was referring to the part about `std::to_array` being in C++20.

gregmarr: If you want to keep it based on the old `make_array` that's fine too, just pointing out the current standardized name.

gregmarr: Okay, I was looking at being the same as this example `std::to_array({ 0, 2, 1, 3 });`

falbrechtskirchinger:
Wow! This improves performance parsing a 136M document by 0.3%. This was truly worth it! :rofl:

falbrechtskirchinger:
Looks good to me.

nlohmann (approved):
Looks good to me.

PR #3643, opened 2022-08-01
Miscellaneous small fixes

falbrechtskirchinger:
* Fix lgtm issues: https://lgtm.com/projects/g/nlohmann/json/alerts/?mode=list.
* Fix `serve_header.py` always exiting with 0 error code.
* Replace deprecated function use in `serve_header.py`.

I may leave this here for a while, in case I find more issues like this. Chances are, of course, I'll find another one moments after this one has been merged. :-)

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -1,0 +1,4 @@
+path_classifiers:
+  thirdparty:
+    - /tools/amalgamate
+    - /tools/cpplint
--- a/tests/src/unit-readme.cpp
+++ b/tests/src/unit-readme.cpp
@@ -106,12 +106,10 @@ TEST_CASE("README" * doctest::skip())
             json j = "{ \"happy\": true, \"pi\": 3.141 }"_json; // NOLINT(modernize-raw-string-literal)
 
             // or even nicer with a raw string literal
-            auto j2 = R"(
-          {
-            "happy": true,
-            "pi": 3.141
-          }
-        )"_json;
+            auto j2 = R"({
+                "happy": true,
+                "pi": 3.141
+            })"_json;
 
             // or explicitly
             auto j3 = json::parse(R"({"happy": true, "pi": 3.141})");
@@ -160,10 +158,10 @@ TEST_CASE("README" * doctest::skip())
             CHECK(foo == true);
 
             // other stuff
-            j.size();     // 3 entries
-            j.empty();    // false
-            j.type();     // json::value_t::array
-            j.clear();    // the array is empty again
+            CHECK(j.size() == 3);                     // 3 entries
+            CHECK_FALSE(j.empty());                   // false
+            CHECK(j.type() == json::value_t::array);  // json::value_t::array
+            j.clear();                                // the array is empty again
 
             // create an object
             json o;
@@ -172,6 +170,7 @@ TEST_CASE("README" * doctest::skip())
             o["baz"] = 3.141;
 
             // find an entry
+            CHECK(o.find("foo") != o.end());
             if (o.find("foo") != o.end())
             {
                 // there is an entry with key "foo"
@@ -266,9 +265,9 @@ TEST_CASE("README" * doctest::skip())
         {
             // a JSON value
             json j_original = R"({
-          "baz": ["one", "two", "three"],
-          "foo": "bar"
-        })"_json;
+                "baz": ["one", "two", "three"],
+                "foo": "bar"
+            })"_json;
 
             // access members with a JSON pointer (RFC 6901)
             j_original["/baz/1"_json_pointer];
@@ -276,10 +275,10 @@ TEST_CASE("README" * doctest::skip())
 
             // a JSON patch (RFC 6902)
             json j_patch = R"([
-          { "op": "replace", "path": "/baz", "value": "boo" },
-          { "op": "add", "path": "/hello", "value": ["world"] },
-          { "op": "remove", "path": "/foo"}
-        ])"_json;
+                { "op": "replace", "path": "/baz", "value": "boo" },
+                { "op": "add", "path": "/hello", "value": ["world"] },
+                { "op": "remove", "path": "/foo"}
+            ])"_json;
 
             // apply the patch
             json j_result = j_original.patch(j_patch);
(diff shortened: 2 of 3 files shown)


tools/cpplint/cpplint.py
```diff
@@ -4852,7 +4852,6 @@ def CheckStyle(filename, clean_lines, linenum, file_extension, nesting_state,
   # if(match(prev, " +for \\(")) complain = 0;
   # if(prevodd && match(prevprev, " +for \\(")) complain = 0;
   scope_or_label_pattern = r'\s*(?:public|private|protected|signals)(?:\s+(?:slots\s*)?)?:\s*\\?$'
-  classinfo = nesting_state.InnermostClass()
```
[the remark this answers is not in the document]
falbrechtskirchinger: Because of this lgtm issue:

https://lgtm.com/projects/g/nlohmann/json/snapshot/4ce8114e8bcc36495bed4b4505157e9bdf984156/files/tools/cpplint/cpplint.py?sort=name&dir=ASC&mode=heatmap#x24ea1ee68e875673:1

I know it'll be overwritten next time you update if upstream hasn't fixed it.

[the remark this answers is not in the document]
nlohmann: I think we should exclude external code from lgtm.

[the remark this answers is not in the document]
falbrechtskirchinger: If you can do that, that would of course be preferred.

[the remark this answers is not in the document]
nlohmann: It can be configured with a `.lgtm.yml` file, see https://lgtm.com/help/lgtm/lgtm.yml-configuration-file.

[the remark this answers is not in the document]
falbrechtskirchinger: Ah, I'll do that then.

falbrechtskirchinger:
Looks good to me.

@nlohmann I'll leave it up to you whether you'd prefer collecting more small stuff here or if you'd want to merge it now.
I'd only tweak the title once I know the final content.

I've tested `server_header.py` (runs and serves requests).
GDB pretty printer received some limited testing as part of the inline namespace PR, it still works on a simple JSON object.

.lgtm.yml
```diff
@@ -0,0 +1,4 @@
+path_classifiers:
+  thirdparty:
+    - /tools/amalgamate
+    - /tools/cpplint
```
falbrechtskirchinger, line 4: I'm not 100% sure that absolute paths are valid. It is used in one example like so: `exclude: /` (exclude everything)

We could customize path classification completely, but I don't know if anyone cares to keep this up-to-date. This just hides the `cpplint.py` alert.

The other issue was not – as I suspected – because of `single_include/` but because of the addition of the 3.10.5 header as part of the ABI test. This was fixed by simply renaming the header guard to match the file name (`json_v3_10_5.hpp`).

tests/src/unit-readme.cpp
```diff
@@ -105,11 +105,11 @@ TEST_CASE("README" * doctest::skip())
 
             // or even nicer with a raw string literal
             auto j2 = R"(
-          {
-            "happy": true,
-            "pi": 3.141
-          }
-        )"_json;
+                {
```
nlohmann, line 108: This is a curious whitespace change.

tests/src/unit-readme.cpp
```diff
@@ -158,10 +158,10 @@ TEST_CASE("README" * doctest::skip())
             CHECK(foo == true);
 
             // other stuff
-            j.size();     // 3 entries
-            j.empty();    // false
-            j.type();     // json::value_t::array
-            j.clear();    // the array is empty again
+            CHECK(j.size() == 3);                     // 3 entries
```
nlohmann, line 161: Just FYI: I copied the code snippets from the README here to make sure they compile in the first place. In an ideal world, the README would be shorter and have more pointers to the documentation - or we'd have a means to strip code snippets from the README and check them automatically.

Anyway, I appreciate this approach of converting the README code to proper tests.

tools/serve_header/serve_header.py
```diff
@@ -240,7 +240,7 @@ def stop(self):
         self.observer.stop()
         self.observer.join()
 
-class HeaderRequestHandler(SimpleHTTPRequestHandler):
+class HeaderRequestHandler(SimpleHTTPRequestHandler): # lgtm[py/missing-call-to-init]
```
nlohmann, line 243: Why don't you call `SimpleHTTPRequestHandler`'s `__init__` function?

tests/src/unit-readme.cpp
falbrechtskirchinger: I think I just indented it to bring the closing `)` to the correct level. Makes sense that it was off given you just pasted it from the README.

tests/src/unit-readme.cpp
falbrechtskirchinger: I added the first 3 checks just because of lgtm. It's not really necessary, since lgtm ignore test code, but it seemed like a good idea.

tools/serve_header/serve_header.py
falbrechtskirchinger: I do. `BaseRequestHandler` doesn't and that seems to be a deliberate choice.

> Class HeaderRequestHandler may not be initialized properly as method BaseRequestHandler.__init__ is not called from its __init__ method.

tests/src/unit-readme.cpp
```diff
@@ -160,10 +160,10 @@ TEST_CASE("README" * doctest::skip())
...
+            CHECK_FALSE(j.empty());                   // false
+            CHECK(j.type() == json::value_t::array);  // json::value_t::array
```
gregmarr, line 165: Are these comments needed?  They seem pretty redundant and could potentially become stale.

falbrechtskirchinger: :man_shrugging:

As Niels mentioned earlier, he just copy & pasted the README.

shaischaudhry (approved):

tests/src/unit-readme.cpp
gregmarr: Ah, I missed that the checks were part of the readme too.

falbrechtskirchinger:
Several items in here have made it into other PRs. I've removed them.
Looks good to me.

nlohmann (approved):
Looks good to me.

PR #3651, opened 2022-08-02
Add Discord badge to README

falbrechtskirchinger:
Add the Shields.io Discord badge with a permanent invite link to the README.

> To use the Discord badge a Discord server admin must enable the widget setting on the server.


nlohmann:
I think we should make clear that Discord will not be a support channel, but a means to coordinate the development. A badge is fine, but this should also go to the README.

nlohmann:
FYI: Enabled the widget in the server settings.

nlohmann (changes requested):
We should mention in https://github.com/nlohmann/json#support that the Discord channel is to coordinate library development and not meant as a means to provide support.

falbrechtskirchinger:
If the server starts attracting a healthy number of people, I don't see why we couldn't add a dedicated support channel?

nlohmann:
> If the server starts attracting a healthy number of people, I don't see why we couldn't add a dedicated support channel?

I am hesitant, because I still get a lot of emails from people asking support, and I stopped answering (and only referred them to GitHub), because such a 1:1 communication has no benefit to others.

Let's just see what happens now, and we can still change our policy later.

nlohmann (approved):
Looks good to me.

PR #3660, opened 2022-08-03
Publish documentation on push to develop branch

nlohmann:
## Changes

- add workflow to publish documentation on push to develop branch
- minor changes:
  - make sure bash is used in docset Makefile (not needed here, but calling the Makefile in GitHub actions yielded errors, because `/bin/sh` was used as default shell)
  - bump version of `plantuml-markdown` package

## Notes

The concurrency setting make sure that no two jobs of the workflow run concurrently. Jobs are waiting until previous jobs complete:

Screen Shot 2022-08-03 at 10 11 27

### DIFF
--- a/.github/workflows/publish_documentation.yml
+++ b/.github/workflows/publish_documentation.yml
@@ -17,7 +17,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - name: install PlantUML
+        run: sudo apt-get install -y plantuml
+      - name: install virtual environment
+        run: make install_venv -C docs/mkdocs
       - name: build documentation
-        run: make install_venv build -C docs/mkdocs
+        run: make build -C docs/mkdocs
       - name: publish documentation
         run: make publish -C docs/mkdocs
--- a/.github/workflows/publish_documentation.yml
+++ b/.github/workflows/publish_documentation.yml
@@ -4,8 +4,7 @@ name: Publish documentation
 on:
   push:
     branches:
-      #- develop
-      - documentation_update
+      - develop
 
 # we don't want to have concurrent jobs, and we don't want to cancel running jobs to avoid broken publications
 concurrency:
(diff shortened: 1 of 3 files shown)


falbrechtskirchinger:
>   * make sure bash is used in docset Makefile (not needed here, but calling the Makefile in GitHub actions yielded errors, because `/bin/sh` was used as default shell)

FYI, `/bin/sh` is symlinked to `dash` on Ubuntu.

.github/workflows/publish_documentation.yml
```diff
@@ -0,0 +1,33 @@
...
+  cancel-in-progress: false
+
+jobs:
+  publish_documentation:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      # We need a recent PlantUML version to generate UML diagrams. The version from Ubuntu is outdated, so we
+      # need to replace the JAR file with a release from PlantUML's GitHub page.
```
falbrechtskirchinger, line 21: Is that true for `ubuntu-22.04`?

nlohmann:
> FYI, `/bin/sh` is symlinked to `dash` on Ubuntu.

Alright. It still did not work. I was trying to have an up-to-date docset somewhere, but the artifacts of GitHub are not too useful here.

falbrechtskirchinger (approved):
Looks good to me.

.github/workflows/publish_documentation.yml
nlohmann: Is that newer than ubuntu-latest?

falbrechtskirchinger: In `ubuntu-22.04` the version is 1.2020.2 and in `ubuntu-latest` (20.04), it is 1.2018.13.

nlohmann: I'll give it a try.

nlohmann: Nice - this makes things much easier and with fewer dependencies.

PR #3667, opened 2022-08-03
Add `operator<` to `json_pointer`, to allow use as a map key

khiner:
Hopefully, the commit description is self-explanatory.

This will require new unit tests, which I can write and add to this PR, but I wanted to get a sanity check that there's not some reason this is a bad idea.

Thank you!


falbrechtskirchinger:
FYI, there's another PR in flight, fixing a regression involving equality comparison (#3664). Seeing that the other operators are missing, I was planning on adding the rest eventually. This is a start.

khiner:
I am having trouble finding the actual issue behind the [failing amalgamation test](https://github.com/nlohmann/json/runs/7660216928?check_suite_focus=true). As far as I can tell, the indentation style matches the surrounding code exactly. Could someone more familiar with this test please help decipher? 🙏

falbrechtskirchinger:
`single_include/nlohmann/json.hpp` is missing. You need to run `make amalgamate` and add `single_include/nlohmann/json.hpp` to your commit. See https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md#files-to-change.

khiner:
Ah! Sorry I missed that 🤦‍♂️

falbrechtskirchinger:
This still looks off. Did you let your editor reformat any of the files?

We use Artistic Style which does not create reproducible formatting. I.e., differently formatted input may produce different output.

khiner:
Hrm, there are no changed files in my `lib/json` directory. The only changes from `develop` whatsoever were the `json_pointer.hpp` changes in this PR. I exclude `lib/json/**` from formatting explicitly in my IDE settings. Maybe something about my environment caused `make amalgamate` to run differently? Maybe you could try running with just my non-amalgamate changes?

falbrechtskirchinger:
Fixed it for you. https://github.com/khiner/json/pull/1/

khiner:
Thanks @falbrechtskirchinger ! I rebased that on my branch. Still not sure why the run output is different between us 🤷‍♂️

khiner:
ayy looks like the other equality operators PR got in, so looks like I'll have to rework this one

falbrechtskirchinger:
We're moving fast right now because we want to get v3.11.2 out the door ASAP.

FYI, my Artistic Style version is:
```
$ astyle --version
Artistic Style Version 3.1
```
Don't know if that might be causing the difference.

khiner:
Turns out I didn't have `astyle` installed at all, heh. So that explains that. Doesn't seem like it should run without warning/error in that case though. Either way, after installing, my `amalgamate` doesn't mess with other lines I didn't touch.

I'm opening up a new PR based off current `develop`

PR #3671, opened 2022-08-04
Add labeler action

nlohmann:
Automatically adds labels to PRs:

- size label (S, M, L) according to the number of additions+deletions
- other labels:
  - documentation
  - CI
  - CMake
  - tests

The idea is to make it easier to grasp what is touched by a CI, and make also to see on first sight that a PR without tests could be an issue.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,8 @@
+documentation:
+  - 'README.md'
+  - 'docs/*'
+  - 'docs/**/*'
+
+ci:
+  - '.github/workflows/*'
+  - '.github/external_ci/*'
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -2,16 +2,20 @@ version: 1
 
 labels:
 - label: "documentation"
-  files: "README.md"
+  files:
+    - "README.md"
 
 - label: "documentation"
-  files: "docs/.*"
+  files:
+    - "docs/.*"
 
 - label: "ci"
-  files: "github/workflows/.*"
+  files:
+    - "github/workflows/.*"
 
 - label: "ci"
-  files: "github/external_ci/.*"
+  files:
+    - "github/external_ci/.*"
 
 - label: "S"
   size-below: 10
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -9,6 +9,18 @@ labels:
   files:
     - "docs/.*"
 
+- label: "tests"
+  files:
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -9,6 +9,18 @@ labels:
...
+  files:
+    - "tests/.*"
+
+- label: "CMake"
+  files:
+    - ".*CMakeLists.txt"
+
+- label: "CMake"
+  files:
+    - "cmake/.*"
```
falbrechtskirchinger, line 22: No need to add the label twice?
```suggestion
- label: "CMake"
  files:
    - ".*CMakeLists.txt"
    - "cmake/.*"
```

falbrechtskirchinger (changes requested):

.github/labeler.yml
```diff
@@ -21,11 +21,11 @@ labels:
   files:
     - "cmake/.*"
 
-- label: "ci"
+- label: "CI"
   files:
     - "github/workflows/.*"
 
-- label: "ci"
+- label: "CI"
```
falbrechtskirchinger, line 28: same here.

.github/labeler.yml
nlohmann: No, the files must match *all* listed regular expressions, so it's an implicit conjunction. For disjunctions, we need to repeat the rule.

.github/labeler.yml
nlohmann: See above.

.github/labeler.yml
falbrechtskirchinger: Wait, this is not the official GitHub labeler? I'm looking at the wrong documentation.

nlohmann: See https://github.com/srvaroa/labeler#configuration:

> The action combines all conditions with an AND operation.
> ...
> If you wish to apply an OR, you may set multiple matchers for the same label.

nlohmann: Right, it's not the official one. This one has more features and is actively developed.

falbrechtskirchinger: I'm looking at:
`- uses: actions/labeler@v4`
https://github.com/marketplace/actions/labeler

But you're using:
`- uses: srvaroa/labeler@master`

falbrechtskirchinger (approved):
Looks good to me.

Nhattan9961:
S

PR #3673, opened 2022-08-04
Refine 'Publish documentation' workflow

falbrechtskirchinger:
Refine the `Publish documentation` workflow:

* Only run the workflow when files in `/docs/mkdocs/` have changed.
* Allow triggering the workflow manually. (Can't hurt, can it?)
* Restrict the workflow to the `nlohmann/json` repository. I.e, don't try to publish the documentation on push to a fork.

### DIFF
--- a/.github/workflows/publish_documentation.yml
+++ b/.github/workflows/publish_documentation.yml
@@ -8,6 +8,7 @@ on:
     paths:
       - docs/mkdocs/**
       - docs/examples/**
+  workflow_dispatch:
 
 # we don't want to have concurrent jobs, and we don't want to cancel running jobs to avoid broken publications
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_documentation.yml
```diff
@@ -5,6 +5,9 @@ on:
   push:
     branches:
       - develop
+    paths:
+      - docs/mkdocs/**
```
nlohmann, line 9: What about the examples?

falbrechtskirchinger: Right. That's it, though?

nlohmann (approved):
Looks good to me.

PR #3674, opened 2022-08-04
Attempt to fix labeler permissions

falbrechtskirchinger:
Change trigger to `pull_request_target` for write permissions and limit permissions to minimum required.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,10 +1,14 @@
 name: "Pull Request Labeler"
 
 on:
-  - pull_request
+  pull_request_target:
+    types: [opened, synchronize]
 
 jobs:
-  build:
+  label:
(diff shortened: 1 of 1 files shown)


nlohmann:
The job seems not to be executed.

falbrechtskirchinger:
I don't think this will trigger until it's merged.

Testing here. I've pushed it to `develop` and submitted a PR. (Of course, we have the same issue of not being able to test permissions this way.)
https://github.com/falbrechtskirchinger/json/pull/1

falbrechtskirchinger:
All I can say is that it runs (once merged). If it improves the situation over the status-quo of not labeling PRs, I don't know.

nlohmann:
Let's give it a try!

nlohmann (approved):
Looks good to me.

PR #3676, opened 2022-08-05
Added simple nlohmann::json operator+

MastProgs:
Add a simple plus operator to easily combine json objects.
It is possible to combine various json objects into one json object.
If the same key value exists, the json object in the back order has a structure that overwrites that value.
I'm a beginner, so I'd appreciate it if you could understand even if there's a lack

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -5074,19 +5074,12 @@ std::string to_string(const NLOHMANN_BASIC_JSON_TPL& j)
     return j.dump();
 }
 
-inline nlohmann::json operator+(const nlohmann::json& lhs, const nlohmann::json& rhs)
+NLOHMANN_BASIC_JSON_TPL_DECLARATION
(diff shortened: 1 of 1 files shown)


falbrechtskirchinger:
Please address these issues:
* You did not read the [contribution guideline](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md#files-to-change) and did not run `make amalgamate`.
* Have you considered the implications of such an operator on implicit conversions?
* The operator only works with `nlohmann::json`, not the general `basic_json` type.
* No checks for whether `lhs` or `rhs` are objects.
* Have you looked at [`update()`](https://json.nlohmann.me/api/basic_json/update/)?
* No unit tests.

nlohmann:
I am not convinced of this addition to the API. It is syntactic sugar for a 4-line function that adds a lot of other expectations: What about arrays? What about strings? What about `operator-`?

MastProgs:
Thank you both for your feedback. The nlohman::json code was so good that I came up with it in a simple way, but I think it was my fault that I didn't think about it more deeply. My lack of experience seems to have caused inconvenience to both of you. I will close the current issue.

falbrechtskirchinger:
Remember, you can always open a [discussion](https://github.com/nlohmann/json/discussions) to share any ideas and gather feedback in the future. :-)

PR #3678, opened 2022-08-05
Fix 'const' qualifier on bool& has no effect

falbrechtskirchinger:
Disable the `to_json` overload for `std::vector<bool>::reference` when `std::vector<bool>::reference` is the same as `basic_json::boolean_t&`.

Thanks, @georgthegreat, for pointing out this issue.

Closes #3677.

### DIFF
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -267,9 +267,15 @@ inline void to_json(BasicJsonType& j, T b) noexcept
     external_constructor<value_t::boolean>::construct(j, b);
 }
 
-template<typename BasicJsonType,
-         enable_if_t<std::is_convertible<const std::vector<bool>::reference&, typename BasicJsonType::boolean_t>::value, int> = 0>
-inline void to_json(BasicJsonType& j, const std::vector<bool>::reference& b) noexcept
+template < typename BasicJsonType, typename BoolRef,
+           enable_if_t <
+               ((std::is_same<std::vector<bool>::reference, BoolRef>::value
+                 && !std::is_same <std::vector<bool>::reference, typename BasicJsonType::boolean_t&>::value)
+                || (std::is_same<std::vector<bool>::const_reference, BoolRef>::value
+                    && !std::is_same <detail::uncvref_t<std::vector<bool>::const_reference>,
+                                      typename BasicJsonType::boolean_t >::value))
+               && std::is_convertible<const BoolRef&, typename BasicJsonType::boolean_t>::value, int > = 0 >
+inline void to_json(BasicJsonType& j, const BoolRef& b) noexcept
 {
     external_constructor<value_t::boolean>::construct(j, static_cast<typename BasicJsonType::boolean_t>(b));
 }
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -5500,9 +5500,15 @@ inline void to_json(BasicJsonType& j, T b) noexcept
     external_constructor<value_t::boolean>::construct(j, b);
 }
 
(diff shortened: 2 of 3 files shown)


nlohmann:
Do we have a chance to add a regression test?

falbrechtskirchinger:
Not without external dependencies. And the warning flag seems to be new.

georgthegreat:
This does not work, unfortunately.
clang12 still issues a warning when he sees `const bool&`:

```
include/nlohmann/detail/conversions/to_json.hpp:274:39: error: 'const' qualifier on reference type 'std::vector<bool>::reference' (aka 'bool &') has no effect [-Werror,-Wignored-qualifiers]
inline void to_json(BasicJsonType& j, const std::vector<bool>::reference& b) noexcept
```

falbrechtskirchinger:
Oh, that's because I made at least one mistake.

georgthegreat:
Still no

```
include/nlohmann/detail/conversions/to_json.hpp:272:36: error: 'const' qualifier on reference type 'std::vector<bool>::reference' (aka 'bool &') has no effect [-Werror,-Wignored-qualifiers]
               std::is_convertible<const std::vector<bool>::reference&, typename BasicJsonType::boolean_t>::value
                                   ^~~~~~
include/nlohmann/detail/conversions/to_json.hpp:274:39: error: 'const' qualifier on reference type 'std::vector<bool>::reference' (aka 'bool &') has no effect [-Werror,-Wignored-qualifiers]
inline void to_json(BasicJsonType& j, const std::vector<bool>::reference& b) noexcept
```

falbrechtskirchinger:
I could try flipping those checks, hoping the compiler short-circuits and doesn't get to the `is_convertible` check which seems to trigger the earning.

georgthegreat:
And why const_reference fix proposed #3677 would not work?

falbrechtskirchinger:
In the end, it would still conflict with the existing overload.

falbrechtskirchinger:
One more try. I'm grasping at straws with this one. If it doesn't work, we may just have to suppress the warning for that function.

georgthegreat:
It looks like the latter version will work for us.

Thanks, @falbrechtskirchinger!

gregmarr:
Probably not a bad idea to also do the change from 3677 and use `const_reference`.

falbrechtskirchinger:
No, that doesn't work. `::reference` and `::const_reference` are different types. In fact, `::const_reference` is `bool` in stdlibc++.

Assuming my last fix attempt works, I was going to give `::const_reference` the same treatment.

falbrechtskirchinger:
I decided against dealing with `std::vector<bool>::const_reference` for now. Chances are, I'd make matters worse for everyone instead of fixing things for some hypothetical STL.

shaischaudhry (approved):

falbrechtskirchinger:
And it turns out, that libc++ does use a different type for `const_reference`:
```
no known conversion from 'std::__1::vector<bool, std::__1::allocator<bool> >::const_reference' (aka '__bit_const_reference<std::__1::vector<bool, std::__1::allocator<bool> > >')
```
Guess I'll have to add it after all.

falbrechtskirchinger:
Let me walk you through the `enable_if` constraint:
Since we already have a `to_json` function for `BasicJsonType::boolean_t`, we need to disable the `vector<bool>::reference`/`vector<bool>::const_reference` overload, if:
* `vector<bool>::reference` is the same as `BasicJsonType::boolean_t &`, or
* `vector<bool>::const_reference`, stripped of cvref qualifiers, is the same as `BasicJsonType::boolean_t`.

This covers all tested and reported STLs.

falbrechtskirchinger (commented):
Looks good to me.

nlohmann (approved):
Looks good to me.

PR #3680, opened 2022-08-05
Add `operator<` to , to allow use as a map key

khiner:
This is a replacement for https://github.com/nlohmann/json/pull/3667

After looking at how extensive [the commit for the other equality operators is](https://github.com/nlohmann/json/commit/9e1a7c85e342a5edec6cc3e96087d4070f55bea7), I see how much this PR needs that's not here (tests/docs/etc). I'm more than happy to cargo code/copy-paste and edit based on that commit to make this one look just like that. Lmk if that would be of value and actually helpful and not just a burden to you vs doing it yourself.

### DIFF
--- a/include/nlohmann/detail/json_pointer.hpp
+++ b/include/nlohmann/detail/json_pointer.hpp
@@ -862,6 +862,15 @@ class json_pointer
     {
         return *this == json_pointer(rhs);
     }
+
+    /// @brief compares JSON pointer and string for less-than
+    template<typename RefStringTypeLhs, typename RefStringTypeRhs>
+    friend bool operator<(json_pointer<RefStringTypeLhs> const& lhs,
+                          json_pointer<RefStringTypeRhs> const& rhs) noexcept
+    {
+        return lhs.reference_tokens < rhs.reference_tokens;
+    }
+
 #else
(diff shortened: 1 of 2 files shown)


falbrechtskirchinger:
I'm not too happy with the current state of the `operator==/!=` string overloads, but not having them means we're not compatible with v3.10.5.
Technically, the same is true for the other operators, but I'm hoping that those are rarely used in practice.

Since you're only looking for map support, maybe we should limit this PR to comparing two JSON pointers for less-than and be done.

If @nlohmann doesn't object, I'd also skip the documentation for now. It's actually easier for me to copy and adjust the documentation from `basic_json`/`json_pointer`, once all operators (including `<=>`) are done.

Also, note that I'm considering one more (small) change to the existing operators. There shouldn't be any merge conflicts or only minor ones. (Waiting on feedback before I proceed.)

shaischaudhry (approved):

falbrechtskirchinger:
I added the bare minimum to get map support into 3.11.2. Hope you don't mind. (#3685)

I'll add documentation when I add the remaining operators if Niels allows it.

khiner:
Thanks @falbrechtskirchinger !

PR #3679, opened 2022-08-05
Amalgamate the forward declaration header

falbrechtskirchinger:
As discussed, we need to amalgamate `json_fwd.hpp` to make it usable as a standalone file (again).

Issue #3657, opened 2022-08-03
extern from/to_json result in linker error


steffenb7333:
### Description

I have a library, which declares `from/to_json` as external in its public header as to hide implementation details like this:

```
#if __has_include(<nlohmann/json_fwd.hpp>)
	#include <nlohmann/json_fwd.hpp>
#else
	#include <nlohmann/json.hpp>
#endif
#include "someclass.hpp"
namespace Library
{
  extern void from_json(const nlohmann::json&, SomeClass&);
  extern void to_json(nlohmann::json&, const SomeClass&);
}
```
This is compiled with some version of nlohmann-json as there is seldom a need to update/recompile.

Anyhow, that library is then used by a heavily and actively developed codebase which now switched to nlohmann-json 3.11.1. The result is that the linker cannot find the conversion functions b/c the type nlohmann::json is now `nlohmann::json_v3_11_1_diag::basic_json` for the codebase. And to make matters worse, if I disable diagnostics, it changes again, b/c that is also encoded the namespace name.

Either I have mis-used this library for ages now (without getting any kind of bug/problem) or "abi_macros.hpp" needs some additional thoughts.

### Reproduction steps

Declare external conversion functions.
Use different version of this library.

### Expected vs. actual results

Should link but does not.

### Compiler and operating system

Linux/gcc-12.1.1

### Library version

3.11.1

### Validation

- [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/nlohmann/json/tree/develop) branch is used.
- [ ] I can successfully [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests).

falbrechtskirchinger:
Mixing versions is indeed a misuse. v3.11.0 makes that more explicit.

Mixing `JSON_DIAGNOSTICS` on/off is even worse (and usually results in a segfault anyway).

I'm dealing with some other regressions now and will give this more thought later.

steffenb7333:
If version mixing was always a misuse how can a library include any public API using nlohmann::json without being compiled for all existing nlohmann-json-versions when exposing what is now in a separate compilation unit is not an option?

falbrechtskirchinger:
Mixing different versions in one codebase was always dubious. v3.11.0 made it possible to use different versions **as long as they don't interact**. Meaning, that you can't pass a `basic_json` instance created by a function compiled with version A to a function compiled with version B. ("Version" in this context includes different settings, i.e., `JSON_DIAGNOSTICS`.)

[...]

PR #3683, opened 2022-08-06
Restructure inline namespace and allow version component to be disabled

falbrechtskirchinger:
Restructure the inline namespace, moving the version component to the end and renaming the prefix to `json_abi`.

Example: `nlohmann::json_abi_diag_v3_11_1`

The prefix of the tag namespace cannot be `json` as that would conflict with the type alias `nlohmann::json`.

The version namespace can be disabled by defining `NLOHMANN_JSON_NAMESPACE_NO_VERSION` to `1`.

### DIFF
--- a/docs/mkdocs/docs/api/macros/index.md
+++ b/docs/mkdocs/docs/api/macros/index.md
@@ -32,7 +32,10 @@ header. See also the [macro overview page](../../features/macros.md).
 ## Library namespace
 
 - [**NLOHMANN_JSON_NAMESPACE**](nlohmann_json_namespace.md) - full name of the `nlohmann` namespace
-- [**NLOHMANN_JSON_NAMESPACE_BEGIN**<br>**NLOHMANN_JSON_NAMESPACE_END**](nlohmann_json_namespace_begin.md) - open and close the library namespace
+- [**NLOHMANN_JSON_NAMESPACE_BEGIN**<br>**NLOHMANN_JSON_NAMESPACE_END**](nlohmann_json_namespace_begin.md) - open and
+  close the library namespace
+- [**NLOHMANN_JSON_NAMESPACE_NO_VERSION**](nlohmann_json_namespace_no_version.md) - disable the version component of
+  the inline namespace
 
 ## Type conversions
 
--- a/docs/mkdocs/docs/api/macros/nlohmann_json_namespace.md
+++ b/docs/mkdocs/docs/api/macros/nlohmann_json_namespace.md
@@ -1,19 +1,18 @@
 # NLOHMANN_JSON_NAMESPACE
 
 ```cpp
-#define NLOHMANN_JSON_NAMESPACE
+#define NLOHMANN_JSON_NAMESPACE /* value */
 ```
 
-This macro evaluates to the full name of the `nlohmann` namespace, including  the name of a versioned and ABI-tagged
-inline namespace. Use this macro to unambiguously refer to the `nlohmann` namespace.
+This macro evaluates to the full name of the `nlohmann` namespace.
 
 ## Default definition
 
-The default value consists of a prefix, a version string, and optional ABI tags depending on whether ABI-affecting
-macros are defined (e.g., [`JSON_DIAGNOSTICS`](json_diagnostics.md), and
-[`JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON`](json_use_legacy_discarded_value_comparison.md)).
+The default value consists of the root namespace (`nlohmann`) and an inline ABI namespace. See
+[`nlohmann` Namespace](../../features/namespace.md#structure) for details.
 
-When the macro is not defined, the library will define it to its default value.
+When the macro is not defined, the library will define it to its default value. Overriding this value has no effect on
+the library.
 
 ## Examples
 
@@ -35,7 +34,8 @@ When the macro is not defined, the library will define it to its default value.
 ## See also
 
 - [`NLOHMANN_JSON_NAMESPACE_BEGIN, NLOHMANN_JSON_NAMESPACE_END`](nlohmann_json_namespace_begin.md)
+- [`NLOHMANN_JSON_NAMESPACE_NO_VERSION`](nlohmann_json_namespace_no_version.md)
 
 ## Version history
 
-- Added in version 3.11.0.
+- Added in version 3.11.0. Changed inline namespace name in version 3.11.2.
--- a/docs/mkdocs/docs/api/macros/nlohmann_json_namespace_begin.md
+++ b/docs/mkdocs/docs/api/macros/nlohmann_json_namespace_begin.md
@@ -1,30 +1,35 @@
 # NLOHMANN_JSON_NAMESPACE_BEGIN, NLOHMANN_JSON_NAMESPACE_END
 
 ```cpp
-#define NLOHMANN_JSON_NAMESPACE_BEGIN  // (1)
-#define NLOHMANN_JSON_NAMESPACE_END    // (2)
+#define NLOHMANN_JSON_NAMESPACE_BEGIN /* value */  // (1)
+#define NLOHMANN_JSON_NAMESPACE_END   /* value */  // (2)
 ```
 
-These macros can be used to open and close the `nlohmann` namespace. They include an inline namespace used to
-differentiate symbols when linking multiple versions (including different ABI-affecting macros) of this library.
+These macros can be used to open and close the `nlohmann` namespace. See
+[`nlohmann` Namespace](../../features/namespace.md#structure) for details.
 
 1. Opens the namespace.
-    ```cpp
-    namespace nlohmann
-    {
-    inline namespace json_v3_11_0
-    {
-    ```
-
 2. Closes the namespace.
-    ```cpp
-    }  // namespace nlohmann
-    }  // json_v3_11_0
-    ```
 
 ## Default definition
 
-The default definitions open and close the `nlohmann` as well as an inline  namespace.
+The default definitions open and close the `nlohmann` namespace. The precise definition of
+[`NLOHMANN_JSON_NAMESPACE_BEGIN`] varies as described [here](../../features/namespace.md#structure).
+
+1. Default definition of `NLOHMANN_JSON_NAMESPACE_BEGIN`:
+
+    ```cpp
+    namespace nlohmann
+    {
+    inline namespace json_abi_v3_11_2
+    {
+    ```
+
+2. Default definition of `NLOHMANN_JSON_NAMESPACE_END`:
+    ```cpp
+    }  // namespace json_abi_v3_11_2
+    }  // namespace nlohmann
+    ```
 
 When these macros are not defined, the library will define them to their default definitions.
 
@@ -32,7 +37,7 @@ When these macros are not defined, the library will define them to their default
 
 ??? example
 
-    The example shows an example how to use `NLOHMANN_JSON_NAMESPACE_BEGIN`/`NLOHMANN_JSON_NAMESPACE_END` from the
+    The example shows how to use `NLOHMANN_JSON_NAMESPACE_BEGIN`/`NLOHMANN_JSON_NAMESPACE_END` from the
     [How do I convert third-party types?](../../features/arbitrary_types.md#how-do-i-convert-third-party-types) page.
 
     ```cpp
@@ -47,8 +52,10 @@ When these macros are not defined, the library will define them to their default
 
 ## See also
 
+- [`nlohmann` Namespace](../../features/namespace.md)
 - [NLOHMANN_JSON_NAMESPACE](nlohmann_json_namespace.md)
+- [`NLOHMANN_JSON_NAMESPACE_NO_VERSION`](nlohmann_json_namespace_no_version.md)
 
 ## Version history
 
-- Added in version 3.11.0.
+- Added in version 3.11.0. Changed inline namespace name in version 3.11.2.
--- a/docs/mkdocs/docs/features/namespace.md
+++ b/docs/mkdocs/docs/features/namespace.md
@@ -1,0 +1,93 @@
+# `nlohmann` Namespace
+
+The 3.11.0 release introduced an
+[inline namespace](https://en.cppreference.com/w/cpp/language/namespace#Inline_namespaces) to allow different parts of
+a codebase to safely use different versions of the JSON library as long as they never exchange instances of library
+types.
+
+## Structure
+
+The complete default namespace name is derived as follows:
+
+- The root namespace is always `nlohmann`.
+- The inline namespace starts with `json_abi` and is followed by serveral optional ABI tags according to the value of
+  these ABI-affecting macros, in order:
+    - [`JSON_DIAGNOSTICS`](../api/macros/json_diagnostics.md) defined non-zero appends `_diag`.
+    - [`JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON`](../api/macros/json_use_legacy_discarded_value_comparison.md)
+      defined non-zero appends `_ldvcmp`.
+- The inline namespace ends with the suffix `_v` followed by the 3 components of the version number separated by
+  underscores. To omit the version component, see [Disabling the version component](#disabling-the-version-component)
+  below.
+
+For example, the namespace name for version 3.11.2 with `JSON_DIAGNOSTICS` defined to `1` is:
+
+```cpp
+nlohmann::json_abi_diag_v3_11_2
+```
+
+## Purpose
+
+Several incompatibilities have been observed. Amongst the most common ones is linking code compiled with different
+definitions of [`JSON_DIAGNOSTICS`](../api/macros/json_diagnostics.md). This is illustrated in the diagram below.
+
+```plantuml
+[**nlohmann_json (v3.10.5)**\nJSON_DIAGNOSTICS=0] as [json]
+[**nlohmann_json (v3.10.5)**\nJSON_DIAGNOSTICS=1] as [json_diag]
+[**some_library**] as [library]
+[**application**] as [app]
+
+[library] ..|> [json]
+[app] ..|> [json_diag]
+[app] ..|>[library]
+```
+
+In releases prior to 3.11.0, mixing any version of the JSON library with different `JSON_DIAGNOSTICS` settings would
+result in a crashing application. If `some_library` never passes instances of JSON library types to the application,
+this scenario became safe in version 3.11.0 and above due to the inline namespace yielding distinct symbol names.
+
+## Limitations
+
+Neither the compiler nor the linker will issue as much as a warning when translation units – intended to be linked
+together and that include different versions and/or configurations of the JSON library – exchange and use library
+types.
+
+There is an exception when forward declarations are used (i.e., when including `json_fwd.hpp`) in which case the linker
+may complain about undefined references.
+
+## Disabling the version component
+
+Different versions are not necessarily ABI-incompatible, but the project does not actively track changes in the ABI and
+recommends that all parts of a codebase exchanging library types be built with the same version. Users can, **at their
+own risk**, disable the version component of the linline namespace, allowing different versions – but not
+configurations – to be used in cases where the linker would otherwise output undefined reference errors.
+
+To do so, define [`NLOHMANN_JSON_NAMESPACE_NO_VERSION`](../api/macros/nlohmann_json_namespace_no_version.md) to `1`.
+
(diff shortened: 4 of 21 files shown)


docs/mkdocs/docs/api/macros/nlohmann_json_namespace_begin.md
```diff
@@ -1,38 +1,48 @@
...
 
 1. Opens the namespace.
+2. Closes the namespace.
+
+## Default definition
+
+The default definitions open and close the `nlohmann` as well as two inline namespaces. The tag namespace encodes the
+state of ABI-affecting options ([`JSON_DIAGNOSTICS](json_diagnostics.md) and
+[`JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON`](json_use_legacy_discarded_value_comparison.md)), the version namespace
+the library version number.
```
gregmarr, line 20: The last part of the sentence feels incomplete.  Maybe something like this?

", and the version namespace encodes the library version number."

docs/mkdocs/docs/api/macros/nlohmann_json_namespace_version_namespace.md
```diff
@@ -0,0 +1,46 @@
+# NLOHMANN_JSON_NAMESPACE
+
+```cpp
+#define NLOHMANN_JSON_NAMESPACE_VERSION_NAMESPACE /* value */
+```
+
+This macro evaluates to the name of the version namespace. See [`NLOHMANN_JSON_NAMESPACE`](nlohmann_json_namespace.md).
+If defined to an empty value, the version namespace will be omitted.
+
+## Default definition
```
gregmarr, line 10: This description sounds like `NLOHMANN_JSON_NAMESPACE`, not `NLOHMANN_JSON_NAMESPACE_VERSION_NAMESPACE`

docs/mkdocs/docs/api/macros/nlohmann_json_namespace_version_namespace.md
```diff
@@ -0,0 +1,46 @@
+# NLOHMANN_JSON_NAMESPACE
+
+```cpp
+#define NLOHMANN_JSON_NAMESPACE_VERSION_NAMESPACE /* value */
+```
+
+This macro evaluates to the name of the version namespace. See [`NLOHMANN_JSON_NAMESPACE`](nlohmann_json_namespace.md).
+If defined to an empty value, the version namespace will be omitted.
```
gregmarr, line 8: Can we do just `NLOHMANN_JSON_NAMESPACE_EXCLUDE_VERSION` and if it's defined to 1, we leave out the version?

docs/mkdocs/docs/api/macros/nlohmann_json_namespace_version_namespace.md
falbrechtskirchinger: No. That is accurate.

But it might change because determining if a preprocessor macro is defined but empty is near impossible. \*sigh\*

docs/mkdocs/docs/api/macros/nlohmann_json_namespace_version_namespace.md
falbrechtskirchinger: That might be the easier option. I just wanted to avoid yet another page of documentation. Although, I think I can document `NLOHMANN_JSON_NAMESPACE_EXCLUDE_VERSION` on the same page.

I'll think about it some more.

falbrechtskirchinger: Thanks. I needed someone to make me stop being stubborn and think for a second. :-)

gregmarr: I was assuming that this would replace this one.

gregmarr:
What is the benefit of having the version part a separate namespace that may or may not be used instead of just a part of the name that may or may not be there?

docs/mkdocs/docs/api/macros/nlohmann_json_namespace_version_namespace.md
falbrechtskirchinger: Yes, exactly.

falbrechtskirchinger:
It just expresses the hierarchy between the truly (and always) ABI-breaking settings and library versions. I don't really know yet if that's useful, but I don't think it hurts either.

falbrechtskirchinger:
ci_test_single_header failing is expected until #3679 is merged.

falbrechtskirchinger (commented):
Looks good to me.

nlohmann:
Is there a way to get back to the original behavior and just have a `nlohmann` namespace with `basic_json` right in it without inline namespaces?

falbrechtskirchinger:
Yes. Since 3.11.0.
https://github.com/nlohmann/json/issues/3657#issuecomment-1207175931

docs/mkdocs/docs/api/macros/nlohmann_json_namespace.md
```diff
@@ -1,27 +1,32 @@
...
 ## Default definition
 
-The default value consists of a prefix, a version string, and optional ABI tags depending on whether ABI-affecting
-macros are defined (e.g., [`JSON_DIAGNOSTICS`](json_diagnostics.md), and
-[`JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON`](json_use_legacy_discarded_value_comparison.md)).
+The default value consists of the root namespace, the tag namespace, and the version namespace.
 
-When the macro is not defined, the library will define it to its default value.
+- The root namespace is always `nlohmann`.
+- The tag namespace starts with `nlohmann_json` and is followed by several optional ABI tags according to the value of
```
nlohmann, line 14: Why not just `json` as start of the tag namespace`

docs/mkdocs/docs/api/macros/nlohmann_json_namespace.md
```diff
@@ -1,27 +1,32 @@
...
+  these ABI-affecting macros, in order:
+    - [`JSON_DIAGNOSTICS`](json_diagnostics.md) defined non-zero appends `_diag`.
+    - [`JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON`](json_use_legacy_discarded_value_comparison.md) defined non-zero
+      appends `_ldvcmp`.
+
```
nlohmann, line 19: I would expect a sentence on the version name space here, because the initial sentence speaks of three parts and this list only has two items.

tools/gdb_pretty_printer/nlohmann-json.py
```diff
@@ -1,7 +1,7 @@
 import gdb
 import re
 
-ns_pattern = re.compile(r'nlohmann::json_v(?P<v_major>\d+)_(?P<v_minor>\d+)_(?P<v_patch>\d+)(?P<tags>\w*)::(?P<name>.+)')
+ns_pattern = re.compile(r'nlohmann(::nlohmann_json(?P<tags>\w*))?(::v(?P<v_major>\d+)_(?P<v_minor>\d+)_(?P<v_patch>\d+))?::(?P<name>.+)')
```
nlohmann, line 4: I think these changes do not belong to this PR.

docs/mkdocs/docs/api/macros/nlohmann_json_namespace.md
falbrechtskirchinger: That would conflict with the type alias `nlohmann::json`.

docs/mkdocs/docs/api/macros/nlohmann_json_namespace.md
falbrechtskirchinger: Right.

tools/gdb_pretty_printer/nlohmann-json.py
falbrechtskirchinger: Without it, the GDB pretty printer isn't usable, but I can factor it out if you prefer.

nlohmann: Oh sorry, I did not look at the code any only assumed that this was not meant for here.

falbrechtskirchinger:
I've added a first, rough version of `features/namespace.md`. Clearly needs more polish in the future, otherwise, I'll incorporate any suggestions you may have.

docs/mkdocs/docs/features/namespace.md
```diff
@@ -0,0 +1,52 @@
...
+The 3.11.0 release introduced an inline namespace to allow different parts of a codebase to use different versions of
+the library as long as they never exchange instances of library types.
+
+In the 3.11.2 release, that namespace was split into the tag namespace – encoding the state of ABI-affecting macros –
+and the version namespace – encoding the version of the library.
+
+The full default namespace name is derived as follows:
+
+- The root namespace is always `nlohmann`.
+- The tag namespace starts with `nlohmann_json` and is followed by serveral optional ABI tags according to the value of
```
gregmarr, line 12: Just thinking out loud here.  I wonder if something like `features` (or `abi_tags`) instead of `nlohmann_json` would make it clearer what is being enumerated here.  `nlohmann::features_diag_ldvcmp::v3_11_2` for example.  Might look a bit weird with no features though: `nlohmann::features::v3_11_2`.  Would it make sense to exclude this namespace if no features were enabled?  That combined with excluding the version namespace would result in just the `nlohmann` namespace as we had before, might not be what we want.

nlohmann: I don't have an opinion about whether I like `features`, `abi_tags` or anything else more - but I do like the idea of "collapsing" everything back to `nlohmann::` when ABI tags and versions are switched off. This way, the new approach feels like an organic addition rather than a drastic change than needs much explanation.

(There, I think `abi_tags` does feel like the better prefix...)

docs/mkdocs/docs/api/macros/nlohmann_json_namespace_begin.md
```diff
@@ -1,38 +1,47 @@
...
 ```cpp
-#define NLOHMANN_JSON_NAMESPACE_BEGIN  // (1)
-#define NLOHMANN_JSON_NAMESPACE_END    // (2)
+#define NLOHMANN_JSON_NAMESPACE_BEGIN /* value */  // (1)
+#define NLOHMANN_JSON_NAMESPACE_END   /* value */  // (2)
 ```
 
-These macros can be used to open and close the `nlohmann` namespace. They include an inline namespace used to
-differentiate symbols when linking multiple versions (including different ABI-affecting macros) of this library.
+These macros can be used to open and close the `nlohmann` namespace. They include two inline namespaces used to
```
nlohmann, line 8: Does "They" stand for "These macros"? As in

> These macros include two inline namespaces

? This confuses me.

docs/mkdocs/docs/features/namespace.md
falbrechtskirchinger: Originally, I wanted `json` to be part of the name, to make it clear what library the types belong to. This is ultimately not important, as the types still appear in the top-level namespace.

While it would be quite elegant to collapse the inline namespaces if the version is disabled and no ABI-altering options are enabled, it would also eliminate the whole point of the tag namespace. A 3.11.x+ version of the library with the aforementioned configuration is potentially different from an old version with an unknown configuration (because it is not part of the symbol names). To prevent linking errors in this scenario, the tag namespace should still be part of the symbol names.

What I can be convinced of is to undo the split of the namespace, which would result in the following hybrid:
```
nlohmann::nlohmann_json[_diag][_ldvcmp][_v3_11_2]
```
I've split the namespace into the current hierarchy in case there's some use for it later but I've not thought of anything yet. That would also simplify the `abi_macros.hpp` code a little.

Do we want an easier option to turn off that inline namespace?
`NLOHMANN_JSON_NAMESPACE_SIMPLE=1` perhaps?
Also, `NLOHMANN_JSON_NAMESPACE_EXCLUDE_VERSION` could be shortened to `NLOHMANN_JSON_NAMESPACE_NO_VERSION`.

docs/mkdocs/docs/api/macros/nlohmann_json_namespace_begin.md
falbrechtskirchinger: Yes, as in the opening macro contains two `inline namespace <NAME> {` parts and the closing macro the matching curly brackets. I may have to rewrite it anyway, let's see.

docs/mkdocs/docs/features/namespace.md
falbrechtskirchinger: I'm leaning towards not making turning off the inline namespaces any easier.

gregmarr: I think this is probably the clearest as to the reason for the namespace existing, and not duplicating the initial `nlohmann` tag, if we can't come up with a reason to have version as a separate namespace.

    nlohmann::abi_tags[_diag][_ldvcmp][_v3_11_2]

It does however seem weird to end up with just this:

    nlohmann::abi_tags

So maybe `_diag` isn't `""/"_diag"` but `_nodiag/_diag`?  That way there's always at least `nlohmann::abi_tags_nodiag` or `nlohmann::abi_tags_diag` (with similar for `_ldvcmp`?)

falbrechtskirchinger: How about `json_abi` instead of `nlohmann_json`? Keeps the name a bit shorter with no options enabled and is less weird than `abi_tags` alone.

gregmarr: Sure.

falbrechtskirchinger: Any objections about shortening `NLOHMANN_JSON_NAMESPACE_EXCLUDE_VERSION` to `NLOHMANN_JSON_NAMESPACE_NO_VERSION`?

gregmarr: No objections.

nlohmann: Go ahead!

docs/mkdocs/docs/features/namespace.md
```diff
@@ -0,0 +1,52 @@
...
+- The version namespace consists of the prefix `v` followed by the 3 components of the version number separated by
+  underscores.
+
+For example, the namespace name for version 3.11.2 with `JSON_DIAGNOSTICS` defined to `1`, is:
+
+```cpp
+nlohmann::nlohmann_json_diag::v3_11_2
+```
+
+## Disabling the version namespace
```
nlohmann, line 26: I had another thought on the documentation.

The current documentation describes what the namespace is and how to manipulate the default behavior. Maybe it should be structured as follows:

- describe that starting in 3.11.0 we changed the namespace from `nlohmann` to a more elaborate version
- describe why this is a benefit - maybe by explicitly listing the incompatibilities that were possible before, but would not have been caught before; maybe also mention `JSON_SKIP_LIBRARY_VERSION_CHECK`
- describe how the namespace is structured
- describe the 3 different styles the namespace can be configured:
	- default - note: this is fine until we have a better means to describe ABI compatibility (right?)
	- no version part - note: this is fine if the library user wants to take the risk of mixing libraries with different versions, but with same ABI tags
	- `nlohmann` - note: use at your own risk

What do you think?

falbrechtskirchinger: I already had planned on addressing this, so thanks for the outline.

falbrechtskirchinger:
Here's how the `plantuml` diagram in `features/namespace.md` renders:

namespace md_uml

docs/mkdocs/docs/features/namespace.md
falbrechtskirchinger: I didn't quite adopt the outline and made a note to take another look and incorporate more of your ideas.

For now, I think what I have is as good as it's going to get in 2 passes.

docs/mkdocs/docs/api/macros/index.md
```diff
@@ -32,7 +32,9 @@ header. See also the [macro overview page](../../features/macros.md).
 ## Library namespace
 
 - [**NLOHMANN_JSON_NAMESPACE**](nlohmann_json_namespace.md) - full name of the `nlohmann` namespace
-- [**NLOHMANN_JSON_NAMESPACE_BEGIN**<br>**NLOHMANN_JSON_NAMESPACE_END**](nlohmann_json_namespace_begin.md) - open and close the library namespace
+- [**NLOHMANN_JSON_NAMESPACE_BEGIN**<br>**NLOHMANN_JSON_NAMESPACE_END**](nlohmann_json_namespace_begin.md) - open and
+  close the library namespace
+- [**NLOHMANN_JSON_NAMESPACE_NO_VERSION**](nlohmann_json_namespace_no_version.md) - disable the version namespace
```
falbrechtskirchinger, line 37: "version component"

docs/mkdocs/docs/api/macros/nlohmann_json_namespace_begin.md
```diff
@@ -1,38 +1,43 @@
...
-differentiate symbols when linking multiple versions (including different ABI-affecting macros) of this library.
+These macros can be used to open and close the `nlohmann` namespace. See
+[`nlohmann` Namespace](../../features/namespace.md#structure) for details.
 
 1. Opens the namespace.
+2. Closes the namespace.
+
+## Default definition
+
+The default definitions open and close the `nlohmann` namespace. The precise definitions vary as described
```
falbrechtskirchinger, line 16: Technically, only `_BEGIN` varies. :man_shrugging:

docs/mkdocs/docs/features/namespace.md
```diff
@@ -0,0 +1,92 @@
...
+this scenario became safe in version 3.11.0 and above due to the inline namespace yielding distinct symbol names.
+
+## Limitations
+
+Neither the compiler nor the linker will issue as much as a warning when translation units – intended to be linked
+together and that include different versions and/or configurations of the JSON library – exchange and use library
+types.
+
+There is an exception when forward declarations are used (i.e., when including `json_fwd.hpp`) in which case the linker
+may complain about undefined references.
```
falbrechtskirchinger, line 54: I want to extend this part with an example but don't have it in me right now. :-)

docs/mkdocs/docs/features/namespace.md
```diff
@@ -0,0 +1,92 @@
...
+
+Different versions are not necessarily ABI-incompatible, but the project does not actively track changes in the ABI and
+recommends that all parts of a codebase exchanging library types be built with the same version. Users can, **at their
+own risk**, disable the version component of the linline namespace, allowing different versions – but not
+configurations – to be used in cases where the linker would otherwise output undefined reference errors.
+
+To do so, define [`NLOHMANN_JSON_NAMESPACE_NO_VERSION`](../api/macros/nlohmann_json_namespace_no_version.md) to `1`.
+
+This applies to version 3.11.2 and above only, versions 3.11.0 and 3.11.1 can apply the technique described in the next
+section to emulate the effect of the `NLOHMANN_JSON_NAMESPACE_NO_VERSION` macro.
```
falbrechtskirchinger, line 66: I did not provide an example here as this will hopefully be an exceedingly rare use case.

docs/mkdocs/docs/features/namespace.md
```diff
@@ -0,0 +1,92 @@
...
+
+The complete default namespace name is derived as follows:
+
+- The root namespace is always `nlohmann`.
+- The inline namespace starts with `json_abi` and is followed by serveral optional ABI tags according to the value of
+  these ABI-affecting macros, in order:
+    - [`JSON_DIAGNOSTICS`](../api/macros/json_diagnostics.md) defined non-zero appends `_diag`.
+    - [`JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON`](../api/macros/json_use_legacy_discarded_value_comparison.md)
+      defined non-zero appends `_ldvcmp`.
+- The inline namespace ends with the suffix `_v` followed by the 3 components of the version number separated by
```
gregmarr, line 18: A comment here about this version string being potentially disabled?

falbrechtskirchinger (commented):
Minor doc issue rest looks good to me.

* ABI test produces namespace names as expected.
* GDB pretty printer works with the new namespace layout.
* mkdocs pass style check.
* Examples outputs regenerated.
* docset index is complete (minus operator pages already in `upstream/develop`).
* Did I forget anything?

docs/mkdocs/docs/features/namespace.md
falbrechtskirchinger: Can do.

gregmarr (approved):

nlohmann (commented):
Looks good to me.

PR #3684, opened 2022-08-07
Deprecate json_pointer/string_t comparisons

falbrechtskirchinger:
#3664 restored comparing `json_pointer` with strings but removed `noexcept`. This is due to no longer comparing by converting the `json_pointer` to string, but by converting the string to a `json_pointer`. Doing so is more accurate (\*) as we no longer allow comparing with strings that aren't valid JSON pointers. The creation of the `json_pointer` temporary is hidden from the user. For that reason I suggest deprecating the string overloads to make users aware of the implications (exceptions, allocation of a `json_pointer`).

\* The accuracy problem is of greater concern for the remaining relational operators, e.g., imagine comparing a `json_pointer` and the string `/foo~` for less-/greater-than.


gregmarr (approved):

nlohmann (approved):
Looks good to me.

PR #3694, opened 2022-08-10
Release draft 3.11.2

nlohmann:
Should be the final state of 3.11.2.


falbrechtskirchinger (approved):
Looks good to me.

docs/examples/basic_json__CompatibleType.output
```diff
@@ -2,7 +2,7 @@
 {"one":1,"three":3,"two":2}
 {"one":1.2,"three":3.4,"two":2.3}
 {"one":true,"three":false,"two":true}
-{"one":true,"three":true,"two":true}
+{"one":true,"three":false,"two":true}
```
gregmarr, line 5: Is it expected that 3.11.0 changed which of the two `"three"` elements were included, and 3.11.2 changed it back?

gregmarr (commented):
That's a lot of version numbers!

docs/examples/basic_json__CompatibleType.output
falbrechtskirchinger: Depends on the STL. :man_shrugging:

gregmarr: Just wondering if the change is an indication of a change in the library, or a change in a CI setting, or if this was a manual change, or if we can expect it to flip-flop in the future.

falbrechtskirchinger: This was generated with libstdc++ (going by `meta.output`) and not libc++.

falbrechtskirchinger: Expect it to flip-flop based on what toolchain Niels uses.

gregmarr: So this file is generated by the release process, and it's reasonable for it to change from time to time during that process.

nlohmann: Sorry about this inconvenience. I had to adjust the release script several times, and it  is still a bit shaky. The examples did not work out of the box so I had to set GCC as compiler. Then there was an update between the releases. Even worse, the behavior differs when I execute the script in my IDE vs. in my terminal (because the paths differs). I hope to get this fixed soon, but for the time being, the behavior will likely stay this way.

gregmarr: As long as it's explainable, that's fine by me.

PR #3695, opened 2022-08-10
FreeBSD (or probably all Linux distributions) compilation fix

Thorek777:
When you try to compile JSON on FreeBSD (for example 12.1) with C++20, GCC10 and libraries like:
DevIL, Cryptopp and MySQL (Oracle) you will get 'no right operand' error.


gregmarr:
What compiler are you using?  Do you have a full sample that shows this issue without any third party libraries?  This library is built regularly on many Linux distributions without an error.

Thorek777:
> What compiler are you using? Do you have a full sample that shows this issue without any third party libraries? This library is built regularly on many Linux distributions without an error.

So that's may be probably my problem. I'm sorry for mistake!

gregmarr:
Yeah, it's probably one of the other libraries causing the issue.  You might be able to change your include order.

PR #3697, opened 2022-08-11
Add Python script for generating Natvis file and update file for 3.11.2

### LINKED ISSUE #3696: MSVC natvis visualizer does not work after introduction of inline ABI namespace
### Description

#844 added a debugger visualizer for Visual Studio.
#3590 introduced an inline namespace under which the basic_json type now lives.

Therefore the type name pattern in the natvis file https://github.com/nlohmann/json/blob/0e61ee8b074ecb40be6439c7470764dc6c3e0743/nlohmann_json.natvis#L3 does not match the actual type anymore which means the visualizer does not work anymore.

With the inline namespace regularly changing, it might be worthwhile to dynamically generate the natvis file from CMake using the same "algorithm" for the type name pattern as implemented with the macros in the source code.

### Reproduction steps

Create a `nlohmann::json` instance, fill it with some data, and break with Visual Studio debugger to inspect the object instance.

Actual: visualizer not active
Expected: visualizer provides pretty printing

### Minimal code example

_No response_

### Error messages

_No response_

Visual Studio

### Library version

3.11.1

### Validation

falbrechtskirchinger:
I updated GDB pretty printer for the new inline namespace but forgot Natvis. This PR adds a small Python script to generate the Natvis file from a Jinja2 template and generates the file for the 3.11.2 release. In the future, the release script should call `./tools/generate_natvis/generate_natvis.py --version X.Y.Z .` (the trailing dot denoting the output directory).

Fixes #3696.

### DIFF
--- a/nlohmann_json.natvis
+++ b/nlohmann_json.natvis
@@ -1,5 +1,9 @@
+<!-- * * * * * * * * AUTO-GENERATED FILE  * * * * * * * * -->
+<!-- Edit ./tools/generate_natvis/nlohmann_json.natvis.j2 -->
+<!-- * * * * * * * * AUTO-GENERATED FILE  * * * * * * * * -->
 <?xml version="1.0" encoding="utf-8"?>
 <AutoVisualizer xmlns="http://schemas.microsoft.com/vstudio/debugger/natvis/2010">
+    <!-- Namespace nlohmann -->
     <Type Name="nlohmann::basic_json&lt;*&gt;">
         <DisplayString Condition="m_type == nlohmann::detail::value_t::null">null</DisplayString>
         <DisplayString Condition="m_type == nlohmann::detail::value_t::object">{*(m_value.object)}</DisplayString>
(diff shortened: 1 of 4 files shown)


nlohmann_json.natvis
```diff
@@ -20,13 +24,253 @@
...
+         Only works in VS 2015 Update 2 and beyond using the new visualization -->
     <Type Name="std::pair&lt;*, nlohmann::basic_json&lt;*&gt;&gt;" IncludeView="MapHelper">
         <DisplayString>{second}</DisplayString>
         <Expand>
             <ExpandedItem>second</ExpandedItem>
         </Expand>
     </Type>
 
+    <!-- Namespace nlohmann::json_abi -->
+    <Type Name="nlohmann::json_abi::basic_json&lt;*&gt;">
```
gregmarr, line 37: Is it possible to use wildcards for the namespace?  `nlohmann::*::basic_json...`?  If so, could we also hardcode the values for `nlohmann::json_abi::detail::value_t::*`?  This would allow a single natvis file that would work for any version?

Otherwise, would it make sense to include the version number in the filename so that users could have multiple version files installed?

falbrechtskirchinger: Wildcards are only supported for template parameters according to the documentation.

Users will have to comment on whether versioning the Natvis file would be helpful. (ping @jheydebrand)

gregmarr: >Wildcards are only supported for template parameters according to the documentation.

I saw documentation that said that they worked for templates, but it didn't specifically say that it was only used there, so I wasn't sure if it was limited to templates or just poorly documented.

falbrechtskirchinger: > The fully qualified name of each Type element is specified in its Name attribute.

and

> ### Templated classes
> The Name attribute of the Type element accepts an asterisk * as a wildcard character that can be used for templated class names.
>
> In the following example, the same visualization is used whether the object is a CAtlArray<int> or a CAtlArray<float>. If there's a specific visualization entry for a CAtlArray<float>, then it takes precedence over the generic one.

The very specific wording suggests to me it only works for templates. :man_shrugging:

falbrechtskirchinger:
@nlohmann This should be included in 3.11.2 if possible. I could open a PR with just the updated Natvis file if you prefer.

nlohmann_json.natvis
gregmarr: Confirmed that it only works for templates. :(

jheydebrand: Thanks for the swift responses.

Versioning the natvis file would not provide any benefit for me.
When consuming the library via its CMake integration the correct / matching natvis file will be included anyway and I have no desire to copy the file to some system-wide directory (which is supported by Visual Studio and might have been the use-case for versioned natvis files you were thinking about?).

gregmarr (approved):

nlohmann (approved):
Looks good to me.

PR #3702, opened 2022-08-14
Add migration guide

nlohmann:
The migration guide collects all steps required to make the code future-proof for upcoming releases.


.github/workflows/labeler.yml
```diff
@@ -13,6 +13,6 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: srvaroa/labeler@master
+      - uses: falbrechtskirchinger/labeler@master
```
nlohmann, line 16: This can be reverted now.

docs/mkdocs/docs/integration/migration_guide.md
```diff
@@ -0,0 +1,248 @@
+# Migration Guide
+
+This page collects some guidelines how to future-proof your code for future versions of this library.
```
falbrechtskirchinger:
```suggestion
This page collects some guidelines on how to future-proof your code for future versions of this library.
```

docs/mkdocs/docs/integration/migration_guide.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Replace deprecated functions
+
+The following functions have been deprecated in earlier versions and will be removed in the next major version (i.e., 
+4.0.0). All deprecations are annotated with
+[`HEDLEY_DEPRECATED_FOR`](https://nemequ.github.io/hedley/api-reference.html#HEDLEY_DEPRECATED_FOR) to report which
+function to use instead.
```
falbrechtskirchinger: "in earlier versions" seems weird if the functions are being deprecated in this version and is already implied by the tense.

```suggestion
The following functions have been deprecated and will be removed in the next major version (i.e., 4.0.0). All
deprecations are annotated with
[`HEDLEY_DEPRECATED_FOR`](https://nemequ.github.io/hedley/api-reference.html#HEDLEY_DEPRECATED_FOR) to report which
function to use instead.
```

docs/mkdocs/docs/integration/migration_guide.md
```diff
@@ -0,0 +1,248 @@
...
+          bool ok = nlohmann::json::accept(s, s + std::strlen(s));
+          ```
+
+#### JSON Pointers
+
+- Comparing JSON Pointers with strings via [`operator==`](../api/json_pointer/operator_eq.md) and
+  [`operator!=`](../api/json_pointer/operator_ne.md) is deprecated since 3.11.2. To compare a
+  [`json_pointer`](../api/json_pointer/index.md) `p` with a string `s`, convert `s` to a `json_pointer` first and use
+  [`json_pointer::operator==`](../api/json_pointer/operator_eq.md) or
+  [`json_pointer::operator!=`](../api/json_pointer/operator_ne.md).
```
falbrechtskirchinger, line 60: This may be confusing.

The member `operator==` does not replace the non-member operators due to deprecation but due to the used C++ standard version.

Until C++20 the non-member operators `operator==(json_pointer, string_t)` and `operator==(string_t, json_pointer)` are deprecated for `operator==(json_pointer, json_pointer)`. (Same applies to `operator!=`.)

Since C++20 the member operator `json_pointer::operator==(string_t)` is deprecated for `json_pointer::operator==(json_pointer)`. `json_pointer::operator!=` does not exist in the source code but is synthesized by the compiler by rewriting the expression using `json_pointer::operator==`, i.e., `a == b` is rewritten as `!(a == b)`.

This is by no means necessary but could reduce code size minimally. (Someone – me? – should verify that on Compiler Explorer.)

docs/mkdocs/docs/integration/migration_guide.md
```diff
@@ -0,0 +1,248 @@
...
+          ```cpp
+          using my_json = nlohmann::basic_json<std::map, std::vector, my_string_type>;
+          nlohmann::json_pointer<my_json> ptr("/foo/bar/1");
+          ```
+  
+    === "Future-proof"
+  
+          ```cpp
+          nlohmann::json_pointer<my_string_type> ptr("/foo/bar/1");
+          ```
```
falbrechtskirchinger, line 109: You could also mention, that `nlohmann::my_json::json_pointer` is an alias for `nlohmann::json_pointer<my_string_type>` and is always an alias to the `json_pointer` with the appropriate string type for all specializations of `basic_json`.

docs/mkdocs/docs/integration/migration_guide.md
```diff
@@ -0,0 +1,248 @@
...
+
+- The legacy comparison behavior for discarded values is deprecated since 3.11.0. It is already disabled by default and
+  can still be enabled by defining
+  [`JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON`](../api/macros/json_use_legacy_discarded_value_comparison.md).
+  
+    === "Deprecated"
+
+          ```cpp
+          #define JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON
```
falbrechtskirchinger:
```suggestion
  [`JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON`](../api/macros/json_use_legacy_discarded_value_comparison.md) to `1`.
  
    === "Deprecated"

          ```cpp
          #define JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON 1
```

docs/mkdocs/docs/integration/migration_guide.md
```diff
@@ -0,0 +1,248 @@
...
+      nlohmann::json j = "Hello, world!";
+      std::string s = j;
+      ```
+
+=== "Future-proof"
+
+      ```cpp
+      nlohmann::json j = "Hello, world!";
+      auto s = j.get<std::string>();
+      ```
```
falbrechtskirchinger, line 191: You could also show `get_to` as an alternative and maybe add a "See also" for `get_ref` and `get_ptr` in the surrounding text.

docs/mkdocs/docs/integration/migration_guide.md
```diff
@@ -0,0 +1,248 @@
...
+
+You can prepare existing code by already defining
+[`JSON_USE_IMPLICIT_CONVERSIONS`](../api/macros/json_use_implicit_conversions.md) to `0` and replace any implicit
+conversions with calls to [`get`](../api/basic_json/get.md).
```
falbrechtskirchinger, line 203: Should this come before the examples? Same applies to the other sections.

docs/mkdocs/docs/integration/migration_guide.md
```diff
@@ -0,0 +1,248 @@
...
+
+To prepare existing code, define [`JSON_USE_GLOBAL_UDLS`](../api/macros/json_use_global_udls.md) to `0` and bring the
+string literals into scope where needed.
+
+## Do not hard-code the complete library namespace
+
+The [`nlohmann` namespace](../features/namespace.md) contains two sub-namespaces to avoid problems when different
+versions or configurations of the library are used in the same project. Always use `nlohmann` as namespace or, when the
+exact version and configuration is relevant, use macro
+[`NLOHMANN_JSON_NAMESPACE`](../api/macros/nlohmann_json_namespace.md) to denote the namespace.
```
falbrechtskirchinger: This section is tricky. I keep oscillating on what is necessary and when.

I'll expand on this later.

(Also, "two sub-namespaces" is definitely wrong now.)

falbrechtskirchinger (changes requested):

docs/mkdocs/docs/integration/migration_guide.md
gregmarr: This should be 1 not 0, right?

falbrechtskirchinger: Fixed. Thanks!

.github/workflows/labeler.yml
nlohmann:
```suggestion
      - uses: srvaroa/labeler@master
```

nlohmann:
What do you think of a preprocessor symbol like `JSON_PEDANTIC` or `JSON_FUTURE_PROOF` that removes (or adds `#error` directives to) all deprecated functions or discouraged behavior like implicit conversions? This could be a mode that people could use to have a rough idea whether their code would be still usable in a future version=

nlohmann:
What do you think of a preprocessor symbol like `JSON_PEDANTIC` or `JSON_FUTURE_PROOF` that removes (or adds `#error` directives to) all deprecated functions or discouraged behavior like implicit conversions? This could be a mode that people could use to have a rough idea whether their code would be still usable in a future version?

gregmarr:
That sounds like a good idea.  Maybe a JSON_REMOVE_DEPRECATED_FUNCTIONS for the functions, and then a single define that sets all of the macros that remove deprecated behaviors, such as implicit conversions and the compare behavior.

nlohmann:
Similar to this: https://github.com/getsentry/responses#deprecations-and-migration-path

docs/mkdocs/docs/integration/migration_guide.md
nlohmann: Done.

docs/mkdocs/docs/integration/migration_guide.md
nlohmann: Done.

docs/mkdocs/docs/integration/migration_guide.md
nlohmann: Done.

docs/mkdocs/docs/integration/migration_guide.md
nlohmann: I fixed the "two sub-namespaces". No idea about the rest.

docs/mkdocs/docs/integration/migration_guide.md
nlohmann: I'm also confused. Can you propose a better wording for the paragraph?

nlohmann:
> That sounds like a good idea. Maybe a JSON_REMOVE_DEPRECATED_FUNCTIONS for the functions, and then a single define that sets all of the macros that remove deprecated behaviors, such as implicit conversions and the compare behavior.

I will not do this in this PR.

docs/mkdocs/docs/integration/migration_guide.md
```diff
@@ -0,0 +1,260 @@
...
+          bool ok = nlohmann::json::accept(s, s + std::strlen(s));
+          ```
+
+#### JSON Pointers
+
+- Comparing JSON Pointers with strings via [`operator==`](../api/json_pointer/operator_eq.md) and
+  [`operator!=`](../api/json_pointer/operator_ne.md) is deprecated since 3.11.2. To compare a
+  [`json_pointer`](../api/json_pointer/index.md) `p` with a string `s`, convert `s` to a `json_pointer` first and use
+  [`json_pointer::operator==`](../api/json_pointer/operator_eq.md) or
+  [`json_pointer::operator!=`](../api/json_pointer/operator_ne.md).
```
falbrechtskirchinger, line 60:
```suggestion
- Comparing JSON pointers with strings using the string overload of [`operator==`](../api/json_pointer/operator_eq.md)
  and [`operator!=`](../api/json_pointer/operator_ne.md) is deprecated since 3.11.2. To compare a
  [`json_pointer`](../api/json_pointer/index.md) `p` with a string `s`, convert `s` to a `json_pointer` first, then
  compare the two JSON pointers.
```

The C++20 change is an implementation detail. It has no bearing on the types involved in the comparison.

docs/mkdocs/docs/integration/migration_guide.md
falbrechtskirchinger: I hesitate because of the advice given here:
https://json.nlohmann.me/features/arbitrary_types/#how-do-i-convert-third-party-types

Currently, I think this section is OK (i.e., the issue is resolved for the purpose of this PR) but I'm conflicted about the advice to use `NLOHMAN_JSON_NAMESPACE_BEGIN` when specializing `adl_serializer`.

PR #3723, opened 2022-09-04
Replace limit macros with std::numeric_limits

### LINKED ISSUE #3722: INT64_MIN/MAX not defined for newer g++
### Description

While cross compiling for arm w/ newer versions of g++ INT64_MAX, INT64_MIN, UINT64, INT32_MIN, INT32_MAX, UINT32_MAX aren't defined.  I found the following:
/* The ISO C99 standard specifies that in C++ implementations these
   macros should only be defined if explicitly requested.  */
#if !defined __cplusplus || defined __STDC_LIMIT_MACROS

Rather than adding the #define to make this work I created a patch that uses the std::numeric_limits<int64_t>::max() 
I guess a "compatibility" include file could be created to work around this issue; but it wasn't obvious to me how to do this.

After applying the patch I ran all the tests and they all passed.
[int64_min_max.patch.txt](https://github.com/nlohmann/json/files/9483669/int64_min_max.patch.txt)

### Reproduction steps

git clone https://github.com/nlohmann/json.git
cd json
git checkout v3.11.2
mkdir build
cd build
cmake -DCMAKE_TOOLCHAIN_FILE=../../aarch64.cmake -DCMAKE_INSTALL_PREFIX=/opt/nxpimx8/usr/local ..
make

See compiler errors below.

### Minimal code example

_No response_

### Error messages

```Shell
json/tests/src/unit-to_chars.cpp:466:23: error: 'INT64_MIN' was not declared in this scope
  466 |         check_integer(INT64_MIN, "-9223372036854775808");
      |                       ^~~~~~~~~
/home/priestwilliaml/newbuild/build/json/tests/src/unit-to_chars.cpp:467:23: error: 'INT64_MAX' was not declared in this scope
  467 |         check_integer(INT64_MAX, "9223372036854775807");
      |                       ^~~~~~~~~

json/tests/src/unit-cbor.cpp: In function 'void _DOCTEST_ANON_FUNC_7()':
/home/priestwilliaml/newbuild/build/json/tests/src/unit-cbor.cpp:176:39: error: 'INT64_MIN' was not declared in this scope
  176 |                     numbers.push_back(INT64_MIN);
      |                                       ^~~~~~~~~

json/tests/src/unit-msgpack.cpp:511:39: error: 'INT64_MIN' was not declared in this scope
  511 |                     numbers.push_back(INT64_MIN);
      |                                       ^~~~~~~~~

json/tests/src/unit-regression1.cpp: In function 'void _DOCTEST_ANON_FUNC_7()':
/home/priestwilliaml/newbuild/build/json/tests/src/unit-regression1.cpp:880:51: error: 'INT64_MIN' was not declared in this scope
  880 |         CHECK(j1.get<json::number_integer_t>() == INT64_MIN);
      |                                                   ^~~~~~~~~
```

linux g++ 11.2.0 crosscompiling for ARM 64

### Library version

v3.11.2 from github

### Validation

falbrechtskirchinger:
Replace uses of `INT_MIN`/`INT_MAX`, etc. with `std::numeric_limits` and consistently use `std` namespaced integer types.

Fixes #3722.

### DIFF
--- a/tests/src/unit-windows_h.cpp
+++ b/tests/src/unit-windows_h.cpp
@@ -7,6 +7,8 @@
 // SPDX-License-Identifier: MIT
 
 #include "doctest_compatibility.h"
+#undef WIN32_LEAN_AND_MEAN
+#undef NOMINMAX
 
 #ifdef _WIN32
     #include <windows.h>
--- a/tests/src/unit-bson.cpp
+++ b/tests/src/unit-bson.cpp
@@ -12,6 +12,7 @@
 using nlohmann::json;
 
 #include <fstream>
+#include <limits>
 #include <sstream>
 #include "make_test_data_available.hpp"
 #include "test_utils.hpp"
@@ -865,7 +866,7 @@ TEST_CASE("BSON numerical data")
             {
                 std::vector<int64_t> const numbers
                 {
-                    INT64_MIN,
+                    (std::numeric_limits<int64_t>::min)(),
                     -1000000000000000000LL,
                     -100000000000000000LL,
                     -10000000000000000LL,
@@ -875,7 +876,7 @@ TEST_CASE("BSON numerical data")
                     -1000000000000LL,
                     -100000000000LL,
                     -10000000000LL,
-                    static_cast<std::int64_t>(INT32_MIN) - 1,
+                    static_cast<std::int64_t>((std::numeric_limits<std::int32_t>::min)()) - 1,
                 };
 
                 for (const auto i : numbers)
@@ -923,7 +924,7 @@ TEST_CASE("BSON numerical data")
             {
                 std::vector<int32_t> const numbers
                 {
-                    INT32_MIN,
+                    (std::numeric_limits<int32_t>::min)(),
                     -2147483647L,
                     -1000000000L,
                     -100000000L,
@@ -947,7 +948,7 @@ TEST_CASE("BSON numerical data")
                     100000000L,
                     1000000000L,
                     2147483646L,
-                    INT32_MAX
+                    (std::numeric_limits<int32_t>::max)()
                 };
 
                 for (const auto i : numbers)
@@ -990,7 +991,7 @@ TEST_CASE("BSON numerical data")
             {
                 std::vector<int64_t> const numbers
                 {
-                    INT64_MAX,
+                    (std::numeric_limits<int64_t>::max)(),
                     1000000000000000000LL,
                     100000000000000000LL,
                     10000000000000000LL,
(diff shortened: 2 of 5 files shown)


falbrechtskirchinger:
`silkeh/clang:dev` either doesn't install `clang-tools` anymore or `scan-build` isn't found for some other reason.

```
    apt-get install -qqy -t llvm-toolchain-bullseye \
        clang clang-tidy clang-format lld llvm && \
```

Not sure why this worked previously. Maybe `clang-tools` was pulled in as a dependency.

setoye:
There is a problem if you do in this way. In a Windows platform if you just `#include <Windows.h>` before your json library, a syntax error with `std::numeric_limits::max()` will occur because of the notorious macro definition in this Windows.h header file:
```c++
#define max(a,b) (((a) > (b)) ? (a) : (b))
```
and this piece of code will expand into 
```c++
std::numeric_limits<std::int32_t>::(((a) > (b)) ? (a) : (b))
```
Which is an error. 
Add a pair of parentheses will prevent the macro expansion:
```c++
(std::numeric_limits<std::int32_t>::max)()
```

falbrechtskirchinger:
Thanks. I was already aware of this. There's no point in fixing this now. This PR needs to be completely redone due to hundreds of merge conflicts after #3724 has been merged.

> Add a pair of parentheses will prevent the macro expansion:
> 
> ```c++
> (std::numeric_limits<std::int32_t>::max)()
> ```

This is already done in other places for this exact reason. The only surprising thing is that our unit test didn't catch this.

Edit: Not that surprising after all. `doctest.h` defines `NOMINMAX`.

falbrechtskirchinger:
Windows builds are expected to fail now.

falbrechtskirchinger:
# :man_facepalming:

tests/src/unit-windows_h.cpp
```diff
@@ -7,6 +7,8 @@
 // SPDX-License-Identifier: MIT
 
 #include "doctest_compatibility.h"
+#undef WIN32_LEAN_AND_MEAN
+#undef NOMINMAX
```
falbrechtskirchinger, line 11: doctest defines these and doesn't restore them, defeating one of the purposes of the unit test.
(I.e., finding unparanthesized uses of `std::numeric_limits<T>::min()`, etc. in `json.hpp`.)

gregmarr (approved):

tests/src/unit-windows_h.cpp
falbrechtskirchinger: I've submitted a PR to doctest.

nlohmann:
Thanks!

PR #3726, opened 2022-09-06
Add dark mode toggle to documentation

falbrechtskirchinger:
In 2022 every website should have a dark mode toggle. :-)

(Automatic switching is only available in the Insiders edition of `mkdocs-material` at the moment.)


nlohmann:
Thanks!

PR #3743, opened 2022-09-13
Optimizing cast types

likunyur:
void * and const void * types, as if casting the type is not required in most compilers.


nlohmann:
This is thirdparty code. It makes no sense to change it here.

falbrechtskirchinger:
Also, C++ isn't C – it's wrong to omit the cast in C++ code.

PR #3759, opened 2022-09-20
Allow value(json_pointer) to work with arrays as documented

falbrechtskirchinger:
As discovered [here](https://github.com/nlohmann/json/discussions/3755), `value(json_pointer)` throws an exception, when accessing an array. This deviates from the documented behavior:
```cpp
try {
   return at(ptr);
} catch(out_of_range) {
   return default_value;
}
```

Allow `value(json_pointer)` to work with arrays, update documentation, and update unit tests.

### DIFF
--- a/tests/src/unit-element_access2.cpp
+++ b/tests/src/unit-element_access2.cpp
@@ -423,10 +423,33 @@ TEST_CASE_TEMPLATE("element access 2", Json, nlohmann::json, nlohmann::ordered_j
 
                     SECTION("array")
                     {
-                        Json j_nonobject(Json::value_t::array);
-                        const Json j_nonobject_const(Json::value_t::array);
-                        CHECK_THROWS_WITH_AS(j_nonobject.value("/foo"_json_pointer, 1), "[json.exception.type_error.306] cannot use value() with array", typename Json::type_error&);
-                        CHECK_THROWS_WITH_AS(j_nonobject_const.value("/foo"_json_pointer, 1), "[json.exception.type_error.306] cannot use value() with array", typename Json::type_error&);
+                        Json j_array = Json::array({j});
+                        const Json j_array_const = Json::array({j});
+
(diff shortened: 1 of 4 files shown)


tests/src/unit-element_access2.cpp
```diff
@@ -423,10 +423,33 @@ TEST_CASE_TEMPLATE("element access 2", Json, nlohmann::json, nlohmann::ordered_j
 
                     SECTION("array")
                     {
-                        Json j_nonobject(Json::value_t::array);
-                        const Json j_nonobject_const(Json::value_t::array);
-                        CHECK_THROWS_WITH_AS(j_nonobject.value("/foo"_json_pointer, 1), "[json.exception.type_error.306] cannot use value() with array", typename Json::type_error&);
-                        CHECK_THROWS_WITH_AS(j_nonobject_const.value("/foo"_json_pointer, 1), "[json.exception.type_error.306] cannot use value() with array", typename Json::type_error&);
+                        Json j_array = Json::array({j});
+                        const Json j_array_const = Json::array({j});
```
gregmarr, line 427: Since this function isn't varied on const vs non-const object, is having the two sets of tests useful?  There is a different function for lvalue vs rvalue for the default, so that might be a better thing to test.

falbrechtskirchinger: That's a rabbit hole I was hoping to avoid (in this PR). There's more I'd want to change...

gregmarr: Okay.

gregmarr (approved):

nlohmann:
Would it make sense to also have an overload for integral types just like at or operator[]?

falbrechtskirchinger:
Right, I was going to look at that as well.

Also, if needed, that is probably better suited for the planned "parity PR". I've identified a few functions with "missing" overloads. Adding it to the list.

PR #3762, opened 2022-09-22
Reorganize CMake code and add feature summary

falbrechtskirchinger:
This PR cleans up the CMake code a bit in preparation for adding static library support.

* Use "section style" from `test.cmake` in the main list file and have fewer sections with a single line of code. Alternatively, we could use the old section style, but I find the current choice visually clearer and less awkward for single-line sections.
* Add a custom feature summary. CMake's built-in feature summary doesn't produce nice output with our choice of options.
* Move option and configuration handling into a separate file.
* Move system information into a separate file and use CMake's built-in variables and tools instead of manually calling `uname`, etc.
* Merge `download_test_data.cmake` into `test.cmake` and make `JSON_TestDataDirectory` configurable via environment variable.
* Rename `ci.cmake`/`test.cmake` to `json_*.cmake`.
* Some changes are loosely inspired by range-v3.

I've held back a few changes that might break stuff and need careful review for a follow-up PR. There're some cosmetic changes affecting the same code that could still be done in this PR, though.

Notes for the follow-up PR:
* `CMAKE_MEMORYCHECK_COMMAND` is not an official CMake variable? (If so, rename.)


CMakeLists.txt
```diff
@@ -1,105 +1,42 @@
...
-endif()
+# handle options and configuration
+include(json_opts)
 
-if (JSON_Diagnostics)
-    message(STATUS "Diagnostics enabled")
+# add CI targets
+if(JSON_CI)
+    include(json_ci)
 endif()
```
falbrechtskirchinger: Not sure where this belongs.

CMakeLists.txt
```diff
@@ -1,105 +1,42 @@
...
-##
-##
-set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake ${CMAKE_MODULE_PATH})
-include(ExternalProject)
-
-##
-## OPTIONS
-##
+#############################################################################
+# setup
```
falbrechtskirchinger: Use verb form.

CMakeLists.txt
```diff
@@ -1,105 +1,42 @@
...
+# add CI targets
+if(JSON_CI)
+    include(json_ci)
 endif()
 
-if (JSON_SystemInclude)
-    set(NLOHMANN_JSON_SYSTEM_INCLUDE "SYSTEM")
-endif()
+#############################################################################
+# library target
```
falbrechtskirchinger: "add library targets"?

cmake/json_info.cmake
```diff
@@ -0,0 +1,51 @@
+#############################################################################
+# cmake info
+#############################################################################
+
+message(STATUS "[nohmann_json]: CMake ${CMAKE_VERSION}")
```
falbrechtskirchinger, line 5: This style ("[nohmann_json]: ") is used by range-v3. Makes it more clear where the output is coming from when the project is used via `add_subdirectory()`, for example.

cmake/json_info.cmake
```diff
@@ -0,0 +1,51 @@
...
+    if(num_cores)
+        message(STATUS "[nohmann_json]: Processor cores: ${num_cores}${num_threads}")
+    endif()
+endif()
+
+#############################################################################
+# compiler info
+#############################################################################
+
+execute_process(COMMAND ${CMAKE_CXX_COMPILER} --version OUTPUT_VARIABLE cxx_version_result OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_VARIABLE cxx_version_result_error ERROR_STRIP_TRAILING_WHITESPACE)
```
falbrechtskirchinger, line 55: Quote compiler.

cmake/json_opts.cmake
```diff
@@ -0,0 +1,73 @@
+include(CMakeDependentOption)
+include(GNUInstallDirs)
+
+#############################################################################
+# test options
+#############################################################################
+
+# VERSION_GREATER_EQUAL is not available in CMake 3.1
+if(${JSON_MAIN_PROJECT} AND (${CMAKE_VERSION} VERSION_EQUAL 3.13 OR ${CMAKE_VERSION} VERSION_GREATER 3.13))
```
falbrechtskirchinger, line 9: Don't expand variables.

cmake/json_summary.cmake
```diff
@@ -0,0 +1,89 @@
...
+    else()
+        message("   Test data: ${JSON_TEST_DATA_URL} (v${JSON_TEST_DATA_VERSION})")
+    endif()
+
+    json_feature(JSON_Valgrind "Execute test suite with Valgrind?")
+    if(JSON_Valgrind)
+        message("   Valgrind command: ${CMAKE_MEMORYCHECK_COMMAND}" ${CMAKE_MEMORYCHECK_COMMAND_OPTIONS})
+    endif()
+
+    unset(test_cxx_standards)
```
falbrechtskirchinger: Prefer `set(var "")`. What if CMake learns to detect uses of undefined variables someday?

cmake/json_summary.cmake
```diff
@@ -0,0 +1,89 @@
...
+    if(JSON_TEST_CXX_STANDARDS_FORCED)
+        set(test_cxx_standards "${test_cxx_standards} ${bld}(forced)${rst}")
+    endif()
+    message("   Test C++ standards:${test_cxx_standards}")
+endif()
+
+message("")
+
+json_feature(JSON_Diagnostics "Diagnostics enabled?")
+json_feature(JSON_DisableEnumSerialization "Default integer enum serialization enabled?")
```
falbrechtskirchinger: This logic is inverted.

cmake/json_test.cmake
```diff
@@ -4,7 +4,21 @@ set(_json_test_cmake_list_file ${CMAKE_CURRENT_LIST_FILE})
 # download test data
 #############################################################################
 
-include(download_test_data)
+# if variable is set, use test data from given directory rather than downloading them
+if(JSON_TEST_DATA_DIRECTORY)
+    add_custom_target(download_test_data)
+    file(WRITE ${CMAKE_BINARY_DIR}/include/test_data.hpp "#define TEST_DATA_DIRECTORY \"${JSON_TEST_DATA_DIRECTORY}\"\n")
```
falbrechtskirchinger: For follow-up PR: Use `PROJECT_BINARY_DIR`.

cmake/json_test.cmake
```diff
@@ -4,7 +4,21 @@ set(_json_test_cmake_list_file ${CMAKE_CURRENT_LIST_FILE})
...
+else()
+    find_package(Git REQUIRED)
+    # target to download test data
+    add_custom_target(download_test_data
+        COMMAND test -d json_test_data || ${GIT_EXECUTABLE} clone -c advice.detachedHead=false --branch v${JSON_TEST_DATA_VERSION} ${JSON_TEST_DATA_URL}.git --quiet --depth 1
```
falbrechtskirchinger: For follow-up PR: `.git` should be part of the URL variable value so other URLs can be used.

tests/CMakeLists.txt
```diff
@@ -107,21 +101,17 @@ else()
     set(test_force FORCE)
 endif()
 
-# Print selected standards marking unavailable ones with brackets
-set(msg_standards "")
+# create list of selected C++ standards for feature summary
+unset(info)
```
falbrechtskirchinger: See above.

cmake/json_info.cmake
falbrechtskirchinger: Not needed (I think) but a consistent style would be nice…

cmake/json_opts.cmake
falbrechtskirchinger: After some testing, I changed my mind. I'll deal with this when I address the shortcomings of the CMake flags test.

cmake/json_summary.cmake
```diff
@@ -0,0 +1,92 @@
...
+
+    if(JSON_TEST_DATA_DIRECTORY)
+        message("   Test data: ${JSON_TEST_DATA_DIRECTORY}")
+    else()
+        message("   Test data: ${JSON_TEST_DATA_URL} (v${JSON_TEST_DATA_VERSION})")
+    endif()
+
+    json_feature(JSON_Valgrind "Execute test suite with Valgrind?")
+    if(JSON_Valgrind)
+        message("   Valgrind command: ${CMAKE_MEMORYCHECK_COMMAND}" ${CMAKE_MEMORYCHECK_COMMAND_OPTIONS})
```
falbrechtskirchinger: This isn't correct but will be fixed in the follow-up PR.

cmake/json_test.cmake
falbrechtskirchinger: Done. (in this PR)

cmake/json_test.cmake
falbrechtskirchinger: Since this is only relevant if someone wanted to build unit tests in an `add_subdirectory()` scenario, I don't think this matters.

falbrechtskirchinger (commented):
Looks good to me.

falbrechtskirchinger:
I added `NEGATE` to the wrong call of `json_feature()`. Fixed.

etienneyvon (approved):

nlohmann (commented):
I have not yet checked whether the new structure feels simpler to me, but only checked the CMake output in a terminal.

Some observations:

1. I think it is a bit too noisy in case the library is just integrated into a bigger project:

	```
	-- [nohmann_json]: CMake 3.24.2
	-- [nohmann_json]: Host system: Darwin-21.5.0
	-- [nohmann_json]: C++ compiler: Apple clang version 14.0.0 (clang-1400.0.29.102)
	-- [nohmann_json]: Feature summary:
	   Build tests? NO
	
	   Diagnostics enabled? NO
	   Default integer enum serialization enabled? YES
	   Define user-defined string literals globally? YES
	   Implicit conversions enabled? YES
	   Legacy discarded value comparison enabled? NO
	
	   Use the multi-header code? YES
	   Include directory: /Users/niels/Code/repositories/json/include/
	   Include as system headers? NO
	```
	
	To my taste, this is too much output.

2. When using ANSI escapes, one should always make sure if the terminal is supporting them in the first place. I was using `ccmake` and got this output:

	Screen Shot 2022-09-25 at 11 30 59

3. I had my reasons to call tools like `uname` to get better information on the operating system, etc.

	This is what I got on my system before:
	
	```
	-- Operating system: Darwin-21.5.0; ProductName:	macOS; ProductVersion:	12.4; BuildVersion:	21F79; Darwin stewie.fritz.box 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:37 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T6000 arm64
	-- Compiler: Apple clang version 14.0.0 (clang-1400.0.29.102); Target: arm64-apple-darwin21.5.0; Thread model: posix; InstalledDir: /Library/Developer/CommandLineTools/usr/bin
	```
	
	This is what I get with this PR:
	
	```
	-- [nohmann_json]: Host system: Darwin-21.5.0
	-- [nohmann_json]: C++ compiler: Apple clang version 14.0.0 (clang-1400.0.29.102)
	```
	
	To me, "macOS 12.4" is easier to understand than just "Darwin-21.5.0".

falbrechtskirchinger:
> 1. I think it is a bit too noisy in case the library is just integrated into a bigger project:

What do you suggest? Most features produced some output and it appears that for new features this was simply forgotten. The difference is, that the feature summary will always output something instead of only when a feature is "active" (enabled or disabled depending on the feature). I'd argue that this predictable structure is easier to read and obviates any need to guess whether a feature is enabled or not.
Any moderately complex project will print a feature summary at the end of the configuration phase and that's why I've come to expect it. Also, for bigger projects that integrate this library, the amount of log output should proportionally be minuscule.

In any case, we could disable the output in specific situations, for example:
* `JSON_MAIN_PROJECT == FALSE`
* `CMAKE_LOG_LEVEL !=  VERBOSE|DEBUG|TRACE`

>    To my taste, this is too much output.

In this instance, I disagree, but I feel that way about the current OS info output (see below). ;-)

> 2. When using ANSI escapes, one should always make sure if the terminal is supporting them in the first place. I was using `ccmake` and got this output:

True. Unfortunately, this is probably impossible to do in CMake and I relied on most terminals supporting ANSI codes these days (even Windows supports ANSI now, I learned today) and accepted garbled output to text files. It *could* be done using `cmake -E cmake_echo_color`, but I'd have to break up writing of lines into more than one step and lines are already separated by output to the other channel (stdout or stderr). This is especially visible in CI and I intend to fix it by outputting the whole feature summary in one go.

Guess I'll remove the escape sequences. Too bad we can't have nice things. :-(

> 3. I had my reasons to call tools like `uname` to get better information on the operating system, etc.
> (...)
>    To me, "macOS 12.4" is easier to understand than just "Darwin-21.5.0".

I agree wholeheartedly. I noticed it in CI and then forgot to put in the fix. Here's what the output was supposed to look like:
```
[nohmann_json]: Host system: macOS 11.7 20G817 x86_64 (Darwin-20.6.0)
[nohmann_json]: Host system: Debian GNU/Linux 11 (bullseye) x86_64 (Linux-5.15.0-1020-azure)
[nohmann_json]: Host system: Windows DataCenter Server (Build 17763) AMD64 (Windows-10.0.20348)
```
(On Linux we see some inaccuracies because of virtualization, but that is independent of the tools used. It would be easy to add information about the virtualized environment on most modern distributions using `systemd-detect-virt`, but I don't think that's needed. Just something to be aware of.)

Is that acceptable?

In this instance, I find the old output way too verbose and noisy. When is the added information *actually* helpful and do you not have to ask people to provide the log output anyway, in which case you could also ask for specific command output?

Note that the new output will also include a `Target system:` line if code is being cross-compiled, which should be useful given the unique issues on embedded platforms.

falbrechtskirchinger:
I've made some modification to the `json_feature()` function:
* Output is now padded instead of bolding the value via ANSI escapes. (`json_feature(KEY_VALUE)` overload)
* Output can be conditionally enabled via the `SWITCH <conditonal_args>...` parameter. Example:
    ```cmake
    json_feature(KEY_VALUE "Feature enabled?" JSON_Feature SWITCH ${JSON_MAIN_PROJECT})`
    ```
    Output will only appear if `${JSON_MAIN_PROJECT}` evaluates to `TRUE`.
* Newlines will be collapsed or removed to avoid redundant newlines surrounding empty sections.

Using these tools, the output can be adjusted as desired.

PR #3790, opened 2022-10-10
use get_to instead of get for array as inplace assignment to avoid ca…

yushb0602:
use get_to instead of get for array as inplace assignment to avoid calling constructor and losing the data

Inplace assignment could keep the value of data member as the below use case
```
class Person
{
  public:
    //Person(){ std::cout<<"Person constructor\n";}
    int age;
    std::string name;
    NLOHMANN_DEFINE_TYPE_INTRUSIVE(Person, name, age);

    int count{1};   // the data must not be reset
};

class SchoolA
{
  public:
    //SchoolA(){ std::cout<<"School constructor\n";}
    std::array<Person, 2> persons;

    NLOHMANN_DEFINE_TYPE_INTRUSIVE(SchoolA, persons);
};

  json obj = R"({"persons":[{"age":100, "name":"alex"}, {"age":200, "name":"edmond"}]})"_json;


 SchoolA   s;
 from_json(obj, s);
 CHECK(s.persons[0].age == 100);

 s.persons[0].count = 88;

 from_json(obj, s);
 CHECK(s.persons[0].count == 88);

```

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -186,7 +186,9 @@ auto from_json(const BasicJsonType& j, T (&arr)[N])  // NOLINT(cppcoreguidelines
 {
     for (std::size_t i = 0; i < N; ++i)
     {
-        arr[i] = j.at(i).template get<T>();
+        // inplace assignment to avoid calling constructor and losing the data of its  members
+        // arr[i] = j.at(i).template get<T>();
+        j.at(i).get_to(arr[i]);
     }
 }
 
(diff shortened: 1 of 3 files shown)


gregmarr:
I'm not sure that this is a good idea.  The `from_json` and `to_json` functions aren't really intended to be called that way.  I imagine that this would be a surprising change in behavior.  Also, this would result in these two parameter types having different behaviors than all other parameter types, such as `std::vector` or `std::list`.

yushb0602:
> I'm not sure that this is a good idea. The `from_json` and `to_json` functions aren't really intended to be called that way. I imagine that this would be a surprising change in behavior. Also, this would result in these two parameter types having different behaviors than all other parameter types, such as `std::vector` or `std::list`.

**from_json_inplace_array_impl**  already exists in library, but don't how to use it.
This feature is what I want. I'll follow this.  I close this pull request.

PR #3816, opened 2022-11-02
Add serialization-only user defined type macros

ZeronSix:
This pull requests adds `NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_ONLY_SERIALIZE` and `NLOHMANN_DEFINE_TYPE_INTRUSIVE_ONLY_SERIALIZE` macros.

Suppose we have a non-default-constructible type `person` and another type that contains a field of type `person`. For example:
```cpp
#include <nlohmann/json.hpp>

#include <string>
#include <vector>

struct person {
    std::string name;
    int age;

    person(std::string name, int age)
        : name{std::move(name)}
        , age{age}
    {}
};

struct persons {
    std::vector<person> list;
};

NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(person, name, age);
NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(persons, list);
```
An attempt to use the `NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE` macro to generate serialization/deserialization code will lead to a compilation error:
```
test.cpp: In function ‘void from_json(const json&, persons&)’:
include/nlohmann/detail/macro_scope.hpp:385:81: error: no matching function for call to ‘nlohmann::json_abi_v3_11_2::basic_json<>::get_to(std::vector<person>&) const’
  385 | #define NLOHMANN_JSON_FROM(v1) nlohmann_json_j.at(#v1).get_to(nlohmann_json_t.v1);
```

`README.md` provides an example how to fix this by providing a specialization of `adl_serializer`. However, this requires writing quite a bit of code and might even be unnecessary in cases when we only need serialization without deserialization.

In this situations, we can use added macros to generate only `to_json` functions.
```cpp
NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_ONLY_SERIALIZE(person, name, age);
NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_ONLY_SERIALIZE(persons, list);
```

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/docs/examples/nlohmann_define_type_intrusive_only_serialize_explicit.cpp
+++ b/docs/examples/nlohmann_define_type_intrusive_only_serialize_explicit.cpp
@@ -1,0 +1,38 @@
+#include <iostream>
+#include <nlohmann/json.hpp>
+
+using json = nlohmann::json;
+using namespace nlohmann::literals;
+
+namespace ns
+{
+class person
+{
+  private:
+    std::string name = "John Doe";
+    std::string address = "123 Fake St";
+    int age = -1;
+
+  public:
+    // No default constructor
+    person(std::string name_, std::string address_, int age_)
+        : name(std::move(name_)), address(std::move(address_)), age(age_)
+    {}
+
+    friend void to_json(nlohmann::json& nlohmann_json_j, const person& nlohmann_json_t)
+    {
+        nlohmann_json_j["name"] = nlohmann_json_t.name;
+        nlohmann_json_j["address"] = nlohmann_json_t.address;
+        nlohmann_json_j["age"] = nlohmann_json_t.age;
+    }
+};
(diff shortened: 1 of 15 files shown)


nlohmann (changes requested):
Please add documentation. See https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/api/macros/index.md for a starting point.

nlohmann:
(Also, please update from the develop branch as it has a working CI now.)

ZeronSix:
I also noticed that the source code highlighting is a bit off in these pages, see https://github.com/nlohmann/json/pull/4169.

nlohmann (approved):
Looks good to me.

PR #3891, opened 2022-12-18
Upgrade Python packages

nlohmann:
- upgrade packages to fix security warning
- also bring mkdocs and other packages up to date


PDeets:
Thanks for making this change. However, I wanted to bring up that as a user of nlohnamm-json via vcpkg, I regularly get security warnings due to our security scanners finding the requirements.txt file and identifying what it believes are insecure dependencies of my source code. I'm just writing C++ code that uses nlohmann-json though, so I always have to dismiss these. However, it is a nuisance to deal with.

Do you think maybe we should add a step in the vcpkg portfile at https://github.com/microsoft/vcpkg/blob/master/ports/nlohmann-json/portfile.cmake to delete the docs folder after it downloads the source code for the tagged release so these dependencies don't show up in security scans.

PR #3898, opened 2022-12-26
Add helper types to make it easier to create a basic_json type with modified template parameters

barcode:
(First part of the points listed in https://github.com/nlohmann/json/pull/3110#issuecomment-1229141478, the rest is done in a different PR to prevent mixing of separate issues / features)

    Make specifying a base class easier/less verbose.

This PR adds member templates to `nlohmann::json` which can be used to create a `basic_json`  type with one replaced template parameter (e.g. `nlohmann::json::with_changed_json_serializer_t<T>`).

Tests and examples were modified to use those templates (except for regression tests).

Open questions from my side:

* Should we add something like this? (I think so, but what is your opinion?)
* Should we go with the current names (`with_changed_*`) or rather use something else (e.g. `with_*`)?

In case this will be added, i will write the documentation.

* * *

## Pull request checklist

### DIFF
--- a/docs/examples/json_base_class_t.cpp
+++ b/docs/examples/json_base_class_t.cpp
@@ -13,19 +13,7 @@ class visitor_adaptor_with_metadata
     void do_visit(const Ptr& ptr, const Fnc& fnc) const;
 };
 
-using json = nlohmann::basic_json <
-             std::map,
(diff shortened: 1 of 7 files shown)


barcode:

PR #3901, opened 2022-12-27
Prevent memory leak when exception is thrown in adl_serializer::to_json

### LINKED ISSUE #3881: Memory leak when exception is thrown in adl_serializer::to_json
### Description

Throwing an exception in serializer function after the output json object is already initialized results in a memory leak (detected with both ASan and valgrind):
```c++
void adl_serializer<Foo>::to_json(json& j, Foo const& f) {
    j["a"] = 42;
    throw std::runtime_error("error");
}
```

### Reproduction steps

Reproducer here: https://godbolt.org/z/dvorMExWh

Expected: no memory leak

### Minimal code example

_No response_

### Error messages

_No response_

linux, clang 15, gcc 12.2

### Library version

3.11.2, godbolt trunk, 7f72eedc2d4fc196d389f5aa0b2659f70dabe278

### Validation

barcode:
(problem described in issue #3881)

There is a memory leak when throwing inside an adl serializers `to_json` method. 

- [x] Add unit test
- [x] Implement fix
- [x] Fix Ci

* * *

## Pull request checklist

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -4209,9 +4209,9 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
 
         data() noexcept = default;
         data(data&&) noexcept = default;
-        data(const data&) noexcept = default;
-        data& operator=(data&&) noexcept = default;
-        data& operator=(const data&) noexcept = default;
+        data(const data&) noexcept = delete;
+        data& operator=(data&&) noexcept = delete;
+        data& operator=(const data&) noexcept = delete;
 
         ~data() noexcept
         {
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -555,7 +555,12 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
 
         void destroy(value_t t)
         {
-            if (object == nullptr)
+            if (
+                (t == value_t::object && object == nullptr) ||
+                (t == value_t::array && array == nullptr) ||
+                (t == value_t::string && string == nullptr) ||
+                (t == value_t::binary && binary == nullptr)
+            )
             {
                 //not initialized (e.g. due to exception in the ctor)
                 return;
@@ -678,7 +683,23 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     */
     void assert_invariant(bool check_parents = true) const noexcept
     {
-        m_data.assert_invariant(check_parents ? this : nullptr);
+        JSON_ASSERT(m_data.m_type != value_t::object || m_data.m_value.object != nullptr);
+        JSON_ASSERT(m_data.m_type != value_t::array || m_data.m_value.array != nullptr);
+        JSON_ASSERT(m_data.m_type != value_t::string || m_data.m_value.string != nullptr);
+        JSON_ASSERT(m_data.m_type != value_t::binary || m_data.m_value.binary != nullptr);
+
+#if JSON_DIAGNOSTICS
+        JSON_TRY
+        {
+            // cppcheck-suppress assertWithSideEffect
+            JSON_ASSERT(!check_parents || !is_structured() || std::all_of(begin(), end(), [this](const basic_json & j)
+            {
+                return j.m_parent == this;
+            }));
+        }
+        JSON_CATCH(...) {} // LCOV_EXCL_LINE
+#endif
+        static_cast<void>(check_parents);
     }
 
     void set_parents()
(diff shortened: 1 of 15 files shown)


barcode:
The test detects the memleak: https://github.com/nlohmann/json/actions/runs/3792554026/jobs/6448842970#step:5:316

barcode:
I went with the solution using a member var, since it requires less changes to the existing code and the resulting code is simpler to understand.

gregmarr:
Since this keeps most of the handling in the json class rather than the new data class, such as the copy constructor and move constructor, should the assert invariant function stay there as well?  Also, should the copy and move constructors of this class be defaulted?  Seems like they should be deleted.

barcode:
should the assert invariant function stay there as well?

in that case the assert invariant method is not called if an exception in the ctor occurred. Then we do not need to adapt that method to deal with broken objects. Since the dtor of an element is called after the objects dtor finishes, the call order would be OK as well (assert before destroy).

    Also, should the copy and move constructors of this class be defaulted? Seems like they should be deleted.

we need at least the move ctor
https://github.com/nlohmann/json/blob/df39768dd0f0a488f363fabfb8a3c495263f751c/include/nlohmann/json.hpp#L1196

The others we could delete.

include/nlohmann/json.hpp
```diff
@@ -673,16 +683,16 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+        JSON_ASSERT(m_data.m_type != value_t::array || m_data.m_value.array != nullptr);
+        JSON_ASSERT(m_data.m_type != value_t::string || m_data.m_value.string != nullptr);
+        JSON_ASSERT(m_data.m_type != value_t::binary || m_data.m_value.binary != nullptr);
 
 #if JSON_DIAGNOSTICS
         JSON_TRY
         {
             // cppcheck-suppress assertWithSideEffect
-            JSON_ASSERT(!check_parents || !is_structured() || std::all_of(begin(), end(), [this](const basic_json & j)
+            JSON_ASSERT(check_parents || !is_structured() || std::all_of(begin(), end(), [this](const basic_json & j)
```
gregmarr: I think this loss of `!` is unintentional and incorrect.  If `check_parents` is false, then the assert should always succeed here.

include/nlohmann/json.hpp
```diff
@@ -1007,9 +1017,9 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     /// @brief construct an array with count copies of given value
     /// @sa https://json.nlohmann.me/api/basic_json/basic_json/
     basic_json(size_type cnt, const basic_json& val)
-        : m_type(value_t::array)
```
gregmarr, line 1010: Would it make sense for `m_data` to have a constructor that just takes the type, so this can stay as an initializer?  Don't really have a good reason for this other than "that's the way it was before", "it should always be initialized to something" (which isn't necessary since it has a default constructor and initializer) and "this prevents it from being initialized to null type and then changed to array type" (which the compiler should be optimizing out anyway).  Don't have a strong feeling either way.

include/nlohmann/json.hpp
barcode: You are right.

include/nlohmann/json.hpp
barcode: I don't like partial initialization via ctors (e.g. only setting the type, but not the value). Hence i think we either should use the default ctor for `m_data` or use one initializing both members.
Therefore I added a ctor `data(size_type, const basic_json&)` and used it for initialization.

gregmarr: Agreed that this is better.

tests/src/unit-regression2.cpp
```diff
@@ -36,7 +36,7 @@ using ordered_json = nlohmann::ordered_json;
     #include <variant>
 #endif
 
-#ifdef JSON_HAS_CPP_20
+#ifdef JSON_HAS_CPP_20 && __has_include(<span>)
```
gregmarr: The proper form for this is `#if defined(JSON_HAS_CPP_20) && ...`

barcode: Maybe I should sleep instead of writing code while tired...

barcode:
I looked a bit at the [ci_test_compilers_clang (10)](https://github.com/nlohmann/json/actions/runs/3835776970/jobs/6529309030#logs) issue. 
This PR did not change the affected test and it worked in [earlier builds](https://github.com/nlohmann/json/actions/runs/3777809193/jobs/6421841164). 

It seems the compiler and changed to a newer version.

old:

    Compiler: clang version 10.0.0; Target: x86_64-unknown-linux-gnu; Thread model: posix; InstalledDir: /usr/local/bin

new:

    Compiler: Debian clang version 10.0.1-++20210313014605+ef32c611aa21-1~exp1~20210313125208.190; Target: x86_64-pc-linux-gnu; Thread model: posix; InstalledDir: /usr/bin

Since this issue has nothing to do with this PR. it probably should be fixed in a separate PR.

barcode:

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #3906, opened 2023-01-04
Fix CI issues

barcode:
I looked a bit at the [ci_test_compilers_clang (10)](https://github.com/nlohmann/json/actions/runs/3835776970/jobs/6529309030#logs) issue.
This PR did not change the affected test and it worked in [earlier builds](https://github.com/nlohmann/json/actions/runs/3777809193/jobs/6421841164).

It seems the compiler and changed to a newer version.

old:

    Compiler: clang version 10.0.0; Target: x86_64-unknown-linux-gnu; Thread model: posix; InstalledDir: /usr/local/bin

new:

    Compiler: Debian clang version 10.0.1-++20210313014605+ef32c611aa21-1~exp1~20210313125208.190; Target: x86_64-pc-linux-gnu; Thread model: posix; InstalledDir: /usr/bin

* * *

## Pull request checklist

Read the [Contribution Guidelines](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md) for detailed information.

- [ x]  Changes are described in the pull request, or an [existing issue is referenced](https://github.com/nlohmann/json/issues).
- [ ]  The test suite [compiles and runs](https://github.com/nlohmann/json/blob/develop/README.md#execute-unit-tests) without error.
- [x ]  [Code coverage](https://coveralls.io/github/nlohmann/json) is 100%. Test cases can be added by editing the [test suite](https://github.com/nlohmann/json/tree/develop/test/src).
- [x ]  The source code is amalgamated; that is, after making changes to the sources in the `include/nlohmann` directory, run `make amalgamate` to create the single-header files `single_include/nlohmann/json.hpp` and `single_include/nlohmann/json_fwd.hpp`. The whole process is described [here](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md#files-to-change).

PR #3929, opened 2023-01-25
Fix for printing long doubles bug in dump_float

RUSLoker:
When you use long double as a floating point type with the current version of include/nlohmann/detail/output/serializer.hpp and try to dump json it prints trash instead of actual number. This if-else fixes the problem. On using long double you just need to add an 'L' modifier before 'g' in format string.


nlohmann (changes requested):
The library targets C++11. Please adjust the code.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -832,7 +832,12 @@ class serializer
 
         // the actual conversion
         // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg,hicpp-vararg)
-        std::ptrdiff_t len = (std::snprintf)(number_buffer.data(), number_buffer.size(), "%.*g", d, x);
+        std::ptrdiff_t len;
+        if (std::is_same<number_float_t, long double>::value) {
+            len = (std::snprintf)(number_buffer.data(), number_buffer.size(), "%.*Lg", d, x);
```
gregmarr: I wonder if it would be better like this, to eliminate the duplication of code which could get out of sync:

    std::ptrdiff_t len = (std::snprintf)(number_buffer.data(), number_buffer.size(), 
        std::is_same<number_float_t, long double>::value ? "%.*Lg" : "%.*g", d, x);

RUSLoker: Yeah, your solution is better

gregmarr (approved):
LGTM, @nlohmann will need to merge.

t-b:
I think a test is missing or?

nlohmann:
Yes, a test is definitely required.

There is also #3906 that I need to take care of to finally have a green CI again. I'm busy right now, so sorry for delays.

nlohmann:
Please update to recent develop version - the CI is fixed there (see #3906).

RUSLoker:
@nlohmann, it asks for your approval to run CI workflows.

RUSLoker:
@nlohmann, @gregmarr, what's wrong with [**Cirrus CI / check**](https://github.com/nlohmann/json/pull/3929/checks?check_run_id=11030030659)?

nlohmann:
> @nlohmann, @gregmarr, what's wrong with [**Cirrus CI / check**](https://github.com/nlohmann/json/pull/3929/checks?check_run_id=11030030659)?

Looks like a glitch.

nlohmann (changes requested):
Please add tests.

nlohmann:
Any update on this? Tests are still missing.

nlohmann:
Please update to the latest develop branch and add a test.

nlohmann:
@RUSLoker It would be great if you could add a test so we can merge this.

nlohmann:
@RUSLoker I'd really like to merge this. Can you please add a test?

nlohmann:
I think this can be closed in favor of #5169.

RUSLoker:
@nlohmann, I haven't found any tests in #5169. Maybe you should take mine?

nlohmann:
Sounds good, so please update to the latest develop branch.

tests/src/unit-serialization.cpp
```diff
@@ -11,8 +11,13 @@
 #include <nlohmann/json.hpp>
 using nlohmann::json;
 
+#include <cstdint>
```
gregmarr: All of these new includes are already included by `json.hpp` above.

RUSLoker: If implicit includes are okey for the project, I remove.

RUSLoker:
@nlohmann, added fix to your solution. Reason described in the code comment.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -835,8 +835,10 @@ class serializer
     JSON_HEDLEY_NON_NULL(1)
     static int snprintf_float(char* buf, std::size_t size, int d, long double x)
     {
+        // %Lg (not %lg): 'l' has no effect on g/G; only 'L' applies to
+        // long double. ISO C11 §7.21.6.1 p7; otherwise UB (p9).
         // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg,hicpp-vararg)
-        return (std::snprintf)(buf, size, "%.*lg", d, x);
+        return (std::snprintf)(buf, size, "%.*Lg", d, x);
```
nlohmann, line 839: Good catch! Thanks for the fix!

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -835,8 +835,10 @@ class serializer
     JSON_HEDLEY_NON_NULL(1)
     static int snprintf_float(char* buf, std::size_t size, int d, long double x)
     {
+        // %Lg (not %lg): 'l' has no effect on g/G; only 'L' applies to
+        // long double. ISO C11 §7.21.6.1 p7; otherwise UB (p9).
```
nlohmann: This comment can be removed.

tests/src/unit-serialization.cpp
```diff
@@ -295,3 +295,73 @@ TEST_CASE("dump with binary values")
...
+    // Linux/macOS x86_64), dump() goes through the snprintf path in
+    // serializer::dump_float(x, std::false_type). That branch must use the
+    // "%.*Lg" format specifier; using "%.*g" with a long double argument is
+    // undefined behavior and corrupts the output.
+    using long_double_json = nlohmann::basic_json<std::map, std::vector, std::string,
+          bool, std::int64_t, std::uint64_t, long double>;
+
+    SECTION("round-trip dump/parse")
+    {
+        const long double values[] =
```
nlohmann: Clang-Tidy complains here:

```
/__w/json/json/tests/src/unit-serialization.cpp:312:15: error: do not declare C-style arrays, use 'std::array' instead [cppcoreguidelines-avoid-c-arrays,hicpp-avoid-c-arrays,modernize-avoid-c-arrays,-warnings-as-errors]
  312 |         const long double values[] =
      |               ^~~~~~~~~~~~~~~~~~~~
```

include/nlohmann/detail/output/serializer.hpp
RUSLoker: Are you sure? I think it's not an obvious thing. So it's better to remain it in place for future contributors.

nlohmann: Yes. It was a typo from my end, but we should not add a comment on `%Lg` that `%Lg` is correct.

tests/src/unit-serialization.cpp
```diff
@@ -295,3 +296,73 @@ TEST_CASE("dump with binary values")
...
+    {
+        constexpr std::array<long double, 13> values =
+        {
+            0.0L, -0.0L, 1.0L, -1.0L,
+            0.5L, -0.5L, 1.5L, -2.25L,
+            1.23e45L, 1.23e-45L,
+            (std::numeric_limits<long double>::min)(),
+            std::numeric_limits<long double>::lowest(),
+            (std::numeric_limits<long double>::max)()
+        };
```
nlohmann: GCC complains - and I am not sure if the `constexpr` is allowed in C++11.

```
/__w/json/json/tests/src/unit-serialization.cpp:321:9: error: missing braces around initializer for 'std::__array_traits<long double, 13>::_Type' {aka 'long double [13]'} [-Werror=missing-braces]
  321 |         };
      |         ^
```

RUSLoker: std::array<T, N> has a constexpr constructor since C++11

  std::array is an aggregate with no user-declared constructors - only implicitly-declared ones ([array.overview]/2: "An array is an aggregate that can be 
  list-initialized with up to N elements...", https://eel.is/c++draft/array.overview).

  The core language guarantees these implicit constructors are constexpr when the element type permits:
  
  - [class.default.ctor]/4: "The implicitly-defined default constructor is constexpr if the constructor would satisfy the requirements of a constexpr function..."
  (https://eel.is/c++draft/class.default.ctor)
  - [class.copy.ctor]/13: "The implicitly-defined copy/move constructor is constexpr if X is a literal type..." (https://eel.is/c++draft/class.copy.ctor)

  These rules have been in the standard since C++11 (N3337, §12.1/6 and §12.8/12). The claim that std::array "only became constexpr in C++14/17/20" refers to its
  member functions (operator[], data(), etc.); construction has been constexpr from C++11.

nlohmann:
Please check the [DCO requirements](https://github.com/nlohmann/json/pull/3929/checks?check_run_id=76175447652).

nlohmann (approved):
Looks good to me.

RUSLoker:
Yeeeeah, finally my fix would be merged after 3 years :) Danke!

nlohmann:
Thank you!

PR #3951, opened 2023-02-22
Fix typo in test.cmake

### LINKED ISSUE #3902: Typo in `cmake/test.cmake`
### Description

At line 260 in [`cmake/test.cmake`]() there is this code 

```cmake
        include(CheckTypeSize)
        check_type_size("size_t" sizeof_size_t LANGUAGE CXX)
        if(sizeof_size_t AND ${sizeof_size_t} EQUAL 4)
            message(STATUS "Auto-enabling 32bit unit test.")
            set(${build_32bit_var} ON)
        else()
            set(${build_32bit_var} OFF)
        endif()
```

The conditional check 

```
        if(sizeof_size_t AND ${sizeof_size_t} EQUAL 4)
```

The first condition `if(sizeof_size_t` is always TRUE, since it is comparing against the string value "sizeof_size_t". To check if the string is empty (meaning check_type_size could not return a value for size_t)  it should be 

```cmake
        if(${sizeof_size_t} AND ${sizeof_size_t} EQUAL 4)
```

If for some reason the size of size_t is unknown (for example, using exotic toolchains where stddef.h might not be visible during the initial cmake evaluation) this will cause the invocation of `cmake` to fail.

### Reproduction steps

Replace the call to `check_type_size` with

```
set(sizeof_size_t "")
```

Running `cmake .` should not fail. 

### Minimal code example

_No response_

### Error messages

```Shell
if given arguments:

    "sizeof_size_t" "AND" "EQUAL" "4"

  Unknown arguments specified
```
```


gcc 11.2.0, Linux

### Library version

3.11.2

### Validation

theevilone45:
Ths fixes typo in cmake/test.cmake

Issue: #3902

### DIFF
--- a/cmake/test.cmake
+++ b/cmake/test.cmake
@@ -259,7 +259,7 @@ function(json_test_should_build_32bit_test build_32bit_var build_32bit_only_var 
         # check if compiler is targeting 32bit by default
         include(CheckTypeSize)
         check_type_size("size_t" sizeof_size_t LANGUAGE CXX)
-        if(sizeof_size_t AND ${sizeof_size_t} EQUAL 4)
+        if(${sizeof_size_t} AND ${sizeof_size_t} EQUAL 4)
             message(STATUS "Auto-enabling 32bit unit test.")
             set(${build_32bit_var} ON)
         else()


nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #3964, opened 2023-03-05
refactor(tests): use prefix instead of postfix operators

zEuS0390:
Using prefix operator over postfix operator when it is not used is more efficient because it only changes the current value without creating a copy of it.


nlohmann:
I am currently only looking at this from my mobile, but it seems the tests call that specific operator to ensure all functions are properly tested. So it could be that this optimization could reduce test coverage.

zEuS0390:
I'm sorry, this is my first time contributing, and I didn't notice it. Well, maybe I should close this PR. Thanks!

PR #3969, opened 2023-03-07
Refactor amalgamation workflow to avoid dangerous use of pull_request_target

### LINKED ISSUE #3945: Dangerous use of pull_request_target
### Description

Hi, I am Joyce working on behalf of Google and the OpenSSF.

The project has one workflow with dangerous or risky behavior (identified by using the [Scorecard tool](https://github.com/ossf/scorecard)), which is the https://github.com/nlohmann/json/blob/develop/.github/workflows/check_amalgamation.yml.

The [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) should not be used with a checkout  as can be seen in the following warning from the Github Documentation Page:

image

### Reproduction steps

None

Analysing the workflow and looking at the more secure alternatives, I could think in two approaches:

#### Use `pull_request`

Instead of using `pull_request_target`, we could use the `pull_request` which would change the current behavior in two ways:

- The verifications would be done in a "merge commit"
- The workflow would not be granted the read-all and write-all permission that make it dangerous.
- EDIT: Create a workflow with a workflow_run trigger in another workflow to add the comment 

#### Use label verification

We can use a `type: [labeled]` and a condition of `if: ${{ github.event.label.name == 'is ok to test' }}` to check for a label "is ok to test" for example, that you would manually add once you saw that nothing potentially dangerous would be running.

That's the two solutions I though, but let me know if you know another one we can explore. I can open a PR with any solution we agree on this issue.

### Minimal code example

_No response_

### Error messages

_No response_

None

### Library version

None

### Validation

joycebrum:
Closes #3945 

I've followed the approach recommended by [Github Security Lab - Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). This way the pull_request workflow won't have any write privilege that could be exploit.

I have tested the new workflow to add the comment in case of failure in the amalgamation workflow here https://github.com/joycebrum/json/pull/3, and it worked fine.

## Pull request checklist

## Please don't

### DIFF
--- a/.github/workflows/check_amalgamation.yml
+++ b/.github/workflows/check_amalgamation.yml
@@ -1,9 +1,24 @@
 name: "Check amalgamation"
 
 on:
-  pull_request_target:
+  pull_request:
+
+permissions: read-all
 
 jobs:
+  save:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Save PR number
+        run: |
+          mkdir -p ./pr
+          echo ${{ github.event.number }} > ./pr/number
+          echo ${{ github.event.pull_request.user.login }} > ./pr/author
+      - uses: actions/upload-artifact@v2
+        with:
+          name: pr
+          path: pr/
+
   check:
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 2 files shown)


nlohmann (approved):
Looks good to me.

nlohmann:
Thanks a lot! I will merge once the CI runs through.

PR #3975, opened 2023-03-11
Do not use `#pragma once`

### LINKED ISSUE #3974: `#pragma once` not supported with C++20 modules in clang
### Description

If compiling with modules enabled with clang, `#pragma once` is used across all headers -- in other words, only the first header with `#pragma once` in it will be included, and the rest ignored. Headers in `nlohmann` use `#pragma once`, and thus the library is difficult or impossible to be used in a modular codebase.

The upstream bug is https://github.com/llvm/llvm-project/issues/38554.

The workaround is to use traditional `#ifndef` include guards instead.

### Reproduction steps

```cpp
module;

#include <nlohmann/json.hpp>

export module a;
```

is sufficient to expose the issue if not using the "single header" version. Otherwise, combining it with any other header using `#pragma once` is sufficient.

The library should compile, but it errors out with bogus messages.

### Minimal code example

_No response_

### Error messages

_No response_

clang

### Library version

b2306145e1789368e6f261680e8dc007e91cc986

### Validation

davidstone:
Fixes https://github.com/nlohmann/json/issues/3974 by using the standard-compliant `#ifndef #define #endif` pattern instead of `#pragma once`.


nlohmann:
Do you know of a compiler that does not support pragma once?

nlohmann:
(Ignore my comment before. Thanks for looking into the issue!)

davidstone:
> (Ignore my comment before. Thanks for looking into the issue!)

Yeah, it's an unfortunate change. `#pragma once` is definitely cleaner and less bug-prone. :(

nlohmann (changes requested):
Can you please address the changes required by cpplint?

davidstone:
The remaining failures are unrelated to my changes. clang 16.0 was just released and it adds a new warning, `Wunsafe-buffer-usage`, and that warning is being triggered by existing code.

nlohmann:
The `ci_cpplint` step is still failing.

gregmarr:
@davidstone You probably need to merge in the latest changes to fix the CI.

nlohmann (changes requested):
Please update from the develop branch as it has a working CI now.

davidstone:
Newer versions of clang have fixed this bug. If someone else wants to take over this to get the PR working for versions of clang without the fix (I don't know which numbered versions are broken vs. working) feel free, but given that I prefer the way the code is with `#pragma once` and there are plenty of other modules bugs fixed in newer versions of clang, I'm happy to just abandon this.

Issue #3992, opened 2023-03-23
NLOHMANN_JSON_SERIALIZE_ENUM defaulting to the first enum value is surprising behavior


bear24rw:
### Description

I had a json document with an invalid enum string. I would like an exception to be raised if the string is not a known conversion back to the enum. I don't want to have to add and check an `Unknown` or `Invalid` enum value which would only be used for json serialization.

This is well documented: "undefined JSON values will default to the first specified conversion." but is IMO is surprising behavior since we throw `json.exception.type_error`'s for other type conversions if they fail (like converting an integer in the json document to a `std::string`)

This is maybe only a good idea for strongly typed `enum class`'s.

### Reproduction steps

```cpp
namespace ns
{
enum class Color
{
    red, green, blue // I don't want to define an unknown item just used for json deserialization
};

NLOHMANN_JSON_SERIALIZE_ENUM(Color,
{
    { Color::red, "red" },
    { Color::green, "green" },
    { Color::blue, "blue" }
})
}

json some_color_json = "pink";
try {
    auto color = some_color_json.get<ns::Color>();
} catch (json::type_error& e) {
    printf("Invalid color!\n");
}
```

### Expected vs. actual results

Expected: It prints 'invalid color'
Actual: `color` gets set to `Color::red`

I believe the behavior I am after can be achieved by with this patch
```patch
diff --git a/include/nlohmann/detail/macro_scope.hpp b/include/nlohmann/detail/macro_scope.hpp
index 6248bea..9916aac 100644
--- a/include/nlohmann/detail/macro_scope.hpp
+++ b/include/nlohmann/detail/macro_scope.hpp
@@ -215,7 +215,9 @@
         {                                                                                       \
             return ej_pair.first == e;                                                          \
         });                                                                                     \
-        j = ((it != std::end(m)) ? it : std::begin(m))->second;                                 \
+        if (it == std::end(m))
+            throw ...
+        j = it;
     }                                                                                           \
     template<typename BasicJsonType>                                                            \
     inline void from_json(const BasicJsonType& j, ENUM_TYPE& e)                                 \
@@ -227,7 +229,9 @@
         {                                                                                       \
             return ej_pair.second == j;                                                         \
         });                                                                                     \
-        e = ((it != std::end(m)) ? it : std::begin(m))->first;                                  \
+        if (it == std::end(m))
+            throw ...
+        e = it;
     }
 
 // Ugly macros to avoid uglier copy-paste when specializing basic_json. They
```

[...]

PR #3996, opened 2023-03-25
alternative macro for throwing on enum serialization errors

pabloariasal:
#3992 

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ You can sponsor this library at [GitHub Sponsors](https://github.com/sponsors/nl
 - [Steve Sperandeo](https://github.com/homer6)
 - [Robert Jefe Lindstädt](https://github.com/eljefedelrodeodeljefe)
 - [Steve Wagner](https://github.com/ciroque)
+- [Lion Yang](https://github.com/LionNatsu)
 
 Thanks everyone!
 


docs/examples/nlohmann_json_serialize_enum.cpp
```diff
@@ -26,7 +26,7 @@ enum class Color
     red, green, blue, unknown
 };
 
-NLOHMANN_JSON_SERIALIZE_ENUM(Color,
```
nlohmann: Please add a new example instead of editing this one.

nlohmann (changes requested):

docs/examples/nlohmann_json_serialize_enum.cpp
pabloariasal: Done.

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum.md
```diff
@@ -75,6 +77,22 @@ inline void from_json(const BasicJsonType& j, type& e);
     --8<-- "examples/nlohmann_json_serialize_enum_2.output"
     ```
 
+??? example "Example 3: Strict error handling"
+
+    The example shows how to define strict error handling for enum serialization.
+    In the example conversion between `Color` and json throw an exception in case of erros (instead of defauting to the first value).
```
gregmarr: typo: erros instead of errors

docs/mkdocs/docs/features/enum_conversion.md
```diff
@@ -27,6 +27,8 @@ NLOHMANN_JSON_SERIALIZE_ENUM( TaskState, {
 The [`NLOHMANN_JSON_SERIALIZE_ENUM()` macro](../api/macros/nlohmann_json_serialize_enum.md) declares a set of
 `to_json()` / `from_json()` functions for type `TaskState` while avoiding repetition and boilerplate serialization code.
 
+An alternative macro [`NLOHMANN_JSON_SERIALIZE_ENUM_STRICT()` macro](../api/macros/nlohmann_json_serialize_enum.md) can be used when a more strict error handling is preffered, throwing in case of serialization errors instead of defaulting to the first enum value defined in the macro.
```
gregmarr: typo: preffered instead of preferred

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -230,6 +230,46 @@
         e = ((it != std::end(m)) ? it : std::begin(m))->first;                                  \
     }
 
+#if (defined(__cpp_exceptions) || defined(__EXCEPTIONS) || defined(_CPPUNWIND)) && !defined(JSON_NOEXCEPTION)
```
gregmarr: If this only works when `JSON_NOEXCEPTION` isn't defined, then that should be noted in the docs.  However, I think you can just use `JSON_THROW` instead of `throw` and use the normal exception override behavior.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -230,6 +230,46 @@
...
+        static_assert(std::is_enum<ENUM_TYPE>::value, #ENUM_TYPE " must be an enum!");                                        \
+        static const std::pair<ENUM_TYPE, BasicJsonType> m[] = __VA_ARGS__;                                                   \
+        auto it = std::find_if(std::begin(m), std::end(m), [e](const std::pair<ENUM_TYPE, BasicJsonType>& ej_pair) -> bool {  \
+            return ej_pair.first == e;                                                                                        \
+        });                                                                                                                   \
+        if (it == std::end(m))                                                                                                \
+        {                                                                                                                     \
+            throw nlohmann::detail::type_error::create(302,                                                                   \
+                    nlohmann::detail::concat("can't serialize enum value ", std::to_string(static_cast<int>(e)),              \
```
gregmarr: Would it make sense to say something about the value not being a enum value listed in `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT`?

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -230,6 +230,46 @@
...
+        static_assert(std::is_enum<ENUM_TYPE>::value, #ENUM_TYPE " must be an enum!");                                        \
+        static const std::pair<ENUM_TYPE, BasicJsonType> m[] = __VA_ARGS__;                                                   \
+        auto it = std::find_if(std::begin(m), std::end(m), [&j](const std::pair<ENUM_TYPE, BasicJsonType>& ej_pair) -> bool { \
+            return ej_pair.second == j;                                                                                       \
+        });                                                                                                                   \
+        if (it == std::end(m))                                                                                                \
+        {                                                                                                                     \
+            throw nlohmann::detail::type_error::create(302,                                                                   \
+                    nlohmann::detail::concat("can't deserialize json: '", j, "' to enum '", #ENUM_TYPE "'"), &j);             \
```
gregmarr: As above, and add the string that couldn't be found.

include/nlohmann/detail/macro_scope.hpp
pabloariasal: agree, will note it in the docs. I think there are some arguments for not using `JSON_NOEXCEPTION`:

- nested macros don't work (the preprocessor doesn't expand a macro inside another one)
- it can lead to silent and hard to debug erros if `JSON_NOEXCEPTION` is defined and the macro is used -> what should we do in that case?
- the macro itself doens't make any sense if `JSON_NOEXCEPTION` is defined (hence better not define it in the first place.

include/nlohmann/detail/macro_scope.hpp
pabloariasal: yes, this is a great suggestion.

include/nlohmann/detail/macro_scope.hpp
pabloariasal: added same hint as above. The string that could not be found is already being printed, it is `j`

docs/mkdocs/docs/features/enum_conversion.md
pabloariasal: fixed

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum.md
pabloariasal: changed

include/nlohmann/detail/macro_scope.hpp
gregmarr: Ah, right, missed that.

include/nlohmann/detail/macro_scope.hpp
gregmarr: >nested macros don't work (the preprocessor doesn't expand a macro inside another one)

It certainly does.  There are many nested macros in the library.  https://godbolt.org/z/bbWv8Ye7r

>it can lead to silent and hard to debug erros if JSON_NOEXCEPTION is defined and the macro is used -> what should we do in that case?

`JSON_THROW` calls `std::abort` when `JSON_NOEXCEPTION` is defined, unless the user has overridden that by defining `JSON_THROW_USER`.

>the macro itself doens't make any sense if JSON_NOEXCEPTION is defined (hence better not define it in the first place.

The library already has well-defined behavior for what happens to throw calls when `JSON_NOEXCEPTION` is defined.

pabloariasal: ok, very interesting. Thanks for the godbolt link! Now I'm confused, my compiler (gcc 12, fedora 37) doesn't like that:
```
/home/pablo/workspace/json/tests/src/unit-conversions.cpp:1665:9:   required from here
/home/pablo/workspace/json/include/nlohmann/detail/macro_scope.hpp:249:23: error: ‘JSON_THROW’ was not declared in this scope; did you mean ‘JSON_TRY’?
  249 |             JSON_THROW(nlohmann::detail::type_error::create(302,                                                                   \
      |             ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  250 |                     nlohmann::detail::concat("can't serialize enum value ", std::to_string(static_cast<int>(e)),              \
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  251 |                                              " of enum '", #ENUM_TYPE, "' to json. ",                                          \
      |                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  252 |                                              "Did you list the value in NLOHMANN_JSON_SERIALIZE_ENUM_STRICT?"), &j));          \
      |                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/pablo/workspace/json/tests/src/unit-conversions.cpp:1618:1: note: in expansion of macro ‘NLOHMANN_JSON_SERIALIZE_ENUM_STRICT’
 1618 | NLOHMANN_JSON_SERIALIZE_ENUM_STRICT(game,
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I wonder what I'm doing wrong

gregmarr: My first thought was that `JSON_THROW` wasn't defined yet, but that's not the case.  It appears right before that macro, and it's not possible to get there without it being defined.  I'm not sure why gcc would choke on that.  The only other thing I can think of is that somehow you have something equivalent to `#define JSON_THROW JSON_THROW`.  Maybe `#define JSON_THROW_USER JSON_THROW`.  Can you reproduce this on compiler explorer?

docs/examples/nlohmann_json_serialize_enum_3.cpp
```diff
@@ -0,0 +1,47 @@
...
+enum class Color
+{
+    red, green, blue, yellow
+};
+
+NLOHMANN_JSON_SERIALIZE_ENUM_STRICT(Color,
+{
+    { Color::green, "green" },
+    { Color::blue, "blue" },
+    { Color::red, "rot" } // note that serialization for yellow is missing
```
bear24rw: Should this say `"red"` instead of `"rot"`?

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum.md
```diff
@@ -1,14 +1,16 @@
...
 #define NLOHMANN_JSON_SERIALIZE_ENUM(type, conversion...)
+#define NLOHMANN_JSON_SERIALIZE_ENUM_STRICT(type, conversion...)
 ```
 
 By default, enum values are serialized to JSON as integers. In some cases this could result in undesired behavior. If an
 enum is modified or re-ordered after data has been serialized to JSON, the later de-serialized JSON data may be
 undefined or a different enum value than was originally intended.
 
-The `NLOHMANN_JSON_SERIALIZE_ENUM` allows to define a user-defined serialization for every enumerator.
+The `NLOHMANN_JSON_SERIALIZE_ENUM` and `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` macros allow to define a user-defined serialization for every enumerator.
```
bear24rw: The wording of this sentence reads really weird to me, a possible alternative:
```suggestion
The `NLOHMANN_JSON_SERIALIZE_ENUM` and `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` macros allow specifying a user-defined serialization for every enumerator.
```
This then matches the language in the following sentence: "first value specified in the macro".

include/nlohmann/detail/macro_scope.hpp
pabloariasal: I'm really clueless, would it be possible for you to try the branch out with the proposed changes locally? Looks like my preprocessor doesn't see `JSON_THROW` and lets it through to the compiler...

docs/examples/nlohmann_json_serialize_enum_3.cpp
pabloariasal: yes, changed

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum.md
pabloariasal: much better, thank you

PR #4015, opened 2023-04-17
PlatformIO integration

pidpawel:
First of all **thank you** for this amazing library. I really do enjoy working with it.

Secondly this pull request is my stab at adding this library to a PlatformIO environment. PlatformIO is a tool that streamlines environment setup for embedded development. It takes care of things like getting proper compiler variant, setting up all the required compiler flags, linker scripts, etc. It even does some IDE setup, if needed. If that weren't enough it also manages libraries and project dependencies reasonably well. This is where this PR comes in.

Each of the PlatformIO libraries needs a `library.json` file (which this PR includes) and may be uploaded to PlatformIO's registry (https://registry.platformio.org/) (which this PR does not). Technically uploading to said registry is recommended and makes the libraries slightly easier for users to use, but it requires setting up accounts and regular uploads to said service (which would require a CI changes to be sustainable) so I decided not to include it right now. If the current solution will gain traction among users I can help with setting all the "proper" channels.

I also edited `README.md` - my changes are in the section Package managers despite PlatformIO technically being closer to a build system due to the said "registry" feature above, which I hope this library make into one day.

Sorry for the multiple commits in the PR - I kinda couldn't resist testing the GitHub's online editor for a single JSON ;-) file change.

Let me know what you think and have a wonderful day
Paweł

### DIFF
--- a/library.json
+++ b/library.json
@@ -29,5 +29,9 @@
   "examples": [
     "https://github.com/nlohmann/json/tree/develop/docs/examples"
   ],
-  "include": "single_include/nlohmann/json.hpp"
+  "include": "single_include/nlohmann/json.hpp",
+  "build": {
+    "flags": ["-std=c++11"],
+    "srcFilter": ["-<**/*>"]
+  }
 }
--- a/library.json
+++ b/library.json
@@ -1,0 +1,33 @@
+{
+  "name": "nlohmann-json",
+  "version": "3.11.2",
+  "description": "JSON for Modern C++",
+  "keywords": [
+    "json",
+    "serialization",
+    "deserialization",
+    "parser",
+    "generator",
+    "C++"
+  ],
+  "authors": [
+    {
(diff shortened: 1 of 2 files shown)


library.json
```diff
@@ -0,0 +1,36 @@
+{
```
nlohmann, line 1: Is this additional file really needed? I don't want to spam my repository.

nlohmann (changes requested):

library.json
pidpawel: As far as I'm aware - it is :(

puzrin: Yes, I confirm. Without the proposed config, PIO tries to build files in all subfolders (including examples) and ends with errors. As a temporary workaround, I had to make a local copy of the headers.

nlohmann:
Closing, as I don't want to put yet another file to the repository root.

PR #4036, opened 2023-05-20
feat: Rebase `feature/optional` to `develop`

### LINKED ISSUE #1749: [C++17] Allow std::optional to convert to nlohmann::json

fsandhei:
Rebasing `feature/optional` in attempt to revive this PR, referring to https://github.com/nlohmann/json/pull/2117.

Sorry for long time between responses.

I may have been going forward with this the "wrong" way, so I apologize in advance.

Please let me know if this should be handled differently.

## TODO
There is one test that is failing: Conversions from a default initialized `nlohmann::json` to `std::optional<T>` seems to fail. In the test case below, the last assertion fails with an exception. This is tested on Arch Linux with GCC 13.1.1.

I was not able to figure out how to fix this so I need some assistance here.

```
/home/fredrik/dev/github/cpp/fredriksd/json/tests/src/unit-conversions.cpp:1573:
TEST CASE:  std::optional
  null

/home/fredrik/dev/github/cpp/fredriksd/json/tests/src/unit-conversions.cpp:1573: ERROR: test case THREW exception: exception thrown in subcase - will translate later when the whole test case has been exited (cannot translate while there is an active exception)

/home/fredrik/dev/github/cpp/fredriksd/json/tests/src/unit-conversions.cpp:1573:
TEST CASE:  std::optional

DEEPEST SUBCASE STACK REACHED (DIFFERENT FROM THE CURRENT ONE):
  null

/home/fredrik/dev/github/cpp/fredriksd/json/tests/src/unit-conversions.cpp:1573: ERROR: test case THREW exception: [json.exception.type_error.302] type must be string, but is null
```

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -53,7 +53,7 @@ void from_json(const BasicJsonType& j, std::optional<T>& opt)
     }
     else
     {
-        opt = j.template get<T>();
+        opt.emplace(j.template get<T>());
     }
 }
 #endif
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -44,6 +44,7 @@ inline void from_json(const BasicJsonType& j, typename std::nullptr_t& n)
 }
 
 #ifdef JSON_HAS_CPP_17
+#ifndef JSON_USE_IMPLICIT_CONVERSIONS
 template<typename BasicJsonType, typename T>
 void from_json(const BasicJsonType& j, std::optional<T>& opt)
 {
@@ -56,7 +57,9 @@ void from_json(const BasicJsonType& j, std::optional<T>& opt)
         opt.emplace(j.template get<T>());
     }
 }
-#endif
+
+#endif // JSON_USE_IMPLICIT_CONVERSIONS
+#endif // JSON_HAS_CPP_17
 
 // overloads for basic_json template parameters
 template < typename BasicJsonType, typename ArithmeticType,
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -21,7 +21,11 @@
 #include <valarray> // valarray
 
 #ifdef JSON_HAS_CPP_17
-    #include <optional> // optional
+    #if __has_include(<optional>)
+        #include <optional>
+    #elif __has_include(<experimental/optional>)
+        #include <experimental/optional>
+    #endif
 #endif
 
 #include <nlohmann/detail/exceptions.hpp>
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -13,12 +13,16 @@
 #include <forward_list> // forward_list
 #include <iterator> // inserter, front_inserter, end
 #include <map> // map
+#ifdef JSON_HAS_CPP_17
+    #include <optional> // optional
+#endif
 #include <string> // string
 #include <tuple> // tuple, make_tuple
 #include <type_traits> // is_arithmetic, is_same, is_enum, underlying_type, is_convertible
 #include <unordered_map> // unordered_map
 #include <utility> // pair, declval
 #include <valarray> // valarray
+
 
 #include <nlohmann/detail/exceptions.hpp>
 #include <nlohmann/detail/macro_scope.hpp>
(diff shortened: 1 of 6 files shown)


fredriksd:
Think something got messed up on comparison, so closing to not make noise. Sorry about that.

fsandhei:
Changed the base branch to `develop`.

I'm sorry if I'm making it a bit difficult here; I forked the repository and rebased `feature/optional` to `develop` and created this PR against `nlohmann:feature/develop` but I noticed that included 590 commits (which makes sense) but it cluttered the actual change here, so I instead changed it towards `nlohmann:develop`.

lightyear15:
I am also interested in having ``std::optional`` support landing in future versions of this library.
I see there was a lot of discussion on how to properly support and convert c++ optional fields vs json nullable fields.
Is there a final decision on the matter? How can we revive this PR and push forward support of ``std::optional`` fields?
@nlohmann

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -43,6 +43,21 @@ inline void from_json(const BasicJsonType& j, typename std::nullptr_t& n)
...
+template<typename BasicJsonType, typename T>
+void from_json(const BasicJsonType& j, std::optional<T>& opt)
+{
+    if (j.is_null())
+    {
+        opt = std::nullopt;
+    }
+    else
+    {
+        opt = j.template get<T>();
```
lightyear15:
```suggestion
        opt = std::make_optional(j.template get<T>());
```

tests/src/unit-conversions.cpp
```diff
@@ -1569,4 +1569,70 @@ TEST_CASE("JSON to enum mapping")
...
+        CHECK(std::optional<std::string>(j_null) == std::nullopt);
+    }
+
+    SECTION("string")
+    {
+        json j_string = "string";
+        std::optional<std::string> opt_string = "string";
+
+        CHECK(json(opt_string) == j_string);
+        CHECK(std::optional<std::string>(j_string) == opt_string);
```
lightyear15, line 1613:
```suggestion
        CHECK(j_null.get<std::optional<std::string>>() == opt_string);
```

fsandhei:
See that clang-tidy is failing on some checks. Apparently this is on develop, too. I see that it is related to a supposed new check from clang-tidy v19: https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-designated-initializers.html.

nlohmann:
> See that clang-tidy is failing on some checks. Apparently this is on develop, too. I see that it is related to a supposed new check from clang-tidy v19: [clang.llvm.org/extra/clang-tidy/checks/modernize/use-designated-initializers.html](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-designated-initializers.html).

I'm on it. May take some more time.

lightyear15:
has this PR stalled again?

fsandhei:
I'm currently blocked by https://github.com/nlohmann/json/pull/4311.

fsandhei:
There has not been any activity in #4311 in a while. I wonder if this would be okay to merge?

nlohmann:
Thanks for keeping this alive. I will have a look at this on Saturday.

fsandhei:
> Thanks for keeping this alive. I will have a look at this on Saturday.

Happy to help! Hopefully this can go in soon.

nlohmann:
Please suppress the last clang-tidy warning and not change the code.

fsandhei:
> Please suppress the last clang-tidy warning and not change the code.

Reverted change with [296549c](https://github.com/nlohmann/json/pull/4036/commits/296549c3ad3941ccf007db543c8e0d3f908a67d0).

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #4039, opened 2023-05-22
Use template get instead of get in examples

### LINKED ISSUE #3827: Compile Error on g++ using get() function
### Description

When iterating over a `nlohmann::json` object and then casting the **values** using `value.get<std::string>()`, we observe that `g++` gives a compile error. On `clang` and `msvc` everything compiles just fine.

Replacing `value.get<std::string>()` with `std::string(value)` fixes the problem on `g++`. But I wonder why the `get` function does not work?

### Reproduction steps

See https://godbolt.org/z/Gxv7dbsjW

I would expect the "minimal code example" to compile as is on `g++`. But it does not.

### Minimal code example

```Shell
// https://godbolt.org/z/Gxv7dbsjW

#include <nlohmann/json.hpp>
#include <iostream>

template<class K, class V>
void convert(const nlohmann::json& sourceData, std::map<K, V>& targetMap)
{
    for (const auto& [key, value] : sourceData.items())
    {
        if constexpr(std::is_same_v<V, std::string>)
            targetMap.emplace(key, value.get<std::string>());
            //targetMap.emplace(key, std::string(value));  // Only way to fix g++ is to use this line
        else
            throw std::runtime_error("Case not covered");
    }
}

int main()
{
    const nlohmann::json data = {
        {"Year2", "2022" },
        {"Year3", "2023" },
        {"Year9", "2029" }
    };
    std::map<std::string, std::string> m;
    convert(data, m);

    for (const auto& [k, v] : m)
        std::cout << " - " << k << ": " << v << std::endl;

    return 0;
}
```

### Error messages

```Shell
<source>:10:42: warning: expected 'template' keyword before dependent template name [-Wmissing-template-keyword]
   10 |             targetMap.emplace(key, value.get<std::string>());
      |                                          ^~~
      |                                          template

<source>:10:57: error: expected primary-expression before '>' token
   10 |             targetMap.emplace(key, value.get<std::string>());
      |                                                         ^

<source>:10:59: error: expected primary-expression before ')' token
   10 |             targetMap.emplace(key, value.get<std::string>());
      |                                                           ^
```

g++ (Linux)

### Library version

3.11.1 & trunk

### Validation

### LINKED ISSUE #4038: Explicit conversion example in docs should use `template get`
### Description

Follow up of https://github.com/nlohmann/json/issues/3827 .

As `get<>()` may not always work, the documentation SHOULD use `template get<>()` instead of `get<>()` in the examples to avoid confusion.

### Reproduction steps

use the explicit conversion described in the example

expected: works

actual: compilation failure

### Minimal code example

_No response_

### Error messages

_No response_

g++ 12 on GNU/Linux

### Library version

3.11.2

### Validation

tusooa:
https://github.com/nlohmann/json/issues/4038

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,7 +301,7 @@ Note the difference between serialization and assignment:
 json j_string = "this is a string";
 
 // retrieve the string value
-auto cpp_string = j_string.get<std::string>();
+auto cpp_string = j_string.template get<std::string>();
 // retrieve the string value (alternative when a variable already exists)
 std::string cpp_string2;
 j_string.get_to(cpp_string2);
@@ -310,7 +310,7 @@ j_string.get_to(cpp_string2);
 std::string serialized_string = j_string.dump();
 
 // output of original string
-std::cout << cpp_string << " == " << cpp_string2 << " == " << j_string.get<std::string>() << '\n';
+std::cout << cpp_string << " == " << cpp_string2 << " == " << j_string.template get<std::string>() << '\n';
 // output of serialized value
 std::cout << j_string << " == " << serialized_string << std::endl;
 ```
@@ -483,7 +483,7 @@ for (auto& element : j) {
 }
 
 // getter/setter
-const auto tmp = j[0].get<std::string>();
+const auto tmp = j[0].template get<std::string>();
 j[1] = 42;
 bool foo = j.at(2);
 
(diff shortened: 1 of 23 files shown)


nlohmann (changes requested):
Please also check the documentation (e.g., in https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/api/basic_json/get.md) and the related examples.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #4046, opened 2023-06-04
Fix compile error with _HAS_STATIC_RTTI=0

ALF-ONE:
Fix compile error with _HAS_STATIC_RTTI=0
See: https://github.com/microsoft/STL/blob/main/stl/inc/any#L15

* * *

## Pull request checklist

## Please don't


nlohmann:
Can you please add some documentation to this (see https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/api/macros/json_no_io.md for example). Furthermore, is there an issue that this MR is closing?

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -133,6 +133,14 @@
     #endif
 #endif
 
+#ifndef JSON_HAS_STATIC_RTTI
+    #if defined(_HAS_STATIC_RTTI) && _HAS_STATIC_RTTI != 0
+        #define JSON_HAS_STATIC_RTTI 1
+    #else
+        #define JSON_HAS_STATIC_RTTI 0
```
gregmarr, line 140: It looks like `_HAS_STATIC_RTTI` is a MS extension and when it's set to `0`, certain functionality is disabled.  Shouldn't this then default to 1:

    #if !defined(_HAS_STATIC_RTTI) || _HAS_STATIC_RTTI != 0
        #define JSON_HAS_STATIC_RTTI 1
    #else
        #define JSON_HAS_STATIC_RTTI 0
    #endif

nlohmann:
Please update to the latest `develop` to have a fixed CI.

nlohmann:
Thanks!

PR #4047, opened 2023-06-06
Fix Clang-Tidy warnings

nlohmann:
The CI errored in latest merge requests due to Clang-Tidy errors.


include/nlohmann/ordered_map.hpp
```diff
@@ -122,7 +122,7 @@ template <class Key, class T, class IgnoredLess = std::less<Key>,
     {
         for (auto it = this->begin(); it != this->end(); ++it)
         {
-            if (m_compare(it->first, key))
+            if (m_compare(it->first, std::forward<KeyType>(key)))
```
gregmarr: I'm not sure these are valid, since it's in a loop.  We don't want to move from this more than once.  (`std::equal_to.operator()` doesn't move, so that's not an issue with the default, but could be with another function.)

nlohmann: I'll check.

gregmarr (approved):

PR #4055, opened 2023-06-16
Improve ordered map

Tomerkm:

1. I updated the following method:
 T& at(const key_type& key)

it will call to the following method:
 **const** T& at(const key_type& key) **const**
The idea is to remove duplicate code

2. For each function that has a loop call to the method this->end() several times, I did the following changes:
For preventing it be called several times, the returned variable is stored in a variable.

* * *

## Pull request checklist

## Please don't


include/nlohmann/ordered_map.hpp
```diff
@@ -43,17 +42,24 @@ template <class Key, class T, class IgnoredLess = std::less<Key>,
...
+        : Container{first, last, alloc}
+    {}
+    ordered_map(std::initializer_list<value_type> init, const Allocator& alloc = Allocator())
+        : Container{init, alloc}
+    {}
 
     std::pair<iterator, bool> emplace(const key_type& key, T&& t)
     {
-        for (auto it = this->begin(); it != this->end(); ++it)
+        auto endItr = this->end();
```
nlohmann, line 61: The [Container requirements](https://en.cppreference.com/w/cpp/named_req/Container) require `end()` to have constant complexity. I wonder whether this change would actually create different assembler code. Did you profile this?

nlohmann (changes requested):
Please add a summary of the changes you made and the reason for the changes.

include/nlohmann/ordered_map.hpp
Tomerkm: I did not profile this.
The idea was to save the value in a variable instead of calling the same function several times
I saved the value end() in a variable.

https://stackoverflow.com/questions/47401691/does-calling-a-function-multiple-times-decrease-the-programm-speed

Tomerkm:
> Please add a summary of the changes you made and the reason for the changes.

I added summary and explain why I did the changes.
Thanks

include/nlohmann/ordered_map.hpp
```diff
@@ -96,31 +101,23 @@ template <class Key, class T, class IgnoredLess = std::less<Key>,
...
-        {
-            if (m_compare(it->first, key))
-            {
-                return it->second;
-            }
-        }
-
-        JSON_THROW(std::out_of_range("key not found"));
+        const ordered_map<Key, T, IgnoredLess, Allocator>& cThis = *this;
+        return const_cast<T&>(cThis.at(key));
```
nlohmann, line 113: Please don't do this. Casting away `const` is way more dangerous than the benefit of avoiding some code repetition.

include/nlohmann/ordered_map.hpp
nlohmann: Yes, but the `end()` function is cheap. Furthermore, the change could change the semantics of the code if the range is changed during iteration (which it is not here, but in other functions.

nlohmann (changes requested):

include/nlohmann/ordered_map.hpp
Tomerkm: @nlohmann 
Thanks for correcting me.
I closed the PR.
I would like to know why the using of const cast is dangerous.
Thanks in advance

PR #4166, opened 2023-09-28
Conditional Compilation for C++17 Compatibility

iconnor:
This pull request wraps two specific files in a conditional compilation block to allow builds using the C++17 standard. The goal is to enhance compatibility without affecting existing functionality.

Motivation:
I am utilizing this library in a PlatformIO project that uses C++17 for both native CI/CD and testing purposes. While I understand that the library aims to use the latest C++ standards, there are scenarios where projects may need to use older but still widely-used standards like C++17.

Changes:
Wrapped the offending code in #if __cplusplus >= 202002L ... #endif blocks to allow compilation under C++17.
Benefits:
Increases the library's compatibility with projects that are not yet able to migrate to C++20.
Allows for more diverse usage in larger codebases that may be using multiple C++ standards.
No impact on existing functionality when used in a C++20 environment.
Alternative Considered:
An alternative could be to completely exclude these files during the build process, but conditional compilation offers a less invasive approach.
* * *

## Pull request checklist

## Please don't

### DIFF
--- a/docs/examples/operator_spaceship__const_reference.c++20.cpp
+++ b/docs/examples/operator_spaceship__const_reference.c++20.cpp
@@ -1,3 +1,4 @@
+#if __cplusplus >= 202002L
 #include <compare>
 #include <iostream>
 #include <nlohmann/json.hpp>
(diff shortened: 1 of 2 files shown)


gregmarr:
This library is actually 100% C++11 compatible.  It can take advantage of later standards if you compile with them, but it doesn't require anything later than that.

The files you modified are examples used in the documentation, not part of the library itself.

nlohmann:
What your code is proposing is already done in the library: any feature that is not pure C++11 is guarded by macros. The MR like this makes no sense, because the example files are just documentation.

Issue #4165, opened 2023-09-26
gcc13.2 compile fail when using std::string to compare the json object


zhangjian3032:
### Description

Kindly refer to this link: `https://godbolt.org/z/c3fzjnqzG`,
okay with gcc 12.3 
failed with gcc 13.1 and 13.2

### Reproduction steps

refer to description

### Expected vs. actual results

expect using gcc13.2 can be compiled successfully
```
    if (std::string("bar") == j["xxx"])
    {
       
    }
 ```

### Minimal code example

```Shell
#include <iostream>
#include <string>
#include <nlohmann/json.hpp>

int main()
{
    nlohmann::json j;


    if (std::string("bar") == j["xxx"])
    {
       
    }
    
    return 0;
}
```
```


### Error messages

_No response_

### Compiler and operating system

gcc12.3 and gcc13.1

### Library version

3.11.1

### Validation

- [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/nlohmann/json/tree/develop) branch is used.
- [ ] I can successfully [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests).

**Comment by gregmarr** [comment]
This only happens with C++20 mode enabled, and appears to be a result of the updated C++ compare code.  It works in one direction but not the other.

This fails

    if (std::string("bar") == j["xxx"])

This is good

    if (j["xxx"] == std::string("bar"))

**Comment by colbychaskell** [comment]
I can confirm I get the same behavior running on ubuntu 23.04 with gcc 13.2. 

It looks like the reason why this only appears for c++20 is the template requirement for the operator==

```
template<typename ScalarType>
requires std::is_scalar_v<ScalarType>
bool operator==(ScalarType rhs) const noexcept
{
    return *this == basic_json(rhs);
}
```

If the 'requires' line is commented out it appears to use this version of the == operator for this case. 

I believe the reason why it works when lhs and rhs are swapped is because of the operator value_t() definition which lets it work with the std::string == operator.

I'm not sure why this function isn't just setup to work with all compatible types which would work with the basic_json(rhs) conversion?

**Comment by nlohmann** [comment]
I did not write this code myself, and did not find the time to read into these operators. Help is more than welcome here!

**Comment by fsandhei** [comment]
Without having worked on the library details I cannot give good rationales, but I decided to try out changing the constraint so `operator==` is called for the compatible types, like the equivalent catch - all constructor.

```
[...]

PR #4181, opened 2023-10-08
Change `operator==` overload that constrains to `is_scalar_v` to constrain to compatible types.

fsandhei:
https://github.com/nlohmann/json/issues/4165 shows an issue with the overload resolution when attempting to compare a `std::string` with `json` from left to right. This can be fixed by changing the constraint to instead of focusing on scalar types, we constrain the comparison operator to "compatible types", in par with the catch-call constructor.

## Pull request checklist

## Please don't


nlohmann:
Can you please try to amalgamate with an older version of astyle, like astyle 3.2? There was a breaking change in astyle which makes the output differ unfortunately.

fsandhei:
> Can you please try to amalgamate with an older version of astyle, like astyle 3.2? There was a breaking change in astyle which makes the output differ unfortunately.

Sure, I'll do it some time later today after work.

fsandhei:
First, sorry, @nlohmann, for the trouble I'm causing here.

So, I tried amalgamating again. There seems to be some issues or inconsistencies with the astyle version 3.2.1 from the AUR versus 3.2.1-build from apt:
   * --squeeze-lines is not defined in the one from Arch, so I had to
     comment that one out in order to make the amalgamate step run.
  * As far as I see, `squeeze-line` was introduced in 3.3.x.
   * A bunch of files are affected by the amalgamate, which does not
     seem right.

Would you want me to perhaps unstage all the other unrelated files that got affected by this?

nlohmann:
This is what I am using locally (patched Makefile):

```Makefile
# call the Artistic Style pretty printer on all source files
pretty:
	/Users/niels/Downloads/astyle/build/astyle \
	    --style=allman \
	    --indent=spaces=4 \
	    --indent-modifiers \
	    --indent-switches \
	    --indent-preproc-block \
	    --indent-preproc-define \
	    --indent-col1-comments \
	    --pad-oper \
	    --pad-header \
	    --align-pointer=type \
	    --align-reference=type \
	    --add-braces \
	    --convert-tabs \
	    --close-templates \
	    --lineend=linux \
	    --preserve-date \
	    --suffix=none \
	    --formatted \
	   $(SRCS) $(TESTS_SRCS) $(AMALGAMATED_FILE) $(AMALGAMATED_FWD_FILE) docs/examples/*.cpp
```

with Astyle 3.1.

fsandhei:
Thank you for so quick response. The amalgamation seems happy now.

nlohmann:
> Thank you for so quick response. The amalgamation seems happy now.

Thanks for your effort!

fsandhei:
I see that the GCC 10 CI run failed. It seems to be not able to parse the concept, which is odd as it [should be supported in GCC 10](https://en.cppreference.com/w/cpp/compiler_support/20).

For the sake of getting this in and have the least amount of hassle I don't use `concept`.

However I was pondering on the idea of having a `concept` which could replace these type traits definitions for recognizing "compatible types". I tinkered with this simple concept below. It worked perfectly fine. I don't know if it is missing any details, though. Would appreciate hearing your opinion on this, as I'm far from an expert :)

```cpp
#ifdef JSON_HAS_CPP_20
   #include <concepts>
    template <typename T, typename BasicJsonType, typename U = uncvref_t<T>>
    concept CompatibleType = requires(const U& type, BasicJsonType& j) {
        { to_json(j, type) } -> std::convertible_to<void>;
    };
#endif
```

The use of this would be quite simple, as we'd then can constrict constructors / operators with this concept:
```
template <CompatibleType<basic_json_t> T>
operator==(T rhs) const noexcept
{
   ...
}
```

EDIT: I realize I goofed myself. It is supported but `JSON_HAS_CPP_20` is not defined because `g++10` requires `-std=c++2a`. I'm able to get past this by also checking for `__cpp_concepts` to be defined at where we define the concept.

fsandhei:
I'm not really sure what is Codacy is actually complaining about. The struct member is being used in the type trait. It might be a bug in Codacy. Do you have any suggestions @nlohmann?

nlohmann:
> I'm not really sure what is Codacy is actually complaining about. The struct member is being used in the type trait. It might be a bug in Codacy. Do you have any suggestions @nlohmann?

Indeed looks like a false positive. I set it to "ignore".

d3matt:
@nlohmann I was surprised to run into this when trying to move a repo to C++20. Is there anything else need to get this merged?

nlohmann:
Thanks a lot for working on this, and sorry it sat for so long! The root cause of #4165 turned out to be the hand-written `operator!=`, which under C++20's P2468R2 rules stops `==` from being tried with the arguments reversed. This was fixed on `develop` in #5253, which removed `operator!=` entirely. `std::string == json` now compiles with current GCC and Clang in C++20 mode. Since this PR re-introduces a member `operator!=`, it would bring the problem back (for example, `int != json` becomes ambiguous on Clang and `unit-comparison.cpp` no longer compiles), so I'm closing it as superseded by #5253. Thanks again for the analysis in the issue; it pointed in the right direction.

*(This comment was written by Claude Code.)*

PR #4226, opened 2023-11-29
Use Clang-Format

nlohmann:
This PR replaces Artistic Style by Clang-Format as tool to indent the code. Astyle introduced a breaking change a while ago which made working with the tool very cumbersome.


include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -44,10 +43,7 @@ inline void from_json(const BasicJsonType& j, typename std::nullptr_t& n)
 }
 
 // overloads for basic_json template parameters
-template < typename BasicJsonType, typename ArithmeticType,
-           enable_if_t < std::is_arithmetic<ArithmeticType>::value&&
-                         !std::is_same<ArithmeticType, typename BasicJsonType::boolean_t>::value,
-                         int > = 0 >
+template<typename BasicJsonType, typename ArithmeticType, enable_if_t<std::is_arithmetic<ArithmeticType>::value && !std::is_same<ArithmeticType, typename BasicJsonType::boolean_t>::value, int> = 0>
```
gregmarr: Ugh, not sure this is a good change.  Is there a setting that will prevent it?

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -100,13 +96,12 @@ inline void from_json(const BasicJsonType& j, typename BasicJsonType::string_t&
...
-        std::is_assignable<StringType&, const typename BasicJsonType::string_t>::value
-        && is_detected_exact<typename BasicJsonType::string_t::value_type, value_type_t, StringType>::value
-        && !std::is_same<typename BasicJsonType::string_t, StringType>::value
-        && !is_json_ref<StringType>::value, int > = 0 >
+template<
+    typename BasicJsonType,
+    typename StringType,
+    enable_if_t<
+        std::is_assignable<StringType&, const typename BasicJsonType::string_t>::value && is_detected_exact<typename BasicJsonType::string_t::value_type, value_type_t, StringType>::value && !std::is_same<typename BasicJsonType::string_t, StringType>::value && !is_json_ref<StringType>::value,
+        int> = 0>
```
gregmarr: Interesting that this one stayed multi-line.  Perhaps there is a max line length setting?

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -197,99 +185,83 @@ inline void from_json_array_impl(const BasicJsonType& j, typename BasicJsonType:
...
 
-template < typename BasicJsonType, typename ConstructibleArrayType,
-           enable_if_t <
-               is_constructible_array_type<BasicJsonType, ConstructibleArrayType>::value&&
-               !is_constructible_object_type<BasicJsonType, ConstructibleArrayType>::value&&
-               !is_constructible_string_type<BasicJsonType, ConstructibleArrayType>::value&&
-               !std::is_same<ConstructibleArrayType, typename BasicJsonType::binary_t>::value&&
-               !is_basic_json<ConstructibleArrayType>::value,
-               int > = 0 >
+template<typename BasicJsonType, typename ConstructibleArrayType, enable_if_t<is_constructible_array_type<BasicJsonType, ConstructibleArrayType>::value && !is_constructible_object_type<BasicJsonType, ConstructibleArrayType>::value && !is_constructible_string_type<BasicJsonType, ConstructibleArrayType>::value && !std::is_same<ConstructibleArrayType, typename BasicJsonType::binary_t>::value && !is_basic_json<ConstructibleArrayType>::value, int> = 0>
```
gregmarr: This one is much longer, must be some other trigger for single line vs multi-line.

include/nlohmann/detail/conversions/to_chars.hpp
```diff
@@ -381,96 +382,95 @@ inline cached_power get_cached_power_for_binary_exponent(int e)
...
-            { 0xAC2820D9623BF429,   880,  284 },
-            { 0x80444B5E7AA7CF85,   907,  292 },
-            { 0xBF21E44003ACDD2D,   933,  300 },
-            { 0x8E679C2F5E44FF8F,   960,  308 },
-            { 0xD433179D9C8CB841,   986,  316 },
-            { 0x9E19DB92B4E31BA9,  1013,  324 },
-        }
-    };
+            {
+                {0xAB70FE17C79AC6CA, -1060, -300},
```
gregmarr: Another questionable change.  I find the whitespace inside the braces is helpful.  Interesting that it did multi-line column alignment of comments on the includes and removed column alignment of the parameters here.

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -30,8 +30,7 @@
 #include <nlohmann/detail/value_t.hpp>
 
 NLOHMANN_JSON_NAMESPACE_BEGIN
-namespace detail
-{
+namespace detail {
```
gregmarr: `.clang-format` has BraceWrapping/AfterNamespace set to false, which is the opposite of the general format of this repo.

include/nlohmann/detail/conversions/from_json.hpp
nlohmann: There is, but I have not tried it yet.

include/nlohmann/detail/conversions/to_chars.hpp
nlohmann: I've seen it, but I have not found the config key to fix this yet.

include/nlohmann/detail/conversions/from_json.hpp
nlohmann: Thanks - I'll adjust.

include/nlohmann/json.hpp
```diff
@@ -124,20 +125,15 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     using basic_json_t = NLOHMANN_BASIC_JSON_TPL;
     using json_base_class_t = ::nlohmann::detail::json_base_class<CustomBaseClass>;
 
-  JSON_PRIVATE_UNLESS_TESTED:
-    // convenience aliases for types residing in namespace detail;
-    using lexer = ::nlohmann::detail::lexer_base<basic_json>;
+    JSON_PRIVATE_UNLESS_TESTED :
```
gregmarr, line 128: You can tell clang-format what this macro means, so it knows it's an access specifier: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#macros

include/nlohmann/detail/abi_macros.hpp
```diff
@@ -47,54 +47,40 @@
...
 
 #ifndef NLOHMANN_JSON_NAMESPACE_BEGIN
-#define NLOHMANN_JSON_NAMESPACE_BEGIN                \
-    namespace nlohmann                               \
-    {                                                \
-    inline namespace NLOHMANN_JSON_NAMESPACE_CONCAT( \
-                NLOHMANN_JSON_ABI_TAGS,              \
-                NLOHMANN_JSON_NAMESPACE_VERSION)     \
-    {
+    #define NLOHMANN_JSON_NAMESPACE_BEGIN                                                                                                                      \
```
gregmarr, line 75: Looks like `AlignEscapedNewlines: Right` aligns to `ColumnLimit` if it's non-zero.  That's unfortunate.

gregmarr (commented):
There are some interesting changes.  I wonder if the result would be different if you went back to the astyle code and formatted again with the current options rather than incremental changes with different options.

include/nlohmann/detail/conversions/to_chars.hpp
gregmarr: `Cpp11BracedListStyle` affects the whitespace.

include/nlohmann/json.hpp
nlohmann: Interesting! I'll give it a try.

include/nlohmann/detail/conversions/to_chars.hpp
nlohmann: I will try without.

include/nlohmann/detail/conversions/from_json.hpp
nlohmann: I think it is a bit better with max width of ColumnLimit=160.

nlohmann:
I call it a day. There seems to be some work needed to move the `NOLINT` comments to the right place after reformatting. Could have been crazy to expect Clang-Format to know what Clang-Tidy is doing...

Furthermore, there seems to be the same version nightmare as with astyle. I am using Clang-Format 17.0.6 locally and Clang-Format 18.0.0 in the CI, and already the results are not the same. Not sure how to make it easy for contributors to use the "right" version.

Finally, Clang-Format is incredibly slow. Like 7 minutes to format `single_include/json.hpp`. Astyle took some seconds.

I'll pick this up next week.

t-b:
>  I am using Clang-Format 17.0.6 locally and Clang-Format 18.0.0 in the CI, and already the results are not the same. 

This is to be expected unforunately. Due to the complexity the devs decided not to be backwards compatible for every edge case.

> Finally, Clang-Format is incredibly slow. Like 7 minutes to format single_include/json.hpp. Astyle took some seconds.

clang-format 17.03 takes here ~19minutes locally.

PR #4111, opened 2023-08-15
Bazel: Improve support for bzlmod

Vertexwahn:
The Bazel build system now supports a new way of handling 3rd party dependencies, such as JSON for Modern C++. For this new way, a `MODULE.bazel` file is needed - in the long term, the `WORKSPACE.bazel` file will not be needed anymore and will be deleted in the future.

PR #4314, opened 2024-03-17
chore(bazel): add MODULE.bazel files for bzlmod

mmorel-35:

Integrate MODULE.bazel from https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/nlohmann_json
Then it shall be possible to publish automatically with the new version with the help of bcr bot : https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md#contribute-a-bazel-module

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,4 @@
+module(
+    name = "nlohmann_json",
+    compatibility_level = 1,
+)
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,5 @@
 
 # Swift Package Manager build directory
 /.build
+
+MODULE.bazel.lock


nlohmann:
Is adding file `MODULE.bazel` really necessary?

mmorel-35:
To my understanding this will at least avoid generate [a patch file](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/nlohmann_json/3.11.3/patches/module_dot_bazel.patch) every time a new release is published to the BCR. 

Having this file in the repository seems to be a trouble, can you explain why ?

nlohmann:
I personally do not maintain any of the package manager integrations. Hence I am reluctant to having these files in the repository.

mmorel-35:
@Vertexwahn did the configuration in the BCR maybe he has an opinion on that.
Maybe MODULE.bazel shall even replace WORKSPACE.bazel

MODULE.bazel
```diff
@@ -0,0 +1,5 @@
+module(
+    name = "nlohmann_json",
+    version = "3.11.3",
```
Vertexwahn: Do no include a version here - this has still to be done via BCR patch file - abseil uses version = "head" but this does also no make to much sense since if you go back to an very old version it will still head - I would not include a version here since its always somehow debatable

Vertexwahn:
@mmorel-35 I proposed to add a `MODULE.bazel` file to this repo already here -> https://github.com/nlohmann/json/pull/4111

I personally would suggest to add a `MODULE.bazel` it to this repo since this makes life with Bazel easier (no patch at BCR, usage without a BCR is easier, etc.) 

In the current state of Bazel (7.x) we need a `WORKSPACE`/`WORKSPACE.bazel` (which can be empty) and a `MODULE.bazel` file. The empty `WORKSPACE`/`WORKSPACE.bazel` is there only for legacy reasons - in the future this file will not be needed anymore.

Currently, we are in a transition phase from Bazel 7 -> 8 (maybe end of this year) -> 9 (maybe end of next year) - i.e. it will take some time until we can remove the WORKSPACE file.

@nlohmann For me it seems that there are people around that care about the Catch2 Bazel support.
You have the cost of having one additional  Bazel related file vs. a community of Bazel users that can easier pick Catch2 for the development of their own projects + Avoid this discussion in the future ;) - anyways it your decision - close the PR if you do not want to have more Bazel files ;)

MODULE.bazel
mmorel-35: Do you mean this ?
```suggestion
    version = "",
```

mmorel-35: Or this ?
```suggestion
    version = "0.0.0",
```

Vertexwahn: would delete the whole version line (since it can easily happen that Catch2 version is updated, but one is forgetting about the Bazel build):
```
module(
    name = "nlohmann_json",
    compatibility_level = 1,
)
```
The correct version can than be introduced via a patch file in BCR. But this is opinionated. Different project handle it differently (e.g. rules_go, abeil-cpp, etc.) - was just an idea to leave version out of this file to not get it wrong

Vertexwahn (approved):

PR #4324, opened 2024-03-27
Allow preventing implicit converions to basic_json

### LINKED ISSUE #2226: std::tuple dangling reference - implicit conversion

I created a nlohmann::json, which I passed to a std::apply. When I wanted to get the json value at a parameter specified function (const nlohmann::json&), it referenced some undefined memory (g++). 

I tried to simplify the code, when I came across the following:
`std::tuple<const nlohmann::json&>::tuple(std::tuple<nlohmann::json&>&&)` constructor creates a temporary object and a dangling reference.
This happens because `nlohmann::json` implicit conversion exists from `std::tuple<nlohmann::json&>`, and the tuple [constructor](https://en.cppreference.com/w/cpp/utility/tuple/tuple) calls the variant (3) instead of (5). 

```
#include <iostream>
#include <tuple>
#include <cassert>

#include "json.hpp"

int main() {
    nlohmann::json j = true;
    std::tuple<const nlohmann::json&> tup(std::forward_as_tuple(j)); 
        // std::forward_as_tuple(std::as_const(j)) works well.
    assert(&j == &std::get<0>(tup));
}
```

[try g++](https://wandbox.org/permlink/RMi9jkaqZdfefePo)

With clang it doesn't compile:
`tuple:232:24: error: reference member '__value_' binds to a temporary object whose lifetime would be shorter than the lifetime of the constructed object`

g++: no assertion happens on the example code.
clang: compiles.

g++: tuple creates a temporary object and creates a dangling reference.
clang: not compiles.

- Compiler: g++ / clang (in wandbox)
- Operating system: ubuntu based (18.04) + wandbox

- [ ] latest release version 3.7.3
- [X] the `develop` branch

(but all version is affected)

- [X] yes

iFreilicht:
Previously, `JSON_USE_IMPLICIT_CONVERSIONS` only prevented implicit conversions _from_ `basic_json` to other types.

With this change, implicit conversions _to_ `basic_json` are also prevented by that macro.

Fixes #2226.

I tested this change also in https://github.com/iFreilicht/json-questions/commit/c0b2f53ebe279b6efd933a73e2b26f1bf2030261, the original repro repo from https://github.com/nlohmann/json/issues/2226#issuecomment-2023142182.

## Pull request checklist


schaumb:
As I thought, this will not work easily because a lot of code depends on implicit construction.

iFreilicht:
Ah yes, it seems that the CI is running more tests that I didn't run locally. I'll have a look at that.

gregmarr:
Could we use a new define `JSON_USE_IMPLICIT_CONSTRUCTORS` for this that defaults to true so that users can opt in when they're ready?

nlohmann:
Good idea.

iFreilicht:
So, after playing whack-a-mole with the tests for 2 hours, I think just making the constructor explicit is not enough, it breaks an incredible amount of features, including safe ones (see the latest WIP commit).

A lot of functions already rely on the implicit conversion, so it seems necessary to add overloads for `CompatibleTypes` to many functions that then convert internally. I'm not sure that's much better than the current state of things.

Maybe there's a smarter way to prevent nested temporary values from being implicitly converted? Maybe if they're required to be moved instead of allowing const references?

cmake/ci.cmake
```diff
@@ -491,6 +491,20 @@ add_custom_target(ci_test_noimplicitconversions
     COMMENT "Compile and test with implicit conversions switched off"
 )
 
+###############################################################################
+# Disable implicit converstructors.
```
gregmarr, line 495: constructors

docs/mkdocs/docs/api/macros/json_use_implicit_constructors.md
```diff
@@ -0,0 +1,95 @@
...
+    When `JSON_USE_IMPLICIT_CONSTRUCTORS` is defined to `0`, the code above no longer compiles. Instead, it must be
+    written like this:
+
+    ```cpp
+    json j = json("Hello, world!");
+    ```
+
+??? example
+
+    Disabling constructors is particularly useful to avoid implicit allocations that can lead to invalid memory access:
```
gregmarr, line 45: Disabling implicit constructors

gregmarr (commented):
Yeah, that is a lot of casting needed.  Not very attractive.

iFreilicht:
I will not continuing work on this. If someone else is interested in picking it up, feel free to do so.

PR #4331, opened 2024-04-04
Updated exception handling to catch const reference in out_of_range

LeilaShcheglova:
Updated exception handling to catch const reference in out_of_range


nlohmann (approved):
Looks good to me.

PR #4335, opened 2024-04-08
Fixed an error in the `Custom data source` example.

philip-paul-mueller:
The `target` member of the `MyIterator` class was used as if it was a reference, but it is a pointer.
Furthermore, the increment operator used `advance()` as a static function, which it is not.


nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #4362, opened 2024-04-29
Add lgtm explanation

### LINKED ISSUE #4361: CodeQL suppressions lack justification
### Description

In this repo there are a handful of old-style "lgtm" suppressions for issues raised by CodeQL. As we use it at Microsoft in internal tooling for Time Travel Debugging, this code goes through internal security scanning, and this issue was raised up with these suppressions.

The policy that was violated requires having a justification on all such suppressions. This justification needs to use the new format `\\ CodeQL[id] Justification goes here` which needs to be in its own line just above the line that had the issue. And the justification needs to be at least 25 characters long.

The issue we're having is specifically with lexer.hpp:

```C++
    token_type scan_number()  // lgtm [cpp/use-of-goto]
```

A possible solution (untested as I write this - I can't run the policy scan locally):

```C++
    // CodeQL [cpp/use-of-goto] `goto` is used here safely to implement a simple lexer's winding control flow
    token_type scan_number()
```

That might not be the best justification. Ideally this would be provided by the person who wrote the `scan_number` function.

Note that I failed to find any actual documentation for the format of CodeQL suppressions. I opened an issue there too: https://github.com/microsoft/codeql/issues/63

### Reproduction steps

This seems to be a new internal security policy over here, so I don't particularly have repro steps to give.

Expected: we can use this wonderful library without triggering security policies.
Actual: we trigger a security policy.

### Minimal code example

_No response_

### Error messages

_No response_

Microsoft Visual C++, Visual Studio 2022 17.10, C++20

### Library version

Used via vcpkg, with commit 0ca0fe433eb70cea0d5761079c0c5b47b736565b

### Validation

nlohmann:
Closes #4361 by adding a comment why `goto` should not be considered harmful in the number parser.

### DIFF
--- a/include/nlohmann/detail/input/lexer.hpp
+++ b/include/nlohmann/detail/input/lexer.hpp
@@ -967,7 +967,7 @@ class lexer : public lexer_base<BasicJsonType>
           locale's decimal point is used instead of `.` to work with the
           locale-dependent converters.
     */
-    token_type scan_number()  // lgtm [cpp/use-of-goto]
+    token_type scan_number()  // lgtm [cpp/use-of-goto] `goto` is used in this function to implement the number-parsing state machine described above. By design, any finite input will eventually reach the "done" state or return token_type::parse_error. In each intermediate state, 1 byte of the input is appended to the token_buffer vector, and only the already initialized variables token_buffer, number_type, and error_message are manipulated.
     {
         // reset token_buffer to store the number's bytes
         reset();
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -8339,7 +8339,7 @@ class lexer : public lexer_base<BasicJsonType>
           locale's decimal point is used instead of `.` to work with the
           locale-dependent converters.
     */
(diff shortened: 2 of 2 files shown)


TheJCAB:
Thanx!

Note that I don't know if a justification in the old style suppression format will be accepted. The internal guide to resolve the issue just mentions the "// CodeQL [query-id] Justification" format, which IIUC needs to be in a comment on its own line (not appended to the code). From my searches in the codeql public documentation, the only mention of suppressions (old or new) is in the release notes of CodeQL 2.12.0 announcing support for the new format, and from looking at documentation for the CodeQL scanning offered here on GitHub, they seem to be using 2.17.0, so that should work fine. I'd encourage you to try it.

TheJCAB:
I opened support requests on both CodeQL here on github and on the internal team managing CodeQL policies, but I've heard nothing from either.

I say go ahead and complete this as-is. It's an improvement and I expect it will suffice. If changing the suppression style becomes needed, it'll be a problem for another day.

Thank you!

PR #4372, opened 2024-05-10
Make iterator_proxy_value a forward_iterator (#4371)

### LINKED ISSUE #4371: iteration_proxy has limited usefulness in C++20 range views
### Description

This is a sort-of continuation, sort-of reopening of #3130 - it's not _exactly_ the same use case, but certain applications of `items()` on a json object still don't play nicely with C++20 `std::ranges`.

In particular, while `items()` can be used as an input for views like `std::views::transform`, that result isn't usable in situations that require a `forward_iterator`, such as constructing a container from the output range's iterators.  See the code below for what I mean.

The compilation error suggests the view iterator's `iterator_category` is not defined, which [is the case](https://en.cppreference.com/w/cpp/ranges/transform_view/iterator) if the base range doesn't model `forward_range`.

I did some debugging using `static_assert`s, and found that while `iteration_proxy_value` _does_ meet [all the requirements](https://en.cppreference.com/w/cpp/iterator/forward_iterator) for a `forward_iterator`, its `iterator_category` is explicitly exposed as only `std::input_iterator_tag`.  Simply changing that to `std::forward_iterator_tag` (or removing it and letting `iterator_traits` deduce it) in my testing makes it fully model `forward_iterator`, so that `items()` models `forward_range` and the view is usable in this case.

### Reproduction steps

- Crate a `json` object
- Run its `items()` through an `std::views::transform`
- Try to construct a container via that result's iterators, e.g. `std::vector<std::string> s{view.begin(), view.end()};`

`iteration_proxy` is just a wrapper around the actual iterator of `json` itself, which _does_ satisfy `forward_iterator`, so I expect `iteration_proxy_value` to also satisfy `forward_iterator`.  Instead, it only satisfies `input_iterator`.

### Minimal code example

```cpp
#include <nlohmann/json.hpp>

#include <ranges>
#include <string>
#include <vector>

int main()
{
    // This works
    nlohmann::json arr { 1, 2, 3 };
    auto arrTransform = std::views::transform([](auto&& element){ return element.template get<int>() * 2; });
    auto arrView = arr | arrTransform;
    std::vector<int> arrVec{arrView.begin(), arrView.end()};

    // This doesn't work
    nlohmann::json obj {
        { "one", 1 },
        { "two", 2 },
        { "three", 3 }
    };
    auto objItems = obj.items();
    auto objTransform = std::views::transform([](auto&& element){ return element.key(); });
    auto objView = objItems | objTransform;
    std::vector<std::string> keys{objView.begin(), objView.end()};  // Fails to compile
}
```

### Error messages

```
json-range.cpp: In function ‘int main()’:
json-range.cpp:24:65: error: no matching function for call to ‘std::vector<std::__cxx11::basic_string<char>, std::allocator<std::__cxx11::basic_string<char> > >::vector(<brace-enclosed initializer list>)’
   23 |     std::vector<std::string> keys{objView.begin(), objView.end()};  // Fails to compile
      |                                                                 ^
In file included from /usr/lib/gcc/x86_64-pc-linux-gnu/13/include/g++-v13/vector:66,
                 from /usr/lib/gcc/x86_64-pc-linux-gnu/13/include/g++-v13/functional:64,
                 from /usr/include/nlohmann/json.hpp:23,
                 from json-range.cpp:1:
/usr/lib/gcc/x86_64-pc-linux-gnu/13/include/g++-v13/bits/stl_vector.h:707:9: note: candidate: ‘template<class _InputIterator, class> constexpr std::vector<_Tp, _Alloc>::vector(_InputIterator, _InputIterator, const allocator_type&) [with <template-parameter-2-2> = _InputIterator; _Tp = std::__cxx11::basic_string<char>; _Alloc = std::allocator<std::__cxx11::basic_string<char> >]’
  707 |         vector(_InputIterator __first, _InputIterator __last,
      |         ^~~~~~
/usr/lib/gcc/x86_64-pc-linux-gnu/13/include/g++-v13/bits/stl_vector.h:707:9: note:   template argument deduction/substitution failed:
In file included from /usr/lib/gcc/x86_64-pc-linux-gnu/13/include/g++-v13/bits/stl_algobase.h:65,
                 from /usr/lib/gcc/x86_64-pc-linux-gnu/13/include/g++-v13/algorithm:60,
                 from /usr/include/nlohmann/json.hpp:21:
/usr/lib/gcc/x86_64-pc-linux-gnu/13/include/g++-v13/bits/stl_iterator_base_types.h: In substitution of ‘template<class _InIter> using std::_RequireInputIter = std::__enable_if_t<std::is_convertible<typename std::iterator_traits< <template-parameter-1-1> >::iterator_category, std::input_iterator_tag>::value> [with _InIter = std::ranges::transform_view<std::ranges::ref_view<nlohmann::json_abi_v3_11_3::detail::iteration_proxy<nlohmann::json_abi_v3_11_3::detail::iter_impl<nlohmann::json_abi_v3_11_3::basic_json<> > > >, main()::<lambda(auto:24&&)> >::_Iterator<false>]’:
/usr/lib/gcc/x86_64-pc-linux-gnu/13/include/g++-v13/bits/stl_vector.h:705:9:   required from here
/usr/lib/gcc/x86_64-pc-linux-gnu/13/include/g++-v13/bits/stl_iterator_base_types.h:250:11: error: no type named ‘iterator_category’ in ‘struct std::iterator_traits<std::ranges::transform_view<std::ranges::ref_view<nlohmann::json_abi_v3_11_3::detail::iteration_proxy<nlohmann::json_abi_v3_11_3::detail::iter_impl<nlohmann::json_abi_v3_11_3::basic_json<> > > >, main()::<lambda(auto:24&&)> >::_Iterator<false> >’
  250 |     using _RequireInputIter =
      |           ^~~~~~~~~~~~~~~~~
```

Gentoo, GCC 13.2

### Library version

3.11.3

### Validation

captaincrutches:
See #4371 for all the details - but long story short, there seems to be no reason not to expose `iterator_proxy_value` as a `forward_iterator` instead of just an `input_iterator`.  This allows more use cases when `items()` is fed into C++20 `std::views`.

If there _is_ a reason not to do so, please let me know - but the `iterator_category` on this type currently seems to [long predate](https://github.com/nlohmann/json/commit/8d8f8907715798dcca106509e4d9056a2b9f784e) the [more recent change](https://github.com/nlohmann/json/pull/3446) to fix it for C++20 ranges initially.  As well, the actual json iterator which the proxy wraps is also a forward iterator, so this keeps them on par with each other in terms of capabilities.

Another option is to remove this `iterator_category` typedef entirely, and let [`iterator_traits`](https://en.cppreference.com/w/cpp/iterator/iterator_traits) deduce it - since the type otherwise satisfies `ForwardIterator`, it should (and does, in my motivating case) do the right thing.  But since the other required typedefs are here, might as well keep it but change it IMO.

I'm not sure what, if any, tests can/should be updated/added in this case - I'm open to suggestions, though it seems like all the relevant `static_assert`s and such should still apply.

* * *

## Pull request checklist

- [X]  Changes are described in the pull request, or an [existing issue is referenced](https://github.com/nlohmann/json/issues). (#4371)
- [X]  The source code is amalgamated; that is, after making changes to the sources in the `include/nlohmann` directory, run `make amalgamate` to create the single-header files `single_include/nlohmann/json.hpp` and `single_include/nlohmann/json_fwd.hpp`. The whole process is described [here](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md#files-to-change).

Note on amalgamation:  I did run `make amalgamate`, and that touched a bunch of lines unrelated to my change.  The actual change I've made is to `iterator_proxy.hpp`, and hopefully hiding whitespace in the diff will narrow that down.  If you'd rather I revert the other amalgamate changes to keep this diff confined to the 2 (well, 4, including generated code) lines I changed, I'm more than happy to do so - but I figured I'd err on the side of caution w.r.t. any checks that run in CI.

### DIFF
--- a/include/nlohmann/detail/iterators/iteration_proxy.hpp
+++ b/include/nlohmann/detail/iterators/iteration_proxy.hpp
@@ -9,7 +9,7 @@
 #pragma once
 
 #include <cstddef> // size_t
-#include <iterator> // input_iterator_tag
+#include <iterator> // forward_iterator_tag
 #include <string> // string, to_string
 #include <tuple> // tuple_size, get, tuple_element
 #include <utility> // move
@@ -40,7 +40,7 @@ template<typename IteratorType> class iteration_proxy_value
     using value_type = iteration_proxy_value;
     using pointer = value_type *;
     using reference = value_type &;
-    using iterator_category = std::input_iterator_tag;
+    using iterator_category = std::forward_iterator_tag;
     using string_type = typename std::remove_cv< typename std::remove_reference<decltype( std::declval<IteratorType>().key() ) >::type >::type;
 
   private:
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -5181,7 +5181,7 @@ NLOHMANN_JSON_NAMESPACE_END
 
 
 #include <cstddef> // size_t
-#include <iterator> // input_iterator_tag
+#include <iterator> // forward_iterator_tag
 #include <string> // string, to_string
 #include <tuple> // tuple_size, get, tuple_element
 #include <utility> // move
@@ -5215,7 +5215,7 @@ template<typename IteratorType> class iteration_proxy_value
     using value_type = iteration_proxy_value;
     using pointer = value_type *;
     using reference = value_type &;
-    using iterator_category = std::input_iterator_tag;
+    using iterator_category = std::forward_iterator_tag;
     using string_type = typename std::remove_cv< typename std::remove_reference<decltype( std::declval<IteratorType>().key() ) >::type >::type;
 
   private:
--- a/tests/src/unit-readme.cpp
+++ b/tests/src/unit-readme.cpp
@@ -171,7 +171,7 @@ TEST_CASE("README" * doctest::skip())
 
             // find an entry
             CHECK(o.find("foo") != o.end());
-            if (o.find("foo") != o.end())
+            if (o.find("foo") != o.end()) // NOLINT(readability-container-contains)
             {
                 // there is an entry with key "foo"
             }


captaincrutches:
Ahh - I see from the build logs that these jobs use astyle 3.1; when I ran `make amalgamate`, I installed astyle 3.4.

Re-running with the matching version of astyle produces much saner results :) Though it still does remove a bunch of blank lines in the single-include headers.

captaincrutches:
Okay, I reset the single-include files entirely and tried again, and now we've got a diff that actually looks correct.  Sorry for the noise!  I'm happy to rebase and clean up these extra bad commits if that would be preferable.

captaincrutches:
[Clang-tidy failures](https://productionresultssa4.blob.core.windows.net/actions-results/a0cc31dc-52bf-4c5b-8763-4828dc1dab7c/workflow-job-run-d89f80ed-9ad9-5f4a-dcc0-4920bef73f10/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-05-11T17%3A10%3A46Z&sig=lQ%2FnsQzpYFXKUNNgGdm4fbgnSqC0bjnuJiREAC3LSu0%3D&sp=r&spr=https&sr=b&st=2024-05-11T17%3A00%3A41Z&sv=2021-12-02) look unrelated to my change - looks like https://github.com/nlohmann/json/pull/4311 already exists trying to fix them.

I'm not sure what to make of the [xcode 12.4 build](https://productionresultssa11.blob.core.windows.net/actions-results/0cc6a91e-282e-4241-8f13-d2838bc47ea7/workflow-job-run-cfdecfa7-91cc-5483-c08a-6b39a1aeb8d9/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-05-11T17%3A01%3A57Z&sig=SHFHA%2Brz%2BCR7%2FSR36O%2BdU45k1QvUfBDMmFqg7AkYRe8%3D&sp=r&spr=https&sr=b&st=2024-05-11T16%3A51%3A52Z&sv=2021-12-02) - seems like something stalled out and hit a timeout, but I can't tell what... the rest of the build looks like it was going fine up until it stopped in its tracks.

nlohmann:
I'll check the Clang-Tidy warnings.

captaincrutches:
Updated now that CI is fixed, to pick up those changes.  Also squashed down all the amalgamate weirdness from before.

gregmarr (approved):
LGTM

nlohmann (approved):
Looks good to me.

captaincrutches:
Clang-tidy failure looks like a `readability-container-contains` in [`unit-readme.cpp`](https://github.com/nlohmann/json/blob/develop/tests/src/unit-readme.cpp#L174), which may indeed have been introduced by this change if making this a forward iterator causes the json type to more closely model [`Container`](https://en.cppreference.com/w/cpp/named_req/Container) and therefore [`AssociativeContainer`](https://en.cppreference.com/w/cpp/named_req/AssociativeContainer).

Looks like there's already a case of this warning that's suppressed [by the recent CI-fixing change](https://github.com/nlohmann/json/pull/4489/files#diff-baa7997e8b6df7d098cfcc978ab2280be2c2529bbc404e2b263a188b214f3de4), I assume because `contains` is only in C++20 for std containers that have it.

On the other hand, this one looks like it's operating on a json object, which does have `contains`, and the line being pointed to is many years old, so maybe updating it to use `contains` is the right thing to do.

I'll defer to you guys here, since it seems a bit of a hot topic: should I fix the warning, or suppress it for now?

nlohmann:
Please add a NOLINT comment.

nlohmann:
FYI: The same issue occurred on the CI of #4036, so it was not introduced by this PR. I have no idea what changed since merging #4489. Nonetheless, if you suppress the line, we have a green CI and can merge this.

nlohmann:
Thanks a lot!

PR #4380, opened 2024-05-24
Add support for BSON element 11: uint64

### LINKED ISSUE #3894: json::to_bson does not handle unsigned 64 bit integers
### Description

Trying to convert a nlohmann::json object which contains an unsigned int of 64 bit length breaks BSON conversion.
I do consider this to be a bug, as BSON should be capable of handling those values [[1](https://bsonspec.org/spec.html)] , but the corresponding header-bits are not set in [write_bson_unsigned](https://github.com/nlohmann/json/blob/develop/single_include/nlohmann/json.hpp#L16028). Only entry headers `0x10` (for `int32`) and `0x12` (for `int64`) are written, but `11` (for `uint_64`) is not.

### Reproduction steps

This bug is triggered whenever one uses the `json::to_bson` function on any object that contains an uint64_t, that really uses all 64 bits.

I expect BSON to generated, but I get an error message (see below)

### Minimal code example

```Shell
#include <map>
#include <string>
#include <nlohmann/json.hpp>

int main(){
  std::map<size_t,std::string> m={{14631357879723465589u /*really just any large enough uint64_number*/,"To_Big"}};
  nlohmann::json j;
  j["map"]=m;
  auto bson=nlohmann::json::to_bson(j); //can not be converted
}
```

### Error messages

```Shell
terminate called after throwing an instance of 'nlohmann::detail::out_of_range'
  what():  [json.exception.out_of_range.407] integer number 14631357879723465589 cannot be represented by BSON as it does not fit int64
```

Clang-13 on Mint 21 64bit, Clang Trunk (via godbolt)

### Library version

3.11.2 (develop branch via godbolt) , 3.10.5

### Validation

slowriot:
This PR adds support for decoding and encoding BSON element 11: 64bit unsigned int.

This resolves https://github.com/nlohmann/json/issues/3894.

## Changes
- Decode element `0x11` correctly as `uint64_t`.
- Update existing `write_bson_unsigned` to encode unsigned 64bit values correctly (using `0x11` instead of signed `0x12`).
- Update existing tests to expect support for `uint64_t`.

## Reference
- https://bsonspec.org/spec.html

## Pull request checklist


nlohmann (changes requested):
Thanks for the patience. The code looks good, but the documentation needs adjustment:

- https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/home/exceptions.md: remove BSON from the description of `json.exception.out_of_range.407`
- https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/features/binary_formats/bson.md: adjust mapping tables

slowriot:
> Thanks for the patience. The code looks good, but the documentation needs adjustment:
> 

Thanks for the review.  Documentation updated in latest.  Please let me know if I've missed anything else.

nlohmann:
Coverage went down. Can you please check if a test is missing?

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1080,14 +1080,14 @@ class binary_writer
...
         {
-            write_bson_entry_header(name, 0x12 /* int64 */);
-            write_number<std::int64_t>(static_cast<std::int64_t>(j.m_data.m_value.number_unsigned), true);
+            write_bson_entry_header(name, 0x11 /* uint64 */);
+            write_number<std::uint64_t>(static_cast<std::uint64_t>(j.m_data.m_value.number_unsigned), true);
         }
         else
         {
-            JSON_THROW(out_of_range::create(407, concat("integer number ", std::to_string(j.m_data.m_value.number_unsigned), " cannot be represented by BSON as it does not fit int64"), &j));
+            JSON_THROW(out_of_range::create(407, concat("unsigned integer number ", std::to_string(j.m_data.m_value.number_unsigned), " cannot be represented by BSON as it does not fit into uint64"), &j));
```
nlohmann: If I read the coverage information correctly, then this line is not covered in a test. Please check and add a test.

nlohmann (changes requested):

nlohmann:
(And also, please update to the latest develop branch so that the CI can be green)

nlohmann:
Btw: have you looked at #4356?

nlohmann:
We recently merged #4590. I think this PR can be closed.

slowriot:
> We recently merged #4590. I think this PR can be closed.

Thanks, just catching up with the changes since I last pushed to this PR - took a moment to resolve the conflicts as #4590 duplicated most of what I did, but it seems there may still have been a few changes in this PR that were missed in the other (unit test for example, and slightly different approach to choosing a format when encoding, documentation grammar, case ordering and a few other minor details).

Please review and let me know if any of it is still of use, and if it can be merged on top or if anything needs to be corrected or removed from this change.  Otherwise of course it can be closed.

nlohmann:
Sure, go ahead.

slowriot:
@nlohmann Having rebased, I think the [remaining changes](https://github.com/nlohmann/json/pull/4380/files) in this PR's diffs may still be useful to merge - please have a look.  Otherwise, please close.

include/nlohmann/detail/output/binary_writer.hpp
gregmarr: I think this can only happen if `number_unsigned` has a larger size than 64 bits.  Does the library support that?  If not, then this can't be hit.

nlohmann:
@codenut Can you please have a look at the remaining diff?

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1076,7 +1076,7 @@ class binary_writer
     {
         return (value <= static_cast<std::uint64_t>((std::numeric_limits<std::int32_t>::max)()))
                ? sizeof(std::int32_t)
-               : sizeof(std::int64_t);
+               : sizeof(std::uint64_t);
```
nlohmann, line 1079: This change seems reasonable.

include/nlohmann/detail/output/binary_writer.hpp
nlohmann: So far, I failed to compile the library with 128-bit integers. For CBOR, we assume all unsigned integers to fit into 64 bits, so I think it's fair to do the same here. (Any code like the one above could not be tested anyway.)

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1090,15 +1090,14 @@ class binary_writer
             write_bson_entry_header(name, 0x10 /* int32 */);
             write_number<std::int32_t>(static_cast<std::int32_t>(j.m_data.m_value.number_unsigned), true);
         }
-        else if (j.m_data.m_value.number_unsigned <= static_cast<std::uint64_t>((std::numeric_limits<std::int64_t>::max)()))
+        else if (j.m_data.m_value.number_unsigned <= std::numeric_limits<std::uint64_t>::max())
         {
-            write_bson_entry_header(name, 0x12 /* int64 */);
-            write_number<std::int64_t>(static_cast<std::int64_t>(j.m_data.m_value.number_unsigned), true);
+            write_bson_entry_header(name, 0x11 /* uint64 */);
```
nlohmann, line 1095: This change seems reasonable.

tests/src/unit-bson.cpp
```diff
@@ -1132,7 +1132,7 @@ TEST_CASE("BSON numerical data")
                     std::vector<std::uint8_t> const expected_bson =
                     {
                         0x14u, 0x00u, 0x00u, 0x00u, // size (little endian)
-                        0x12u, /// entry: int64
+                        0x11u, /// entry: uint64
```
nlohmann, line 1135: This is a consequence of the change in `write_bson_unsigned` above.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1090,15 +1090,14 @@ class binary_writer
             write_bson_entry_header(name, 0x10 /* int32 */);
             write_number<std::int32_t>(static_cast<std::int32_t>(j.m_data.m_value.number_unsigned), true);
         }
-        else if (j.m_data.m_value.number_unsigned <= static_cast<std::uint64_t>((std::numeric_limits<std::int64_t>::max)()))
+        else if (j.m_data.m_value.number_unsigned <= std::numeric_limits<std::uint64_t>::max())
```
nlohmann: The call to `max` needs to be put in parentheses to work with MSVC. This is the reason the Windows jobs are currently failing.

nlohmann (changes requested):
The remaining changes seem reasonable, but I would also like to get @codenut 's take on this.

- Please revert the whitespace changes in the documentation. I think there is only one grammar fix left then.
- There is an uncovered branch that can be removed as it would only be triggered with 128-bit integers.
- A call to `max` needs to be put in parentheses to fix the Windows builds.

slowriot:
> The remaining changes seem reasonable, but I would also like to get @codenut 's take on this.
> 
>     * Please revert the whitespace changes in the documentation. I think there is only one grammar fix left then.
> 
>     * There is an uncovered branch that can be removed as it would only be triggered with 128-bit integers.
> 
>     * A call to `max` needs to be put in parentheses to fix the Windows builds.

Thanks for reviewing.  I've merged in the latest develop branch which already includes equivalent changes for a lot of the original PR, and tidied up the remaining diffs.  Please let me know if there's anything outstanding to address.

nlohmann:
Please check  https://github.com/nlohmann/json/pull/4380/checks?check_run_id=35473431508on how to fix the DCO warning.

include/nlohmann/detail/output/binary_writer.hpp
codenut: This change is logical, but I have some concerns about backward compatibility. It may break the following scenario:

```
void write()
{
    const uint64_t = 9223372036854775807L;
    json const j = {
            {"entry", l}
    };
    const std::vector<uint8_t> bson = json::to_bson(j);
    saveToTable1(bson);
    saveToTable2(bson);
}

write(); // was called before the code changes in this PR
```

```
void read()
{
    const std::vector<uint8_t> bson1 = loadFromTable1();
    json const j = json::from_bson(bson);
    const std::vector<uint8_t> bson1_roundtrip = json::to_bson(j);

    const std::vector<uint8_t> bson2 = loadFromTable2();

    if (equals(bson1_roundtrip, bson2)) { 
        ...
    }
}

read(); 
```

With the changes in this PR, the comparison between `bson1_roundtrip` and `bson2` will fail. It is hard to tell if clients rely on this behavior, but I would like to highlight this potential issue.

nlohmann:
(The coverage job is broken also in other branches. I am looking into this)

include/nlohmann/detail/output/binary_writer.hpp
nlohmann: For compatibility, we could serialize unsigned integers until int64_max with 0x12 and all numbers larger with 0x11.

nlohmann:

nlohmann (changes requested):

nlohmann:
Please update to the latest `develop` branch and check https://github.com/nlohmann/json/pull/4380/checks?check_run_id=35473431508 on how to fix the DCO warning.

nlohmann:
#4590 was merged.

PR #4396, opened 2024-06-11
Remove alwayslink=True Bazel flag

mering:
Remove the `alwayslink=True` flag from `BUILD.bazel`.

This should have no effect for header only libraries as mentioned in https://github.com/googleapis/google-cloud-cpp/pull/14310#issuecomment-2154986153.


nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #4439, opened 2024-08-23
Make serializer dump() non-recursive

vii:
This allows printing very deep trees without running out of stack.

### DIFF
--- a/include/nlohmann/detail/output/serializer.hpp
+++ b/include/nlohmann/detail/output/serializer.hpp
@@ -68,8 +68,8 @@ class serializer
                error_handler_t error_handler_ = error_handler_t::strict)
         : o(std::move(s))
         , loc(std::localeconv())
-        , thousands_sep(loc->thousands_sep == nullptr ? '\0' : std::char_traits<char>::to_char_type(* (loc->thousands_sep)))
-        , decimal_point(loc->decimal_point == nullptr ? '\0' : std::char_traits<char>::to_char_type(* (loc->decimal_point)))
+        , thousands_sep(loc->thousands_sep == nullptr ? '\0' : std::char_traits<char>::to_char_type( * (loc->thousands_sep)))
+        , decimal_point(loc->decimal_point == nullptr ? '\0' : std::char_traits<char>::to_char_type( * (loc->decimal_point)))
         , indent_char(ichar)
         , indent_string(512, indent_char)
         , error_handler(error_handler_)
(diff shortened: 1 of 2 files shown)


vii:
Unfortunately has some compile errors on MSVC etc. Can update and fix if you like this direction - 

> error C7555: use of designated initializers requires at least '/std:c++20'

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -104,272 +104,320 @@ class serializer
...
+            const unsigned int offset;
+            const decltype(start_val.m_data.m_value.object->cbegin()) iterator;
+        };
+        std::vector<dump_state> stack;
+        stack.push_back({start_val, start_indent, 0});
 
-                if (pretty_print)
+        while (!stack.empty())
+        {
+            const dump_state& state = stack.back();
```
gregmarr, line 131: By taking a reference here, you end up with a dangling reference when you do `pop_back()`.  Even if you don't use `state` again after the `pop_back()`, you are referring to the `BasicJsonType` inside it with `val`.  It's probably better to store a `BasicJsonType *` in the state, so you don't have to worry about referenceds to references, and then you can do this:

    const BasicJsonType* val = stack.back().val;
    const unsigned int current_indent = stack.back().new_indent;
    const unsigned int offset = stack.back().offset;
    auto iterator = stack.back().iterator;
    stack.pop_back();

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -104,272 +104,320 @@ class serializer
...
-                {
-                    o->write_characters("{}", 2);
-                    return;
-                }
+            const BasicJsonType& val;
+            const unsigned int new_indent;
+            const unsigned int offset;
+            const decltype(start_val.m_data.m_value.object->cbegin()) iterator;
+        };
+        std::vector<dump_state> stack;
```
gregmarr, line 126: A `std::stack` would probably be a better choice.

gregmarr (commented):
I'm not saying whether this is something that should be pursued or not, but in case it is, a couple initial observations.

t-b:
I very much like this fix here! It was on my todo list as well.

nlohmann:
Thank you very much for this PR, @vii, and sorry it never got a proper review. Making `dump()` non-recursive was the right call. The same goal has since been reached in #5285, which rewrote the serializer so it no longer overflows the stack on deeply nested values. It is already on `develop`, so I'm closing this PR as superseded. Thanks again for the work and for flagging the problem early!

PR #4456, opened 2024-09-11
CMake: generate a pkg-config file that follow pkg-config conventions

dcbaker:
The pkg-config file generated by Meson matches the convention of using a `prefix` variable, and an `includedir` variable that is relative to `prefix` (`includedir=${prefix}/include`). The one generated by CMake does not do this, but hard codes the include directory straight into the `Cflags` field. There are advantages to having the `prefix` (thus the convention)`, especially when cross compiling.

This PR makes the CMake generated pkg-config file meat convention, and match the one generated by Meson.

## Pull request checklist

- [x]  Code coverage](https://coveralls.io/github/nlohmann/json) is 100%. Test cases can be added by editing the [test suite](https://github.com/nlohmann/json/tree/develop/test/src).


nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

Issue #4271, opened 2024-01-19
to_json(std::filesystem::path) can create invalid UTF-8 chars on windows


MHebes:
### Description

This conversion function:

https://github.com/nlohmann/json/blob/7efe875495a3ed7d805ddbb01af0c7725f50c88b/include/nlohmann/detail/conversions/to_json.hpp#L416C1-L420C2

```c++
template<typename BasicJsonType>
inline void to_json(BasicJsonType& j, const std_fs::path& p)
{
    j = p.string();
}
```

uses `p.string()`, which does not give a UTF-8-encoded string on windows (in some cases, maybe?). Trying to `dump()` the resultant JSON throws a "invalid UTF-8 byte" exception.

### Reproduction steps

Convert a `std::filesystem::path`, which contains a unicode "Right Single Quotation Mark" character (U+2019), to a `json` implicitly or with `to_json`.

Inspect the new `json (string_t)`'s bytes, either by `dump()`ing, or converting to BSON.

### Expected vs. actual results

Expected: "Strings are stored in UTF-8 encoding." per https://json.nlohmann.me/api/basic_json/string_t/

Actual: The string gets converted by `std::filesystem::path::string()`, which appears to convert it to [Windows-1252 encoding](https://en.wikipedia.org/wiki/Windows-1252). Its bytes end up as `\x92` rather than `\xe2\x80\x99`.

### Minimal code example

```c++
#include <filesystem>
#include <iostream>
#include <nlohmann/json.hpp>

int main() {
  try {
    wchar_t wide_unicode_right_quote[2] = {0x2019, 0};  // came from a directory_iterator in reality
    nlohmann::json apost = std::filesystem::path(wide_unicode_right_quote);
    std::cout << apost << std::endl;
    return 0;
  } catch (const std::exception& e) {
    std::cerr << e.what() << std::endl;
    return 1;
  }
}
```

Workaround I'm using is to use `WideCharToMultiByte` + `.native()` to get the string in UTF-8 before passing to nlohmann:

```c++
inline std::string Narrow(std::wstring_view wstr) {
  if (wstr.empty()) return {};
  int len = ::WideCharToMultiByte(CP_UTF8, 0, &wstr[0], wstr.size(), nullptr, 0, nullptr, nullptr);
  std::string out(len, 0);
  ::WideCharToMultiByte(CP_UTF8, 0, &wstr[0], wstr.size(), &out[0], len, nullptr, nullptr);
  return out;
}

int main() {
  try {
[...]

PR #4502, opened 2024-11-19
#4271: Use OS independent utf8 string representation of std_fs::path at to_j…

zel1b08a:
#

According to [#4271](https://github.com/nlohmann/json/issues/4271).

`std::filesystem::path::string` method uses system-dependent native format. `String` type is always `UTF8` format, by the json docs. So we force encoding to be `UTF8`.

### DIFF
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -435,7 +435,7 @@ inline void to_json(BasicJsonType& j, const T& t)
 template<typename BasicJsonType>
 inline void to_json(BasicJsonType& j, const std_fs::path& p)
 {
-    j = p.string();
+    j = p.u8string();
 }
 #endif
 
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -440,7 +440,6 @@ inline void to_json(BasicJsonType& j, const std_fs::path& p)
 #else
     j = p.u8string();
 #endif
-
 }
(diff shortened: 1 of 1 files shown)


nlohmann:
Please run `make amalgamate` with Astyle (http://astyle.sourceforge.net) 3.1 installed.

include/nlohmann/detail/conversions/to_json.hpp
```diff
@@ -435,7 +435,7 @@ inline void to_json(BasicJsonType& j, const T& t)
 template<typename BasicJsonType>
 inline void to_json(BasicJsonType& j, const std_fs::path& p)
 {
-    j = p.string();
+    j = p.u8string();
```
nlohmann, line 441: Note that from C++20, `u8string` will return `std::u8string` instead of `std::string` - I am not sure if this assignment works in this case.

zel1b08a: I see the copy with `reinterpret_cast` to `char*` there only..)

nlohmann: I saw a remark on cppreference.com

zel1b08a: I mean the u8string to string conversion here in that issue

nlohmann:
The exception could come from the conversion of `std::u8string` to JSON. As `std::u8string` is not supported, the string is treated like a sequence of bytes and converted to an array.

risa2000:
I believe this has been resolved by #4631 and could be closed.

PR #4505, opened 2024-11-20
Make SAX output locale-independent

### LINKED ISSUE #4084: SAX interface unexpectedly gets locale-altered float representation.
### Description

I didn't expect (neither from experience, nor from the documentation) that the float string representation would be altered according to the locale when feeded into the SAX parser.
I understand, why it has to be altered internally, but I'd expect the original string in the callback.

### Reproduction steps

`LC_NUMERIC=de_DE.UTF-8`

`{ "aFloat": 42.0 }`

```c++
bool ParserImpl::number_float( number_float_t val, const string_t &s )
{
    std::cout << s << '\n';
}
```

Expected: 42.0
Actual: 42,0

### Minimal code example

```c++
#include <nlohmann/json.hpp>
#include <iostream>
#include <string_view>
#include <clocale>

using Json = nlohmann::json;

struct ParserImpl : public nlohmann::json_sax<Json>
{
    bool null() override { return true; };
    bool boolean( bool val ) override { return true; };
    bool number_integer( number_integer_t val ) override { return true; };
    bool number_unsigned( number_unsigned_t val ) override { return true; };
    bool number_float( number_float_t val, const string_t &s ) override {
        std::cout << s << '\n';
        return true; 
    };
    bool string( string_t &val ) override { return true; };
    bool binary( binary_t &val ) override { return true; };
    bool start_object( std::size_t elements ) override { return true; };
    bool key( string_t &val ) override { return true; };
    bool end_object() override { return true; };
    bool start_array( std::size_t elements ) override { return true; };
    bool end_array() override { return true; };
    bool parse_error( std::size_t position, const std::string &last_token,
                      const nlohmann::detail::exception &ex ) override { return true; };
};

int main() {
    if(!setlocale(LC_NUMERIC, "de_DE.utf8"))
        return -1;
    std::string_view data { R"json({ "aFloat": 42.0 })json" };
    ParserImpl sax {};
    Json::sax_parse( data, &sax );
    return 0;
}

```
`g++-12 -o jsontest -I /path/to/nlohmann_json-src/single_include/ --std=c++20 jsontest.cpp`

### Error messages

_No response_

Linux/gcc-12, Windows/MSVC19.35

### Library version

3.11.2

### Validation

nlohmann:
Closes #4084

### DIFF
--- a/tests/thirdparty/doctest/doctest.h
+++ b/tests/thirdparty/doctest/doctest.h
@@ -5493,7 +5493,7 @@ namespace {
                         .writeAttribute("testsuite", in.data[i]->m_test_suite)
                         .writeAttribute("filename", skipPathFromFilename(in.data[i]->m_file.c_str()))
                         .writeAttribute("line", line(in.data[i]->m_line))
-                        .writeAttribute("skipped", in.data[i]->m_skip);
+                        .writeAttribute("Dped", in.data[i]->m_skip);
                 }
                 xml.scopedElement("OverallResultsTestCases")
                         .writeAttribute("unskipped", in.run_stats->numTestCasesPassingFilters);
--- a/tests/thirdparty/doctest/doctest.h
+++ b/tests/thirdparty/doctest/doctest.h
@@ -5493,7 +5493,7 @@ namespace {
                         .writeAttribute("testsuite", in.data[i]->m_test_suite)
                         .writeAttribute("filename", skipPathFromFilename(in.data[i]->m_file.c_str()))
                         .writeAttribute("line", line(in.data[i]->m_line))
-                        .writeAttribute("Dped", in.data[i]->m_skip);
+                        .writeAttribute("skipped", in.data[i]->m_skip);
                 }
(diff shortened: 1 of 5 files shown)


nlohmann (commented):
Todo:

- [x] add a test suite for locale-dependent issues

jzakrzewski (approved):

tests/thirdparty/doctest/doctest.h
```diff
@@ -5493,7 +5493,7 @@ namespace {
                         .writeAttribute("testsuite", in.data[i]->m_test_suite)
                         .writeAttribute("filename", skipPathFromFilename(in.data[i]->m_file.c_str()))
                         .writeAttribute("line", line(in.data[i]->m_line))
-                        .writeAttribute("skipped", in.data[i]->m_skip);
+                        .writeAttribute("Dped", in.data[i]->m_skip);
```
gregmarr: @nlohmann This looks like an accidental change.

nlohmann: Good catch!

PR #4506, opened 2024-11-21
fix: integer parsed as float when EINTR set in errno

StuartGorman:
**Problem:** Occasionally the OS will set errno to `EINTR` if some system call has been interrupted. If that happens in `lexer:scan_number()` it will be treated the same as errno being set to `ERANGE` due to the check being `errno == 0`. This will cause valid integers to be parsed as floats.

**Fix:** Changing the errno check to explicitly check for ERANGE, as that is the condition that the parsing code cares about.

**Environment:** 
- gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
- cmake version 3.26.3
* * *

## Pull request checklist

### DIFF
--- a/include/nlohmann/detail/input/lexer.hpp
+++ b/include/nlohmann/detail/input/lexer.hpp
@@ -1256,7 +1256,7 @@ scan_number_done:
             // we checked the number format before
             JSON_ASSERT(endptr == token_buffer.data() + token_buffer.size());
 
-            if (errno == 0)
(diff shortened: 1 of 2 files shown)


nlohmann:
Please run the amalgamation with astyle 3.1.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #4513, opened 2024-11-25
BJData optimized binary array type

nebkat:
See https://github.com/NeuroJSON/bjdata/pull/6 for further information.

Introduces a dedicated `B` marker for bytes.

This is used as the strong type marker in optimized array format to encode binary data such that it can also be decoded back to binary data (instead of wrongly decoding as an integer array).

Draft while awaiting the release of BJData draft 3.

Would a `legacy_binary` flag be desirable to continue supporting draft 2 expectations of uint8 typed arrays for binary?

* * *

## Pull request checklist

### DIFF
--- a/docs/mkdocs/docs/api/basic_json/to_bjdata.md
+++ b/docs/mkdocs/docs/api/basic_json/to_bjdata.md
@@ -4,13 +4,16 @@
 // (1)
 static std::vector<std::uint8_t> to_bjdata(const basic_json& j,
                                            const bool use_size = false,
-                                           const bool use_type = false);
+                                           const bool use_type = false,
+                                           const bjdata_version_t version = bjdata_version_t::draft2);
 
 // (2)
 static void to_bjdata(const basic_json& j, detail::output_adapter<std::uint8_t> o,
-                      const bool use_size = false, const bool use_type = false);
+                      const bool use_size = false, const bool use_type = false,
+                      const bjdata_version_t version = bjdata_version_t::draft2);
 static void to_bjdata(const basic_json& j, detail::output_adapter<char> o,
-                      const bool use_size = false, const bool use_type = false);
+                      const bool use_size = false, const bool use_type = false,
+                      const bjdata_version_t version = bjdata_version_t::draft2);
 ```
 
 Serializes a given JSON value `j` to a byte vector using the BJData (Binary JData) serialization format. BJData aims to
@@ -34,6 +37,9 @@ The exact mapping and its limitations is described on a [dedicated page](../../f
 
 `use_type` (in)
 :   whether to add type annotations to container types (must be combined with `#!cpp use_size = true`); optional,
+
+`version` (in)
+:   which version of BJData to use (see [draft 3](../../features/binary_formats/bjdata.md#draft-3-binary-format)); optional,
 `#!cpp false` by default.
 
 ## Return value
@@ -68,3 +74,4 @@ Linear in the size of the JSON value `j`.
 ## Version history
 
 - Added in version 3.11.0.
+- BJData version parameter (for draft3 binary encoding) added in version 3.12.0.
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -28,6 +28,13 @@ NLOHMANN_JSON_NAMESPACE_BEGIN
 namespace detail
 {
 
+/// how to encode BJData
+enum class bjdata_version_t
+{
+    draft2,
+    draft3,
+};
+
 ///////////////////
 // binary writer //
 ///////////////////
@@ -735,11 +742,14 @@ class binary_writer
     @param[in] use_type    whether to use '$' prefixes (optimized format)
     @param[in] add_prefix  whether prefixes need to be used for this value
     @param[in] use_bjdata  whether write in BJData format, default is false
+    @param[in] bjdata_version  which BJData version to use, default is draft2
     */
     void write_ubjson(const BasicJsonType& j, const bool use_count,
                       const bool use_type, const bool add_prefix = true,
-                      const bool use_bjdata = false)
+                      const bool use_bjdata = false, const bjdata_version_t bjdata_version = bjdata_version_t::draft2)
     {
+        const bool bjdata_draft3 = bjdata_version == bjdata_version_t::draft3;
+
         switch (j.type())
         {
             case value_t::null:
@@ -829,7 +839,7 @@ class binary_writer
 
                 for (const auto& el : *j.m_data.m_value.array)
                 {
-                    write_ubjson(el, use_count, use_type, prefix_required, use_bjdata);
+                    write_ubjson(el, use_count, use_type, prefix_required, use_bjdata, bjdata_version);
                 }
 
                 if (!use_count)
@@ -847,11 +857,11 @@ class binary_writer
                     oa->write_character(to_char_type('['));
                 }
 
-                if (use_type && !j.m_data.m_value.binary->empty())
+                if (use_type && ((use_bjdata && bjdata_draft3) || !j.m_data.m_value.binary->empty()))
                 {
                     JSON_ASSERT(use_count);
                     oa->write_character(to_char_type('$'));
-                    oa->write_character('U');
+                    oa->write_character(use_bjdata && bjdata_draft3 ? 'B' : 'U');
                 }
 
                 if (use_count)
@@ -870,7 +880,7 @@ class binary_writer
                 {
                     for (size_t i = 0; i < j.m_data.m_value.binary->size(); ++i)
                     {
-                        oa->write_character(to_char_type('U'));
+                        oa->write_character(to_char_type((use_bjdata && bjdata_draft3) ? 'B' : 'U'));
                         oa->write_character(j.m_data.m_value.binary->data()[i]);
                     }
                 }
@@ -887,7 +897,7 @@ class binary_writer
             {
                 if (use_bjdata && j.m_data.m_value.object->size() == 3 && j.m_data.m_value.object->find("_ArrayType_") != j.m_data.m_value.object->end() && j.m_data.m_value.object->find("_ArraySize_") != j.m_data.m_value.object->end() && j.m_data.m_value.object->find("_ArrayData_") != j.m_data.m_value.object->end())
                 {
-                    if (!write_bjdata_ndarray(*j.m_data.m_value.object, use_count, use_type))  // decode bjdata ndarray in the JData format (https://github.com/NeuroJSON/jdata)
+                    if (!write_bjdata_ndarray(*j.m_data.m_value.object, use_count, use_type, bjdata_version))  // decode bjdata ndarray in the JData format (https://github.com/NeuroJSON/jdata)
                     {
                         break;
                     }
@@ -931,7 +941,7 @@ class binary_writer
                     oa->write_characters(
                         reinterpret_cast<const CharType*>(el.first.c_str()),
                         el.first.size());
-                    write_ubjson(el.second, use_count, use_type, prefix_required, use_bjdata);
+                    write_ubjson(el.second, use_count, use_type, prefix_required, use_bjdata, bjdata_version);
                 }
 
                 if (!use_count)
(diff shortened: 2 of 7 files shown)


nlohmann:
Please run `make amalgamate` with AStyle 3.1.

tests/src/unit-bjdata.cpp
```diff
@@ -1514,11 +1542,8 @@ TEST_CASE("BJData")
                     // create expected byte vector
                     std::vector<std::uint8_t> expected;
                     expected.push_back(static_cast<std::uint8_t>('['));
-                    if (N != 0)
```
nlohmann, line 1517: Seeing existing test cases change - is this a breaking change for existing clients?

nebkat: Yes - this is a breaking change for encoding, but fixes roundtrip encoding/decoding.

Before:
- `json::binary` gets encoded as a typed array of `U`
- typed array of `U` gets decoded as `json::array` ❗does not match encoded data

After:
- `json::binary` gets encoded as a typed array of `B`
- typed array of `B` gets decoded as `json::binary` ✅ matches encoded data

I could introduce a new flag to continue encoding binary as `uint8` if the breaking change is undesirable?

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -847,11 +847,11 @@ class binary_writer
                     oa->write_character(to_char_type('['));
                 }
 
-                if (use_type && !j.m_data.m_value.binary->empty())
+                if (use_type && (use_bjdata || !j.m_data.m_value.binary->empty()))
```
nebkat: @nlohmann 
And that specific test-case you commented on has changed due to this line. In order to ensure that encoding an empty `json::binary` still decodes to `json::binary` I force the insertion of the type here regardless of whether there are contents or not.

That said this problem will still exist if other libraries are used (e.g. python encodings an empty `bytearray` and doesn't include the type).

tests/src/unit-bjdata.cpp
nlohmann: We must not break existing code, so yes, a parameter would make sense.

nebkat:
`draft3_binary` parameter added to `to_bjdata`, defaulting to false, along with tests for both versions.

`from_bjdata` previously wasn't able to parse `uint8` binary arrays so it has just gained the ability to parse `byte` arrays as binary with no parameter.

Would it make sense to introduce a `to_bjdata_draft3()` function and deprecate `to_bjdata()` to encourage use of the newer version?

nlohmann:
Thanks a lot! Feels much better without a breaking change!

Regarding the draft version - would it make sense to add an int parameter `draft` to the function and defaulting it to 2 and allow to set it to 3 instead of the boolean? Then we have some room for the future there.

Regarding deprecation - adding a new function to the API just for the sake of deprecating another feels odd. I think we should rather deprecate the default parameter for the mentioned `draft` parameter and force client to choose the value in the future.

What do you think?

nebkat:
I like it, didn't think of that, much better solution!

So perhaps something like this?

```c++
/// how to encode BJData
enum class bjdata_version_t
{
    draft2 JSON_HEDLEY_DEPRECATED_FOR(3.12.0, draft3),
    draft3,
};

...

static std::vector<std::uint8_t> to_bjdata(const basic_json& j,
        const bool use_size = false,
        const bool use_type = false,
        const bjdata_version_t version = bjdata_version_t::draft2)
{
    std::vector<std::uint8_t> result;
    to_bjdata(j, result, use_size, use_type, version);
    return result;
}

// warning: 'nlohmann::json_abi_v3_11_3::detail::bjdata_version_t::draft2' is deprecated: Since 3.12.0; use draft3 [-Wdeprecated-declarations]
// 4315 |             const bjdata_version_t version = bjdata_version_t::draft2)
```

Could technically even put `ubjson` as a version as they are so close and use it internally.

nlohmann:
> Could technically even put `ubjson` as a version as they are so close and use it internally.

You mean using BJData as parameter for UBJSON output/input?

nebkat:
Yes, internally.

`json/include/nlohmann/detail/output/binary_writer.hpp`:
```diff
    // json/include/nlohmann/detail/output/binary_writer.hpp
    void write_ubjson(const BasicJsonType& j, const bool use_count,
                      const bool use_type, const bool add_prefix = true,
-                      const bool use_bjdata = false)
+                      const bjdata_version_t version = bjdata_version_t::ubjson) // or "draft0", which could be argued is equivalent
    {
```

Then we can do:

`include/nlohmann/json.hpp`: 
```diff
    /// @brief create a UBJSON serialization of a given JSON value
    /// @sa https://json.nlohmann.me/api/basic_json/to_ubjson/
    static void to_ubjson(const basic_json& j, detail::output_adapter<std::uint8_t> o,
                          const bool use_size = false, const bool use_type = false)
    {
-        binary_writer<std::uint8_t>(o).write_ubjson(j, use_size, use_type);
+         binary_writer<std::uint8_t>(o).write_bjdata(j, use_size, use_type, bjdata_version_t::ubjson); // Or defaulted as above
    }

    /// @brief create a BJData serialization of a given JSON value
    /// @sa https://json.nlohmann.me/api/basic_json/to_bjdata/
    static void to_bjdata(const basic_json& j, detail::output_adapter<std::uint8_t> o,
-                         const bool use_size = false, const bool use_type = false)
+                         const bool use_size = false, const bool use_type = false, const bjdata_version_t version = bjdata_version_t::draft2)
    {
-        binary_writer<std::uint8_t>(o).write_ubjson(j, use_size, use_type, true, true);
+        binary_writer<std::uint8_t>(o).write_ubjson(j, use_size, use_type, true, version);
    }
```

Would cut down on the amount of `if (use_bjdata && bjdata_version == ...)` and instead just be `if (version == ...)`.

nlohmann:
Yes, what I meant: would this be something that could - in the long run - be exposed to the customer? As in: deprecating `to_bjdata` in favor of `to_ubjson` with a version parameter.

nebkat:
Ah right - yes, but if anything I would deprecate `to_ubjson` in favour of `to_bjdata` considering UBJSON has not seen any meaningful updates in ~8 years. If we can play our part in moving people towards BJData that seems like a positive thing in the long run! It is the closest thing we have to a 1-to-1 binary mapping of JSON without the extra fluff.

nlohmann:
Well, I thought of BJData to be a dialect of UBJSON, so I don't think replacing the UBJSON with those of BJData is a good idea.

nebkat:
The change in endianness between UBJSON/BJData makes it feel like more than just a dialect to me so I'd fear labelling it as such (and having both use `to_ubjson`) would cause some confusion if we do not explicitly deprecate UBJSON as a legacy format.

I suppose that applies to my proposal of a `ubjson` in `bjdata_version_t` too, so in retrospect I would stick to [this](https://github.com/nlohmann/json/pull/4513#issuecomment-2521578597) version - but let me know if you'd still like a common version enum.

A third option would be a public facing `bjdata_version_t` as above and a hidden `ubjson_dialect_t` for use in the common converter functions:

```c++
enum class ubjson_dialect_t
{
    ubjson_spec12,
    bjdata_draft2,
    bjdata_draft3,
};
```

nlohmann:
> The change in endianness between UBJSON/BJData makes it feel like more than just a dialect to me so I'd fear labelling it as such (and having both use `to_ubjson`) would cause some confusion if we do not explicitly deprecate UBJSON as a legacy format.

You are right - let's keep the two formats separated in the public API (while they may of course share code in the detail namespace).

I think the idea of a version enum from https://github.com/nlohmann/json/pull/4513#issuecomment-2521578597 makes sense here in order not to break existing code.

nlohmann:
Please see https://github.com/nlohmann/json/pull/4513/checks?check_run_id=35168472051 on how to sign off your commits.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -28,6 +28,13 @@ NLOHMANN_JSON_NAMESPACE_BEGIN
 namespace detail
 {
 
+/// how to encode BJData
+enum class bjdata_version_t
+{
+    draft2,
```
nlohmann, line 34: Didn't you want to use

```cpp
draft2 JSON_HEDLEY_DEPRECATED_FOR(3.12.0, draft3),
```

here?

nebkat: Ah yes, forgot to comment on this - I was getting deprecation warnings just from the default parameter value 😐 Seems the only way to achieve what we want is to overload the function and deprecate the original signature where version isn't specified too. I'll add that later today.

nebkat:
That unfortunately leaves `to_bjdata(j, true, true, nlohmann::json::bjdata_version_t::draft3)` looking rather verbose.

Possible alternatives are:

1. `to_bjdata(j, bjdata_config_t { size, type, version } )`, nice because we can default version to 3 if unspecified, and the config can be stored in a variable.
2. `to_bjdata(j, use_size, use_type, draft3_binary = false)`, less room for expansion, though I doubt there will be many more breaking versions.
3. `to_bjdata(version, j, use_size, use_type)`, at least don't have to specify `use_size` and `use_type` every time.
4. `to_bjdata_v3()` with `to_bjdata()` deprecated.
5. Shortened `bjdata_version_t`, maybe `bjdata::draft3`?

nlohmann:
> That unfortunately leaves `to_bjdata(j, true, true, nlohmann::json::bjdata_version_t::draft3)` looking rather verbose.

ADL should allow you to write `to_bjdata(j, true, true, bjdata_version_t::draft3)`. Still not nice, but at least shorter.
 
> 1. `to_bjdata(j, bjdata_config_t { size, type, version } )`, nice because we can default version to 3 if unspecified, and the config can be stored in a variable.

This could be a way for the future - also the `dump` function currently suffers from too many parameters.

> 2. `to_bjdata(j, use_size, use_type, draft3_binary = false)`, less room for expansion, though I doubt there will be many more breaking versions.

I see. But adding yet another function to the public API is also not too nice.

> 3. `to_bjdata(version, j, use_size, use_type)`, at least don't have to specify `use_size` and `use_type` every time.

I see, though I wonder how many people this actually affects.

> 4. `to_bjdata_v3()` with `to_bjdata()` deprecated.

No. The public API is already so large.

> 5. Shortened `bjdata_version_t`, maybe `bjdata::draft3`?

I see your point, but `bjdata_version_t::draft3` is expressive and to the point.

I think we should extend the current `to_bson` function. I will create a discussion to collect the respective options for each serialization function. Maybe some aspects can be shared between the formats. Once we have these in place, we can deprecate the existing APIs and use cleaner ones where all options are expressed in dedicated structs.

What do you think?

nebkat:
> ADL should allow you to write to_bjdata(j, true, true, bjdata_version_t::draft3). Still not nice, but at least shorter.

Didn't realize ADL let you do this, that's not so bad.

> I think we should extend the current to_bson function. I will create a discussion to collect the respective options for each serialization function. Maybe some aspects can be shared between the formats. Once we have these in place, we can deprecate the existing APIs and use cleaner ones where all options are expressed in dedicated structs.

Agreed, I think that is the best way forward. 

I've reverted back to the commit where nothing is deprecated, so no current users will be affected, and power users (probably just me for the next while) can specify `draft3` as last argument explicitly.

nlohmann (approved):
Looks good to me.

docs/mkdocs/docs/api/basic_json/to_bjdata.md
```diff
@@ -68,3 +74,4 @@ Linear in the size of the JSON value `j`.
 ## Version history
 
 - Added in version 3.11.0.
+- BJData version parameter (for draft3 binary encoding) added in version 3.12.0.
```
nebkat, line 77:
```suggestion
- BJData version parameter (for draft3 binary encoding) added in version 3.11.4.
```

nlohmann: No, the next version number ist not yet clear. Could also be 3.12.0.

nlohmann:
Thanks!!

nebkat:
Thanks for the assistance!

docs/mkdocs/docs/api/basic_json/to_bjdata.md
```diff
@@ -34,6 +37,9 @@ The exact mapping and its limitations is described on a [dedicated page](../../f
 
 `use_type` (in)
 :   whether to add type annotations to container types (must be combined with `#!cpp use_size = true`); optional,
+
+`version` (in)
+:   which version of BJData to use (see [draft 3](../../features/binary_formats/bjdata.md#draft-3-binary-format)); optional,
```
gregmarr, line 42: This new parameter should have gone after the next line, which is the default value for `use_type`.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -735,11 +742,14 @@ class binary_writer
...
     @param[in] add_prefix  whether prefixes need to be used for this value
     @param[in] use_bjdata  whether write in BJData format, default is false
+    @param[in] bjdata_version  which BJData version to use, default is draft2
     */
     void write_ubjson(const BasicJsonType& j, const bool use_count,
                       const bool use_type, const bool add_prefix = true,
-                      const bool use_bjdata = false)
+                      const bool use_bjdata = false, const bjdata_version_t bjdata_version = bjdata_version_t::draft2)
     {
+        const bool bjdata_draft3 = bjdata_version == bjdata_version_t::draft3;
```
gregmarr, line 751: If you make this `const bool bjdata_draft3 = use_bjdata && bjdata_version == bjdata_version_t::draft3;` then you can eliminate the `use_bjdata &&` on each of the checks below, and in some cases the extra parens around that.

nlohmann:
@nebkat Can you create another PR for the  https://github.com/nlohmann/json/pull/4513#pullrequestreview-2535026794?

include/nlohmann/detail/output/binary_writer.hpp
nlohmann: Fixed in https://github.com/nlohmann/json/pull/4588

docs/mkdocs/docs/api/basic_json/to_bjdata.md
nlohmann: Fixed in https://github.com/nlohmann/json/pull/4588

Issue #4514, opened 2024-11-25
parse json OOM 


ddxxsx:
### Description

I have a string, which is about 10MB in memory. However, when I use json::parse(), the memory cost is 1.6GB.
I wonder if it's a bug?

### Reproduction steps

A long string with chinese charater encoded with utf-8.
call
json j = json::parse(str);
can trigger this bug.

### Expected vs. actual results

lower memory cost.

### Compiler and operating system

windows

### Library version

branch: develop, commit hash: ee32bfc

### Validation

- [X] The bug also occurs if the latest version from the [`develop`](https://github.com/nlohmann/json/tree/develop) branch is used.
- [X] I can successfully [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests).

nlohmann:
Please share the input to check. Please also make sure you compile in release mode.

ddxxsx:
The input file is too large to upload(github limit 25MB), but the file is 64MB large.
The release mode costs less memory(600MB), while debug mode costs almost 1GB memory.

My code is like this:

    ifstream file("C:\\input.txt");
    std::stringstream buffer;
    buffer << file.rdbuf();

    std::string content = buffer.str();
    json j = json::parse(content);



Is there a way to upload the input file here?

nlohmann:
I would assume using xz would make the file small enough.

ddxxsx:
[input.zip](https://github.com/user-attachments/files/17911718/input.zip)

Github doesn't support xz file, I attach a zip file instead.

Memory cost:
![image](https://github.com/user-attachments/assets/67165301-cecb-4c6e-b046-03d92f9489b4)

nlohmann:
<img width="729" alt="Screenshot 2024-11-27 at 18 12 00" src="https://github.com/user-attachments/assets/4b300442-c31e-4224-b1e8-9a9bbb8f2acd">

I see 250 MB for that file. Not sure what your compiler is doing there.

My code for reference:

```cpp
#include <fstream>
#include <iostream>
#include <nlohmann/json.hpp>

int main() {
	std::ifstream f("/Users/niels/Downloads/input.txt");
	auto x = nlohmann::json::parse(f);

	 do  {
	 	std::cout << '\n' << "Press a key to continue...";
	 } while (std::cin.get() != '\n');
}
```

```shell
g++ -I include -O3 -DNDEBUG=1 -std=c++11 parse.cpp
```

PR #4517, opened 2024-11-25
json start/end position implementation

sushshring:
# Abstract
Referring to discussion: [4455](https://github.com/nlohmann/json/discussions/4455), this pull request introduces the implementation to retrieve the start and end positions of nested objects within the JSON during parsing.

# Motivation
We have a service implementation with JSON schema where a field within the nested objects contains the hash value for that object. The service verifies the hash value of each of the nested objects before operating on the rest of the data sent.

For example, consider the following JSON:
```
{
    "name": "foo",
    "data":
    {
        "type": "typeA",
        "value": 1,
        "details": {
            "nested_type": "nested_typeA",
            "nested_value": 2
        }
    },
    "data_hash": "hashA"
}
```
Here, data_hash contains the hash of the object "details". In order to verify the data hash, we need to be able to retrieve the exact string that parsed out "details" including the spaces and newlines. Currently there is no way to achieve this using nlohmann/json parser.

# Changes proposed
- Add two fields to basic_json: `size_t start_position` and `size_t end_position`. 
- Add a reference to the lexer in json_sax_parser to retrieve the current position in the input string.
- Whenever a BasicJsonType is created by the parser, calculate the start and end positions for that object from the original string and store those values.

# Memory considerations
We considered storing substrings in the output JSON objects and sub-objects directly as well, however, considering the memory footprint increase that it would create, we opted for the option where only two size_t fields are stored per basic_json created.

# Validation
We have added tests to the class_parser test suite that cover the following cases:
- Array inside an object
- Objects inside arrays
- Doubly nested objects
- String fields
- Integer and float fields
- Float values with insignificant digits
- Boolean fields
- Null fields

Since the change affects the sax_parser, for each of these test cases we validate scenarios where no callback is passed, a callback is passed that accepts all fields, and a callback is passed that filters specific fields. 
* * *

## Pull request checklist

## Please don't


nlohmann:
Thanks for the effort!

However, adding two `size_t` members is a lot of overhead. When we introduced diagnostics, we hid a single pointer behind a preprocessor macro to avoid every single client to suffer from the overhead. Issues like #4514 show that the memory efficiency is already quite bad.

I am hesitant how to continue here.

gregmarr:
I wonder if this is something that could be done with the data in the custom base class, so that only those that want to opt in to this behavior could enable it.  That does make it a custom class, and not `nlohmann::json` or `nlohmann::ordered_json`.

sushshring:
Taking that advice, I'm gonna add a new class like below to use as a json custom base class

```
class json_base_class_with_start_end_markers {
    size_t start_position = std::string::npos;
    size_t end_position = std::string::npos;

public:
    size_t get_start_position() const noexcept
    {
        return start_position;
    }

    size_t get_end_position() const noexcept
    {
        return end_position;
    }

    void set_start_position(size_t start) noexcept
    {
        start_position = start;
    }

    void set_end_position(size_t end) noexcept
    {
        end_position = end;
    }
};
```
We will use `if (std::is_base_of<json_base_class_with_start_end_markers, BasicJsonType>){}` whenever the start_position and end_position setters are called within json_sax.hpp.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -351,11 +439,13 @@ class json_sax_dom_callback_parser
...
+    using lexer_t = lexer<BasicJsonType, InputAdapterType>;
 
     json_sax_dom_callback_parser(BasicJsonType& r,
-                                 parser_callback_t cb,
-                                 const bool allow_exceptions_ = true)
-        : root(r), callback(std::move(cb)), allow_exceptions(allow_exceptions_)
+                                 const parser_callback_t cb,
+                                 const bool allow_exceptions_ = true,
+                                 lexer_t* lexer_ = nullptr)
+        : root(r), callback(cb), allow_exceptions(allow_exceptions_), m_lexer_ref(lexer_)
```
sushshring: bad merge: put back std::move()

tests/src/unit-disabled_exceptions.cpp
```diff
@@ -20,10 +20,10 @@ using json = nlohmann::json;
 // for #2824
 /////////////////////////////////////////////////////////////////////
 
-class sax_no_exception : public nlohmann::detail::json_sax_dom_parser<json>
+class sax_no_exception : public nlohmann::detail::json_sax_dom_parser<json, decltype(nlohmann::detail::input_adapter(""))>
```
gregmarr: Why the `decltype()` here when you know the type?

tests/src/unit-deserialization.cpp
```diff
@@ -588,7 +588,8 @@ TEST_CASE("deserialization")
                 auto first = str.begin();
                 auto last = str.end();
                 json j;
-                json_sax_dom_parser<json> sax(j, true);
+                auto ia = nlohmann::detail::input_adapter(str);
+                json_sax_dom_parser<json, decltype(ia)> sax(j, true);
```
gregmarr: Why the extra unused object and the `decltype()` here when you know the type?

include/nlohmann/detail/json_base_class_with_start_end_markers.hpp
```diff
@@ -0,0 +1,52 @@
...
+/*!
+@brief Custom base class of the @ref basic_json class.
+
+*/
+class json_base_class_with_start_end_markers
+{
+    size_t start_position = std::string::npos;
+    size_t end_position = std::string::npos;
+
+  public:
```
gregmarr: These are basically public members.  Any reason for these being private other than it lets you set breakpoints?

include/nlohmann/detail/json_base_class_with_start_end_markers.hpp
```diff
@@ -0,0 +1,52 @@
...
+// |  |  |__   |  |  | | | |  version 3.11.3
+// |_____|_____|_____|_|___|  https://github.com/nlohmann/json
+//
+// SPDX-FileCopyrightText: 2013-2023 Niels Lohmann <https://nlohmann.me>
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+#include <type_traits> // conditional, is_same
+#include <string>
```
gregmarr: Doesn't look like these headers are used here.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -62,7 +62,7 @@ static inline bool little_endianness(int num = 1) noexcept
 /*!
 @brief deserialization of CBOR, MessagePack, and UBJSON values
 */
-template<typename BasicJsonType, typename InputAdapterType, typename SAX = json_sax_dom_parser<BasicJsonType>>
+template<typename BasicJsonType, typename InputAdapterType, typename SAX = json_sax_dom_parser<BasicJsonType, InputAdapterType>>
```
nlohmann, line 68: Why do you need `InputAdapterType` here?

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -548,6 +680,67 @@ class json_sax_dom_callback_parser
     }
 
   private:
+
+    template<typename Value>
+    void set_start_end_pos_for_json_value(Value& v)
```
nlohmann: Can we use some SFINAE magic to have two versions of this function - one for the case that `std::is_base_of<::nlohmann::detail::json_base_class_with_start_end_markers, BasicJsonType>::value)` and if not with the latter doing nothing? Then this if is done at compile time.

nlohmann (changes requested):
Some comments from my side. I am honestly not 100% sold to the use case this PR is solving. As I stated in https://github.com/nlohmann/json/discussions/4455#discussioncomment-11421816, I *think* we preserve every relevant part of the raw input except whitespace. Did I miss anything?

include/nlohmann/detail/input/json_sax.hpp
sushshring: Added SFINAE with use of macros

include/nlohmann/detail/input/binary_reader.hpp
sushshring: I've added a reference to the lexer in the json_sax_dom_parser, which is how I retrieve the specific positions to be added as start_pos and end_pos. The lexer reference requires the type of InputAdapterType to be passed in, so thats been added to the json_sax_dom_parser as well.

sushshring:
> Some comments from my side. I am honestly not 100% sold to the use case this PR is solving. As I stated in [#4455 (comment)](https://github.com/nlohmann/json/discussions/4455#discussioncomment-11421816), I _think_ we preserve every relevant part of the raw input except whitespace. Did I miss anything?

Whitespace is key to our use case - my colleague will provide a more detailed explanation in #4455 soon.

include/nlohmann/detail/json_base_class_with_start_end_markers.hpp
```diff
@@ -0,0 +1,32 @@
...
+{
+
+/*!
+@brief Custom base struct of the @ref basic_json class.
+This class exposes the start and end positions for all fields of a JSON object
+with reference to the parsed input.
+*/
+struct json_base_class_with_start_end_markers
+{
+  public:
```
gregmarr: `public` is the default for a `struct`.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -227,19 +244,18 @@ class json_sax_dom_parser
         return true;
     }
 
+    HANDLE_START_END_POS_DEFINITION(start_object,
+    {
+        reinterpret_cast<::nlohmann::detail::json_base_class_with_start_end_markers*>(ref_stack.back())->start_position = m_lexer_ref->get_position() - 1;
```
gregmarr: Due to the SFINAE changes, the `reinterpret_cast<::nlohmann::detail::json_base_class_with_start_end_markers*>()` is not needed.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -259,35 +275,35 @@ class json_sax_dom_parser
         return true;
     }
 
+    HANDLE_START_END_POS_DEFINITION(end_object,
+    {
+        reinterpret_cast<::nlohmann::detail::json_base_class_with_start_end_markers*>(ref_stack.back())->end_position = m_lexer_ref->get_position();
```
gregmarr: As above

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -259,35 +275,35 @@ class json_sax_dom_parser
...
+        CALL_HANDLE_START_END_POS_FUNCTION(end_object);
 
         ref_stack.back()->set_parents();
         ref_stack.pop_back();
         return true;
     }
 
+    HANDLE_START_END_POS_DEFINITION(start_array,
+    {
+        reinterpret_cast<::nlohmann::detail::json_base_class_with_start_end_markers*>(ref_stack.back())->start_position = m_lexer_ref->get_position() - 1;
```
gregmarr: As above

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -297,16 +313,17 @@ class json_sax_dom_parser
         return true;
     }
 
+    HANDLE_START_END_POS_DEFINITION(end_array,
+    {
+        reinterpret_cast<::nlohmann::detail::json_base_class_with_start_end_markers*>(ref_stack.back())->end_position = m_lexer_ref->get_position();
```
gregmarr: As above

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -333,65 +350,60 @@ class json_sax_dom_parser
 
   private:
 
-    void set_start_end_pos_for_json_value(BasicJsonType& v)
+    HANDLE_START_END_POS_DEFINITION(json_value,
     {
-        if (JSON_HEDLEY_UNLIKELY((std::is_base_of<::nlohmann::detail::json_base_class_with_start_end_markers, BasicJsonType>::value)))
+        auto* v_base = reinterpret_cast<::nlohmann::detail::json_base_class_with_start_end_markers*>(&v);
```
gregmarr: Due to SFINAE, don't need `v_base`, just use `v`.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -510,6 +522,11 @@ class json_sax_dom_callback_parser
         return true;
     }
 
+    HANDLE_START_END_POS_DEFINITION(start_object,
+    {
+        reinterpret_cast<::nlohmann::detail::json_base_class_with_start_end_markers*>(ref_stack.back())->start_position = m_lexer_ref->get_position() - 1;
```
gregmarr: Not required, as above.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -555,6 +568,11 @@ class json_sax_dom_callback_parser
         return true;
     }
 
+    HANDLE_START_END_POS_DEFINITION(end_object,
+    {
+        reinterpret_cast<::nlohmann::detail::json_base_class_with_start_end_markers*>(ref_stack.back())->end_position = m_lexer_ref->get_position();
```
gregmarr: As above

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -595,6 +610,11 @@ class json_sax_dom_callback_parser
         return true;
     }
 
+    HANDLE_START_END_POS_DEFINITION(start_array,
+    {
+        reinterpret_cast<::nlohmann::detail::json_base_class_with_start_end_markers*>(ref_stack.back())->start_position = m_lexer_ref->get_position() - 1;
```
gregmarr: As above

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -625,6 +639,11 @@ class json_sax_dom_callback_parser
         return true;
     }
 
+    HANDLE_START_END_POS_DEFINITION(end_array,
+    {
+        reinterpret_cast<::nlohmann::detail::json_base_class_with_start_end_markers*>(ref_stack.back())->end_position = m_lexer_ref->get_position();
```
gregmarr: As above

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -681,66 +697,60 @@ class json_sax_dom_callback_parser
 
   private:
 
-    template<typename Value>
-    void set_start_end_pos_for_json_value(Value& v)
+    HANDLE_START_END_POS_DEFINITION(json_value,
     {
-        if (JSON_HEDLEY_UNLIKELY((std::is_base_of<::nlohmann::detail::json_base_class_with_start_end_markers, BasicJsonType>::value)))
+        auto* v_base = reinterpret_cast<::nlohmann::detail::json_base_class_with_start_end_markers*>(&v);
```
gregmarr: Due to SFINAE, don't need `v_base`, just use `v`.

tests/src/unit-deserialization.cpp
```diff
@@ -588,8 +588,7 @@ TEST_CASE("deserialization")
                 auto first = str.begin();
                 auto last = str.end();
                 json j;
-                auto ia = nlohmann::detail::input_adapter(str);
-                json_sax_dom_parser<json, decltype(ia)> sax(j, true);
+                json_sax_dom_parser<json, typename nlohmann::detail::container_input_adapter_factory_impl::container_input_adapter_factory<std::string>::adapter_type> sax(j, true);
```
gregmarr: Well, that's a bit more ugly than I expected.  Maybe an `input_adapter_type` alias template would be useful here.

raphgianotti:
> Some comments from my side. I am honestly not 100% sold to the use case this PR is solving. As I stated in [#4455 (comment)](https://github.com/nlohmann/json/discussions/4455#discussioncomment-11421816), I _think_ we preserve every relevant part of the raw input except whitespace. Did I miss anything?

The main issue we are trying to solve is when needing to sign the string corresponding to a nested object in a json, example:
{
    "object_to_sign": {
        "int_value": 23,
        "float_value": 1.234,
        "string_value": "hello"
    },
    "signature": "<signature>"
}
In this scenario, for us to calculate the signature using a public key to validate the contents of "object_to_sign" are correct depends fully on every thing about the string representing "object_to_sign" in the original payload (including spaces/line breaks). When defining a protocol for this, it's not possible (since the sender could be employing any json library) to simply use the nlohmann::json object generated from parsing to generate the string for "object_to_sign" again, as it could differ slightly, even if its contents are equivalent.
 
Even thought this is a somewhat niche scenario, it's mostly exposing information nlohmann has access to and, as an optional feature, I think it can add value to the library, as doing it outside it would require some form of parsing the contents of the string again, doubling the complexity and involving using two separate json parsers where this change allows the use of a single one.

tests/src/unit-deserialization.cpp
sushshring: Just to confirm, the suggestion is to add a `using string_input_adapter = decltype(input_adapter(std::declval<std::string>()));` in input_adapters.hpp or keep the `input_adapter_type` alias scoped to the two tests: unit-deserializer.cpp and unit-disabled_exceptions.cpp?

gregmarr: If it would be useful (and works) to use it as the return type of `input_adapter`, that would be good, I think.

gregmarr:
I think it's going to be difficult to use the macro for `handle_start_end_pos_for_json_value` due to the unused parameters when the base class is used.  It may make sense to write those two functions out individually.

I had another thought about these:

```
#define HANDLE_START_END_POS_DEFINITION(__handlecase, __statement,...)   \
    template <class Q = BasicJsonType>                                                                                                                  \
    typename std::enable_if<std::is_base_of<::nlohmann::detail::json_base_class_with_start_end_markers, Q>::value, void>::type                          \
    handle_start_end_pos_for_##__handlecase(__VA_ARGS__)                                                                                                \
    {                                                                                                                                                   \
        if (m_lexer_ref)                                                                                                                                \
        {                                                                                                                                               \
            __statement                                                                                                                                 \
        }                                                                                                                                               \
    }                                                                                                                                                   \
    \
    template <class Q = BasicJsonType>                                                                                                                  \
    typename std::enable_if<!std::is_base_of<::nlohmann::detail::json_base_class_with_start_end_markers, Q>::value, void>::type                         \
    handle_start_end_pos_for_##__handlecase(__VA_ARGS__){}

    HANDLE_START_END_POS_DEFINITION(start_object,
    {
        ref_stack.back()->start_position = m_lexer_ref->get_position() - 1;
    })
```

What if we created two set functions, and put them in the base class, with no-ops in the main class if the base class isn't used?  Then the function can always be called, and you don't need to protect the one-line callers, but you would want to protect the more involved ones because of the extra code involved.
```
struct json_base_class_with_start_end_markers {
    size_t start_position = ...;
    size_t end_position = ...;
    // functions for storing the start and end positions when this is the basic_json base class.
    void set_start_position(size_t i) { start_position = i; }
    void set_end_position(size_t i) { end_position = i; }
}

class basic_json {
    // no-op functions for storing the start and end positions when the base class isn't used.
    typename std::enable_if<!std::is_base_of<detail::json_base_class_with_start_end_markers, decltype(*this)>::value, void>::type
    void set_start_position(size_t){}
    typename std::enable_if<!std::is_base_of<detail::json_base_class_with_start_end_markers, decltype(*this)>::value, void>::type
    void set_end_position(size_t){}

};

    void handle_start_end_pos_for_start_object()
    {
        if (m_lexer_ref != nullptr)
        {
            ref_stack.back()->set_start_position(m_lexer_ref->get_position() - 1);
        }
    }
```

sushshring:
@gregmarr I'd considered that, but decided against it to minimize changes to basic_json. Since it's a no-op I'm guessing the compiler might just optimize those away. I will update with that change

gregmarr:
Yeah, since this is header-only, the call itself should definitely get eliminated, and then any computations that are just related to that will hopefully be eliminated too.

include/nlohmann/json_fwd.hpp
```diff
@@ -70,6 +71,20 @@ struct ordered_map;
 /// @sa https://json.nlohmann.me/api/ordered_json/
 using ordered_json = basic_json<nlohmann::ordered_map>;
 
+/// @brief a minimal specialization that uses the base class json_base_class_with_start_end_markers
+using json_with_start_end_markers = nlohmann::basic_json <
```
nlohmann: I hope we can find a better name for this feature...

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -298,6 +349,61 @@ class json_sax_dom_parser
...
+    HANDLE_START_END_POS_DEFINITION(json_value,
+    {
+        auto* v_base = reinterpret_cast<::nlohmann::detail::json_base_class_with_start_end_markers*>(&v);
+        v_base->end_position = m_lexer_ref->get_position();
+
+        switch (v.type())
+        {
+            case value_t::boolean:
+            {
+                v_base->start_position = v_base->end_position - (v.m_data.m_value.boolean ? 4 : 5);
```
nlohmann: Please comment the magic numbers (4=true, 5=false).

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -548,6 +696,61 @@ class json_sax_dom_callback_parser
...
```
nlohmann: Again, please add a comment for the magic numbers.

include/nlohmann/json_fwd.hpp
sushshring: Open to discussion for names :). I'm terrible at coming up with good names

nlohmann: If I had a better one, I'd have dropped it here. :-)

sushshring:
@gregmarr After some prototyping I have realized that to correctly enable SFINAE in the derived class, we need to add the following:
```
    template<class T = CustomBaseClass>
    typename std::enable_if<!std::is_same<T, detail::json_base_class_with_start_end_markers>::value, void>::type
    set_start_position(size_t){}
    template<class T = CustomBaseClass>
    typename std::enable_if<!std::is_same<T, detail::json_base_class_with_start_end_markers>::value, void>::type
    set_end_position(size_t){}
 
    template<class T = CustomBaseClass>
    typename std::enable_if<std::is_same<T, detail::json_base_class_with_start_end_markers>::value, void>::type
    set_start_position(size_t pos)
    {
        reinterpret_cast<detail::json_base_class_with_start_end_markers*>(this)->set_start_position(pos);
    }
    template<class T = CustomBaseClass>
    typename std::enable_if<std::is_same<T, detail::json_base_class_with_start_end_markers>::value, void>::type
    set_end_position(size_t pos)
    {
        reinterpret_cast<detail::json_base_class_with_start_end_markers*>(this)->set_end_position(pos);
    }
```
The methods need the alternate implementation in order to work with SFINAE. The already implemented versions in the base class do not count for the SFINAE pattern's error ignore. This results in a rather ugly `reinterpret_cast<>` within basic_json that will certainly violate https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c153-prefer-virtual-function-to-casting. The safer alternative is to add virtual methods to both base classes and override those in basic_json, which will likely affect performance slightly.

The macros that I've added in the recent commit don't work with C++11, however, that can be fixed by adding a new version of the macros that takes no arguments and removes the varargs, which makes the CI checks happy. I will implement that in my next commit instead.

include/nlohmann/json_fwd.hpp
sushshring: For the base class, how about `json_base_with_positions`? or `json_base_positional`?
Then this could be `json_with_positions`, or `json_positional`.

nlohmann:
A very wild idea: why not adding another macro like DIAGNOSTIC_POSITIONS and implement this without inheritance?

gregmarr:
> The methods need the alternate implementation in order to work with SFINAE. The already implemented versions in the base class do not count for the SFINAE pattern's error ignore. 

There is no SFINAE for the simple functions with this method.

```
    void handle_start_end_pos_for_end_array()
    {
        if (m_lexer_ref != nullptr)
        {
            ref_stack.back()->set_end_position(m_lexer_ref->get_position());
        }
    }

    bool end_array()
    {
        bool keep = true;

        if (ref_stack.back())
        {
            keep = callback(static_cast<int>(ref_stack.size()) - 1, parse_event_t::array_end, *ref_stack.back());
            if (keep)
            {
                handle_start_end_pos_for_end_array();
```

Since object and array do the same thing, that could also be simplified to reduce the number of functions:
```
    void record_start_pos()
    {
        if (m_lexer_ref != nullptr)
        {
            ref_stack.back()->set_start_position(m_lexer_ref->get_position() - 1);
        }
    }
    void record_end_pos()
    {
        if (m_lexer_ref != nullptr)
        {
            ref_stack.back()->set_end_position(m_lexer_ref->get_position());
        }
    }
```

>why not adding another macro like DIAGNOSTIC_POSITIONS and implement this without inheritance?

That's certainly a possibility.  I thought the base class might be cleaner, but it changes the types, so maybe it isn't.  We'd need to encode this define into the intermediate namespace selection so that we don't link against something with the wrong definition, as we do for `JSON_DIAGNOSTICS`.

https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/abi_macros.hpp#L33-L37

sushshring:
> A very wild idea: why not adding another macro like DIAGNOSTIC_POSITIONS and implement this without inheritance?

This seems feasible, thanks for the note on the intermediate namespace @gregmarr. We are prototyping this locally as an option and will post the next iteration based on our findings.

include/nlohmann/detail/json_base_class_with_start_end_markers.hpp
sushshring: Implemented the macros, so base class is gone.

include/nlohmann/detail/input/json_sax.hpp
sushshring: Implemented the macros, so no need for SFINAE

tests/src/unit-deserialization.cpp
sushshring: Added

include/nlohmann/json_fwd.hpp
sushshring: Implemented the macros, so base class is gone. basic_json now exposes two property getters: `start_pos()` and `end_pos()`, if the `DIAGNOSTIC_POSITIONS` macro is set. Otherwise there is no additional memory footprint.

include/nlohmann/detail/input/json_sax.hpp
sushshring: Added

include/nlohmann/detail/input/json_sax.hpp
sushshring: Added

sushshring:
@nlohmann pushed a new commit that implements the diagnostic style macro. I personally think this looks cleaner as well, PTAL.

include/nlohmann/detail/abi_macros.hpp
```diff
@@ -26,6 +26,10 @@
     #define JSON_DIAGNOSTICS 0
 #endif
 
+#ifndef DIAGNOSTIC_POSITIONS
```
gregmarr: Should this define start with JSON_ like the others?

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -12,11 +12,12 @@
 #include <string> // string
 #include <utility> // move
 #include <vector> // vector
+#include <type_traits> // enable_if_t
 
 #include <nlohmann/detail/exceptions.hpp>
 #include <nlohmann/detail/macro_scope.hpp>
 #include <nlohmann/detail/string_concat.hpp>
-
+#include <nlohmann/detail/input/lexer.hpp>
```
gregmarr: Lost a blank line here.

include/nlohmann/json.hpp
```diff
@@ -893,6 +893,11 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
                 JSON_ASSERT(false); // NOLINT(cert-dcl03-c,hicpp-static-assert,misc-static-assert) LCOV_EXCL_LINE
         }
         JSON_ASSERT(m_data.m_type == val.type());
+
+#if DIAGNOSTIC_POSITIONS
+        start_position = val.start_position;
+        end_position = val.end_position;
+#endif
```
gregmarr: blank line after?

include/nlohmann/json.hpp
```diff
@@ -1204,15 +1209,24 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
             default:
                 break;
         }
+#if DIAGNOSTIC_POSITIONS
```
gregmarr: blank line before?

include/nlohmann/json.hpp
```diff
@@ -1394,6 +1417,24 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+        return start_position;
+    }
+
+    constexpr size_t end_pos() const noexcept
+    {
+        return end_position;
+    }
+  private:
+    /// the start position of the value
+    size_t start_position = std::string::npos;
```
gregmarr: I think these data members should be next to the other data members, down around 4265, probably after the existing ones so as to not unnecessarily change the order of data members.  We also need to make sure that the members are initialized in the constructors in same order, as compilers will warn if they're not in the same order.

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -417,6 +417,9 @@ typename container_input_adapter_factory_impl::container_input_adapter_factory<C
     return container_input_adapter_factory_impl::container_input_adapter_factory<ContainerType>::create(container);
 }
 
+// specialization for std::string
+using string_input_adapter = decltype(input_adapter(std::declval<std::string>()));
```
gregmarr: `using string_input_adapter_type =`?

tests/src/unit-cbor.cpp
```diff
@@ -1631,7 +1631,8 @@ TEST_CASE("CBOR")
                 };
 
                 json j;
-                auto cbp = nlohmann::detail::json_sax_dom_callback_parser<json>(j, callback, true);
+                auto ia = nlohmann::detail::input_adapter(input);
+                auto cbp = nlohmann::detail::json_sax_dom_callback_parser<json, decltype(ia)>(j, callback, true);
```
gregmarr: `string_input_adapter_type`?

tests/src/unit-class_parser.cpp
```diff
@@ -219,7 +219,8 @@ json parser_helper(const std::string& s)
     CHECK(j_nothrow == j);
 
     json j_sax;
-    nlohmann::detail::json_sax_dom_parser<json> sdp(j_sax);
+    auto ia = nlohmann::detail::input_adapter(s);
+    nlohmann::detail::json_sax_dom_parser<json, decltype(ia)> sdp(j_sax);
```
gregmarr: `string_input_adapter_type`?

tests/src/unit-class_parser_diagnostic_positions.cpp
```diff
@@ -0,0 +1,1959 @@
...
+
+    // if this line was reached, no exception occurred
+    // -> check if result is the same without exceptions
+    json j_nothrow;
+    CHECK_NOTHROW(json::parser(nlohmann::detail::input_adapter(s), nullptr, false).parse(true, j_nothrow));
+    CHECK(j_nothrow == j);
+
+    json j_sax;
+    auto ia = nlohmann::detail::input_adapter(s);
+    nlohmann::detail::json_sax_dom_parser<json, decltype(ia)> sdp(j_sax);
```
gregmarr: `string_input_adapter_type`?

tests/src/unit-regression2.cpp
```diff
@@ -162,11 +162,11 @@ struct adl_serializer<NonDefaultConstructible>
 // for #2824
 /////////////////////////////////////////////////////////////////////
 
-class sax_no_exception : public nlohmann::detail::json_sax_dom_parser<json>
+class sax_no_exception : public nlohmann::detail::json_sax_dom_parser<json, decltype(nlohmann::detail::input_adapter(""))>
 {
   public:
     explicit sax_no_exception(json& j)
-        : nlohmann::detail::json_sax_dom_parser<json>(j, false)
+        : nlohmann::detail::json_sax_dom_parser<json, decltype(nlohmann::detail::input_adapter(""))>(j, false)
```
gregmarr: `string_input_adapter_type`?

sushshring:
The  ci_clang_tidy job keeps complaining about the parser callback being passed by value in json_sax, but that’s not really any code we’re modifying in this PR. What’s the suggestion for that?

include/nlohmann/json.hpp
sushshring: Made a judgement call on the location, thanks for providing the appropriate one. I’ll make sure to run the ci tools locally for the compiler checks

include/nlohmann/json_fwd.hpp
nlohmann: For this PR, changing the name to `JSON_DIAGNOSTIC_POSITION` would be sufficient. As possible next step, we should also pick up the positions to provide better error message since we now know the positions of each part of the JSON.

include/nlohmann/detail/abi_macros.hpp
nlohmann: Yes, and we should also have a brief first version of a documentation ready like https://raw.githubusercontent.com/nlohmann/json/develop/docs/mkdocs/docs/api/macros/json_diagnostics.md - I can take care about the rest.

nlohmann:
> The ci_clang_tidy job keeps complaining about the parser callback being passed by value in json_sax, but that’s not really any code we’re modifying in this PR. What’s the suggestion for that?

I think everywhere else we pass it as non-const parameter, which seems to be OK for Clang-Tidy.

include/nlohmann/json_fwd.hpp
gregmarr: I like the better error message idea, when this is enabled.

include/nlohmann/detail/abi_macros.hpp
```diff
@@ -40,10 +40,10 @@
     #define NLOHMANN_JSON_ABI_TAG_DIAGNOSTICS
 #endif
 
-#if DIAGNOSTIC_POSITIONS
-    #define NLOHMANN_JSON_ABI_TAG_DIAGNOSTIC_POSITIONS _dp
+#if JSON_DIAGNOSTIC_POSITIONS
+    #define NLOHMANN_JSON_ABI_TAG_JSON_DIAGNOSTIC_POSITIONS _dp
```
gregmarr: Don't need to add `JSON_` in the middle of this one, see `NLOHMANN_JSON_ABI_TAG_DIAGNOSTICS` for `JSON_DIAGNOSTICS`.

include/nlohmann/json.hpp
```diff
@@ -1209,25 +1210,25 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
     /// @sa https://json.nlohmann.me/api/basic_json/basic_json/
     basic_json(basic_json&& other) noexcept
         : json_base_class_t(std::forward<json_base_class_t>(other)),
-#if DIAGNOSTIC_POSITIONS
-          start_position(other.start_position),
-          end_position(other.end_position),
-#endif
           m_data(std::move(other.m_data))
+#if JSON_DIAGNOSTIC_POSITIONS
+        , start_position(other.start_position),
```
gregmarr: Ew.  This is ugly.  Maybe it should be before `m_data` even though that moves the member.  @nlohmann thoughts?

Probably easier to just put the commas first.
```
          m_data(std::move(other.m_data))
#if JSON_DIAGNOSTIC_POSITIONS
          , start_position(other.start_position)
          , end_position(other.end_position)
#endif

File: include/nlohmann/detail/input/json_sax.hpp [hunk 35]
```diff
@@ -12,10 +12,12 @@
 #include <string> // string
 #include <utility> // move
 #include <vector> // vector
+#include <type_traits> // enable_if_t
```
gregmarr: These includes are sorted.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -12,10 +12,12 @@
...
 
 #include <nlohmann/detail/exceptions.hpp>
 #include <nlohmann/detail/macro_scope.hpp>
 #include <nlohmann/detail/string_concat.hpp>
+#include <nlohmann/detail/input/lexer.hpp>
```
gregmarr: Looks like these are sorted too.

include/nlohmann/json.hpp
```diff
@@ -893,6 +893,12 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
                 JSON_ASSERT(false); // NOLINT(cert-dcl03-c,hicpp-static-assert,misc-static-assert) LCOV_EXCL_LINE
         }
         JSON_ASSERT(m_data.m_type == val.type());
+
+#if JSON_DIAGNOSTIC_POSITIONS
+        start_position = val.start_position;
+        end_position = val.end_position;
```
gregmarr: @nlohmann This is a constructor, should this be in the member initialization list?  Nothing else in this constructor is in that list, but that seems to just be because they're complicated.

include/nlohmann/json.hpp
```diff
@@ -1205,6 +1211,11 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
                 break;
         }
 
+#if JSON_DIAGNOSTIC_POSITIONS
+        start_position = other.start_position;
+        end_position = other.end_position;
+#endif
```
gregmarr: There is a member initialization list in this one, so perhaps these should be there, though that will require having the commas before the elements:

```
#if JSON_DIAGNOSTIC_POSITIONS
    , start_position(other.start_position)
    , end_position(other.end_position)
#endif
```

include/nlohmann/json.hpp
sushshring: Compiler doesn’t like it being before m_data, any suggestions?

gregmarr: It would require reordering the members in the class definition, not just here in this constructor.  I think it's probably better to just put the commas first and align them with `m_data`.

nlohmann: Yes, it's ugly, but changing the order everywhere else is even uglier.

docs/mkdocs/docs/api/macros/json_diagnostic_positions.md
```diff
@@ -6,9 +6,13 @@
...
+When enabled, two new properties: `start_pos()` and `end_pos()` are added to `nlohmann::json` objects and fields. `start_pos()` returns the start position of
+that JSON object/field in the original string the object was parsed from. Likewise, `end_pos()` returns the end position of that JSON object/field in the
+original string the object was parsed from.
 
-For objects and arrays, the start and end positions represent the positions of the opening and closing braces or brackets, respectively. For fields, the start and end positions represent either the opening and closing quotes for that field's value or the first and character after last in the field's numerical or predefined true/false/null values.
+For objects and arrays, the start and end positions represent the positions of the opening and closing braces or brackets, respectively. For fields, the start
+and end positions represent either the opening and closing quotes for that field's value or the first and character after last in the field's numerical or
```
gregmarr: Now that I read this, this seems odd to me.  The end for array and object is the last character that represents the array or object, but for a value it is the character after the last?  Is that how it's implemented?  Is it intended to be implemented that way?

sushshring: Good point catching this. The documentation is slightly off - we always return one past the end of the substring. I will update the doc here and add a comment to the code as well.

docs/mkdocs/docs/api/macros/json_diagnostic_positions.md
```diff
@@ -0,0 +1,56 @@
+# JSON_DIAGNOSTIC_POSITIONS
+
+```cpp
+#define JSON_DIAGNOSTIC_POSITIONS /* value */
+```
+
+This macro enables position diagnostics for generated JSON objects.
+
+When enabled, two new properties: `start_pos()` and `end_pos()` are added to `nlohmann::json` objects and fields. `start_pos()` returns the start position of
```
nlohmann: Replace `nlohmann::json` with `nlohmann::basic_json`, because the properties are also available for `nlohmann::ordered_json`.

docs/mkdocs/docs/api/macros/json_diagnostic_positions.md
```diff
@@ -0,0 +1,56 @@
...
+that JSON object/field in the original string the object was parsed from. Likewise, `end_pos()` returns the end position of that JSON object/field in the
+original string the object was parsed from.
+
+For objects and arrays, the start and end positions represent the positions of the opening and closing braces or brackets, respectively. For fields, the start
+and end positions represent either the opening and closing quotes for that field's value or the first and character after last in the field's numerical or
+predefined true/false/null values.
+
+`start_pos()` and `end_pos()` are only set if the JSON object was parsed using `json::parse()`. For all other cases, `std::string::npos` will be returned.
```
nlohmann: Replace `json::parse()` with `parse()`.

docs/mkdocs/docs/api/macros/json_diagnostic_positions.md
```diff
@@ -0,0 +1,56 @@
...
+
+Note that enabling this macro increases the size of every JSON value by two size_t fields and adds
```
nlohmann: Replace size_t with `std::size_t` (with backticks).

nlohmann (changes requested):

docs/mkdocs/docs/api/macros/json_diagnostic_positions.md
sushshring: Updated

docs/mkdocs/docs/api/macros/json_diagnostic_positions.md
sushshring: Updated

docs/mkdocs/docs/api/macros/json_diagnostic_positions.md
sushshring: Updated

docs/mkdocs/docs/api/macros/json_diagnostic_positions.md
sushshring: Updated

docs/examples/diagnostic_positions.cpp
```diff
@@ -0,0 +1,51 @@
+#include <iostream>
+
+# define JSON_DIAGNOSTIC_POSITIONS 1
```
nlohmann:
```suggestion
#define JSON_DIAGNOSTIC_POSITIONS 1
```

docs/mkdocs/docs/api/macros/json_diagnostic_positions.md
```diff
@@ -0,0 +1,57 @@
...
+    ```
+    
+    Output:
+
+    ```
+    --8<-- "examples/diagnostic_positions.output"
+    ```
+
+    The output shows the start/end positions of all the objects and fields in the JSON string.
+## Version history
```
nlohmann, line 60:
```suggestion

## Version history
```

include/nlohmann/json.hpp
```diff
@@ -4225,6 +4249,23 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     basic_json* m_parent = nullptr;
 #endif
 
+#if JSON_DIAGNOSTIC_POSITIONS
+    /// the start position of the value
+    size_t start_position = std::string::npos;
```
nlohmann:
```suggestion
    std::size_t start_position = std::string::npos;
```

include/nlohmann/json.hpp
```diff
@@ -4225,6 +4249,23 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+    /// the end position of the value
+    size_t end_position = std::string::npos;
```
nlohmann:
```suggestion
    std::size_t end_position = std::string::npos;
```

include/nlohmann/json.hpp
```diff
@@ -4225,6 +4249,23 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+  public:
+    constexpr size_t start_pos() const noexcept
```
nlohmann:
```suggestion
    constexpr std::size_t start_pos() const noexcept
```

include/nlohmann/json.hpp
```diff
@@ -4225,6 +4249,23 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+    {
+        return start_position;
+    }
+
+    constexpr size_t end_pos() const noexcept
```
nlohmann:
```suggestion
    constexpr std::size_t end_pos() const noexcept
```

docs/examples/diagnostic_positions.cpp
sushshring: Updated

include/nlohmann/json.hpp
sushshring: Updated

docs/mkdocs/docs/api/macros/json_diagnostic_positions.md
sushshring: Updated

sushshring:
> ## 🔴 Amalgamation check failed! 🔴
> The source code has not been amalgamated. @sushshring Please read and follow the [Contribution Guidelines](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md#files-to-change).

Not sure why it's saying this, i've run amalgamate on the recent commit.

nlohmann:
> > ## 🔴 Amalgamation check failed! 🔴
> > The source code has not been amalgamated. @sushshring Please read and follow the [Contribution Guidelines](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md#files-to-change).
> 
> Not sure why it's saying this, i've run amalgamate on the recent commit.

The pipeline was quite busy, so maybe this was just delayed.

nlohmann:
Once the pipeline is through, please check why the coverage went down. Note you can download an artifact from the coverage job which contains HTML pages showing which lines are not covered. As always, coverage information is a bit fuzzy and sometimes you see the closing braces of functions in red which makes no sense. Nonetheless you should make sure every added code is covered by a test.

sushshring:
The coverage check is also odd. The commit here https://github.com/nlohmann/json/pull/4517/commits/8c67186a32f4cb7f36363c697a507fa1e35068bc has the same coverage, which seemed acceptable to it.

Regardless, I can add one more test that improves the coverage for json_type_t::discarded, but the remaining missing coverage check is for the default switch branch which has an assert(false) since it should never be hit.

nlohmann:
These lines can be skipped by adding `// LCOV_EXCL_LINE`.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -298,6 +336,79 @@ class json_sax_dom_parser
...
+                    // skip setting the values here.
+                    break;
+                }
+                // LCOV_EXCL_START
+                default:
+                {
+                    // Handle all possible types discretely, default handler should never be reached.
+                    JSON_ASSERT(false); // NOLINT(cert-dcl03-c,hicpp-static-assert,misc-static-assert,-warnings-as-errors)
+                }
+                    // LCOV_EXCL_STOP
```
gregmarr: This comment is indented one step too far.

sushshring:
@nlohmann looks like CI is all green. Any other blockers before this can be checked in?

include/nlohmann/detail/input/json_sax.hpp
sushshring: Amalgamate seems to be indenting this locally. Is there some way to skip the tidy command and only run the amalgamation for single include headers?

nlohmann (approved):
Looks good to me.

include/nlohmann/detail/input/json_sax.hpp
nlohmann: This is a limitation from Astyle. In other parts, we only added `// LCOV_EXCL_LINE` after the `default:` and the `JSON_ASSERT(false)` to avoid this. In any case - run `make amalgamate` to have the CI accept the code. If that does not work, delete the files in `single_include` to force a regeneration with `make amalgamate`.

nlohmann:
Please update to the latest `develop` branch which contains a fix for Clang-Tidy, see #4558.

include/nlohmann/detail/input/json_sax.hpp
sushshring: I will switch to using `// LCOV_EXCL_LINE` instead then

gregmarr (approved):

nlohmann:
Thanks a lot!

sushshring:
Woot woot 🎉!

nlohmann:
@sushshring I overworked the documentation of this feature a bit. Could you please have a look at

- https://github.com/nlohmann/json/blob/cleanup/docs/mkdocs/docs/api/macros/json_diagnostic_positions.md
- https://github.com/nlohmann/json/blob/cleanup/docs/mkdocs/docs/api/basic_json/start_pos.md
- https://github.com/nlohmann/json/blob/cleanup/docs/mkdocs/docs/api/basic_json/end_pos.md

If you have any comments, please add them to https://github.com/nlohmann/json/pull/4560. Thanks!

PR #4520, opened 2024-11-28
Adjust template parameters for ordered_map

### LINKED ISSUE #4518: `json_fwd.hpp`  don't define default template arguments for ordered_map
### Description

Unlike `json` and `ordered_json`, the default template arguments for `ordered_map` is not defined in `json_fwd.hpp` , so when compile failed with template arguments mismatch when use `json_fwd.hpp` with `ordered_map`

### Reproduction steps

compile the below example code 

the below code example compile failed

### Minimal code example

```Shell

void test(nlohmann::ordered_map<std::string, std::string>& val);
```

### Error messages

```Shell
CLANG
> error: too few template arguments for class template 'ordered_map'

GCC
> error: wrong number of template arguments (2, should be 4)
```

gcc/clang

### Library version

3.11.3

### Validation

nlohmann:
Fixes #4518

### DIFF
--- a/include/nlohmann/json_fwd.hpp
+++ b/include/nlohmann/json_fwd.hpp
@@ -10,9 +10,11 @@
 #define INCLUDE_NLOHMANN_JSON_FWD_HPP_
 
 #include <cstdint> // int64_t, uint64_t
+#include <functional> // less
 #include <map> // map
 #include <memory> // allocator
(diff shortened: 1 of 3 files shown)


nlohmann:
Oof. I should have tested this...

```
template parameter redefines default argument
```

errors all over the place.

Issue #4358, opened 2024-04-24
Unexpected conversion from 'true'/'false' to some integer  types


jboelterintc:
### Description

Converting from `"value": true` or `"value": false` behaves unexpectedly and differently for certain integer types.

I was expecting all conversions from a boolean true/false to integer value to fail, however it is stored as a 1 or 0.

### Reproduction steps

Given:
```cpp 
template<typename T>
class Value {
public:
    T value;
private:
    NLOHMANN_DEFINE_TYPE_INTRUSIVE(Value<T>, value);
};
```

This will result in a `1` in `value`.

```cpp
auto v = R"({
    "value" : true
})"_json.get<Value<int>>();
```

Using `uint64_t` will throw a `type_error`.

```cpp
auto v = R"({
    "value" : true
})"_json.get<Value<uint64_t>>();
```

### Expected vs. actual results

I expected all boolean to integer conversions to fail.  Instead a `true` -> `1` and `false` -> `0` for certain integer values.

Various integer conversions - https://godbolt.org/z/7Wrh6EanW

The `uint64_t` case is taking the path through `get_arithmetic_value`

```cpp
template<typename BasicJsonType>
inline void from_json(const BasicJsonType& j, typename BasicJsonType::number_unsigned_t& val)
{
    get_arithmetic_value(j, val);
}
```

The `int` case is taking the path through 

```cpp

struct from_json_fn
{
    template<typename BasicJsonType, typename T>
    auto operator()(const BasicJsonType& j, T&& val) const
    noexcept(noexcept(from_json(j, std::forward<T>(val))))
    -> decltype(from_json(j, std::forward<T>(val)))
    {
        return from_json(j, std::forward<T>(val));
    }
};

// which calls

template < typename BasicJsonType, typename ArithmeticType,
           enable_if_t <
               std::is_arithmetic<ArithmeticType>::value&&
               !std::is_same<ArithmeticType, typename BasicJsonType::number_unsigned_t>::value&&
               !std::is_same<ArithmeticType, typename BasicJsonType::number_integer_t>::value&&
               !std::is_same<ArithmeticType, typename BasicJsonType::number_float_t>::value&&
               !std::is_same<ArithmeticType, typename BasicJsonType::boolean_t>::value,
               int > = 0 >
inline void from_json(const BasicJsonType& j, ArithmeticType& val)
{
//...

        case value_t::boolean:
        {
[...]

PR #4523, opened 2024-11-30
Disallowing implicit conversion from boolean type to some integers type (#4358)

arifghaz:
## Abstract
Referring to issue #4358 , this pull request make some changes that causes implicit conversions from various integer types (listed here - https://godbolt.org/z/7Wrh6EanW) to boolean to fail. Now, instead of inconsistence behavior for different integer types, all implicit conversion from boolean to different integer types will throw a type error.

## Changes Proposed
Remove explicit overload for boolean_t from the from_json function (line 343 from_json.hpp). This will allow all implicit conversions from boolean to integers to fail.

## Possible Issues
This removal will causes two unit tests in unit-conversions.cpp to fail, but both unit tests involve implicit conversion from boolean to uint8_t (line 556-560). Based on issue #4358 , the failure of these two unit tests are expected. Removal of these tests allow the changes to pass all the remaining test cases.

## Validation
Unit test will be added if the propose change is accepted.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -203,7 +203,7 @@ inline void from_json(const BasicJsonType& j, std::valarray<T>& l)
 }
 
 template<typename BasicJsonType, typename T, std::size_t N>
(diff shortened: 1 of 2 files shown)


nlohmann (changes requested):
This change breaks the public API - existing code (such as the tests) must not be affected by the change. For this issue, a preprocessor macro would be needed. Not sure if this is the best approach. This should be discussed in #4358.

PR #4528, opened 2024-12-06
Allow ordered_json in NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE macro

ArashPartow:
Allow for `ordered_json `type to be used with the define macros provided in `macro_scope.hpp`.

Usage:

```
#include <cstdio>

#include <nlohmann/json.hpp>

struct mystruct
{
    int x = 1;
    double y = 2.0;
    std::string z = "three";
};

NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE
(
    mystruct,
    z,
    y,
    x
);

int main()
{
    mystruct ms;

    const auto msjson = nlohmann::to_string(nlohmann::ordered_json{{ "mystruct", ms }});

    const std::string expected = R"({"mystruct":{"z":"three","y":2.0,"x":1}})";

    printf("%s\n",msjson  .c_str());
    printf("%s\n",expected.c_str());

    return 0;
}
```

### DIFF
--- a/include/nlohmann/detail/macro_scope.hpp
+++ b/include/nlohmann/detail/macro_scope.hpp
@@ -420,14 +420,18 @@
 */
 #define NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(Type, ...)  \
     inline void to_json(nlohmann::json& nlohmann_json_j, const Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__)) } \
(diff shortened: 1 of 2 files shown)


gregmarr:
This version in enum handling is what they should really be:

```
```

gregmarr:
See also #2843 and #2532

nlohmann (changes requested):
I agree with https://github.com/nlohmann/json/pull/4528#issuecomment-2523699499 - a templated version would be a cleaner solution. Especially if more specializations are added such as #4517.

ArashPartow:
@nlohmann Happy for you to close this PR, as https://github.com/nlohmann/json/pull/4517 might be the more appropriate PR to make the necessary changes.

nlohmann:
Closed in favor of #4597.

PR #4555, opened 2024-12-18
Add keep parameter to dump to copy invalid UTF-8 bytes as-is

### LINKED ISSUE #4552: UTF-8 invalid characters are not always ignored when dumping with error_handler_t::ignore
### Description

According to this: https://json.nlohmann.me/api/basic_json/dump/#parameters , when passing `error_handler_t::ignore` to `dump()` function, invalid UTF-8 characters should be ignored and copied as-is into the final string.

However, I'm debugging the following minimal code:
```cpp
    std::string test = "test\334\005";
    nlohmann::json node{};
    node["test"] = test;
    auto test_dump = node.dump(-1, ' ', false, nlohmann::json::error_handler_t::ignore);
```

and the final `test_dump` string contains `test\005` (byte `\334` is gone).

image

Is this expected? Am I missing something?

### Reproduction steps

Just try to run/debug the following:

```cpp
    std::string test = "test\334\005";
    nlohmann::json node{};
    node["test"] = test;
    auto test_dump = node.dump(-1, ' ', false, nlohmann::json::error_handler_t::ignore);
```

Actual: `test_dump` contains `test\005`
Expected: `test_dump` contains `test\334\005`

### Minimal code example

```Shell
std::string test = "test\334\005";
nlohmann::json node{};
node["test"] = test;
auto test_dump = node.dump(-1, ' ', false, nlohmann::json::error_handler_t::ignore);
```

### Error messages

_No response_

gcc (Alpine 12.2.1_git20220924-r10) 12.2.1 20220924

### Library version

3.11.2

### Validation

nlohmann:
Add an enumerator `keep` to `error_handler_t` to allow to keep the input as-is in case of UTF-8 errors.

Fixes #4552


jordan-hoang:
It looks good to me. I thought you would do json_string === s_kept, but you took it's substring at position 1 instead, I guess dump gives you an extra character in the beginning or something?

nlohmann:
Yes, dump adds quotes.

docs/mkdocs/docs/api/basic_json/dump.md
```diff
@@ -25,10 +25,11 @@ and `ensure_ascii` parameters.
     result consists of ASCII characters only.
 
 `error_handler` (in)
-:   how to react on decoding errors; there are three possible values (see [`error_handler_t`](error_handler_t.md):
-    `strict` (throws and exception in case a decoding error occurs; default), `replace` (replace invalid UTF-8 sequences
-    with U+FFFD), and `ignore` (ignore invalid UTF-8 sequences during serialization; all bytes are copied to the output
-    unchanged)).
+:   how to react on decoding errors; there are three possible values (see [`error_handler_t`](error_handler_t.md)):
```
gentooise:
```suggestion
:   how to react on decoding errors; there are four possible values (see [`error_handler_t`](error_handler_t.md)):
```

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -398,6 +399,13 @@ class serializer
         std::size_t bytes_after_last_accept = 0;
         std::size_t undumped_chars = 0;
 
+        // copy string as-is if error handler is set to keep, and we don't want to ensure ASCII
+        if (error_handler == error_handler_t::keep && !ensure_ascii)
+        {
+            o->write_characters(s.data(), s.size());
```
gentooise, line 405: Just for me to understand, how would this behave exactly? If there is a `0×dc` byte for example, it will be escaped as `\334` octal string (or `\xdc` hex, or similar)?

I think the important thing is to not break the json format.

And also what about other UTF-8 accepted chars? Like `\b` or `\t` handled below: how will they be dumped in this case?

I have limited access these days (from mobile), and I don't know exactly the purpose of `ensure_ascii` and its default value. If you could provide some hint it would be helpful.

Thank you

nlohmann: Oh, you're right! Just copying the input to the output is wrong here, because valid characters like LF that must be escaped to `\n` would not be escaped and the resulting JSON would be invalid. Thanks for noting. I will fix this.

gregmarr:
@nlohmann Does this still need work?

nlohmann:
Yes. And clarification.

PR #4558, opened 2024-12-18
Suppress modernize-use-integer-sign-comparison

nlohmann:
The [`modernize-use-integer-sign-comparison`](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-integer-sign-comparison.html) was added to Clang-Tidy and is only applicable to C++20 code. Therefore, we suppress this check.


gregmarr:
Is this showing a lot of hits?  Should we create a helper function for pre-c++20 that makes those comparisons safer?

nlohmann:
> Is this showing a lot of hits? Should we create a helper function for pre-c++20 that makes those comparisons safer?

See https://github.com/nlohmann/json/actions/runs/12396405659/job/34604358473?pr=4557.

The first hit is this:

```
/__w/json/json/include/nlohmann/detail/input/binary_reader.hpp:1112:13: error: comparison between 'signed' and 'unsigned' integers [modernize-use-integer-sign-comparison,-warnings-as-errors]
 1112 |         if (len != static_cast<std::size_t>(-1))
      |             ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~   
      |             std::cmp_not_equal( ,              )
```

`len` is of type `std::size_t` here. Not sure what to make out of it.

Yes, one approach could be to have our own implementation of cmp_equal, cmp_not_equal, cmp_less, cmp_greater, cmp_less_equal, and cmp_greater_equal and use it throughout. I would still like to merge this first, because all pipelines will fail now without suppressing this issue.

Are you OK with this and having an issue to fix this in the future?

gregmarr:
Yeah, since all the current usages are safe, and the compiler is apparently looking right through the cast that we're using to make it "safe", then I think it's fine to log it for the future.  Maybe we can use `std::numeric_limits<std::size_t>::max()` instead of `static_cast<size_t>(-1)`.

Issue #4559, opened 2024-12-18
Modernize integer comparison


nlohmann:
Yeah, since all the current usages are safe, and the compiler is apparently looking right through the cast that we're using to make it "safe", then I think it's fine to log it for the future.  Maybe we can use `std::numeric_limits<std::size_t>::max()` instead of `static_cast<size_t>(-1)`.

_Originally posted by @gregmarr in https://github.com/nlohmann/json/issues/4558#issuecomment-2551796276_

nlohmann:
See #4558 for more information.

shivenducs1136:
Hi @nlohmann I want to take this up!

nlohmann:
Go ahead! Let me know if you need anything.

PR #4577, opened 2024-12-30
Modernize integer comparison

codenut:
Replace static_cast<size_t>(-1) with std::numeric_limits<std::size_t>::max() as suggested in https://github.com/nlohmann/json/issues/4559

Builds are failing in MSVS because of  `<Windows.h>` header file that includes macro definitions named max and min:
```
#define max(a,b) (((a) > (b)) ? (a) : (b))
```

one way to solve this issue is to wrap `max` with parenthesis. 

Source: https://stackoverflow.com/questions/27442885/syntax-error-with-stdnumeric-limitsmax

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -172,7 +172,7 @@ class binary_reader
         std::int32_t document_size{};
         get_number<std::int32_t, true>(input_format_t::bson, document_size);
 
-        if (JSON_HEDLEY_UNLIKELY(!sax->start_object(static_cast<std::size_t>(-1))))
+        if (JSON_HEDLEY_UNLIKELY(!sax->start_object(detail::unknown_size())))
         {
             return false;
         }
@@ -394,7 +394,7 @@ class binary_reader
         std::int32_t document_size{};
         get_number<std::int32_t, true>(input_format_t::bson, document_size);
 
-        if (JSON_HEDLEY_UNLIKELY(!sax->start_array(static_cast<std::size_t>(-1))))
+        if (JSON_HEDLEY_UNLIKELY(!sax->start_array(detail::unknown_size())))
         {
             return false;
         }
@@ -654,7 +654,7 @@ class binary_reader
             }
 
             case 0x9F: // array (indefinite length)
-                return get_cbor_array(static_cast<std::size_t>(-1), tag_handler);
+                return get_cbor_array(detail::unknown_size(), tag_handler);
 
             // map (0x00..0x17 pairs of data items follow)
             case 0xA0:
@@ -708,7 +708,7 @@ class binary_reader
             }
 
             case 0xBF: // map (indefinite length)
-                return get_cbor_object(static_cast<std::size_t>(-1), tag_handler);
+                return get_cbor_object(detail::unknown_size(), tag_handler);
 
             case 0xC6: // tagged item
             case 0xC7:
@@ -1096,7 +1096,7 @@ class binary_reader
     }
 
     /*!
-    @param[in] len  the length of the array or static_cast<std::size_t>(-1) for an
+    @param[in] len  the length of the array or detail::unknown_size() for an
                     array of indefinite size
     @param[in] tag_handler how CBOR tags should be treated
     @return whether array creation completed
@@ -1109,7 +1109,7 @@ class binary_reader
             return false;
         }
 
-        if (len != static_cast<std::size_t>(-1))
+        if (len != detail::unknown_size())
         {
             for (std::size_t i = 0; i < len; ++i)
             {
@@ -1134,7 +1134,7 @@ class binary_reader
     }
 
     /*!
-    @param[in] len  the length of the object or static_cast<std::size_t>(-1) for an
+    @param[in] len  the length of the object or detail::unknown_size() for an
                     object of indefinite size
     @param[in] tag_handler how CBOR tags should be treated
     @return whether object creation completed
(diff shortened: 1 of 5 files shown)


nlohmann:
> Builds are failing in MSVS because of `<Windows.h>` header file that includes macro definitions named max and min:
> 
> ```
> #define max(a,b) (((a) > (b)) ? (a) : (b))
> ```
> 
> one way to solve this issue is to wrap `max` with parenthesis.
> 
> Source: [stackoverflow.com/questions/27442885/syntax-error-with-stdnumeric-limitsmax](https://stackoverflow.com/questions/27442885/syntax-error-with-stdnumeric-limitsmax)

Thanks, this is the way we do it in the other parts of the library. The failing AppVeyor test was explicitly added to detect these problems.

nlohmann (changes requested):
Looking at the diff, it becomes obvious how frequent we use `-1` as magic number for "unknown size". As it is mostly used in the context of the SAX interface, it could be a good idea to define a constant `unknown_size` in that header and use this throughout the code.

nlohmann:
Note this PR only addresses one aspect of #4559 - the bigger part is about using `cmp_not_equal` and the likes as #4558.

codenut:
> Note this PR only addresses one aspect of #4559 - the bigger part is about using `cmp_not_equal` and the likes as #4558.

Thank you for reviewing my PR. ~~Would it be possible to address your feedback in a separate PR? I will create a new PR and issue for that purpose.~~

Edit: Sorry, I misunderstood your comment. I will make the changes as suggested by @gregmarr

codenut:
> Note this PR only addresses one aspect of #4559 - the bigger part is about using `cmp_not_equal` and the likes as #4558.

Do you think implementing the comparison helper functions is a "good first issue"? If it is, I am interested in working on that too.

gregmarr:
> As it is mostly used in the context of the SAX interface, it could be a good idea to define a constant unknown_size in that header and use this throughout the code.

It's used in the parser, and in the main class in the `set_parent` function, so I don't think burying it in the SAX interface is a good idea.  Also, having it as a constant, even a constexpr constant, is potentially opening up storage and linkage issues, so it's probably better to just have it as a `constexpr` function in the `detail` namespace.

```
NLOHMANN_JSON_NAMESPACE_BEGIN
namespace detail
{
constexpr std::size_t unknown_size() { return (std::numeric_limits<std::size_t>::max)(); }
////////////
// parser //
////////////
```

gregmarr:
> > Note this PR only addresses one aspect of #4559 - the bigger part is about using `cmp_not_equal` and the likes as #4558.
> 
> Do you think implementing the comparison helper functions is a "good first issue"? If it is, I am interested in working on that too.

It depends.  If by "first issue" you mean first in this repository, but you have significant experience in template metaprogramming and SFINAE, then maybe.  If you mean a good issue for an inexperienced programmer, then no.

The `cmp_not_equal` function requires C++20 or later.  This would require writing a helper function for C++11 through C++17.  The "possible implementation" of `cmp_not_equal` shown at cppreference.com uses `if constexpr`, which is only available in C++17 and later.  So there would be one version with `if constexpr` for C++17, and one set of functions with SFINAE for C++11 and C++14.

However, I think that using `max` instead of `std::size_t(-1)` will actually eliminate the need for `cmp_not_equal`, as that was a way to eliminate the warnings with the signed and unsigned comparisons, which will no longer happen because we're using an alternate method to get the value.

nlohmann:
If this PR is sufficient to fix the Clang-Tidy warning `modernize-use-integer-sign-comparison´, then I am more than happy to leave the `cmp_not_equal` alone.

nlohmann:
> > As it is mostly used in the context of the SAX interface, it could be a good idea to define a constant unknown_size in that header and use this throughout the code.
> 
> It's used in the parser, and in the main class in the `set_parent` function, so I don't think burying it in the SAX interface is a good idea. Also, having it as a constant, even a constexpr constant, is potentially opening up storage and linkage issues, so it's probably better to just have it as a `constexpr` function in the `detail` namespace.
> 
> ```
> NLOHMANN_JSON_NAMESPACE_BEGIN
> namespace detail
> {
> constexpr std::size_t unknown_size() { return (std::numeric_limits<std::size_t>::max)(); }
> ////////////
> // parser //
> ////////////
> ```

`set_parent()` is never called with the `size_t(-1)` argument explicitly. It is the default argument. So it would be sufficient to change line

```cpp
reference set_parent(reference j, std::size_t old_capacity = static_cast<std::size_t>(-1))
```

to

```cpp
reference set_parent(reference j, std::size_t old_capacity = std::numeric_limits<std::size_t>::max()))
```

with a respective comment about this value.

Then we have the following appearances of `static_cast<std::size_t>(-1)`:

- binary_reader.hpp: in `get_cbor_array` and `get_cbor_object`
- binary_reader.hpp: calling the SAX interface for `start_array`/`start_object`
- json_sax.hpp: checking the `len` value in `start_array`/`start_object`
- json_sax.hpp: default parameter in `json_sax_acceptor`
- parser.hpp: calling the SAX interface for `start_array`/`start_object`

Interestingly, we also have in binary_reader:

```cpp
static JSON_INLINE_VARIABLE constexpr std::size_t npos = static_cast<std::size_t>(-1);
```

which is used at some places.

* * *

While this comment should make the case for moving the value closer to the parser, I think I over-engineered this in my head. So I think having a `detail::unknown_size()` function is sufficient to get rid of the magic values.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

codenut:
> Thanks!

Thank you for your thorough feedback. I look forward to collaborating on more "first good issue" changes!

nlohmann:
You could help by leaving a review at #4581.

PR #4581, opened 2025-01-01
Enable modernize-use-integer-sign-comparison check

### LINKED ISSUE #4559: Modernize integer comparison

nlohmann:
Close #4559 after merging #4577.

Background: A new version of Clang-Tidy added a `modernize-use-integer-sign-comparison` check which failed on the code - mostly, because conversions like `static_cast<std::size_t>(-1)`. These have been replaced with `std::numeric_limits<std::size_t>::max()` in #4577. This PR en-enabled the Clang-Tidy check and fixes some test files.


tests/src/unit-alt-string.cpp
```diff
@@ -28,7 +28,7 @@ class alt_string
   public:
     using value_type = std::string::value_type;
 
-    static constexpr auto npos = static_cast<std::size_t>(-1);
+    static constexpr auto npos = (std::numeric_limits<std::size_t>::max)();
```
codenut, line 31: I think we can also use the `unknown_size()` function here and on there usages of `std::numeric_limits<std::size_t>::max`

i.e.

`nlohmann::detail::unknown_size()`

nlohmann: Yes, but I don't want to rely too much on content from the `detail` namespace as it is allowed to be changed any time.

Issue #4561, opened 2024-12-19
Use diagnostic positions in exceptions


nlohmann:
We recently merged #4517 which added a macro `JSON_DIAGNOSTIC_POSITIONS` that, when defined, adds two `std::size_t` members to JSON values indicating the position of the first and last byte of this element in the input text. We should use this information to provide better diagnostic messages in exceptions.

harytary:
@nlohmann I have been looking into this, I wonder if the below change makes sense for this issue
```
diff --git a/include/nlohmann/detail/exceptions.hpp b/include/nlohmann/detail/exceptions.hpp
index 1fabeb16..6cc85221 100644
--- a/include/nlohmann/detail/exceptions.hpp
+++ b/include/nlohmann/detail/exceptions.hpp
@@ -132,6 +132,8 @@ class exception : public std::exception
             return concat(a, '/', detail::escape(b));
         });
         return concat('(', str, ") ");
+#elif JSON_DIAGNOSTICS_POSITION
+        return concat(" (at start byte: ", std::to_string(leaf_element->start_pos()), "; last read: byte ", std::to_string(leaf_element->end_pos()), ") ");        
 #else
         static_cast<void>(leaf_element);
         return "";

```

hnampally:
@nlohmann I have been looking into this, I was wondering if Json_Diagnostic_Positions and Json_Diagnostics options are mutually exclusive, also please provide some hints for this task.

nlohmann:
No, the two options are not mutually exclusive:

- `JSON_DIAGNOSTICS` allows to add a JSON Pointer to the exception messages that can help to locate the part of the JSON that lead to the error. See https://json.nlohmann.me/api/macros/json_diagnostics/ for more information.
- `JSON_DIAGNOSTIC_POSITIONS` allows to query for each parse JSON value the byte positions in the input (see https://json.nlohmann.me/api/macros/json_diagnostic_positions/). The goal of this issue is to add this to the exception messages.

So an example for the default behavior would be:

```
[json.exception.type_error.302] type must be number, but is string
```

for `JSON_DIAGNOSTICS`, this can be:

```
[json.exception.type_error.302] (/address/housenumber) type must be number, but is string
```

Now for `JSON_DIAGNOSTIC_POSITIONS` we can query the position in the input and could do something like this:

```
[json.exception.type_error.302] (byte 34-39) type must be number, but is string
```

And if both options are set, we could have both:

```
[...]

PR #4583, opened 2025-01-05
Issue #4561 -  use diagnostic positions in exceptions

hnampally:
This PR is for #4561, updates the diagnostics function in the exceptions.hpp to use diagnostic positions in exceptions.

* * *

## Pull request checklist


docs/examples/diagnostic_positions.cpp
```diff
@@ -5,6 +5,30 @@
 
 using json = nlohmann::json;
 
+/* Demonstration of type error exception with diagnostic postions support enabled */
```
nlohmann: Please do not edit existing examples, but add new ones. Each example should demonstrate one and only one aspect of the library.

docs/examples/diagnostics_extended.cpp
```diff
@@ -1,15 +1,22 @@
 #include <iostream>
 
 # define JSON_DIAGNOSTICS 1
+# define JSON_DIAGNOSTIC_POSITIONS 1
```
nlohmann: See above. Please also add one example where both `JSON_DIAGNOSTICS` and `JSON_DIAGNOSTIC_POSITIONS` are defined.

include/nlohmann/detail/exceptions.hpp
```diff
@@ -131,6 +130,12 @@ class exception : public std::exception
         {
             return concat(a, '/', detail::escape(b));
         });
+#if JSON_DIAGNOSTIC_POSITIONS
+        str += ", byte " + std::to_string(leaf_element->start_pos()) + "-" + std::to_string(leaf_element->end_pos());
```
nlohmann: Note `start_pos()` and `end_pos()` may return `std::string::npos` if the position is not known. In these cases, do not add a byte range.

nlohmann (changes requested):
Please also check the failing test. Just as for `JSON_DIAGNOSTICS`, you may need to add `#ifdef`s to the other test files to accommodate for the different exception outputs.

docs/examples/diagnostic_positions.cpp
```diff
@@ -48,4 +48,4 @@ int main()
     std::cout << "1" << "\n";
     std::cout << "Parsed string: \n";
     std::cout << json_string.substr(j["address"]["housenumber"].start_pos(), j["address"]["housenumber"].end_pos() - j["address"]["housenumber"].start_pos()) << "\n\n";
-}
+}
```
nlohmann, line 51: Please revert this file.

docs/examples/diagnostic_positions.output
```diff
@@ -47,4 +47,3 @@ Original string:
 1
 Parsed string: 
 1
-
```
nlohmann, line 50: Please revert this file.

docs/examples/diagnostics_extended.cpp
```diff
@@ -19,4 +19,4 @@ int main()
     {
         std::cout << e.what() << '\n';
     }
-}
+}
```
nlohmann, line 22: Please revert this file.

docs/examples/diagnostics_extended.output
```diff
@@ -1 +1 @@
-[json.exception.type_error.302] (/address/housenumber) type must be number, but is string
+[json.exception.type_error.302] (/address/housenumber) type must be number, but is string
```
nlohmann, line 1: Please revert this file.

include/nlohmann/detail/exceptions.hpp
```diff
@@ -131,7 +130,20 @@ class exception : public std::exception
...
             return concat(a, '/', detail::escape(b));
         });
-        return concat('(', str, ") ");
+#if JSON_DIAGNOSTIC_POSITIONS
+        if ((leaf_element->start_pos() != std::string::npos) && (leaf_element->end_pos() != std::string::npos))
+        {
+            str += ", byte " + std::to_string(leaf_element->start_pos()) + "-" + std::to_string(leaf_element->end_pos());
+            return concat('(', str, ") ");
+        }
+#endif
```
nlohmann, line 139: This logic looks weird:

```cpp
#if JSON_DIAGNOSTIC_POSITIONS
#endif
elif JSON_DIAGNOSTIC_POSITIONS
```

tests/src/unit-json_patch.cpp
```diff
@@ -60,14 +60,20 @@ TEST_CASE("JSON patch")
             json const doc2 = R"({ "q": { "bar": 2 } })"_json;
 
             // because "a" does not exist.
+#if defined(JSON_DIAGNOSTIC_POSITIONS)
```
nlohmann, line 63: Due to https://github.com/nlohmann/json/blob/60c48755e3f717eace7830d7bbc0d8f1a5e0cc8a/include/nlohmann/detail/abi_macros.hpp#L29-L31, `JSON_DIAGNOSTIC_POSITIONS` will always be defined. Either to `0` or `1`. So checking for `defined(JSON_DIAGNOSTIC_POSITIONS)` will always be true.

nlohmann (changes requested):

include/nlohmann/detail/exceptions.hpp
```diff
@@ -131,7 +130,20 @@ class exception : public std::exception
         {
             return concat(a, '/', detail::escape(b));
         });
-        return concat('(', str, ") ");
+#if JSON_DIAGNOSTIC_POSITIONS
+        if ((leaf_element->start_pos() != std::string::npos) && (leaf_element->end_pos() != std::string::npos))
```
codenut, line 134: Line 134-139 appears to be similar to 140-146. Should we consider extracting it into a helper function for better code reusability? Like below..

```
std::string get_byte_positions(const BasicJsonType* leaf_element)
{
#if JSON_DIAGNOSTIC_POSITIONS
    if ((leaf_element->start_pos() != std::string::npos) && (leaf_element->end_pos() != std::string::npos))
    {
        return concat('(', "byte ", std::to_string(leaf_element->start_pos()), "-", std::to_string(leaf_element->end_pos()), ") ");
    }
    return "";
#else
    static_cast<void>(leaf_element);
    return "";
#endif
}
```

include/nlohmann/detail/exceptions.hpp
gregmarr: Yes, a line was lost, and it should still be between the `#endif` and the `#elif`:

```
        return concat('(', str, ") ");
```

Actually, that should just be moved out of the `if`.

include/nlohmann/detail/exceptions.hpp
```diff
@@ -131,7 +130,20 @@ class exception : public std::exception
...
+        {
+            str += ", byte " + std::to_string(leaf_element->start_pos()) + "-" + std::to_string(leaf_element->end_pos());
+            return concat('(', str, ") ");
+        }
+#endif
+#elif JSON_DIAGNOSTIC_POSITIONS
+        if ((leaf_element->start_pos() != std::string::npos) && (leaf_element->end_pos() != std::string::npos))
+        {
+            auto str = concat("byte ", std::to_string(leaf_element->start_pos()), "-", std::to_string(leaf_element->end_pos()));
+            return concat('(', str, ") ");
```
gregmarr, line 144: Shouldn't need two separate `concat()` calls here.

include/nlohmann/detail/exceptions.hpp
```diff
@@ -131,7 +130,20 @@ class exception : public std::exception
         {
             return concat(a, '/', detail::escape(b));
         });
-        return concat('(', str, ") ");
+#if JSON_DIAGNOSTIC_POSITIONS
+        if ((leaf_element->start_pos() != std::string::npos) && (leaf_element->end_pos() != std::string::npos))
+        {
+            str += ", byte " + std::to_string(leaf_element->start_pos()) + "-" + std::to_string(leaf_element->end_pos());
```
gregmarr, line 136: Should these be `bytes` instead of `byte`?

tests/src/unit-json_patch.cpp
hnampally: I understand, unfortunately my workspace got messed up and I was unable to test it thoroughly, I didn't intend to push these changes, anyway I will set up a fresh workspace, I wonder if its okay to close this PR and open a new PR with fresh branch?

hnampally:
I am closing this PR and creating a new one.

include/nlohmann/detail/exceptions.hpp
hnampally: thanks for the suggestion, yes I will do this, it looks like a good idea to me.

PR #4585, opened 2025-01-07
use diagnostic positions in exceptions

### LINKED ISSUE #4561: Use diagnostic positions in exceptions

hnampally:
This PR is for https://github.com/nlohmann/json/issues/4561, updates the diagnostics function in the exceptions.hpp to use diagnostic positions in exceptions.

* * *

## Pull request checklist


docs/examples/diagnostics_extended_positions.cpp
```diff
@@ -0,0 +1,30 @@
...
+    json j = json::parse(json_invalid_string);
+    try
+    {
+        int housenumber = j["address"]["housenumber"];
+    }
+    catch (const json::exception& e)
+    {
+        std::cout << e.what() << '\n';
+    }
+}
```
nlohmann: Please add a new line to the end of the file.

docs/mkdocs/docs/api/macros/json_diagnostic_positions.md
```diff
@@ -54,8 +54,33 @@ When the macro is not defined, the library will define it to its default value.
     ```
     --8<-- "examples/diagnostic_positions.output"
     ```
-
     The output shows the start/end positions of all the objects and fields in the JSON string.
 
+??? example "Example 2: using extended diagnostics with positions enabled in exceptions"
```
nlohmann: Please move this example down.

nlohmann: Please also add this example to docs/mkdocs/docs/api/macros/json_diagnostics.md

include/nlohmann/detail/exceptions.hpp
```diff
@@ -131,16 +131,30 @@ class exception : public std::exception
...
-        static_cast<void>(leaf_element);
-        return "";
+        return get_byte_positions(leaf_element);
 #endif
     }
 
   private:
     /// an exception object as storage for error messages
     std::runtime_error m;
+    template<typename BasicJsonType>
```
nlohmann, line 145: The coverage report says this code is not covered:

image

nlohmann (changes requested):

include/nlohmann/detail/exceptions.hpp
```diff
@@ -131,16 +131,30 @@ class exception : public std::exception
         {
             return concat(a, '/', detail::escape(b));
         });
-        return concat('(', str, ") ");
+
+        return concat('(', str, ") ") + get_byte_positions(leaf_element);
```
codenut: Nit: you can also include the `get_byte_positions...` in the `concat` call instead of concatenating with `+` operator. i.e. `concat('(', str, ") ", get_byte_positions(leaf_element))`

codenut: Aside from readability, I believe using `concat` also has a performance benefit.

include/nlohmann/detail/exceptions.hpp
```diff
@@ -131,16 +131,30 @@ class exception : public std::exception
...
+        if ((leaf_element->start_pos() != std::string::npos) && (leaf_element->end_pos() != std::string::npos))
+        {
+            return concat('(', "bytes ", std::to_string(leaf_element->start_pos()), "-", std::to_string(leaf_element->end_pos()), ") ");
+        }
+        return "";
+#else
+        static_cast<void>(leaf_element);
+        return "";
+#endif
+    }
```
gregmarr, line 160: You can merge `'(', "bytes "` into `"(bytes "`.

@nlohmann Is there a reason that the existing function uses `return "";` over `return {};`?

I wonder if this would be clearer like this:

```
   static std::string get_byte_positions(const BasicJsonType* leaf_element)
    {
#if JSON_DIAGNOSTIC_POSITIONS
        if ((leaf_element->start_pos() != std::string::npos) && (leaf_element->end_pos() != std::string::npos))
        {
            return concat("(bytes ", std::to_string(leaf_element->start_pos()), "-", std::to_string(leaf_element->end_pos()), ") ");
        }
#else
        static_cast<void>(leaf_element);
#endif
        return {};
    }
```
or like this:
```
#if JSON_DIAGNOSTIC_POSITIONS
    static std::string get_byte_positions(const BasicJsonType* leaf_element)
    {
        if ((leaf_element->start_pos() != std::string::npos) && (leaf_element->end_pos() != std::string::npos))
        {
            return concat("(bytes ", std::to_string(leaf_element->start_pos()), "-", std::to_string(leaf_element->end_pos()), ") ");
        }
        return {};
    }
#else
    static std::string get_byte_positions(const BasicJsonType*)
    {
        return {};
    }
#endif
```

nlohmann: I personally think returning "" as empty string is clearer than {}.

include/nlohmann/detail/exceptions.hpp
gregmarr: The coverage report doesn't build with `JSON_DIAGNOSTIC_POSITIONS` so there needs to be a file like `tests/src/unit-diagnostics.cpp` that always builds with it to get it into the coverage report.

tests/src/unit-json_patch.cpp
```diff
@@ -60,14 +60,20 @@ TEST_CASE("JSON patch")
...
             CHECK_THROWS_WITH_AS(doc2.patch(patch1), "[json.exception.out_of_range.403] key 'a' not found", json::out_of_range&);
+#endif
 
             json const doc3 = R"({ "a": {} })"_json;
             json const patch2 = R"([{ "op": "add", "path": "/a/b/c", "value": 1 }])"_json;
 
             // should cause an error because "b" does not exist in doc3
 #if JSON_DIAGNOSTICS
             CHECK_THROWS_WITH_AS(doc3.patch(patch2), "[json.exception.out_of_range.403] (/a) key 'b' not found", json::out_of_range&);
+#elif JSON_DIAGNOSTIC_POSITIONS
```
gregmarr, line 75: Should these tests be run with both defined as well?

docs/examples/diagnostic_positions_exception.cpp
```diff
@@ -0,0 +1,29 @@
...
+        "address": {
+            "street": "Fake Street",
+            "housenumber": "1"
+        }
+    }
+    )";
+    json j = json::parse(json_invalid_string);
+    try
+    {
+        int housenumber = j["address"]["housenumber"];
```
gregmarr: @nlohmann Does it make sense to run Codacy on the examples?  If so, should this just remove the assignment?

docs/examples/diagnostics_extended_positions.cpp
```diff
@@ -0,0 +1,30 @@
...
+        "address": {
+            "street": "Fake Street",
+            "housenumber": "1"
+        }
+    }
+    )";
+    json j = json::parse(json_invalid_string);
+    try
+    {
+        int housenumber = j["address"]["housenumber"];
```
gregmarr: As above.

include/nlohmann/detail/exceptions.hpp
hnampally: @gregmarr @nlohmann thanks for your suggestions and inputs, this "(bytes " is better and your second approach is easier to read, I will go with below snippet if you guys are okay with it
```
#if JSON_DIAGNOSTIC_POSITIONS
    static std::string get_byte_positions(const BasicJsonType* leaf_element)
    {
        if ((leaf_element->start_pos() != std::string::npos) && (leaf_element->end_pos() != std::string::npos))
        {
            return concat("(bytes ", std::to_string(leaf_element->start_pos()), "-", std::to_string(leaf_element->end_pos()), ") ");
        }
        return "";
    }
#else
    static std::string get_byte_positions(const BasicJsonType*)
    {
        return "";
    }
#endif
```

tests/src/unit-json_patch.cpp
hnampally: @gregmarr I thought about that but I am not sure if there is ci_test target with both defined, @nlohmann please let me know what you think about this?

docs/examples/diagnostic_positions_exception.cpp
nlohmann: I have no idea if I can deactivate Codacy for specific folders. I'd rather see this issue fixed - maybe by `std::cout << j["address"]["housenumber"]` or something.

tests/src/unit-json_patch.cpp
nlohmann: In the CI, we have a dedicated test for `JSON_DIAGNOSTICS` and `JSON_DIAGNOSTIC_POSITIONS`. I don't think starting with combinations is helpful here. We maybe should rather have a dedicated test file where `JSON_DIAGNOSTICS` and `JSON_DIAGNOSTIC_POSITIONS` are defined together. This is then also automatically part of the coverage report.

include/nlohmann/detail/exceptions.hpp
nlohmann: Yes. And for the combination with `JSON_DIAGNOSTICS`, see https://github.com/nlohmann/json/pull/4585#discussion_r1906530495.

docs/examples/diagnostic_positions_exception.cpp
gregmarr: Since the intent is to just show the exception `j["address"]["housenumber"];` may be enough, but then you might get a warning about a discarded return value, so the `std::cout` is probably better.

tests/src/unit-json_patch.cpp
gregmarr: Normally I wouldn't think that having all combinations would be useful, but this is one case where the options interact.  A dedicated combined test might work just as well without increasing test run time too much.

hnampally: Yes, I am creating a new test case with both defined in a new test file similar to unit-diagnostics.cpp

include/nlohmann/detail/exceptions.hpp
```diff
@@ -141,20 +141,24 @@ class exception : public std::exception
...
         if ((leaf_element->start_pos() != std::string::npos) && (leaf_element->end_pos() != std::string::npos))
         {
-            return concat('(', "bytes ", std::to_string(leaf_element->start_pos()), "-", std::to_string(leaf_element->end_pos()), ") ");
+            return concat("(bytes ", std::to_string(leaf_element->start_pos()), "-", std::to_string(leaf_element->end_pos()), ") ");
         }
         return "";
+    }
 #else
+    template<typename BasicJsonType>
+    static std::string get_byte_positions(const BasicJsonType* leaf_element)
```
gregmarr, line 156: @nlohmann Do you have a preference on providing the parameter name and using `static_cast` to mark it as used, or leaving the parameter name off to say it's always unused?

Since this isn't isn't being used for documentation, my preference would be to leave the name off.

tests/src/unit-diagnostic-positions.cpp
```diff
@@ -0,0 +1,33 @@
...
+        {
+            "address": {
+                "street": "Fake Street",
+                "housenumber": "1"
+            }
+        }
+        )";
+        json j = json::parse(json_invalid_string);
+        int housenumber;
+        CHECK_THROWS_WITH_AS(housenumber = j["address"]["housenumber"], "[json.exception.type_error.302] (/address/housenumber) (bytes 108-111) type must be number, but is string", json::type_error);
```
gregmarr: I think just this:

```
        json j = json::parse(json_invalid_string);
        CHECK_THROWS_WITH_AS(j["address"]["housenumber"], "[json.exception....
```

hnampally: I understand, but this won't cause the type error exception we need for this test case, I think I need to static_cast<int> like below:
```
json j = json::parse(json_invalid_string);        
CHECK_THROWS_WITH_AS(static_cast<int>(j["address"]["housenumber"]), ....
```

gregmarr: Oh, right, you want to call `get<int>();`

gregmarr: The static cast won't work in explicit mode, so probably better to use the explicit version.

include/nlohmann/detail/exceptions.hpp
nlohmann: Yes, leaving the name off should be fine.

include/nlohmann/detail/exceptions.hpp
hnampally: I added tests/src/unit-diagnostic-positions.cpp to cover this function.

include/nlohmann/detail/exceptions.hpp
gregmarr: > error: all parameters should be named in a function [hicpp-named-parameter,readability-named-parameter,-warnings-as-errors]

Okay, who thought THAT was a good clang tidy check???

>The check implements the following rule originating in the Google C++ Style Guide:

Oh, well that explains it.  I find myself disagreeing with their style guide on a regular basis.  At least they let you get around it by putting the name in a comment in that position. (*eyeroll*)

include/nlohmann/detail/exceptions.hpp
nlohmann: The case of `return ""` is still uncovered:

image

You need a test case where a JSON value that was not parsed, but user generated, created an error. Like

```cpp
json j = "foo";
int i = j.get<std::string>();
```

which should throw.

nlohmann:

include/nlohmann/detail/exceptions.hpp
hnampally: image
It looks like it is still not covered after I added a new test case!

hnampally: I first tried the below code snippet you suggested

```
json j = "foo";
int i = j.get<std::string>();
```
It was failing to compile as shown below

image

so I did this https://github.com/nlohmann/json/pull/4585/files#diff-c46120f8da0119bb3e0143d7ef60d3c893202f3b44e7e481fe02ca8669db55f6R37
it seems like the correct exception was thrown but coverage still fails!

gregmarr: See https://github.com/nlohmann/json/pull/4585#discussion_r1906060127 for why the coverage test isn't hitting it.

nlohmann:
There is a problem building the documentation:

```
Error reading page 'api/macros/json_diagnostic_positions.md': Snippet at path 'examples/diagnostic_positions_exception.cpp' could not be found
```

Pleae check if you can build the documentation locally.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

Issue #4535, opened 2024-12-11
uint64_t being converted to int64_t in to_json


adlai-shawareb:
### Description

I am getting this runtime error:

Error converting JSON to BSON ([json.exception.out_of_range.407] integer number 18446744073701195776 cannot be represented by BSON as it does not fit int64

If I comment out use of the uint64_t value in to_json, and from_json, I don't get the error.

This appears to be very similar to Issue #4236, but not sure if it is the same.

If it is the same issue, do we know when 3.11.4 will be released?

### Reproduction steps

This code will reproduce it:   

 uint64_t m_Data0 = 0;

    inline void to_json( nlohmann::json& j, const Info& cInfo )
    {
        j = nlohmann::json{ 
               { "Data0", cInfo.m_Data0 } };

### Expected vs. actual results

I expect to_json to bring the uint64_t into the json object as a uint64_t.

### Compiler and operating system

Ubuntu 22.04 LTS

### Library version

3.10.5

### Validation

- [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/nlohmann/json/tree/develop) branch is used.
- [ ] I can successfully [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests).

nlohmann:
Looks like #3894

nlohmann:
Fixed by merging #4590.

PR #4590, opened 2025-01-09
Support BSON uint64 de/serialization

codenut:
This implements support for BSON uint64 de/serialization.

https://github.com/nlohmann/json/issues/4535

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/docs/mkdocs/docs/features/binary_formats/bson.md
+++ b/docs/mkdocs/docs/features/binary_formats/bson.md
@@ -73,7 +73,7 @@ The library maps BSON record types to JSON value types as follows:
 | Symbol                | 0x0E             | *unsupported*   |
 | JavaScript Code       | 0x0F             | *unsupported*   |
 | int32                 | 0x10             | number_integer  |
-| Timestamp             | 0x11             | number_unsigned |
+| uint64(Timestamp)     | 0x11             | number_unsigned |
 | 128-bit decimal float | 0x13             | *unsupported*   |
 | Max Key               | 0x7F             | *unsupported*   |
 | Min Key               | 0xFF             | *unsupported*   |
@@ -94,3 +94,8 @@ The library maps BSON record types to JSON value types as follows:
     ```json
     --8<-- "examples/from_bson.output"
     ```
+
+!!! note "Handling of BSON type 0x11"
+
+    BSON type 0x11 is used to represent uint64 numbers. This library treats these values purely as uint64 numbers 
+    and does not parse them into date-related formats.
--- a/docs/mkdocs/docs/features/binary_formats/bson.md
+++ b/docs/mkdocs/docs/features/binary_formats/bson.md
@@ -23,7 +23,7 @@ The library uses the following mapping from JSON values types to BSON types:
 | number_integer  | 2147483648..9223372036854775807           | int64     | 0x12   |
 | number_unsigned | 0..2147483647                             | int32     | 0x10   |
 | number_unsigned | 2147483648..9223372036854775807           | int64     | 0x12   |
-| number_unsigned | 9223372036854775808..18446744073709551615 | --        | --     |
+| number_unsigned | 9223372036854775808..18446744073709551615 | uint64    | 0x11   |
 | number_float    | *any value*                               | double    | 0x01   |
 | string          | *any value*                               | string    | 0x02   |
 | array           | *any value*                               | document  | 0x04   |
@@ -73,7 +73,7 @@ The library maps BSON record types to JSON value types as follows:
 | Symbol                | 0x0E             | *unsupported*   |
 | JavaScript Code       | 0x0F             | *unsupported*   |
 | int32                 | 0x10             | number_integer  |
-| Timestamp             | 0x11             | *unsupported*   |
+| Timestamp             | 0x11             | number_unsigned |
 | 128-bit decimal float | 0x13             | *unsupported*   |
 | Max Key               | 0x7F             | *unsupported*   |
 | Min Key               | 0xFF             | *unsupported*   |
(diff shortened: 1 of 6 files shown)


docs/mkdocs/docs/home/exceptions.md
```diff
@@ -839,7 +839,7 @@ A parsed number could not be stored as without changing it to NaN or INF.
 
 ### json.exception.out_of_range.407
 
-UBJSON and BSON only support integer numbers up to 9223372036854775807.
+UBJSON only support integer numbers up to 9223372036854775807.
```
nlohmann, line 842: If the exception 407 cannot be thrown for BSON any more, then https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/home/exceptions.md#jsonexceptionout_of_range407 also needs to be adjusted.

docs/mkdocs/docs/features/binary_formats/bson.md
```diff
@@ -23,7 +23,7 @@ The library uses the following mapping from JSON values types to BSON types:
 | number_integer  | 2147483648..9223372036854775807           | int64     | 0x12   |
 | number_unsigned | 0..2147483647                             | int32     | 0x10   |
 | number_unsigned | 2147483648..9223372036854775807           | int64     | 0x12   |
-| number_unsigned | 9223372036854775808..18446744073709551615 | --        | --     |
+| number_unsigned | 9223372036854775808..18446744073709551615 | int64     | 0x11   |
```
nlohmann: Shouldn't this be `uint64`?

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1097,7 +1097,8 @@ class binary_writer
         }
         else
         {
-            JSON_THROW(out_of_range::create(407, concat("integer number ", std::to_string(j.m_data.m_value.number_unsigned), " cannot be represented by BSON as it does not fit int64"), &j));
+            write_bson_entry_header(name, 0x11 /* int64 */);
```
nlohmann: Shouldn't the comment be `uint64`?

nlohmann (changes requested):
Isn't `0x11` a timestamp and not just a `uint64` number? It may be the same on the wire, but maybe the documentation should mention that we do not parse timestamps into anything date-related.

Also, please check https://github.com/nlohmann/json/pull/4590/checks?check_run_id=35367015620 on how to sign-off your commits.

codenut:
> Isn't `0x11` a timestamp and not just a `uint64` number? It may be the same on the wire, but maybe the documentation should mention that we do not parse timestamps into anything date-related.
> 
> Also, please check https://github.com/nlohmann/json/pull/4590/checks?check_run_id=35367015620 on how to sign-off your commits.

Please if the commit [Treat 0x11 as uint64 and not timestamp specific](https://github.com/nlohmann/json/commit/12d99be7cc75f2f6d3c6c01686bc256be3f75729) is aligned with your suggestion.

docs/mkdocs/docs/features/binary_formats/bson.md
```diff
@@ -73,7 +73,7 @@ The library maps BSON record types to JSON value types as follows:
 | Symbol                | 0x0E             | *unsupported*   |
 | JavaScript Code       | 0x0F             | *unsupported*   |
 | int32                 | 0x10             | number_integer  |
-| Timestamp             | 0x11             | *unsupported*   |
+| uint64                | 0x11             | number_unsigned |
```
nlohmann: Maybe "uint64 (Timestamp)" to not confuse BSON people too much.

docs/mkdocs/docs/features/binary_formats/bson.md
```diff
@@ -94,3 +94,7 @@ The library maps BSON record types to JSON value types as follows:
     ```json
     --8<-- "examples/from_bson.output"
     ```
+
+!!! note "Handling of BSON type 0x11"
+BSON type 0x11 is used to represent uint64 numbers. This library treats these values purely as uint64 numbers
```
nlohmann: After a `!!!` block,

- please add an empty line
- indent the text below with 4 spaces

Only then, a box is rendered in the final documentation.

docs/mkdocs/docs/home/exceptions.md
nlohmann: Oh, sorry. You already did this - I thought this was still in the `bson.md`...

codenut: What's the reason why the library doesn't support UBJSON uint64? 

Edit: I was looking at the code and it looks like UBJSON uint64 is supported. I will write a test later on to verify that.

gregmarr: https://ubjson.org/type-reference/value-types/#numeric doesn't list `uint64` as a supported type.

nlohmann (approved):
Looks good to me.

nlohmann:
Thank you!

docs/mkdocs/docs/home/exceptions.md
nlohmann: That is correct, but (learning from our documentation):

> Since version 3.9.0, integer numbers beyond int64 are serialized as high-precision UBJSON numbers, and this exception does not further occur.

nlohmann: (I'll update the documentation accordingly - json.exception.out_of_range.407 now cannot happen any more)

nlohmann: PTAL at https://github.com/nlohmann/json/commit/7bf9a06e0eb5a3090627a194ed12b5a8e50072c0

PR #4591, opened 2025-01-09
make number serialzation precision configurable

maddanio:
This PR adds the possibility to control floating point serialization precision via std::setprecision. I also have a version with an explicit annotator instead if you absolutely want to avoid behaviour changes (as users may have inadvertedly modified the iostate in their code right now), but I find this way much more natural (I was quite suprised that std::setprecision had no effect).

* * *

## Pull request checklist

### DIFF
--- a/include/nlohmann/detail/conversions/to_chars.hpp
+++ b/include/nlohmann/detail/conversions/to_chars.hpp
@@ -978,10 +978,11 @@ notation. Otherwise it will be printed in exponential notation.
 JSON_HEDLEY_NON_NULL(1)
 JSON_HEDLEY_RETURNS_NON_NULL
 inline char* format_buffer(char* buf, int len, int decimal_exponent,
-                           int min_exp, int max_exp)
+                           int min_exp, int max_exp, size_t precision)
 {
     JSON_ASSERT(min_exp < 0);
     JSON_ASSERT(max_exp > 0);
+    precision = (std::min)<size_t>(precision, std::numeric_limits<FloatType>::max_digits10);
 
     const int k = len;
     const int n = len + decimal_exponent;
@@ -1009,6 +1010,9 @@ inline char* format_buffer(char* buf, int len, int decimal_exponent,
 
         JSON_ASSERT(k > n);
 
+        // truncate the digits by the precision
+        k = (std::min)(static_cast<size_t>(n) + precision, static_cast<size_t>(k));
+
         std::memmove(buf + (static_cast<size_t>(n) + 1), buf + n, static_cast<size_t>(k) - static_cast<size_t>(n));
         buf[n] = '.';
         return buf + (static_cast<size_t>(k) + 1U);
@@ -1023,7 +1027,8 @@ inline char* format_buffer(char* buf, int len, int decimal_exponent,
         buf[0] = '0';
         buf[1] = '.';
         std::memset(buf + 2, '0', static_cast<size_t>(-n));
-        return buf + (2U + static_cast<size_t>(-n) + static_cast<size_t>(k));
+        // truncate the reported buffer end by the precision
+        return buf + (std::min)(precision + 2, (2U + static_cast<size_t>(-n) + static_cast<size_t>(k)));
     }
 
     if (k == 1)
@@ -1040,7 +1045,7 @@ inline char* format_buffer(char* buf, int len, int decimal_exponent,
 
         std::memmove(buf + 2, buf + 1, static_cast<size_t>(k) - 1);
         buf[1] = '.';
-        buf += 1 + static_cast<size_t>(k);
+        buf += 1 + (std::min)(precision, static_cast<size_t>(k));
     }
 
     *buf++ = 'e';
@@ -1062,7 +1067,7 @@ format. Returns an iterator pointing past-the-end of the decimal representation.
 template<typename FloatType>
 JSON_HEDLEY_NON_NULL(1, 2)
 JSON_HEDLEY_RETURNS_NON_NULL
-char* to_chars(char* first, const char* last, FloatType value)
+char* to_chars(char* first, const char* last, FloatType value, size_t precision = std::numeric_limits<FloatType>::max_digits10)
 {
     static_cast<void>(last); // maybe unused - fix warning
     JSON_ASSERT(std::isfinite(value));
@@ -1111,7 +1116,7 @@ char* to_chars(char* first, const char* last, FloatType value)
     JSON_ASSERT(last - first >= 2 + (-kMinExp - 1) + std::numeric_limits<FloatType>::max_digits10);
     JSON_ASSERT(last - first >= std::numeric_limits<FloatType>::max_digits10 + 6);
 
-    return dtoa_impl::format_buffer(first, len, decimal_exponent, kMinExp, kMaxExp);
+    return dtoa_impl::format_buffer(first, len, decimal_exponent, kMinExp, kMaxExp, precision);
 }
 
 }  // namespace detail
--- a/include/nlohmann/detail/output/serializer.hpp
+++ b/include/nlohmann/detail/output/serializer.hpp
@@ -64,13 +64,14 @@ class serializer
     @param[in] ichar  indentation character to use
     @param[in] error_handler_  how to react on decoding errors
     */
-    serializer(output_adapter_t<char> s, const char ichar,
+    serializer(output_adapter_t<char> s, const char ichar, std::size_t prec = 1000,
                error_handler_t error_handler_ = error_handler_t::strict)
         : o(std::move(s))
         , loc(std::localeconv())
         , thousands_sep(loc->thousands_sep == nullptr ? '\0' : std::char_traits<char>::to_char_type(* (loc->thousands_sep)))
         , decimal_point(loc->decimal_point == nullptr ? '\0' : std::char_traits<char>::to_char_type(* (loc->decimal_point)))
         , indent_char(ichar)
+        , precision(prec)
         , indent_string(512, indent_char)
         , error_handler(error_handler_)
     {}
@@ -820,7 +821,7 @@ class serializer
     void dump_float(number_float_t x, std::true_type /*is_ieee_single_or_double*/)
     {
         auto* begin = number_buffer.data();
-        auto* end = ::nlohmann::detail::to_chars(begin, begin + number_buffer.size(), x);
+        auto* end = ::nlohmann::detail::to_chars(begin, begin + number_buffer.size(), x, precision);
 
         o->write_characters(begin, static_cast<size_t>(end - begin));
     }
@@ -828,10 +829,10 @@ class serializer
     void dump_float(number_float_t x, std::false_type /*is_ieee_single_or_double*/)
     {
         // get number of digits for a float -> text -> float round-trip
-        static constexpr auto d = std::numeric_limits<number_float_t>::max_digits10;
+        static constexpr int d_max = std::numeric_limits<number_float_t>::max_digits10;
+        int d = static_cast<int>((std::min)(precision, static_cast<std::size_t>(d_max)));
 
         // the actual conversion
-        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg,hicpp-vararg)
         std::ptrdiff_t len = (std::snprintf)(number_buffer.data(), number_buffer.size(), "%.*g", d, x);
 
         // negative value indicates an error
@@ -977,6 +978,8 @@ class serializer
 
     /// the indentation character
     const char indent_char;
+    /// precision for floating point output
+    std::size_t precision;
     /// the indentation string
     string_t indent_string;
 
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -1272,10 +1272,11 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     string_t dump(const int indent = -1,
                   const char indent_char = ' ',
                   const bool ensure_ascii = false,
-                  const error_handler_t error_handler = error_handler_t::strict) const
+                  const error_handler_t error_handler = error_handler_t::strict,
+                  const size_t precision = std::numeric_limits<size_t>::max()) const
     {
         string_t result;
-        serializer s(detail::output_adapter<char, string_t>(result), indent_char, error_handler);
+        serializer s(detail::output_adapter<char, string_t>(result), indent_char, precision, error_handler);
 
         if (indent >= 0)
         {
@@ -3982,8 +3983,9 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         o.width(0);
 
         // do the actual serialization
-        serializer s(detail::output_adapter<char>(o), o.fill());
+        serializer s(detail::output_adapter<char>(o), o.fill(), static_cast<size_t>(o.precision())).;
         s.dump(j, pretty_print, false, static_cast<unsigned int>(indentation));
+
         return o;
     }
 
--- a/tests/src/unit-precision.cpp
+++ b/tests/src/unit-precision.cpp
@@ -1,0 +1,57 @@
+/*
+    __ _____ _____ _____
+ __|  |   __|     |   | |  JSON for Modern C++ (test suite)
+|  |  |__   |  |  | | | |  version 3.10.5
+|_____|_____|_____|_|___|  https://github.com/nlohmann/json
+
+Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+SPDX-License-Identifier: MIT
+Copyright (c) 2013-2022 Niels Lohmann <http://nlohmann.me>.
+
+Permission is hereby  granted, free of charge, to any  person obtaining a copy
+of this software and associated  documentation files (the "Software"), to deal
+in the Software  without restriction, including without  limitation the rights
+to  use, copy,  modify, merge,  publish, distribute,  sublicense, and/or  sell
+copies  of  the Software,  and  to  permit persons  to  whom  the Software  is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE  IS PROVIDED "AS  IS", WITHOUT WARRANTY  OF ANY KIND,  EXPRESS OR
+IMPLIED,  INCLUDING BUT  NOT  LIMITED TO  THE  WARRANTIES OF  MERCHANTABILITY,
+FITNESS FOR  A PARTICULAR PURPOSE AND  NONINFRINGEMENT. IN NO EVENT  SHALL THE
+AUTHORS  OR COPYRIGHT  HOLDERS  BE  LIABLE FOR  ANY  CLAIM,  DAMAGES OR  OTHER
+LIABILITY, WHETHER IN AN ACTION OF  CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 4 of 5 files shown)


include/nlohmann/detail/conversions/to_chars.hpp
```diff
@@ -978,10 +978,11 @@ notation. Otherwise it will be printed in exponential notation.
 JSON_HEDLEY_NON_NULL(1)
 JSON_HEDLEY_RETURNS_NON_NULL
 inline char* format_buffer(char* buf, int len, int decimal_exponent,
-                           int min_exp, int max_exp)
+                           int min_exp, int max_exp, size_t precision)
 {
     JSON_ASSERT(min_exp < 0);
     JSON_ASSERT(max_exp > 0);
+    precision = std::min<size_t>(precision, 1000);
```
nlohmann: There is some issue with `<windows.h>` being included before this library. You need to put `min` and `max` in parentheses to have it properly compile. Just grep for "min" and you will find an example.

nlohmann: Is there a better value than `1000` for this? Those magic numbers are hard to understand later on.

include/nlohmann/json.hpp
```diff
@@ -1272,10 +1272,11 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     string_t dump(const int indent = -1,
                   const char indent_char = ' ',
                   const bool ensure_ascii = false,
-                  const error_handler_t error_handler = error_handler_t::strict) const
+                  const error_handler_t error_handler = error_handler_t::strict,
```
nlohmann, line 1275: Similar to https://github.com/nlohmann/json/pull/4513#issuecomment-2574572656 where a `bjdata_config_t` type was discussed to control the BJData serialization, I think it is also long time to have such a time for the JSON serialization as the parameters start piling up. I'm hesitant to say if this is the last parameter to add before such a config type is introduced, or the first one where we stop the madness...

include/nlohmann/json.hpp
```diff
@@ -1272,10 +1272,11 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+                  const size_t precision = std::numeric_limits<size_t>::max()) const
```
nlohmann, line 1276: Why having `std::numeric_limits<size_t>::max()` here as default value?

include/nlohmann/json.hpp
```diff
@@ -3975,15 +3976,17 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     friend std::ostream& operator<<(std::ostream& o, const basic_json& j)
     {
         // read width member and use it as indentation parameter if nonzero
-        const bool pretty_print = o.width() > 0;
-        const auto indentation = pretty_print ? o.width() : 0;
+        const auto width = o.width();
```
nlohmann: Why this refactoring?

include/nlohmann/json.hpp
```diff
@@ -3975,15 +3976,17 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+        const bool pretty_print = width > 0;
+        const auto indentation = pretty_print ? width : 0;
 
         // reset width to 0 for subsequent calls to this stream
         o.width(0);
 
         // do the actual serialization
-        serializer s(detail::output_adapter<char>(o), o.fill());
+        serializer s(detail::output_adapter<char>(o), o.fill(), static_cast<size_t>(std::max<std::streamsize>(0, o.precision())));
```
nlohmann: According to https://en.cppreference.com/w/cpp/io/streamsize,

> Except in the constructors of [std::strstreambuf](https://en.cppreference.com/w/cpp/io/strstreambuf), negative values of std::streamsize are never used.

Why this max here?

tests/src/unit-precision.cpp
```diff
@@ -0,0 +1,58 @@
...
+{
+    json j;
+    j = M_PI;
+    auto to_string = [](const nlohmann::json & j, size_t precision)
+    {
+        std::stringstream ss;
+        ss << std::setprecision(precision) << j;
+        return ss.str();
+    };
+    auto s1 = to_string(j, 1);
```
nlohmann, line 51: For clarity, I would also add a test here for the default behavior of no explicitly set precision.

tests/src/unit-precision.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include "doctest_compatibility.h"
+
+#include <cmath>
+#include <iomanip>
+#include <nlohmann/json.hpp>
+#include <sstream>
+#include <string>
+using json = nlohmann::json;
+using ordered_json = nlohmann::ordered_json;
```
nlohmann: Is this needed?

tests/src/unit-precision.cpp
```diff
@@ -0,0 +1,58 @@
...
+    __ _____ _____ _____
+ __|  |   __|     |   | |  JSON for Modern C++ (test suite)
+|  |  |__   |  |  | | | |  version 3.10.5
+|_____|_____|_____|_|___|  https://github.com/nlohmann/json
+
+Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+SPDX-License-Identifier: MIT
+Copyright (c) 2013-2022 Niels Lohmann <http://nlohmann.me>.
+
+Permission is hereby  granted, free of charge, to any  person obtaining a copy
```
nlohmann, line 11: The MIT license does not need to be quoted verbatim here.

Just take the header that is used in all other tests:

```cpp
//     __ _____ _____ _____
//  __|  |   __|     |   | |  JSON for Modern C++ (supporting code)
// |  |  |__   |  |  | | | |  version 3.11.3
// |_____|_____|_____|_|___|  https://github.com/nlohmann/json
//
// SPDX-FileCopyrightText: 2013 - 2024 Niels Lohmann <https://nlohmann.me>
```

include/nlohmann/detail/conversions/to_chars.hpp
```diff
@@ -1062,7 +1063,7 @@ format. Returns an iterator pointing past-the-end of the decimal representation.
 template<typename FloatType>
 JSON_HEDLEY_NON_NULL(1, 2)
 JSON_HEDLEY_RETURNS_NON_NULL
-char* to_chars(char* first, const char* last, FloatType value)
+char* to_chars(char* first, const char* last, FloatType value, size_t precision = std::numeric_limits<FloatType>::max_digits10)
```
nlohmann, line 1070: Just to understand: why did you provide a default value for `precision` here?

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -832,6 +833,8 @@ class serializer
 
         // the actual conversion
         // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg,hicpp-vararg)
+        char format[100];
```
nlohmann: Where is the magic number `100` coming from?

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -977,6 +980,8 @@ class serializer
 
     /// the indentation character
     const char indent_char;
+    /// precision for floating point output
+    size_t precision;
```
nlohmann:
```suggestion
    std::size_t precision;
```

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -832,6 +833,8 @@ class serializer
 
         // the actual conversion
         // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg,hicpp-vararg)
+        char format[100];
+        snprintf(format, 100, "%%.%dg", precision);
```
nlohmann: Where is `format` used? Is `snprintf` really necessary? Maybe check out `details::concat` from `detail/string_concat.hpp`.

include/nlohmann/detail/conversions/to_chars.hpp
```diff
@@ -1023,7 +1024,7 @@ inline char* format_buffer(char* buf, int len, int decimal_exponent,
         buf[0] = '0';
         buf[1] = '.';
         std::memset(buf + 2, '0', static_cast<size_t>(-n));
-        return buf + (2U + static_cast<size_t>(-n) + static_cast<size_t>(k));
+        return buf + std::min(precision + 2, (2U + static_cast<size_t>(-n) + static_cast<size_t>(k)));
```
nlohmann: Please document this term.

include/nlohmann/detail/conversions/to_chars.hpp
```diff
@@ -1011,7 +1012,7 @@ inline char* format_buffer(char* buf, int len, int decimal_exponent,
 
         std::memmove(buf + (static_cast<size_t>(n) + 1), buf + n, static_cast<size_t>(k) - static_cast<size_t>(n));
         buf[n] = '.';
-        return buf + (static_cast<size_t>(k) + 1U);
+        return buf + (std::min(static_cast<size_t>(n) + precision, static_cast<size_t>(k)) + 1U);
```
nlohmann: Please document this term.

nlohmann (changes requested):

include/nlohmann/detail/conversions/to_chars.hpp
maddanio: yikes, yes, will do

include/nlohmann/json.hpp
maddanio: so mine is one too much? :) but sure, I understand, will look into it

include/nlohmann/json.hpp
maddanio: it will be limited by type later, so this default makes sure the maximum possible value is chosen

include/nlohmann/json.hpp
maddanio: habit, revert?

include/nlohmann/json.hpp
maddanio: defensiveness, I will remove it

tests/src/unit-precision.cpp
maddanio: copy pasted from another test, but yes

tests/src/unit-precision.cpp
maddanio: no, copy paste

tests/src/unit-precision.cpp
maddanio: ok, but tbh I will then just adapt the test to what I observer, since I have no expectation for that case

maddanio:
thanks for looking at this. so does that mean this PR will get merged when I address all of the requests? I was submitting it a bit speculatively and dont want to sink work if thats not the case or more principal discussion is needed first

include/nlohmann/json.hpp
maddanio: ah, hang on, isnt that a public api breaking change? so you want me to change the public api? or add an overload?

include/nlohmann/detail/output/serializer.hpp
gregmarr: I assume it was supposed to replace `"%.*g"` below.

include/nlohmann/detail/output/serializer.hpp
gregmarr: By putting this after the `// NOLINT` comment you've broken its association with the `std::snprintf`

include/nlohmann/detail/output/serializer.hpp
gregmarr: Should instead just do `..."%*.*g", precision, d, x)`

include/nlohmann/detail/conversions/to_chars.hpp
maddanio: because this is the largest meaningful precision, I will use this as default throughtout

include/nlohmann/detail/output/serializer.hpp
gregmarr: This also means that `precision` should be `int` instead of `size_t`.  There is no reason for it to be larger than `int`.

gregmarr: Oh, that's the wrong position, that's width.  We already specify precision as `d`, so a defined `precision` override just replaces the existing `d`.

maddanio: yes, on it

include/nlohmann/json.hpp
gregmarr: The old behavior is `std::numeric_limits<number_float_t>::max_digits10;` so this should do something like have a default value of `-1` meaning use `max_digits10` and have `>= 0` mean use the defined precision instead of the built-in one.
Also see my comments at the actual `snprintf` for why I think it should be `int` instead of `size_t`.

include/nlohmann/json.hpp
gregmarr: >so mine is one too much? :) but sure, I understand, will look into it

Many others before you have been "one too much" and thus haven't been done.  :)

include/nlohmann/detail/output/serializer.hpp
maddanio: I just realized i did not test this path. can you point me to another test that actually provides a non-ieee float?

include/nlohmann/detail/output/serializer.hpp
gregmarr: See my comments at the actual `snprintf` for why I think it should be `int` instead of `size_t`.

include/nlohmann/detail/conversions/to_chars.hpp
gregmarr: I would suggest `int precisionOverride = -1` and any value less than 0 is ignored and the current behavior is used.

include/nlohmann/json.hpp
maddanio: reverte

gregmarr:
@nlohmann has the final say here on whether this can go forward.

Prior to that, I think it would be good to define what exactly precision means here.  Is it the maximum number of digits after the decimal point, or the exact number?  Having it mean exact is more involved, as it can change a current `0.0` to `0.000000`.  This should also define if it is going to use truncation, or rounding.  That is, is PI with precision 3 `3.142` or `3.141`.  Similarly, is `1.999999999` to any precision less than 9 just `2.0`?

One reason I recommended using `-1` as an override is that there is more here than just shortening the number.  If a precision isn't defined, everything just behave exactly as it is now.  If it is defined, then you have to look at rounding, and thus you may not be able to use Grisu2.  I haven't looked into it enough to know if you can implement a rounding precision parameter in Grisu2.

Then the next question is whether `std::setprecision` affects this, or if it's only configured via parameters to `dump` and if that means that it's time to create a configuration method for `dump`.  That is related to the question of whether or not it would affect the `ostream << json` operators.  As mentioned in the initial comment, having `std::setprecision` affect either of these things would be a breaking change.

include/nlohmann/json.hpp
```diff
@@ -3975,15 +3976,16 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     friend std::ostream& operator<<(std::ostream& o, const basic_json& j)
     {
         // read width member and use it as indentation parameter if nonzero
-        const bool pretty_print = o.width() > 0;
-        const auto indentation = pretty_print ? o.width() : 0;
+        const bool pretty_print = o.width > 0;
+        const auto indentation = pretty_print ? o.width : 0;
```
gregmarr: Didn't quite revert this properly, missing the `()`.

gregmarr:
About the maximum or exact and truncation vs rounding:

```
    std::cout << std::setprecision(3) << M_PI << "\n" << 1.9999 << "\n";
```
This prints 
```
3.142
2
```
https://www.godbolt.org/z/8Mne8Y59M

include/nlohmann/json.hpp
nlohmann: There has been so many discussions about tweaking every last aspect of serializations that adding parameters is getting out of hand.

nlohmann:
I am very hesitant, and these are just my thoughts right now:

- I personally don't think the serialization is broken. It does what it should, and the reason we chose the current precision is to enable roundtripping.
- A lot of people are offended by the way the library does serialization. Partly because they don't understand how floating-point numbers work. How can a library change their `1.9` to `1.89999999999999991118`?!
- The other part are concerned about bandwidth and want to limit the number of bytes used to write π to the wire.
- Again others want to control more aspects of the output, like where `\n` are added, how content is aligned, etc.
- All this can be implemented somehow, and it's never about "What C++ code is needed to realize this problem?" but more "How can we put this into an API that people want to use, and other people want to maintain in their free time."
- The questions @gregmarr raises are very valid, and I think there is no single right answer to them. If we define precision = "maximal number of digits after the `.` using truncation rather than rounding", then there will be a feature request later this year asking to change any other point on the list.
- Also, the library is frequently criticized for its performance. Adding all these extension hooks does not really improve this. Also, the other JSON libraries seem not to care about this aspect (or their users don't ask for it). Performance should not be the "no" to all feature requests, but should definitely be on the list of concerns.
- It's hard to find the right answer to all this, because in the end, the library is targeting developers, not just myself.

Sorry for this rough brain-dump. I will not be able to make any decision right now. But I think I will not add this without having a clear plan how serialization can be improved conceptually before adding some more ifs here and there.

nlohmann:
I forgot one thing that other libraries do that avoids some of the confusion I see in issues here: they store the original input string and only convert it to a number on demand. As a result, parsing is cheaper and you can create the *exact* serialization of the input. Not sure if we want to do this here, but a "precision = keep whatever precision I put in" requirement is also thinkable.

maddanio:
I dont think performance difference is measurable, the only action is changing some pointers and doing some comparisons. if any I would surmise performance improves with lower precision.

yes, it is truncating, so if you want rounding that would be more involved. but if desired I would be willing to dig in.

all in all we in my company really want this, because we have some rather large sets of numbers in json, and this saves us a rather large amount of space and noise. but if you don't think this is what the library needs we will just continue on our fork. and I fully understand how float works, but I also know in many cases the number of digits I really need, and what is just noise (in physics when we calculated stuff we rarely kept more than 3 significant digits :))

risa2000:
Several years ago I wrote an app which stored soma data in JSON file (using this library) and hashed part of this file by using a crypto hash function. The hashed part was a JSON serialization using a "canonical" format (i.e. no formatting, indentation, or white spaces).
It worked because of the aforementioned feature of *roundtripping*.

Then I learned the hard way, when using Python JSON implementation on the very same data that `nlohmann::json` produced output that the Python module was not able to "roundtrip". Apparently `nlohmann::json` uses a different rounding rules than Python native `json` module, so some numbers got the last precision bit flipped when read by Python and then serialized again. This would not make a big deal mathematically wise, but made a big deal if the text representation was an input to a hashing function.

At the end I had to revert to writing a custom `float` parser in Python's `simplejson` which preserved the textual representation of the JSON `number` (i.e. floats) and used this for hashing instead of roundtripping the float value.
All this just to say that even when one library is able to roundtrip, it does not guarantee an interoperability with the other. Another implication is that if the library changes the way it works with "floats" it will break scenarios like the one described above (even some would argue it is a poor design).

My understanding of what @maddanio suggests is not in fact changing "precision" (in particular because the proposal does not care about the rounding), but *formatting* (representation). I.e. being able to control the length of the textual representation.

I believe it would be good if the library supported a way to write custom serializer/deserializer the similar way it supports custom conversions for e.g. `JSON number`. This way whoever wanted to write in a particular format would just supply its own (de)serializer without a need to fork the project or impose the required behavior on the library.

PR #4594, opened 2025-01-12
Serialize empty tuple into '[]'  instead of null

### LINKED ISSUE #4530: Empty `std::tuple<>` cannot be serialized and de-serialized
### Description

The behavior of serializing and de-sereailzing an empty tuple changed from `3.9.1` to `3.10.0`.

`null` used to de-serialize without a problem to `std::tuple<>` but since throws an error since `3.10.0`.

The change of behavior happened in https://github.com/nlohmann/json/pull/2576.

Note that I'm not entirely sure this is a bug per-say, but we relied on this behavior in https://github.com/luxonis/depthai-core so I decided to report in case the previous behavior was "more correct".

### Reproduction steps

To reproduce run the MRE with both `3.10.0` and `3.9.1`.

Expected - `null` succesfully de-serializes when `j.get<std::tuple<>>()` is called on `null`.
Actual:
```
terminate called after throwing an instance of 'nlohmann::detail::type_error'
  what():  [json.exception.type_error.302] type must be array, but is null
```

### Minimal code example

```Shell
#include <nlohmann/json.hpp>

int main() {
    auto sourceTuple = std::tuple<>();
    nlohmann::json j = sourceTuple;
    auto returnedTuple = j.get<std::tuple<>>();
    return 0;
}
```

### Error messages

```Shell
terminate called after throwing an instance of 'nlohmann::detail::type_error'
  what():  [json.exception.type_error.302] type must be array, but is null
Aborted (core dumped)
```

Ubuntu 20.04, GCC 9.4.0

### Library version

3.11.3

### Validation

codenut:
This PR fixes https://github.com/nlohmann/json/issues/4530 by serializing an empty tuple into `[]` instead of `null`.

* * *

## Pull request checklist

## Please don't

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -448,6 +448,12 @@ std::tuple<Args...> from_json_tuple_impl_base(BasicJsonType&& j, index_sequence<
     return std::make_tuple(std::forward<BasicJsonType>(j).at(Idx).template get<Args>()...);
 }
 
+template<typename BasicJsonType>
+std::tuple<> from_json_tuple_impl_base(BasicJsonType& /*unused*/, index_sequence<> /*unused*/)
+{
+    return {};
+}
+
 template < typename BasicJsonType, class A1, class A2 >
 std::pair<A1, A2> from_json_tuple_impl(BasicJsonType&& j, identity_tag<std::pair<A1, A2>> /*unused*/, priority_tag<0> /*unused*/)
 {
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -426,6 +426,13 @@ inline void to_json_tuple_impl(BasicJsonType& j, const Tuple& t, index_sequence<
     j = { std::get<Idx>(t)... };
 }
 
+template<typename BasicJsonType, typename Tuple>
+inline void to_json_tuple_impl(BasicJsonType& j, const Tuple& /*unused*/, index_sequence<> /*unused*/)
+{
+    using array_t = typename BasicJsonType::array_t;
+    j = array_t();
+}
+
 template<typename BasicJsonType, typename T, enable_if_t<is_constructible_tuple<BasicJsonType, T>::value, int > = 0>
 inline void to_json(BasicJsonType& j, const T& t)
 {
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -5190,6 +5190,12 @@ std::tuple<Args...> from_json_tuple_impl_base(BasicJsonType&& j, index_sequence<
     return std::make_tuple(std::forward<BasicJsonType>(j).at(Idx).template get<Args>()...);
 }
 
+template<typename BasicJsonType>
+std::tuple<> from_json_tuple_impl_base(BasicJsonType& /*unused*/, index_sequence<> /*unused*/)
+{
+    return {};
+}
+
 template < typename BasicJsonType, class A1, class A2 >
 std::pair<A1, A2> from_json_tuple_impl(BasicJsonType&& j, identity_tag<std::pair<A1, A2>> /*unused*/, priority_tag<0> /*unused*/)
 {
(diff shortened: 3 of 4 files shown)


include/nlohmann/detail/conversions/to_json.hpp
```diff
@@ -423,7 +423,8 @@ inline void to_json(BasicJsonType& j, const T& b)
 template<typename BasicJsonType, typename Tuple, std::size_t... Idx>
 inline void to_json_tuple_impl(BasicJsonType& j, const Tuple& t, index_sequence<Idx...> /*unused*/)
 {
-    j = { std::get<Idx>(t)... };
+    using array_t = typename BasicJsonType::array_t;
+    j = (sizeof...(Idx) == 0) ? array_t() : array_t{ std::get<Idx>(t)... };
```
gregmarr: Could (should?) this be done as an overload instead?

Untested:
```
template<typename BasicJsonType, typename Tuple>
inline void to_json_tuple_impl(BasicJsonType& j, const Tuple& t, index_sequence<> /*unused*/)
{
    j = array_t();
}
```
or
```
inline void to_json(BasicJsonType& j, const tuple<>& t)
{
    j = array_t();
}
```

gregmarr:
Title has typo on empty

include/nlohmann/detail/conversions/to_json.hpp
codenut: That is much better than the ternary operator. Thanks!

nlohmann:

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -448,6 +448,12 @@ std::tuple<Args...> from_json_tuple_impl_base(BasicJsonType&& j, index_sequence<
     return std::make_tuple(std::forward<BasicJsonType>(j).at(Idx).template get<Args>()...);
 }
 
+template<typename BasicJsonType>
+std::tuple<> from_json_tuple_impl_base(BasicJsonType& /*unused*/, index_sequence<> /*unused*/)
+{
+    return {};
```
nlohmann, line 454:
```suggestion
    return array_t();
```

Braced initialization is somewhat cursed in this library (see https://json.nlohmann.me/home/faq/#brace-initialization-yields-arrays). Please avoid it.

nlohmann (changes requested):

include/nlohmann/detail/conversions/from_json.hpp
gregmarr: This isn't returning `json`, it's returning `std::tuple<>`, so it would be `std::tuple<>{}` instead of `{}` if you still want that change.

nlohmann: Oh sorry. Then never mind.

gregmarr (approved):

nlohmann:
I had a look at #4530, and the issue stated

> null used to de-serialize without a problem to std::tuple<> but since throws an error since 3.10.0.

and

> Expected - null succesfully de-serializes when j.get<std::tuple<>>() is called on null.

However, when I execute this code

```cpp
// deserialize null
nlohmann::json j_null;
CHECK_NOTHROW(j_null.get<std::tuple<>>());
```

on this branch, I still get the error

```

[json.exception.type_error.302] type must be array, but is null
```

I am fine with serializing and empty tuple to `[]`, but in the end, we do not fix the linked issue, right?

gregmarr:
The linked issue title and first two sentences:

Title
> Empty std::tuple<> cannot be serialized and de-serialized

1
> The behavior of serializing and de-sereailzing an empty tuple changed from 3.9.1 to 3.10.0.

2
> null used to de-serialize without a problem to std::tuple<> but since throws an error since 3.10.0.

So I would say it fixes the title first sentence (round trip from c++ to json to c++) by fixing the serializing.  I think it's clear that the serializing to `null` is a bug and should be fixed.

It doesn't fix the second sentence (deserialize to `std::tuple<>` from `null`).  I think the "Expected behavior" in that issue is wrong and this is desired behavior.  You can't deserialize to `std::string` or `int` from `null`, so I think it's fine that you can't deserialize it to `std::tuple<>` either.

nlohmann:
Thanks @gregmarr. I totally agree, but I wanted to make sure to what extent we address all aspects of the issue.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #4597, opened 2025-01-13
Generate template functions with NLOHMANN_DEFINE_TYPE macros

kimci86:
Changes:

- make `NLOHMANN_DEFINE_TYPE_*` generate template functions. This allows to use any specialization of `nlohmann::basic_json` such as `nlohmann::ordered_json`, instead of `nlohmann::json` only.
- update and extend related tests
- update related documentation

Existing issues:
- #2528
- #4377

Related PRs:
- #2532 (closed)
- #2843 (stale)
- #4528

Reviews on related PRs have raised the following concerns:

- There might be a risk that generated `to_json`/`from_json` becoming a template change overload resolution is some twisted cases.
- The template type `BasicJsonType` is not restricted to `basic_json` instances.

To be noted: `NLOHMANN_JSON_SERIALIZE_ENUM` already uses that same design of unrestricted `BasicJsonType` template parameter.
If we want to restrict it, I see two options:
- Restructure headers so that macros can be defined after traits to be able to use `is_basic_json`.
  Currently `include/nlohmann/detail/meta/type_traits.hpp` includes `include/nlohmann/detail/macro_scope.hpp` so we cannot use `is_basic_json` in macros.
- Do not `#undef` `NLOHMANN_BASIC_JSON_TPL_DECLARATION` and `NLOHMANN_BASIC_JSON_TPL` in `macro_unscope.hpp` so that we can use them in macros to effectively write the full `basic_json` template type.

What do you think?

Thank you for this awesome library!

## Pull request checklist

Read the [Contribution Guidelines](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md) for detailed information.

- [x]  Changes are described in the pull request, or an [existing issue is referenced](https://github.com/nlohmann/json/issues).
- [x]  The test suite [compiles and runs](https://github.com/nlohmann/json/blob/develop/README.md#execute-unit-tests) without error.
- [ ]  [Code coverage](https://coveralls.io/github/nlohmann/json) is 100%. Test cases can be added by editing the [test suite](https://github.com/nlohmann/json/tree/develop/test/src).
- [x]  The source code is amalgamated; that is, after making changes to the sources in the `include/nlohmann` directory, run `make amalgamate` to create the single-header files `single_include/nlohmann/json.hpp` and `single_include/nlohmann/json_fwd.hpp`. The whole process is described [here](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md#files-to-change).

Issue #3810, opened 2022-10-30
`ordered_json` doesn't support construction from C array of custom type


lethal-guitar:
### Description

I have a C-style array of structs, plus an appropriate `to_json` overload. I can succesfully convert the array to JSON using a regular `nlohmann::json` object.  However, trying to do the same using `nlohmann::ordered_json` fails to compile - see details below.

### Reproduction steps

See the example code below.

### Expected vs. actual results

I'd expect the code to compile, but I get a compiler error.

### Minimal code example

Build with `-std=c++17 -O3`.

```cpp
#include <nlohmann/json.hpp>

struct Example
{
    int bla;
};

NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(Example, bla);

int main()
{
    Example states[45];

    // This works:
    nlohmann::json j;
    j["test"] = states;

    // This doesn't compile:
    nlohmann::ordered_json oj;
    oj["test"] = states;
}
```

### Error messages

gcc 9.4.0 reports:

```Shell
In file included from /opt/compiler-explorer/gcc-9.4.0/include/c++/9.4.0/vector:66,
                 from /opt/compiler-explorer/gcc-9.4.0/include/c++/9.4.0/functional:62,
                 from /opt/compiler-explorer/gcc-9.4.0/include/c++/9.4.0/pstl/glue_algorithm_defs.h:13,
                 from /opt/compiler-explorer/gcc-9.4.0/include/c++/9.4.0/algorithm:71,
                 from /opt/compiler-explorer/libs/nlohmann_json/v3.11.1/single_include/nlohmann/json.hpp:21,
                 from <source>:1:
/opt/compiler-explorer/gcc-9.4.0/include/c++/9.4.0/bits/stl_uninitialized.h: In instantiation of '_ForwardIterator std::uninitialized_copy(_InputIterator, _InputIterator, _ForwardIterator) [with _InputIterator = const Example*; _ForwardIterator = nlohmann::json_v3_11_1::basic_json<nlohmann::json_v3_11_1::ordered_map>*]':
[...]

PR #4608, opened 2025-01-19
Add regression test for #3810

nlohmann:
Issue https://github.com/nlohmann/json/issues/3810 was fixed by merging https://github.com/nlohmann/json/pull/4597. This PR adds a regression test.

### DIFF
--- a/tests/src/unit-regression2.cpp
+++ b/tests/src/unit-regression2.cpp
@@ -377,6 +377,17 @@ inline for_3333::for_3333(const json& j)
     : for_3333(j.value("x", 0), j.value("y", 0))
 {}
 
+/////////////////////////////////////////////////////////////////////
+// for #3810
+/////////////////////////////////////////////////////////////////////
+
+struct Example_3810
+{
+    int bla{};
+};
+
+NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(Example_3810, bla);
+
 TEST_CASE("regression tests 2")
 {
     SECTION("issue #1001 - Fix memory leak during parser callback")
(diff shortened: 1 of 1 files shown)


tests/src/unit-regression2.cpp
```diff
@@ -1004,6 +1015,23 @@ TEST_CASE("regression tests 2")
...
+
+        // This doesn't compile:
+        nlohmann::ordered_json oj;
+        oj["test"] = states;
+        CHECK(oj["test"].dump() == "[{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0},{\"bla\":0}]");
+
+        CHECK(j["test"].dump() == oj["test"].dump());
```
gregmarr: Should the strings in 1026 and 1031 just be a single const string to be absolutely sure that they stay the same?  That would also eliminate the need for the test at line 1033.

nlohmann: You're absolutely right. Before that, I tested

```cpp
CHECK(j == oj);
```

which failed on several compilers. I wanted to first have the CI green again before I cleaned up.

PR #4609, opened 2025-01-19
Add ignore_trailing_commas option

### LINKED ISSUE #1429: Support for bad commas
- Some of the projects I work on use man-made Jsons as config files. At start it is nice, then people start to forget some commas at the end of objects/arrays. When using Newtonsoft's Json.Net, it handles it gracefully, just ignoring. When using yours, it gives me an error. Thus, I have to correct some files from time to time so that they can be read.
I love your lib and that feature would make my life easier. 

- A minimal Json with bad commas:
```
{
    "A": 10,
}
           /\ Bad Comma
```

As I see it, the comma puts the lib in a state where it expects some other member, but it gets a } instead. Thus, the grammar must be slightly altered to allow , } as just being a }. Same goes for ,] being threated as just a ]

Thanks for the Lib :)

### LINKED ISSUE #1734: Allow Tail { "Key": "Value" } Comma [@ READ ONLY]
Example of current use ->

Only `{ "Name": "Michael", "Age": 36 }`

Example of use after this feature request ->

Either `{ "Name": "Michael", "Age": 36 }` or `{ "Name": "Michael", "Age": 36, }`

Note! This feature request should only affect the functions that read the trees. Those functions that write should remain the same.

### LINKED ISSUE #1787: Support for trailing commas
It looks that the current parse function throws an out_of_range exception when a trailing comma is detected e.g.:

```
{
    "var": "value",
    "var2": "value2",
}
```

The above works fine if the comma after "value2" is removed.

I know that the trailing comma is not allowed in the original json specification but I think that it is going to be allowed in future ones and it seems to me an unnecessary inconvenience.

Would you consider allowing this in your library?

chirsz-ever:
This PR adds `ignore_trailing_commas` argument to the `detail::parser` constructor, `basic_json::accept`, `basic_json::parse` and `basic_json::sax_parse`.

## Why

In short: I totally agree with [the JWCC extension](https://nigeltao.github.io/blog/2021/json-with-commas-comments.html).

After adding this, we can parse tsconfig.json, which [allow trailing commas](https://github.com/microsoft/vscode/blob/main/extensions/typescript-language-features/schemas/tsconfig.schema.json). (By the way, VSCode has a `allowTrailingCommas` option in its json schema configuration).

More projects would be attracted to use this library, e.g. <https://github.com/R2Northstar/NorthstarLauncher/issues/715>.

Related issues: #1429 #1787 #1734 

## Detail Changes

- Add `ignore_trailing_commas` argument to the `detail::parser` constructor, `basic_json::accept`, `basic_json::parse` and `basic_json::sax_parse`.
- Update documets and README.md.
- Update the unit test "class_parser". Add `,` to non-empty objects and arrays for testing.

## Pull request checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1801,7 +1801,17 @@ This library does not support comments by default. It does so for three reasons:
   
 3. It is dangerous for interoperability if some libraries would add comment support while others don't. Please check [The Harmful Consequences of the Robustness Principle](https://tools.ietf.org/html/draft-iab-protocol-maintenance-01) on this.
 
-However, you can pass set parameter `ignore_comments` to true in the `parse` function to ignore `//` or `/* */` comments. Comments will then be treated as whitespace.
+However, you can set set parameter `ignore_comments` to true in the `parse` function to ignore `//` or `/* */` comments. Comments will then be treated as whitespace.
+
+### Trailing commas
+
+The JSON specification does not allow trailing commas in arrays and objects, and hence this library is treating them as parsing errors by default.
+
+Like comments, you can set parameter `ignore_trailing_commas` to true in the `parse` function to ignore trailing commas in arrays and objects. Note that a single comma as the only content of the array or object (`[,]` or `{,}`) is not allowed, and multiple trailing commas (`[1,,]`) are not allowed either.
+
+This library does not add trailing commas when serializing JSON data.
+
+For more information, see [JSON With Commas and Comments (JWCC)](https://nigeltao.github.io/blog/2021/json-with-commas-comments.html).
 
 ### Order of object keys
 
--- a/docs/mkdocs/docs/api/basic_json/accept.md
+++ b/docs/mkdocs/docs/api/basic_json/accept.md
@@ -4,12 +4,14 @@
 // (1)
 template<typename InputType>
 static bool accept(InputType&& i,
-                   const bool ignore_comments = false);
+                   const bool ignore_comments = false,
+                   const bool ignore_trailing_commas = false);
 
 // (2)
 template<typename IteratorType>
 static bool accept(IteratorType first, IteratorType last,
-                   const bool ignore_comments = false);
+                   const bool ignore_comments = false,
+                   const bool ignore_trailing_commas = false);
 ```
 
 Checks whether the input is valid JSON.
@@ -48,6 +50,10 @@ Unlike the [`parse()`](parse.md) function, this function neither throws an excep
 
 `ignore_comments` (in)
 :   whether comments should be ignored and treated like whitespace (`#!cpp true`) or yield a parse error
+    (`#!cpp false`); (optional, `#!cpp false` by default)
+
+`ignore_trailing_commas` (in)
+:   whether trailing commas in arrays or objects should be ignored and treated like whitespace (`#!cpp true`) or yield a parse error
     (`#!cpp false`); (optional, `#!cpp false` by default)
 
 `first` (in)
@@ -102,6 +108,7 @@ A UTF-8 byte order mark is silently ignored.
 - Added in version 3.0.0.
 - Ignoring comments via `ignore_comments` added in version 3.9.0.
 - Changed [runtime assertion](../../features/assertions.md) in case of `FILE*` null pointers to exception in version 3.12.0.
+- Added `ignore_trailing_commas` in version 3.12.1.
 
 !!! warning "Deprecation"
 
--- a/docs/mkdocs/docs/examples/comments.cpp
+++ b/docs/mkdocs/docs/examples/comments.cpp
@@ -1,0 +1,31 @@
+
+#include <iostream>
+#include <nlohmann/json.hpp>
+
+using json = nlohmann::json;
+
+int main()
+{
+    std::string s = R"(
+    {
+        // update in 2006: removed Pluto
+        "planets": ["Mercury", "Venus", "Earth", "Mars",
+                    "Jupiter", "Uranus", "Neptune" /*, "Pluto" */]
+    }
+    )";
+
+    try
+    {
+        json j = json::parse(s);
+    }
+    catch (json::exception& e)
+    {
+        std::cout << e.what() << std::endl;
+    }
+
+    json j = json::parse(s,
+                         /* callback */ nullptr,
+                         /* allow exceptions */ true,
+                         /* ignore_comments */ true);
+    std::cout << j.dump(2) << '\n';
+}
--- a/docs/mkdocs/docs/examples/trailing_commas.cpp
+++ b/docs/mkdocs/docs/examples/trailing_commas.cpp
@@ -1,0 +1,37 @@
+#include <iostream>
+#include <nlohmann/json.hpp>
+
+using json = nlohmann::json;
+
+int main()
+{
+    std::string s = R"(
+    {
+        "planets": [
+            "Mercury",
+            "Venus",
+            "Earth",
+            "Mars",
+            "Jupiter",
+            "Uranus",
+            "Neptune",
+        ]
+    }
+    )";
+
+    try
+    {
+        json j = json::parse(s);
+    }
+    catch (json::exception& e)
+    {
+        std::cout << e.what() << std::endl;
+    }
+
+    json j = json::parse(s,
+                         /* callback */ nullptr,
+                         /* allow exceptions */ true,
+                         /* ignore_comments */ false,
+                         /* ignore_trailing_commas */ true);
+    std::cout << j.dump(2) << '\n';
+}
--- a/docs/mkdocs/docs/features/comments.md
+++ b/docs/mkdocs/docs/features/comments.md
@@ -11,7 +11,9 @@ This library does not support comments *by default*. It does so for three reason
 
 3. It is dangerous for interoperability if some libraries add comment support while others do not. Please check [The Harmful Consequences of the Robustness Principle](https://tools.ietf.org/html/draft-iab-protocol-maintenance-01) on this.
 
-However, you can pass set parameter `ignore_comments` to `#!c true` in the parse function to ignore `//` or `/* */` comments. Comments will then be treated as whitespace.
+However, you can set parameter `ignore_comments` to `#!cpp true` in the [`parse`](../api/basic_json/parse.md) function to ignore `//` or `/* */` comments. Comments will then be treated as whitespace.
+
+For more information, see [JSON With Commas and Comments (JWCC)](https://nigeltao.github.io/blog/2021/json-with-commas-comments.html).
 
 !!! example
 
@@ -28,56 +30,11 @@ However, you can pass set parameter `ignore_comments` to `#!c true` in the parse
     When calling `parse` without additional argument, a parse error exception is thrown. If `ignore_comments` is set to `#! true`, the comments are ignored during parsing:
 
     ```cpp
-    #include <iostream>
-    #include "json.hpp"
-    
-    using json = nlohmann::json;
-    
-    int main()
-    {
-        std::string s = R"(
-        {
-            // update in 2006: removed Pluto
-            "planets": ["Mercury", "Venus", "Earth", "Mars",
-                        "Jupiter", "Uranus", "Neptune" /*, "Pluto" */]
-        }
-        )";
-        
-        try
-        {
-            json j = json::parse(s);
-        }
-        catch (json::exception &e)
-        {
-            std::cout << e.what() << std::endl;
-        }
-        
-        json j = json::parse(s,
-                             /* callback */ nullptr,
-                             /* allow exceptions */ true,
-                             /* ignore_comments */ true);
-        std::cout << j.dump(2) << '\n';
-    }
+    --8<-- "examples/comments.cpp"
     ```
 
     Output:
     
     ```
-    [json.exception.parse_error.101] parse error at line 3, column 9:
-    syntax error while parsing object key - invalid literal;
-    last read: '<U+000A>    {<U+000A>        /'; expected string literal
-    ```
-    
-    ```json
-    {
-      "planets": [
-        "Mercury",
-        "Venus",
-        "Earth",
-        "Mars",
-        "Jupiter",
-        "Uranus",
-        "Neptune"
-      ]
-    }
+    --8<-- "examples/comments.output"
     ```
--- a/docs/mkdocs/docs/features/trailing_commas.md
+++ b/docs/mkdocs/docs/features/trailing_commas.md
@@ -1,0 +1,39 @@
+# Trailing Commas
+
+Like [comments](comments.md), this library does not support trailing commas in arrays and objects *by default*.
+
+You can set parameter `ignore_trailing_commas` to `#!cpp true` in the [`parse`](../api/basic_json/parse.md) function to allow trailing commas in arrays and objects. Note that a single comma as the only content of the array or object (`[,]` or `{,}`) is not allowed, and multiple trailing commas (`[1,,]`) are not allowed either.
+
+This library does not add trailing commas when serializing JSON data.
+
+For more information, see [JSON With Commas and Comments (JWCC)](https://nigeltao.github.io/blog/2021/json-with-commas-comments.html).
+
+!!! example
+
+    Consider the following JSON with trailing commas.
+
+    ```json
+    {
+        "planets": [
+            "Mercury",
+            "Venus",
+            "Earth",
+            "Mars",
+            "Jupiter",
+            "Uranus",
+            "Neptune",
+        ]
+    }
+    ```
+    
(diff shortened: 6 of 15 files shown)


README.md
```diff
@@ -1765,6 +1765,14 @@ This library does not support comments by default. It does so for three reasons:
 
 However, you can pass set parameter `ignore_comments` to true in the `parse` function to ignore `//` or `/* */` comments. Comments will then be treated as whitespace.
 
+### Trailing Commas
+
+Trailing commas in arrays and objects are alse not part of the [JSON specification](https://tools.ietf.org/html/rfc8259), and this library does not support it by default.
```
gregmarr: typo: alse

README.md
```diff
@@ -1765,6 +1765,14 @@ This library does not support comments by default. It does so for three reasons:
...
+
+Like comments, you can pass set parameter `allow_trailing_commas` to true in the `parse` function to allow trailing commas in arrays and objects. Trailing commas will then be ignored.
```
gregmarr: If this is added, should this be called `ignore_trailing_commas` to match `ignore_comments`?

docs/mkdocs/docs/features/trailing_commas.md
```diff
@@ -0,0 +1,89 @@
+# Trailing Commas
+
+Like comments, this library does not support trailing commas in arrays and objects *by default*.
+
+You can pass set parameter `allow_trailing_commas` to `#!c true` in the parse function to allow trailing commas in arrays and objects. Trailing commas are ignored during parsing.
```
gregmarr: Either "pass" or "set", not both.  `#!cpp` everywhere else.

docs/mkdocs/docs/features/trailing_commas.md
```diff
@@ -0,0 +1,89 @@
...
+
+Note that this library does not add trailing commas when serializing JSON data. Even when you set `allow_trailing_commas` to true, `{,}` and `[,]` are not valid JSON.
```
gregmarr: I think this second sentence belongs as part of the previous paragraph. Something like

>  ...allow trailing commas after the last element in arrays or objects.  A single comma as the only content of the array or object (`[,]` or `{,}`)  is not allowed."

> are not valid JSON.

In any case, this is not valid JSON, but an extension, or a different file format (JWCC), so I wouldn't use this description.

I would file this whole thing under Postel's law: "Be liberal in what you accept, and conservative in what you send."

README.md
chirsz-ever: I choose "allow" because I think that commas are not like comments: successive commas are invalid (`[1,,]`), but successive comments are valid (`[1 /*, 2*/ /*, 3*/]`).

Do you think `ignore_trailing_commas` is better? If you really think so, I would rename it, or I would try to avoid using "ignore" when describing trailing commas.

docs/mkdocs/docs/features/trailing_commas.md
chirsz-ever: I refer to `comments.md`. If it's wrong here, it's alse wrong there...

README.md
gregmarr: I don't see a difference in meaning between `ignore` and `allow`, and so this looks to me like a needless difference, but @nlohmann has the final say here.

docs/mkdocs/docs/features/trailing_commas.md
gregmarr: You are right, the "pass set" part is wrong in both places.

I don't know that the tagging of `true` is necessarily wrong, I just noticed that it was different than other parts of the PR.

docs/mkdocs/docs/features/trailing_commas.md
```diff
@@ -0,0 +1,88 @@
...
+        }
+        
+        json j = json::parse(s,
+                             /* callback */ nullptr,
+                             /* allow exceptions */ true,
+                             /* ignore_comments */ false,
+                             /* allow_trailing_commas */ true);
+        std::cout << j.dump(2) << '\n';
+    }
+    ```
```
nlohmann: Move this code into an example file (docs/mkdocs/docs/example). The documentation generation will then compile it and create the output. You can then include both in the documentation via

```markdown
--8<-- "examples/meta.output"
```

(see other documentation files for examples)

This has the advantage of making sure the example remains compilable in the future.

README.md
```diff
@@ -1765,6 +1765,14 @@ This library does not support comments by default. It does so for three reasons:
 
 However, you can pass set parameter `ignore_comments` to true in the `parse` function to ignore `//` or `/* */` comments. Comments will then be treated as whitespace.
 
+### Trailing Commas
```
nlohmann:
```suggestion
### Trailing commas
```

nlohmann (changes requested):

nlohmann:
Maybe the documentation should somewhere also mention JWCC - it's good when things have a name.

chirsz-ever:
How do you think about combining the two options `ignore_comments` and `ignore_trailing_commas` into a single option to replace the current `ignore_comments`? For example, after the update, `parse(str, nullptr, true, true)` would allow both comments and trailing commas.

We can name the new combined option `allow_comments_and_trailing_commas` or `enable_jwcc` or `relaxed_syntax` or another better name.

I think that any user who choose to allow comments will not mind allowing trailing commas, and it is hard to imagine a use case where one _must_ be enabled and the other disabled.

Regarding compatibility issues, this will not affect the serialization results, nor will it refuse to accept inputs accepted by previous versions. It will just accept some inputs that would not have been accepted before.

This solution is also conducive to adding new features in the future. But for me this is not important, because I only support adding two features of JWCC.

chirsz-ever:
> Maybe the documentation should somewhere also mention JWCC - it's good when things have a name.

Can I declare that this library would never add extensions (such as allow `NaN` as number, single quoted string, unquoted identifiers as object keys ...) other than JWCC in the future?

nlohmann:
> > Maybe the documentation should somewhere also mention JWCC - it's good when things have a name.
> 
> Can I declare that this library would never add extensions (such as allow `NaN` as number, single quoted string, unquoted identifiers as object keys ...) other than JWCC in the future?

Never say never...

nlohmann:
> How do you think about combining the two options `ignore_comments` and `ignore_trailing_commas` into a single option to replace the current `ignore_comments`? For example, after the update, `parse(str, nullptr, true, true)` would allow both comments and trailing commas.
> 
> We can name the new combined option `allow_comments_and_trailing_commas` or `enable_jwcc` or `relaxed_syntax` or another better name.
> 
> I think that any user who choose to allow comments will not mind allowing trailing commas, and it is hard to imagine a use case where one _must_ be enabled and the other disabled.
> 
> Regarding compatibility issues, this will not affect the serialization results, nor will it refuse to accept inputs accepted by previous versions. It will just accept some inputs that would not have been accepted before.
> 
> This solution is also conducive to adding new features in the future. But for me this is not important, because I only support adding two features of JWCC.

No, please leave it as is as it would change the behavior of existing code. Just like with the `dump` function, we need to figure out how to organize the great number of parameters in the future.

docs/mkdocs/docs/examples/comments.cpp
```diff
@@ -0,0 +1,31 @@
...
+    {
+        std::cout << e.what() << std::endl;
+    }
+    
+    json j = json::parse(s,
+                            /* callback */ nullptr,
+                            /* allow exceptions */ true,
+                            /* ignore_comments */ true);
+    std::cout << j.dump(2) << '\n';
+}
```
nlohmann: Newline at the end of the file is missing.

docs/mkdocs/docs/examples/trailing_commas.cpp
```diff
@@ -0,0 +1,37 @@
...
+        std::cout << e.what() << std::endl;
+    }
+    
+    json j = json::parse(s,
+                            /* callback */ nullptr,
+                            /* allow exceptions */ true,
+                            /* ignore_comments */ false,
+                            /* ignore_trailing_commas */ true);
+    std::cout << j.dump(2) << '\n';
+}
```
nlohmann: Newline at the end of the file is missing.

docs/mkdocs/docs/features/comments.md
```diff
@@ -28,56 +28,11 @@ However, you can pass set parameter `ignore_comments` to `#!c true` in the parse
...
-            std::cout << e.what() << std::endl;
-        }
-        
-        json j = json::parse(s,
-                             /* callback */ nullptr,
-                             /* allow exceptions */ true,
-                             /* ignore_comments */ true);
-        std::cout << j.dump(2) << '\n';
-    }
+    --8<-- "examples/comments.cpp"
```
nlohmann, line 33: Please also add this example to the documentation of the `parse` function in `parse.md`.

docs/mkdocs/docs/features/trailing_commas.md
```diff
@@ -0,0 +1,37 @@
...
+            "Jupiter",
+            "Uranus",
+            "Neptune",
+        ]
+    }
+    ```
+    
+    When calling `parse` without additional argument, a parse error exception is thrown. If `ignore_trailing_commas` is set to `#! true`, the trailing commas are ignored during parsing:
+
+    ```cpp
```
nlohmann, line 31: Please also add this example to the documentation of the `parse` function in `parse.md`.

nlohmann (changes requested):

chirsz-ever:
I combine the documentation about comments and trailing commas. Any more suggestions?

nlohmann:
> I combine the documentation about comments and trailing commas. Any more suggestions?

No, please leave it as is. We may combine this in the future, but right now, I would like the URLs of the documentation to not change (https://www.w3.org/Provider/Style/URI).

chirsz-ever:
> > I combine the documentation about comments and trailing commas. Any more suggestions?
> 
> No, please leave it as is. We may combine this in the future, but right now, I would like the URLs of the documentation to not change (https://www.w3.org/Provider/Style/URI).

Could we just add a redirect rule to `docs/mkdocs/mkdocs.yml` as below?

```diff
       redirect_maps:
         'api/basic_json/operator_gtgt.md': api/operator_gtgt.md
         'api/basic_json/operator_ltlt.md': api/operator_ltlt.md
         'api/basic_json/operator_literal_json.md': api/operator_literal_json.md
         'api/basic_json/operator_literal_json_pointer.md': api/operator_literal_json_pointer.md
         'api/json_pointer/operator_string.md': api/json_pointer/operator_string_t.md
         'home/code_of_conduct.md': community/code_of_conduct.md
+        'features/comments.md': features/comments_and_trailing_commas.md
```

nlohmann:
> > > I combine the documentation about comments and trailing commas. Any more suggestions?
> > 
> > 
> > No, please leave it as is. We may combine this in the future, but right now, I would like the URLs of the documentation to not change (https://www.w3.org/Provider/Style/URI).
> 
> Could we just add a redirect rule to `docs/mkdocs/mkdocs.yml` as below?
> 
> ```diff
>        redirect_maps:
>          'api/basic_json/operator_gtgt.md': api/operator_gtgt.md
>          'api/basic_json/operator_ltlt.md': api/operator_ltlt.md
>          'api/basic_json/operator_literal_json.md': api/operator_literal_json.md
>          'api/basic_json/operator_literal_json_pointer.md': api/operator_literal_json_pointer.md
>          'api/json_pointer/operator_string.md': api/json_pointer/operator_string_t.md
>          'home/code_of_conduct.md': community/code_of_conduct.md
> +        'features/comments.md': features/comments_and_trailing_commas.md
> ```

Yes, but please revert anyway.

README.md
```diff
@@ -1763,7 +1763,17 @@ This library does not support comments by default. It does so for three reasons:
   
 3. It is dangerous for interoperability if some libraries would add comment support while others don't. Please check [The Harmful Consequences of the Robustness Principle](https://tools.ietf.org/html/draft-iab-protocol-maintenance-01) on this.
 
-However, you can pass set parameter `ignore_comments` to true in the `parse` function to ignore `//` or `/* */` comments. Comments will then be treated as whitespace.
+However, you can set set parameter `ignore_comments` to true in the `parse` function to ignore `//` or `/* */` comments. Comments will then be treated as whitespace.
+
+### Trailing commas
+
+Trailing commas in arrays and objects are also not part of the [JSON specification](https://tools.ietf.org/html/rfc8259), and this library does not support it by default.
```
nlohmann: Maybe nitpicking, but "trailing commas" are not a feature, but to ignore them is. So shouldn't this sentence read

> The JSON specification does not allow trailing commas in arrays and objects, and hence this library is treating them as parsing errors.

README.md
```diff
@@ -1763,7 +1763,17 @@ This library does not support comments by default. It does so for three reasons:
...
+
+Like comments, you can set parameter `ignore_trailing_commas` to true in the `parse` function to ignore trailing commas in arrays and objects. Note that a single comma as the only content of the array or object (`[,]` or `{,}`) is not allowed, and multiple trailing commas (`[1,,]`) are not allowed either.
+
+This library does not add trailing commas when serializing JSON data.
+
+For more information, see [JWCC](https://nigeltao.github.io/blog/2021/json-with-commas-comments.html) (JSON With Commas and Comments).
```
nlohmann:
```suggestion
For more information, see [JSON With Commas and Comments (JWCC)](https://nigeltao.github.io/blog/2021/json-with-commas-comments.html).
```

docs/mkdocs/docs/api/basic_json/accept.md
```diff
@@ -50,6 +52,10 @@ Unlike the [`parse()`](parse.md) function, this function neither throws an excep
 :   whether comments should be ignored and treated like whitespace (`#!cpp true`) or yield a parse error
     (`#!cpp false`); (optional, `#!cpp false` by default)
 
+`ignore_trailing_commas` (in)
+:   whether trailing commas in arrays or objects should be allowed (`#!cpp true`) or yield a parse error
```
nlohmann:
```suggestion
:   whether trailing commas in arrays or objects should be ignored and treated like whitespace (`#!cpp true`) or yield a parse error
```

(Same wording as for `ignore_comments`, also in other functions)

docs/mkdocs/docs/examples/comments.cpp
```diff
@@ -0,0 +1,31 @@
...
+    }
+    catch (json::exception& e)
+    {
+        std::cout << e.what() << std::endl;
+    }
+
+    json j = json::parse(s,
+                         /* callback */ nullptr,
+                         /* allow exceptions */ true,
+                         /* ignore_comments */ true);
```
nlohmann, line 29: Why is `ignore_comments` set to `true` here?

docs/mkdocs/docs/features/trailing_commas.md
```diff
@@ -0,0 +1,39 @@
+# Trailing Commas
+
+Like comments, this library does not support trailing commas in arrays and objects *by default*.
```
nlohmann:
```suggestion
Like [comments](comments.md), this library does not support trailing commas in arrays and objects *by default*.
```

docs/mkdocs/docs/examples/comments.cpp
chirsz-ever: Because this is the example code of `ignore_comments`...

nlohmann: Oh, sorry. Mixed the files. You're right.

barroca (approved):
Nice new feature. :)

confluence:
Hi! Is there any ETA for this PR to be merged? Are any of the requested changes still outstanding? I'm very interested in using this feature in a work project (it would make it much easier for us to create a nice human-readable default configuration file). Thanks for everyone's work on this library!

svnscha:
> Hi! Is there any ETA for this PR to be merged? Are any of the requested changes still outstanding? I'm very interested in using this feature in a work project (it would make it much easier for us to create a nice human-readable default configuration file). Thanks for everyone's work on this library!

+1

helgeklein:
Hi there, we would also need this. Is there an ETA?

barroca:
@nlohmann are you ok with the changes now?

nlohmann (changes requested):
An entry like "- Added `ignore_trailing_commas` in version 3.12.1." is missing in the "Version history" section of docs/mkdocs/docs/api/basic_json/sax_parse.md.

nlohmann:
> @nlohmann are you ok with the changes now?

Yes, I just added a minor remark on the documentation.

chirsz-ever:
> An entry like "- Added `ignore_trailing_commas` in version 3.12.1." is missing in the "Version history" section of docs/mkdocs/docs/api/basic_json/sax_parse.md.

Done.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #4612, opened 2025-01-20
add strict enum de/serialization macro

hnampally:
#3992

- [X] The changes are described in detail, both the what and why.
- [X] If applicable, an [#3992 ](https://github.com/nlohmann/json/issues) is referenced.
- [X] If applicable, the [documentation](https://json.nlohmann.me) is updated.
- [X] The source code is amalgamated by running `make amalgamate`.

### DIFF
--- a/docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
+++ b/docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
@@ -4,7 +4,7 @@
 #define NLOHMANN_JSON_SERIALIZE_ENUM_STRICT(type, conversion...)
 ```
 
-The `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` allows to define a user-defined serialization for every enumerator.
+The `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` macro allows defining a user-defined serialization for every enumerator.
 
 This macro declares strict serialization and deserialization functions (`to_json` and `from_json`) for an enum type. 
 Unlike [`NLOHMANN_JSON_SERIALIZE_ENUM`](nlohmann_json_serialize_enum.md), this macro enforces strict validation and 
@@ -16,7 +16,7 @@ throws errors for unmapped values instead of defaulting to the first enum value.
 :   name of the enum to serialize/deserialize
 
 `conversion` (in)
-:   a pair of an enumerator and a JSON serialization; arbitrary pairs can be given as a comma-separated list
+:   A list of parameters alternating between an enumerator value and a string to use in the JSON serialization.
 
 ## Default definition
 
@@ -37,10 +37,9 @@ inline void from_json(const BasicJsonType& j, type& e);
 
 !!! important "Important notes"
 
-    - If an enum value appears more than once in the mapping, only the first occurrence will be used for serialization, 
-        subsequent mappings for the same enum value will be ignored.
-    - If a JSON value appears more than once in the mapping, only the first occurrence will be used for deserialization, 
-        subsequent mappings for the same JSON value will be ignored.
+    - Duplicate enum or JSON values in the mapping are not supported. Only the first occurrence will be used, 
+    and subsequent mappings will be ignored. 
+    This behavior is currently tolerated but may become an error in future versions.
     - Unlike `NLOHMANN_JSON_SERIALIZE_ENUM`, this macro enforces strict validation:
         - Attempting to serialize an unmapped enum value will throw a `type_error.302` exception
         - Attempting to deserialize an unmapped JSON value will throw a `type_error.302` exception
--- a/docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
+++ b/docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
@@ -59,7 +59,7 @@ inline void from_json(const BasicJsonType& j, type& e);
     Expected output:
     
     ```
-    [json.exception.type_error.302] can't serialize - enum value 3 out of range
+    [json.exception.type_error.302] serialization failed: enum value 3 is out of range
     ```
 
 ??? example "Example 2: Strict deserialization"
@@ -73,7 +73,7 @@ inline void from_json(const BasicJsonType& j, type& e);
     Expected output:
     
     ```
-    [json.exception.type_error.302] can't deserialize - invalid json value : "yellow"
+    [json.exception.type_error.302] deserialization failed: invalid JSON value "yellow"
     ```
 
 Both examples demonstrate:
--- a/docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
+++ b/docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
@@ -6,8 +6,9 @@
 
 The `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` allows to define a user-defined serialization for every enumerator.
 
-This macro declares strict serialization and deserialization functions (`to_json` and `from_json`) for an enum type. Unlike [`NLOHMANN_JSON_SERIALIZE_ENUM`](nlohmann_json_serialize_enum.md), this macro enforces strict validation and throws errors for unmapped values instead of defaulting to the first enum value.
-
+This macro declares strict serialization and deserialization functions (`to_json` and `from_json`) for an enum type. 
+Unlike [`NLOHMANN_JSON_SERIALIZE_ENUM`](nlohmann_json_serialize_enum.md), this macro enforces strict validation and 
+throws errors for unmapped values instead of defaulting to the first enum value.
 
 ## Parameters
 
@@ -36,12 +37,14 @@ inline void from_json(const BasicJsonType& j, type& e);
 
 !!! important "Important notes"
 
-- If an enum value appears more than once in the mapping, only the first occurrence will be used for serialization, subsequent mappings for the same enum value will be ignored.
-- If a JSON value appears more than once in the mapping, only the first occurrence will be used for deserialization, subsequent mappings for the same JSON value will be ignored.
-- Unlike `NLOHMANN_JSON_SERIALIZE_ENUM`, this macro enforces strict validation:
-  - Attempting to serialize an unmapped enum value will throw a `type_error.302` exception
-  - Attempting to deserialize an unmapped JSON value will throw a `type_error.302` exception
-  - There is no default value behavior - all values must be explicitly mapped
+    - If an enum value appears more than once in the mapping, only the first occurrence will be used for serialization, 
+        subsequent mappings for the same enum value will be ignored.
+    - If a JSON value appears more than once in the mapping, only the first occurrence will be used for deserialization, 
+        subsequent mappings for the same JSON value will be ignored.
+    - Unlike `NLOHMANN_JSON_SERIALIZE_ENUM`, this macro enforces strict validation:
+        - Attempting to serialize an unmapped enum value will throw a `type_error.302` exception
+        - Attempting to deserialize an unmapped JSON value will throw a `type_error.302` exception
+        - There is no default value behavior - all values must be explicitly mapped
 
 ## Examples
 
--- a/docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
+++ b/docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
@@ -90,4 +90,4 @@ Both examples demonstrate:
 
 ## Version history
 
-- Added in version 3.11.3.
+- Added in version 3.11.4
--- a/docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
+++ b/docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
@@ -1,0 +1,90 @@
+# NLOHMANN_JSON_SERIALIZE_ENUM_STRICT
+
+```cpp
+#define NLOHMANN_JSON_SERIALIZE_ENUM_STRICT(type, conversion...)
+```
+
+The `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` allows to define a user-defined serialization for every enumerator.
+
+This macro declares strict serialization and deserialization functions (`to_json` and `from_json`) for an enum type. Unlike [`NLOHMANN_JSON_SERIALIZE_ENUM`](nlohmann_json_serialize_enum.md), this macro enforces strict validation and throws errors for unmapped values instead of defaulting to the first enum value.
+
+
+## Parameters
+
+`type` (in)
+:   name of the enum to serialize/deserialize
+
+`conversion` (in)
+:   a pair of an enumerator and a JSON serialization; arbitrary pairs can be given as a comma-separated list
+
+## Default definition
+
+The macro adds two functions to the namespace which take care of the serialization and deserialization:
+
+```cpp
+template<typename BasicJsonType>
+inline void to_json(BasicJsonType& j, const type& e);
+template<typename BasicJsonType>
+inline void from_json(const BasicJsonType& j, type& e);
+```
+
+## Notes
+
+!!! info "Prerequisites"
+
+    The macro must be used inside the namespace of the enum.
+
+!!! important "Important notes"
+
+- If an enum value appears more than once in the mapping, only the first occurrence will be used for serialization, subsequent mappings for the same enum value will be ignored.
+- If a JSON value appears more than once in the mapping, only the first occurrence will be used for deserialization, subsequent mappings for the same JSON value will be ignored.
+- Unlike `NLOHMANN_JSON_SERIALIZE_ENUM`, this macro enforces strict validation:
+  - Attempting to serialize an unmapped enum value will throw a `type_error.302` exception
+  - Attempting to deserialize an unmapped JSON value will throw a `type_error.302` exception
+  - There is no default value behavior - all values must be explicitly mapped
+
+## Examples
+
+??? example "Example 1: Strict serialization"
+
+    The example shows how `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` enforces strict validation when serializing an enum value that is not in the mapping:
+
+    ```cpp
+    --8<-- "examples/nlohmann_json_serialize_enum_strict.cpp"
+    ```
+
+    Expected output:
+    
+    ```
+    [json.exception.type_error.302] can't serialize - enum value 3 out of range
+    ```
+
+??? example "Example 2: Strict deserialization"
+
+    The example shows how `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` enforces strict validation when deserializing a JSON value that is not in the mapping:
+
+    ```cpp
+    --8<-- "examples/nlohmann_json_deserialize_enum_strict.cpp"
+    ```
+
+    Expected output:
+    
+    ```
+    [json.exception.type_error.302] can't deserialize - invalid json value : "yellow"
+    ```
+
+Both examples demonstrate:
+
+- Proper error handling using try-catch blocks
+- Clear error messages indicating the cause of failure
+- No default value behavior - all values must be explicitly mapped
+- Exception throwing for unmapped values
+
+## See also
+
+- [Specializing enum conversion](../../features/enum_conversion.md)
+- [`JSON_DISABLE_ENUM_SERIALIZATION`](json_disable_enum_serialization.md)
+
+## Version history
+
+- Added in version 3.11.3.
--- a/docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
+++ b/docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
@@ -1,26 +1,11 @@
 #include <iostream>
 #include <nlohmann/json.hpp>
 
-#ifdef __cpp_exceptions
-    #undef __cpp_exceptions
-    #define __cpp_exceptions 1
-#endif
-
-#ifdef JSON_NOEXCEPTION
-    #define JSON_NOEXCEPTION 0
-#endif
-
 #if (defined(__cpp_exceptions) || defined(__EXCEPTIONS) || defined(_CPPUNWIND)) && !defined(JSON_NOEXCEPTION)
     #define JSON_THROW(exception) throw exception
-    #define JSON_TRY try
-    #define JSON_CATCH(exception) catch(exception)
-    #define JSON_INTERNAL_CATCH(exception) catch(exception)
 #else
     #include <cstdlib>
     #define JSON_THROW(exception) std::abort()
-    #define JSON_TRY if(true)
-    #define JSON_CATCH(exception) if(false)
-    #define JSON_INTERNAL_CATCH(exception) if(false)
 #endif
 
 using json = nlohmann::json;
--- a/docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
+++ b/docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
@@ -20,9 +20,8 @@ NLOHMANN_JSON_SERIALIZE_ENUM_STRICT(Color,
 
 int main()
 {
-
-    // deserialization
     json j_yellow = "yellow";
+    // deserialization
     try
     {
         auto yellow = j_yellow.template get<ns::Color>();
--- a/docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
+++ b/docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
@@ -1,0 +1,57 @@
+#include <iostream>
+#include <nlohmann/json.hpp>
+
+#ifdef __cpp_exceptions
+    #undef __cpp_exceptions
+    #define __cpp_exceptions 1
+#endif
+
+#ifdef JSON_NOEXCEPTION
+    #define JSON_NOEXCEPTION 0
+#endif
+
+#if (defined(__cpp_exceptions) || defined(__EXCEPTIONS) || defined(_CPPUNWIND)) && !defined(JSON_NOEXCEPTION)
+    #define JSON_THROW(exception) throw exception
+    #define JSON_TRY try
+    #define JSON_CATCH(exception) catch(exception)
+    #define JSON_INTERNAL_CATCH(exception) catch(exception)
+#else
+    #include <cstdlib>
+    #define JSON_THROW(exception) std::abort()
+    #define JSON_TRY if(true)
+    #define JSON_CATCH(exception) if(false)
+    #define JSON_INTERNAL_CATCH(exception) if(false)
+#endif
+
+using json = nlohmann::json;
+
+namespace ns
+{
+enum class Color
+{
+    red, green, blue
+};
+
+NLOHMANN_JSON_SERIALIZE_ENUM_STRICT(Color,
+{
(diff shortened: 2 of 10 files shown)


tests/src/unit-conversions.cpp
```diff
@@ -1657,6 +1657,94 @@ TEST_CASE("JSON to enum mapping")
...
+#ifdef __cpp_exceptions
+    #undef __cpp_exceptions
+    #define __cpp_exceptions 1
+#endif
+
+#ifdef JSON_NOEXCEPTION
+    #define JSON_NOEXCEPTION 0
+#endif
+
+#if (defined(__cpp_exceptions) || defined(__EXCEPTIONS) || defined(_CPPUNWIND)) && !defined(JSON_NOEXCEPTION)
```
hnampally, line 1669: for some reason, if I don't have this defined I get the following compilation errors
image

docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
```diff
@@ -0,0 +1,57 @@
...
+#ifdef __cpp_exceptions
+    #undef __cpp_exceptions
+    #define __cpp_exceptions 1
+#endif
+
+#ifdef JSON_NOEXCEPTION
+    #define JSON_NOEXCEPTION 0
+#endif
+
+#if (defined(__cpp_exceptions) || defined(__EXCEPTIONS) || defined(_CPPUNWIND)) && !defined(JSON_NOEXCEPTION)
```
hnampally, line 13: these lines 13 to 24 are copied from json.hpp; this is required as a workaround to get through the following compilation issues
Screenshot from 2025-01-20 00-13-05
@nlohmann  @gregmarr @codenut please advise the proper way to resolve this compilation issue.

hnampally: Compiler: c++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0;

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
```diff
@@ -0,0 +1,93 @@
...
+
+!!! info "Prerequisites"
+
+    The macro must be used inside the namespace of the enum.
+
+!!! important "Important notes"
+
+    - If an enum value appears more than once in the mapping, only the first occurrence will be used for serialization, 
+        subsequent mappings for the same enum value will be ignored.
+    - If a JSON value appears more than once in the mapping, only the first occurrence will be used for deserialization,
```
nlohmann: Just a thought: shouldn't a strict macro also take care of this?

docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
```diff
@@ -0,0 +1,42 @@
+#include <iostream>
+#include <nlohmann/json.hpp>
+
+#if !defined(JSON_NOEXCEPTION) && !defined(JSON_THROW_USER) && !defined(JSON_THROW)
```
nlohmann: Is this needed? The other examples work with exceptions without this.

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
hnampally: Thank you for catching that! Yes, that makes sense. Could you clarify which exception should be thrown in this case?

docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
hnampally: Screenshot from 2025-01-19 20-47-22
Yes, for some reason, I encountered a compilation issue without this. I noticed that JSON_THROW is already defined in json.hpp, but the issue persists. Could you let me know if I'm overlooking something?

hnampally: It looks like it is being undefined in macro_unscope.hpp at the end of json.hpp. Please find the relevant line at [nlohmann/json.hpp](https://github.com/nlohmann/json/blob/8c7dcd3b436dc72d912f525482cd8bf1afcd15a0/include/nlohmann/json.hpp#L5304).

hnampally: I noticed that other examples, such as nlohmann_json_serialize_enum_2.cpp, do not encounter this compilation issue because they use macro functions like NLOHMANN_JSON_SERIALIZE_ENUM, which do not throw exceptions. In macro_scope.hpp, NLOHMANN_JSON_SERIALIZE_ENUM_STRICT is the first macro to trigger the JSON_THROW exception. Therefore, I believe I need to redefine it in this example code.

nlohmann: No, this definitely needs to be fixed - a user must not worry about any macros for exceptions. In this case, the exception from `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` should be thrown via a plain `throw`. The `JSON_THROW` is only used for exceptions that can occur inside of the library whereas it makes no sense for a user to use a different throw mechanism for the the exception in `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT`.

nlohmann (changes requested):

docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
gregmarr: On the other hand, if you don't use `JSON_THROW`, then `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` is not usable with exceptions disabled.  An alternative solution to having the macro escape the library could be to have an internal function like this just for this particular usage of throwing from inside one of the macros:

```
    template<typename T> 
    void json_throw_from_serialize_macro(T&& exception) 
    {
        JSON_THROW std::forward<T>(exception);
    }
```

hnampally: @nlohmann Yes, I completely agree. I will give Greg's proposed solution a try.
@gregmarr thanks for your code snippet.

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
nlohmann: That's a good question! As C++11 most probably does not allow to detect such duplicates at compile time and an exception could only be thrown the first time the `from_json` function is used, such an exception would be rather unexpected. I therefore think we should leave this as is.

gregmarr: It should be possible to do a `static_assert` that there are no duplicate values in this array, but I'm not sure what the compile time cost would be.  The hardest part would be supporting that in C++11 where `constexpr` is pretty limited.

```
       static const std::pair<ENUM_TYPE, BasicJsonType> m[] = __VA_ARGS__;
```

nlohmann: Though intellectually interesting, the question is at what price such a check would come.

gregmarr: > Though intellectually interesting, the question is at what price such a check would come.

It eliminates the need for this documentation, as it then becomes a compile-time error:

> If an enum value appears more than once in the mapping, ...
> If a JSON value appears more than once in the mapping, ...

The price would completely be compile time.  Since this is specifically strict enum support, I think it would probably be reasonable for someone to think that the library should test this, and the use of the macro is completely opt-in.

It would basically be an O(N^2) operation so hopefully there aren't a lot of enum entries.  You'd need to iterate over the array, and for each element, iterate over the array again, and check that when the two iterators aren't equal, the members of the pairs aren't equal either.

nlohmann: Understood. With "price" I meant how C++11-compliant code for this check would look like.

gregmarr: Unfortunately, I can only get this to work in C++20, as in order to `static_assert` that there are no duplicates in the array, the array has to be `constexpr` and `constexpr` prior to C++20 doesn't support `std::pair<ENUM_TYPE, BasicJsonType> m[] = ...`.

nlohmann: I understand. And I guess there is no way to map the pairs to only cope with an array of `ENUM_TYPE`?

gregmarr: For posterity, the implementation that requires C++20:

```
template<typename Iter>
constexpr bool enum_array_has_no_duplicates(Iter begin, Iter end)
{
    for (auto it1 = begin; it1 != end; it1++)
    {
        for (auto it2 = it1 + 1; it2 != end; it2++)
        {
           if (it1->first == it2->first || it1->second == it2->second)
           {
            return false;
           } 
        }
    }

    return true;
}

    static constexpr std::pair<ENUM_TYPE, BasicJsonType> m[] = { ...

    static_assert(enum_array_has_no_duplicates(std::begin(m), std::end(m)));
```

gregmarr: That code prevents this definition:

```
enum Color
{
    RED = 1,
    Red = 1,
    Green = 2
};

NLOHMANN_JSON_SERIALIZE_ENUM_STRICT(Color,
{
    {RED, "RED"},
    {Red, "Red"}, // fails due to matching RED and Red
    {Green, "RED"} // fails to matching "RED" and "RED"
})
```

gregmarr: > I understand. And I guess there is no way to map the pairs to only cope with an array of ENUM_TYPE?

It should check both `ENUM_TYPE` and `BasicJsonType` duplication because there is mapping in both directions.  

This could be a C++20-only feature, but that would be messy, because it's not just the added function and static assert, but you also have to make `m` either `const` or `constexpr` .  That could maybe be done with a `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT_CONSTEXPR` that is `constexpr` in C++20 and `const` in earlier versions, and use that instead of `const` on the definition of `m`.

It could make three different arrays, one of pairs, one of enums, and one of jsons, but you'd need to remove all the braces and do the alternating between enum and json that is done in the new named serialization macro, which means that you'd have limits on the number of enums as you have there.

hnampally: @gregmarr Thank you for sharing this idea and code snippet. NLOHMANN_JSON_SERIALIZE_ENUM_STRICT_CONSTEXPR sounds great to me. Do you think it would be better to implement this in a separate PR?

nlohmann: Definitely not in this one, and I am not sure about a C++20-only feature.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -17,6 +17,36 @@
 
 #include <nlohmann/detail/abi_macros.hpp>
 
+NLOHMANN_JSON_NAMESPACE_BEGIN
+namespace detail
+{
+template<typename T>
+[[noreturn]] inline void throw_if_exceptions_enabled(T&& exception)
+{
+#if defined(__cpp_exceptions) || defined(__EXCEPTIONS) || defined(_CPPUNWIND) || defined(EXCEPTIONS)
```
gregmarr: Can you just use `JSON_THROW(std::forward<T>(exception));`, with perhaps a `static_cast<void>(exception);` before it for when they're disabled?

hnampally: Do you mean something like this?
`template<typename T>
[[noreturn]] inline void throw_if_exceptions_enabled(T&& exception)
{
    JSON_THROW(std::forward<T>(exception));
// #if defined(__cpp_exceptions) || defined(__EXCEPTIONS) || defined(_CPPUNWIND) || defined(EXCEPTIONS)
//     throw std::forward<T>(exception);
// #else
//     // Forward the exception (even if unused) and abort
//     std::forward<T>(exception);
//     std::abort();
// #endif
}
} // namespace detail
`
Above code is failing to compile with following error
`/home/hnampally/json/tests/src/unit-conversions.cpp:1692:9:   required from here
/home/hnampally/json/include/nlohmann/detail/macro_scope.hpp:26:15: error: ‘JSON_THROW’ was not declared in this scope; did you mean ‘JSON_TRY’?
   26 |     JSON_THROW(std::forward<T>(exception));
      |     ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     JSON_TRY
`
I think the issue arises because JSON_THROW is defined in macro_scope.hpp and then undefined at the end of json.hpp in macro_unscope.hpp. As a result, JSON_THROW within the macro is not visible in unit_conversions.cpp

docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
gregmarr: This can go away now, right?

docs/mkdocs/docs/examples/nlohmann_json_serialize_enum_strict.cpp
```diff
@@ -0,0 +1,41 @@
+#include <iostream>
+#include <nlohmann/json.hpp>
+
+#if !defined(JSON_NOEXCEPTION) && !defined(JSON_THROW_USER) && !defined(JSON_THROW)
```
gregmarr: This one can also go away.

docs/mkdocs/docs/features/enum_conversion.md
```diff
@@ -59,3 +59,22 @@ Other Important points:
 - If an enum or JSON value is specified more than once in your map, the first matching occurrence from the top of the
   map will be returned when converting to or from JSON.
 - To disable the default serialization of enumerators as integers and force a compiler error instead, see [`JSON_DISABLE_ENUM_SERIALIZATION`](../api/macros/json_disable_enum_serialization.md).
+
+An alternative macro [`NLOHMANN_JSON_SERIALIZE_ENUM_STRICT()` macro](../api/macros/nlohmann_json_serialize_enum.md) can be used when a more strict error handling is preffered, throwing in case of serialization errors instead of defaulting to the first enum value defined in the macro.
```
gregmarr: typo: preffered should be preferred

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -17,6 +17,20 @@
 
 #include <nlohmann/detail/abi_macros.hpp>
 
+
+// exclude unsupported compilers
+#if !defined(JSON_SKIP_UNSUPPORTED_COMPILER_CHECK)
```
gregmarr: Is this new or moved here?  If new, do we want to not allow these compilers to be used or just not support them?

hnampally: I think it was moved by 'make amalgamate' or something

gregmarr: Oh, it's duplicated.  It appears again right below.

hnampally: oh yeah, I just realized that, thanks for catching.

hnampally: for some reason it is not showing up in my gitk diff!
image

hnampally: anyway sorry for the confusion, I will remove it.

include/nlohmann/detail/macro_scope.hpp
hnampally: When I attempted JSON_THROW, I encountered a "cppcoreguidelines-missing-std-forward,-warnings-as-errors" issue specifically when DJSON_NOEXCEPTION is defined. I'm uncertain about the most effective resolution for this scenario.

However, the following approach has successfully passed all CI checks.
```
template<typename T>
[[noreturn]] inline void throw_if_exceptions_enabled(T&& exception)
{
#if defined(__cpp_exceptions) || defined(__EXCEPTIONS) || defined(_CPPUNWIND) || defined(EXCEPTIONS)
    throw std::forward<T>(exception);
#else
    // Forward the exception (even if unused) and abort
    std::forward<T>(exception);
    std::abort();
#endif
}
```

gregmarr: Ah, fun with clang tidy.  It's probably not worth fighting with it.  :)

docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
gregmarr: Just this remaining and then LGTM.

include/nlohmann/detail/macro_scope.hpp
hnampally: Yeah, I agree :)

docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
hnampally: Yes, somehow I missed it, thanks for catching!

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -242,6 +242,63 @@
         e = ((it != std::end(m)) ? it : std::begin(m))->first;                                  \
     }
 
+NLOHMANN_JSON_NAMESPACE_BEGIN
+namespace detail
+{
+template<typename T>
+[[noreturn]] inline void json_throw_from_serialize_macro(T&& exception)
+{
+#if defined(__cpp_exceptions) || defined(__EXCEPTIONS) || defined(_CPPUNWIND) || defined(EXCEPTIONS)
```
nlohmann: When defining `JSON_THROW` in the same file, we use the following code to detect exceptions. Please use the same here to avoid issues.

```cpp
```

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -242,6 +242,63 @@
...
+        /* NOLINTNEXTLINE(modernize-avoid-c-arrays) we don't want to depend on <array> */       \
+        static const std::pair<ENUM_TYPE, BasicJsonType> m[] = __VA_ARGS__;                     \
+        auto it = std::find_if(std::begin(m), std::end(m),                                      \
+                               [e](const std::pair<ENUM_TYPE, BasicJsonType>& ej_pair) -> bool  \
+        {                                                                                       \
+            return ej_pair.first == e;                                                          \
+        });                                                                                     \
+        if (it == std::end(m)) { \
+            auto value = static_cast<typename std::underlying_type<ENUM_TYPE>::type>(e); \
+            nlohmann::detail::json_throw_from_serialize_macro(nlohmann::detail::type_error::create(302, nlohmann::detail::concat("can't serialize - enum value ", std::to_string(value), " out of range"), &j)); \
```
nlohmann: I don't like the exception message. What about

> serialization failed: enum value ... is out of range

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -242,6 +242,63 @@
...
+        static_assert(std::is_enum<ENUM_TYPE>::value, #ENUM_TYPE " must be an enum!");          \
+        /* NOLINTNEXTLINE(modernize-avoid-c-arrays) we don't want to depend on <array> */       \
+        static const std::pair<ENUM_TYPE, BasicJsonType> m[] = __VA_ARGS__;                     \
+        auto it = std::find_if(std::begin(m), std::end(m),                                      \
+                               [&j](const std::pair<ENUM_TYPE, BasicJsonType>& ej_pair) -> bool \
+        {                                                                                       \
+            return ej_pair.second == j;                                                         \
+        });                                                                                     \
+        if (it == std::end(m))            \
+            nlohmann::detail::json_throw_from_serialize_macro(nlohmann::detail::type_error::create(302, nlohmann::detail::concat("can't deserialize - invalid json value : ", j.dump()), &j)); \
```
nlohmann: Same as above:

> deserialization failed: invalid JSON value '...'

nlohmann (changes requested):

include/nlohmann/detail/macro_scope.hpp
hnampally: oh okay, thanks!

include/nlohmann/detail/macro_scope.hpp
hnampally: Yes that sounds better, thanks!

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
```diff
@@ -0,0 +1,93 @@
+# NLOHMANN_JSON_SERIALIZE_ENUM_STRICT
+
+```cpp
+#define NLOHMANN_JSON_SERIALIZE_ENUM_STRICT(type, conversion...)
+```
+
+The `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` allows to define a user-defined serialization for every enumerator.
```
gregmarr: This is missing some words.

> The `...` macro allows defining a ...
or
> The `...` macro allows one to define a ...

I can see this is basically the same text as in the original, it should probably be updated there too.

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
```diff
@@ -0,0 +1,93 @@
...
+Unlike [`NLOHMANN_JSON_SERIALIZE_ENUM`](nlohmann_json_serialize_enum.md), this macro enforces strict validation and 
+throws errors for unmapped values instead of defaulting to the first enum value.
+
+## Parameters
+
+`type` (in)
+:   name of the enum to serialize/deserialize
+
+`conversion` (in)
+:   a pair of an enumerator and a JSON serialization; arbitrary pairs can be given as a comma-separated list
```
gregmarr: I would say "pairs" rather than "a pair", and "a JSON serialization" is odd, maybe "the string to be used in the JSON serialization.  The "arbitrary pairs can be given as a comma-separated list" sounds weird to me.
Maybe
> A list of parameters alternating between an enumerator value and a string to use in the JSON serialization.

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
```diff
@@ -0,0 +1,93 @@
...
+## Notes
+
+!!! info "Prerequisites"
+
+    The macro must be used inside the namespace of the enum.
+
+!!! important "Important notes"
+
+    - If an enum value appears more than once in the mapping, only the first occurrence will be used for serialization, 
+        subsequent mappings for the same enum value will be ignored.
```
gregmarr: Would it be worth putting in something here like "will currently be ignored but may result in an error in the future"?

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
gregmarr: "If a string value" since it's not an arbitrary json value but a string....

Oh, OH, it doesn't require that it be a string, does it?

😮 

There's nothing in the code that requires that this be a string.

```
NLOHMANN_JSON_SERIALIZE_ENUM_STRICT(Color, {
  { Red, "red" },
  { Green, nlohmann::json(1) },
  { Blue, nlohmann::json({{{"purple", "people eater"}}})}
})


int main()
{
  nlohmann::json j;
  j = { Red, Green, Blue };
  std::cout << j.dump() << "\n";
  auto j2 = nlohmann::json::parse(j.dump());
  std::vector<Color> vec = j2;
  std::cout << vec[0] << vec[1] << vec[2];
}
```

```
["red",1,[{"purple":"people eater"}]]
012
```

@hnampally Is this intended behavior, or did this just accidentally fall out of the second parameter being `BasicJsonType` instead of `std::string`? 

If this is intended behavior, it should definitely be documented before someone trips upon it accidentally, or someone relies on it and then someone else "fixes" it.

If it's not intended behavior, then it should be fixed.

hnampally: @gregmarr thanks for bringing this up but I am not really sure, @nlohmann please weigh in!

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
```diff
@@ -0,0 +1,92 @@
...
+    The example shows how `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` enforces strict validation when serializing an enum value that is not in the mapping:
+
+    ```cpp
+    --8<-- "examples/nlohmann_json_serialize_enum_strict.cpp"
+    ```
+
+    Expected output:
+    
+    ```
+    [json.exception.type_error.302] serialization failed: enum value 3 is out of range
```
nlohmann: Also use an include here:

```
--8<-- "examples/nlohmann_json_serialize_enum_strict.output"
```

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md
```diff
@@ -0,0 +1,92 @@
...
+    The example shows how `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` enforces strict validation when deserializing a JSON value that is not in the mapping:
+
+    ```cpp
+    --8<-- "examples/nlohmann_json_deserialize_enum_strict.cpp"
+    ```
+
+    Expected output:
+    
+    ```
+    [json.exception.type_error.302] deserialization failed: invalid JSON value "yellow"
```
nlohmann: Also use an include here:

```
--8<-- "examples/nlohmann_json_deserialize_enum_strict.output"
```

docs/mkdocs/docs/examples/nlohmann_json_deserialize_enum_strict.cpp
```diff
@@ -0,0 +1,35 @@
...
+{
+    { Color::red, "red" },
+    { Color::green, "green" },
+    { Color::blue, "blue" },
+})
+}
+
+int main()
+{
+
```
nlohmann: Remove the empty line.

Also: move line

```cpp
json j_yellow = "yellow";
```

above the comment

```cpp
// deserialization
```

nlohmann (changes requested):

gregmarr:
It looks like you have a space in your filename here: 
docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum _strict.md

Also, the discussion about strings vs arbitrary json for the value has been marked outdated because there was a change in the file.  @nlohmann did you see that discussion?

jmonticelli:
This looks particularly useful, I was surprised by the behavior of the default enum serialization macro.
Thanks for putting this PR up btw @hnampally

nlohmann:
@hnampally I know it's been a while, but would you be willing to pick this up again?

hnampally:
> @hnampally I know it's been a while, but would you be willing to pick this up again?

@nlohmann Sure, let me give this PR a quick re-read and hopefully we can get it merged/closed.

gregmarr:
@nlohmann Isn't this the same thing as #5151 which is just about ready to go?

hnampally:
https://github.com/nlohmann/json/pull/5151 solves the same issue and ready to go, therefore cloing this PR

nlohmann:
@hnampally Sorry for the confusion!

PR #4628, opened 2025-01-25
Add note on derived return type for value function

### LINKED ISSUE #4626: basic_json::operator[key] and basic_json::value(key) return different values
### Description

When parsing the following json, `operator[key]` and `value(key)` return different results despite 'key' being a valid field.

```
{
	"array": [
		{
			"id": 3024667086235176575
				},
		{
			"id": 5894322762133244432
				},
		{
			"id": 7912588194672682964
				},
		{
			"id": 16577926123393930376
				}
		]
}
```

In this case `j["array"][0]["id"] != j["array"][0].value("id",0)`.

### Reproduction steps

See minimal example below.

If a key exists in a json object, I would expect `j[key]` to return the same value as `j.value(key, default)`.

### Minimal code example

```Shell
Here is a minimal program that reproduces the bug:


#include "json.hpp"
#include <string>

int main(int argc, char* argv[])
{
	std::string s = R"({
                          "array": [
                            {
                              "id": 3024667086235176575
                            },
                            {
                              "id": 5894322762133244432
                            },
                            {
                              "id": 7912588194672682964
                            },
                            {
                              "id": 16577926123393930376
                            }
                          ]
                        })";

	json j = json::parse(s);

	json& arr = j["array"];
	json& arr0 = arr[0];
	uint64_t idTest0 = arr0["id"];
	uint64_t idTest1 = arr0.value("id", 0);

	// Expected idTest0 == idTest1 == 3024667086235176575
	// Got idTest0 == 3024667086235176575
	//     idTest1 == 18446744071677218431

	return 0;
}
```

### Error messages

```Shell

```

msvc (VS2022), windows 11

### Library version

json-3.11.3

### Validation

nlohmann:
This PR adds a note and an example to the documentation of the `value` function that addresses issue #4626.

### DIFF
--- a/docs/mkdocs/docs/api/basic_json/value.md
+++ b/docs/mkdocs/docs/api/basic_json/value.md
@@ -103,6 +103,30 @@ changes to any JSON value.
 2. Logarithmic in the size of the container.
 3. Logarithmic in the size of the container.
 
+## Notes
+
+!!! warning "Return type"
+
+    The return type of the `value` function is determined by the type of the provided default value. This can have
+    unexpected effects. In the example below, we store a 64-bit unsigned integer. We get exactly that value when using
+    [`operator[]`](operator[].md). However, when we call `value` and provide `#!c 0` as default value, then `#!c -1` is
+    returned. The occurs, because `#!c 0` has type `#!c int` which overflows when handling the value
+    `#!c 18446744073709551615`.
+
+    To address this issue, either provide a correctly typed default value or use the template parameter to specify the
+    desired return type. Note that this issue occurs even when a value is stored at the provided key, and the default
+    value is not used as the return value.
+
+    ```cpp
+    --8<-- "examples/value__return_type.cpp"
+    ```
+
+    Output:
+    
+    ```json
+    --8<-- "examples/value__return_type.output"
+    ```
+
 ## Examples
 
 ??? example "Example: (1) access specified object element with default value"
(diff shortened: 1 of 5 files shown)


docs/mkdocs/docs/api/basic_json/value.md
```diff
@@ -103,6 +103,30 @@ changes to any JSON value.
 2. Logarithmic in the size of the container.
 3. Logarithmic in the size of the container.
 
+## Notes
+
+!!! warning "Return type"
+
+    The return type of the `value` function is determined by the type of the provided default value. This can have
```
gregmarr: Maybe like this?

> The `value` function is a  template, and the return type of the function is determined by the type of the provided default value unless otherwise specified.


I thought at first something like this:

> The `value` function is a  template, and the return type of the function is the type of the provided default value unless the type is explicitly specified.

But then I looked into the actual function again, and it's a bit more complicated.  There is a ValueType, and sometimes a KeyType, and sometimes a ReturnType which defaults to being based on the ValueType.

I also realized that it's used to call `get<ValueType>` internally, and maybe that should be mentioned.

Then I thought that's probably way too complicated for this basic intro, and maybe what's there now is as much detail as there should be here, and it doesn't say "is the type of" but rather "is determined by" and that's probably okay.  People can dig into the implementation if they want more information.

nlohmann: Good point. I adjusted it.

gregmarr (approved):

PR #4654, opened 2025-02-17
Fix ~basic_json causing std::terminate

### LINKED ISSUE #4653: ~basic_json can terminate program
### Description

During ~basic_json, it is possible to get an exception from the allocation subsystem (quota/policy violation etc). This should be expected and accounted for.

Either:
- ~basic_json should not be using any allocation, OR
- upon exception a non allocating alternative to destroy should be used.

### Reproduction steps

Throw any exception when allocation request is made during ~basic_json

Expected: `~basic_json` should not terminate a program.
Actual: it terminates the program

### Minimal code example

```Shell
#include <nlohmann/json.hpp>
using json = nlohmann::json;

thread_local bool should_throw = false;

struct QuotaReached : std::exception {};

void* operator new(std::size_t size) {
    if (should_throw){
        throw QuotaReached{};
    }
    void* ptr = std::malloc(size);
    if (!ptr) {
        throw std::bad_alloc();
    }
    return ptr;
}

int main()
{
    {
        auto j = json{1,2,3,4};
        should_throw = true;
        // `j` is destroyed, ~basic_json is noexcept
        // allocation attempted, exception thrown
        // terminate is called
    }
}

https://godbolt.org/z/zh41cxW8j
```

### Error messages

```Shell
terminate called after throwing an instance of 'QuotaReached'
  what():  std::exception
Program terminated with signal: SIGSEGV
```

gcc 14.2 Linux

### Library version

3.11.1

### Validation

Ignition:
During `~basic_json`, it was possible to receive an exception from the allocation subsystem (quota/policy violation etc). This would cause `std::terminate` because destructors are noexcept.
This fix will gracefully handle that case by allowing RAII to handle anything within the current stack (used to avoid recursion), then fallback to the recursion approach for anything that still remains.

fixes [#4653](https://github.com/nlohmann/json/issues/4653)

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -569,51 +569,126 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
             }
             if (t == value_t::array || t == value_t::object)
             {
-                // flatten the current json_value to a heap-allocated stack
-                std::vector<basic_json> stack;
+                // Iteratively flatten nested containers to prevent
+                // stack overflow from recursive destruction. Children
+                // are promoted into the parent array's spare capacity
+                // (no allocation), with overflow to a heap stack. If
+                // the heap stack cannot be allocated, the catch block
+                // lets RAII recurse instead of calling std::terminate.
 
-                // move the top-level items to stack
-                if (t == value_t::array)
+#if (defined(__cpp_exceptions) || defined(__EXCEPTIONS) || defined(_CPPUNWIND))
+                try
                 {
-                    stack.reserve(array->size());
-                    std::move(array->begin(), array->end(), std::back_inserter(stack));
-                }
-                else
-                {
-                    stack.reserve(object->size());
-                    for (auto&& it : *object)
-                    {
-                        stack.push_back(std::move(it.second));
-                    }
-                }
+#endif
+                    // Uses the default allocator, we must not be
+                    // affected by a user allocator that throws or
+                    // copies on move.
+                    std::vector<basic_json> stack;
 
-                while (!stack.empty())
-                {
-                    // move the last item to a local variable to be processed
-                    basic_json current_item(std::move(stack.back()));
-                    stack.pop_back();
-
-                    // if current_item is array/object, move
-                    // its children to the stack to be processed later
-                    if (current_item.is_array())
+                    if (t == value_t::array)
                     {
-                        std::move(current_item.m_data.m_value.array->begin(), current_item.m_data.m_value.array->end(), std::back_inserter(stack));
-
-                        current_item.m_data.m_value.array->clear();
-                    }
-                    else if (current_item.is_object())
-                    {
-                        for (auto&& it : *current_item.m_data.m_value.object)
+                        // Move children from source into parent,
+                        // leaving `remain` behind. Stays within
+                        // spare capacity. Uses resize() rather than
+                        // erase() to avoid instantiating move-
+                        // assignment in MSVC module builds.
+                        auto* parent = array;
+                        auto promote = [parent](basic_json & source, size_type remain)
                         {
-                            stack.push_back(std::move(it.second));
-                        }
+                            if (source.is_array())
+                            {
+                                auto* src = source.m_data.m_value.array;
+                                auto first = src->begin() + static_cast<typename array_t::difference_type>(remain);
+                                std::move(first, src->end(), std::back_inserter(*parent));
+                                src->resize(remain);
+                            }
+                            else if (source.is_object())
+                            {
+                                auto* src = source.m_data.m_value.object;
+                                const auto to_move = src->size() - remain;
+                                for (size_type i = 0; i < to_move; ++i)
+                                {
+                                    parent->push_back(std::move(src->begin()->second));
+                                    src->erase(src->begin());
+                                }
+                            }
+                        };
 
-                        current_item.m_data.m_value.object->clear();
+                        while (!array->empty())
+                        {
+                            if (array->back().is_structured())
+                            {
+                                const auto spare = parent->capacity() - parent->size();
+                                const auto n = array->back().size();
+
+                                if (n <= spare + 1)
+                                {
+                                    // All fit (+1 from freeing container slot)
+                                    basic_json nested(std::move(array->back()));
+                                    array->pop_back();
+                                    promote(nested, 0);
+                                }
+                                else if (spare > 0)
+                                {
+                                    // Partial: revisited once promoted
+                                    // children are processed
+                                    promote(array->back(), n - spare);
+                                }
+                                else
+                                {
+                                    // No capacity: overflow to stack
+                                    stack.push_back(std::move(array->back()));
+                                    array->pop_back();
+                                }
+                            }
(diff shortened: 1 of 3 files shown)


Ignition:
@nlohmann I'm having trouble making the Windows build correct. I don't have windows machine available to debug/diagnose.

include/nlohmann/json.hpp
```diff
@@ -568,51 +568,65 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
-                std::vector<basic_json> stack;
-
-                // move the top-level items to stack
-                if (t == value_t::array)
-                {
-                    stack.reserve(array->size());
-                    std::move(array->begin(), array->end(), std::back_inserter(stack));
-                }
-                else
+#ifdef __cpp_exceptions
```
nlohmann: For what it's worth, this is how we detect exceptions at other parts in the code:

```cpp
#if (defined(__cpp_exceptions) || defined(__EXCEPTIONS) || defined(_CPPUNWIND))
```

nlohmann:
> @nlohmann I'm having trouble making the Windows build correct. I don't have windows machine available to debug/diagnose.

I don't use Windows myself. :-/

include/nlohmann/json.hpp
Ignition: I'll change that and push up the change

gregmarr:
I'm not wild about this idea.  It's swallowing unknown exceptions and setting up a stack overflow.

nlohmann:
I am not sure what to make of this right now. The noexcept annotation was from times where we did hardly anything in the destructor.

gregmarr:
Destructors are `noexcept` by default.

Ignition:
> I'm not wild about this idea. It's swallowing unknown exceptions and setting up a stack overflow.

Normally I would fully agree about swallowing unknown exceptions being bad. But in this case with the current implementation there are only a few potential sources of exceptions.
- allocation subsystem, should only be able to throw an exception via allocate or construct, I have yet to see a use case where destruct or deallocate throw (for the same reasons that destructors are noexcept)
- iteration, which doesn't ATM throw, and a good implementation shouldn't ever need to throw
- moves, which are noexcept so no exceptions there either

So it just a question about is it safe to swallow unknown exception from the allocation subsystem. My feeling is yes, because its actually not swallowing, but handling a corner case even if you don't know what that exception is.

You either:
- Do nothing with the exception, hit `noexcept`, `std::terminate` which kills the program which is **bad**.
- Make `~basic_json` `noexcept(false)`, which I **DO NOT** recommend. It just moves the problem onto the consumer, any type that has json as a data member would also need a `noexcept(false)` destructor.
- Handle the unknown exception, in this case we are confident the exception could have only come from the memory subsystem. Regardless of the actual exception, we can handle it by letting RAII do it job and then fallback to doing regular recursive cleanup. This does risk stack exhaustion, but that is much better than `std::terminate`.

gregmarr:
Do you use the `JSON_DIAGNOSTICS` define?  If not, would you be okay with using that option?  

I think it's possible to rewrite this to not use the vector if the `m_parent` field exists, using an algorithm based on Morris Traversal.  The basic idea is that you have a `current` pointer.  You start at the root (`this`), and walk the current pointer down the tree.  When you find an object/array, you look at the first object element or the last array element.  If it's a leaf node, remove it from the object/array.  By using the last array element instead of the first, you are just doing a pop_back on the array instead of moving everything after it.  If it's an object/array, you update current and repeat.  Once the object/array is empty, you update current to be its parent, and then remove the thing you just processed from its object/array.

I haven't written any of this code, just thinking about what might be possible.

Ignition:
@gregmarr @nlohmann 
I have delivered a patch in my build that prevents std::terminate in my product. Unfortunate I don't have the time to spend on getting this merged here. I maybe will come back to this PR, but don't bet/rely on it.

@gregmarr I agree it is possible to do it without heap allocation. I think that modification is worth doing (assuming benchmarks show that is fine). Here is Herb Sutter talking about destroying trees https://youtu.be/JfmTagWcqoE?si=XyJR0YpYZGm4FGbB&t=970

gregmarr:
@Ignition Yeah, the current method is the bottom row of the table, O(N) heap space with moves.  In order to remove the O(N) heap space, which is the issue that you're having of allocating memory in a destructor, we'd have to go to the second row of the table, which is O(N log N) time.

Ignition:
@gregmarr I came back to this, because it would be nice to upstream this fix rather than having my own patch.

You're right that Morris Traversal like approach with m_parent would give the ideal O(N) time / O(1) space destruction with no allocation and no stack overflow. The limitation is that m_parent is only available when JSON_DIAGNOSTICS is defined, the fix should work for all configurations. 
 
The universal approach: for arrays, children are promoted into the parent vector's spare capacity (zero allocation for the common case), with overflow to a lazily-allocated stack wrapped in try-catch to prevent std::terminate. A JSON_DIAGNOSTICS-only path using the parent pointer traversal you describe could be a worthwhile follow-up optimisation, it would eliminate the overflow stack entirely for that configuration.

include/nlohmann/json.hpp
```diff
@@ -569,51 +569,126 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+                                else if (spare > 0)
+                                {
+                                    // Partial: revisited once promoted
+                                    // children are processed
+                                    promote(array->back(), n - spare);
+                                }
+                                else
+                                {
+                                    // No capacity: overflow to stack
+                                    stack.push_back(std::move(array->back()));
```
gregmarr, line 640: This is still a memory allocation, so you haven't achieved your goal of not allocating memory in the destructor.  I take it you are assuming that any failed allocation would result in a thrown allocation that you can catch.  Even if that were the case, then you have converted the potential issue from a failed allocation into a stack overflow.  It seems to me that the only way to prevent both the stack overflow and the possibility of an allocation failure is the Morris Traversal using the parent pointer.  You could simulate that if the parent pointer isn't available using a local container for the current path from the root, but that is also an allocation that can fail.

Ignition: My goal is not for no allocations. My goal is to avoid std::terminate if an allocator throws. Which can be for more reasons that running out of memory, in my case its because of a quota is exceeded in a subsystem. 

I don't want to change the entire representation of an object to make it zero allocations. This is a bigger change than this fix. I just try to 1) reuse existing buffers to flatten where possible, 2) flatten using heap allocation if needed, 3) allow stack call if required.

Its a best effort approach, without restructuring all of the json library. I've had [a patch](https://github.com/memgraph/memgraph/blob/master/conan_recipes/recipes/nlohmann_json/all/patches/nlohmann_json3.11.3.patch) for over a year that works, I'd just like something delivered upstream so I can cleanup downstream project.

gregmarr: My main point is that you are replacing one terminate with another (stack overflow), which the library had previously avoided due to actual users experiencing it.  I'm just a contributor but to me that is a non-starter, especially when it's possible to avoid both with the addition of a single pointer, and when that pointer is already there with a configuration option.

PR #4663, opened 2025-02-23
Add clang-tidy plugin to convert implicit conversions to explicit ones

mikecrowe:
Add a clang-tidy plugin containing one check to replace implicit conversions (as enabled by default with JSON_USE_IMPLICIT_CONVERSIONS) with explicit ones. This will make it easier for library users to switch away from using implicit conversions which should make it possible for the library to start disallowing them sooner.

The skeleton of the check was generated by the clang-tidy add_new_check.py script. It added headers indicating that the files were under the LLVM Apache-2.0-with-LLVM-exception licence. I've replaced these with the MIT licence to match the rest of the library but would be happy to change them back.

The check was written using LLVM's clang-format style. I can modify that if you'd prefer.

The check_clang_tidy.py script was copied in its entirety from the LLVM repository so it must keep its existing Apache-2.0-with-LLVM-exception licence.

The test file is written in the style of clang-tidy's own tests. This means that it cannot include system headers. We might be able to relax that for a plugin so that the actual single_include header could be used. That might make cross-platform CI harder though and I haven't looked into doing so.

I know very little about CMake or GitHub workflows. What I have implemented appears to work, but I don't know whether both work the way you would like. I have a Linux and Debian bias.

A reviewer of the [original check](https://github.com/llvm/llvm-project/pull/126425) asked about implicit conversions to `std::wstring` and `std::string_view`. I think that neither of those are possible. Is that true?

- [X] The changes are described in detail, both the what and why. **In individual commit messages**
- [ ] If applicable, an [existing issue](https://github.com/nlohmann/json/issues) is referenced. **N/A**
- [ ] The [Code coverage](https://coveralls.io/github/nlohmann/json) remained at 100%. A test case for every new line of code. **It looks like I need to create the pull request before I see the results of this**
- [X] If applicable, the [documentation](https://json.nlohmann.me) is updated.
- [ ] The source code is amalgamated by running `make amalgamate`. **N/A**

### DIFF
--- a/.github/workflows/clang-tidy-plugin.yml
+++ b/.github/workflows/clang-tidy-plugin.yml
@@ -1,0 +1,60 @@
+name: Ubuntu
+
+on:
+  push:
+    branches:
+      - develop
+      - master
+      - release/*
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  ci_test_clang_tidy:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # clang-tidy plugins require at least Clang 16. This means that
+        # anything older than Ubuntu 24.04 and Debian 12 can't work at all.
+        include:
+          - distro: ubuntu:24.04
+            clang_version: 16
+          - distro: ubuntu:24.04
+            clang_version: 17
+          - distro: ubuntu:24.04
+            clang_version: 18
+          - distro: debian:12
+            clang_version: 16
+          - distro: debian:12
+            clang_version: 19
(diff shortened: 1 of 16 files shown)


mikecrowe:
I've made a small improvement to the check, improved the documentation and tried to placate the REUSE licence checks.

I think that the Codacy check failures probably just need to be ignored somehow. We don't really want to modify LLVM's `check_clang_tidy.py` and the checks are inappropriate for the `modernize-nlohmann-json-explicit-conversions.cpp` test file.

I'm afraid that I don't really understand the Windows failures.

Issue #2189, opened 2020-06-14
3.8.0: unit-cbor.cpp test failures


dvzrv:
Hi! I package nlohmann-json for Arch Linux. Upon building version 3.8.0 I ran the unit tests, but some fail.

For the last stable release (3.7.3) I used the below described workflow and the unit tests pass. Maybe something has changed in the meantime?

#### What is the issue you have?

```
Executing test-algorithms...
[doctest] doctest version is "2.3.7"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases:      1 |      1 passed |      0 failed |      0 skipped
[doctest] assertions:     39 |     39 passed |      0 failed |
[doctest] Status: SUCCESS!
Executing test-allocator...
[doctest] doctest version is "2.3.7"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases:      3 |      3 passed |      0 failed |      0 skipped
[doctest] assertions:     17 |     17 passed |      0 failed |
[doctest] Status: SUCCESS!
Executing test-alt-string...
[doctest] doctest version is "2.3.7"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases:      1 |      1 passed |      0 failed |      0 skipped
[doctest] assertions:     28 |     28 passed |      0 failed |
[doctest] Status: SUCCESS!
Executing test-bson...
[doctest] doctest version is "2.3.7"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases:      6 |      6 passed |      0 failed |      1 skipped
[doctest] assertions:    486 |    486 passed |      0 failed |
[doctest] Status: SUCCESS!
Executing test-capacity...
[doctest] doctest version is "2.3.7"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases:      1 |      1 passed |      0 failed |      0 skipped
[doctest] assertions:    120 |    120 passed |      0 failed |
[doctest] Status: SUCCESS!
Executing test-cbor...
[doctest] doctest version is "2.3.7"
[doctest] run with "--help" for options
===============================================================================
src/unit-cbor.cpp:1912:
TEST CASE:  single CBOR roundtrip
  sample.json

src/unit-cbor.cpp:1912: ERROR: test case THREW exception: exception thrown in subcase - will translate later when the whole test case has been exited (cannot translate while there is an active exception)

===============================================================================
src/unit-cbor.cpp:1912:
TEST CASE:  single CBOR roundtrip

[...]

PR #4669, opened 2025-02-27
tests: Mark the tests that don't need to downoad test files with the …

LocutusOfBorg:
This way in Debian and Ubuntu we can easily skip them when no internet is available

### DIFF
--- a/tests/abi/config/CMakeLists.txt
+++ b/tests/abi/config/CMakeLists.txt
@@ -6,6 +6,9 @@ target_link_libraries(abi_config_default PRIVATE abi_compat_main)
 add_test(
     NAME test-abi_config_default
     COMMAND abi_config_default ${DOCTEST_TEST_FILTER})
+set_tests_properties(test-abi_config_default PROPERTIES
+    LABELS "offline"
+)
 
 # test no version namespace
 add_executable(abi_config_noversion noversion.cpp)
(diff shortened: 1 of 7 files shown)


gregmarr:
Is it better to mark the ones that don't need to download things, or the ones that do?

nlohmann:
From the README:

> Note that during the `ctest` stage, several JSON test files are downloaded from an [external repository](https://github.com/nlohmann/json_test_data). If policies forbid downloading artifacts during testing, you can download the files yourself and pass the directory with the test files via `-DJSON_TestDataDirectory=path` to CMake. Then, no Internet connectivity is required. See [issue #2189](https://github.com/nlohmann/json/issues/2189) for more information.

Would using `JSON_TestDataDirectory` be an option?

LocutusOfBorg:
Can we merge and improve later maybe?

nlohmann:
Can you check if JSON_TestDataDirectory works for you?

LocutusOfBorg:
Hello, so some questions:
1) why is this test data on different repo?
(I presume due to $SIZE)
2) adding it as external tarball will increase the code complexity, forget to update, make source download heavier, as well as having it as git submodule.
I'll think a little bit about it, but maybe instead of having a 100MB tarball additional, we can better exclude some tests...

LocutusOfBorg:
ok we can package the json test data separately in Debian as standalone project, this will also help your ci test environment (if you want to use apt)
Something like this looks working, but I now have errors in cmake_fetch and fetch2 tests (we need to disable them too)
```diff
--- nlohmann-json3-3.11.3.orig/cmake/download_test_data.cmake
+++ nlohmann-json3-3.11.3/cmake/download_test_data.cmake
@@ -1,6 +1,12 @@
 set(JSON_TEST_DATA_URL     https://github.com/nlohmann/json_test_data)
 set(JSON_TEST_DATA_VERSION 3.1.0)
 
+set(JSON_TEST_DATA_DIR /usr/share/json_test_data-${JSON_TEST_DATA_VERSION})
+
+if(EXISTS ${JSON_TEST_DATA_DIR})
+  set(JSON_TestDataDirectory ${JSON_TEST_DATA_DIR})
+endif()
+
 # if variable is set, use test data from given directory rather than downloading them
 if(JSON_TestDataDirectory)
     message(STATUS "Using test data in ${JSON_TestDataDirectory}.")
```

LocutusOfBorg:
88 - cmake_fetch_content_configure (Failed)            git_required not_reproducible
	 89 - cmake_fetch_content_build (Not Run)               git_required not_reproducible
	 90 - cmake_fetch_content2_configure (Failed)           git_required not_reproducible
	 91 - cmake_fetch_content2_build (Not Run)              git_required not_reproducible

nlohmann:
> 1. why is this test data on different repo?
>    (I presume due to $SIZE)

Exactly.

PR #4670, opened 2025-02-27
tests: Fix ignored attributes warning during build

LocutusOfBorg:
Fix build warning.

### DIFF
--- a/tests/src/make_test_data_available.hpp
+++ b/tests/src/make_test_data_available.hpp
@@ -18,7 +18,8 @@ namespace utils
 
 inline bool check_testsuite_downloaded()
 {
-    const std::unique_ptr<std::FILE, decltype(&std::fclose)> file(std::fopen(TEST_DATA_DIRECTORY "/README.md", "r"), &std::fclose);
+    using FilePtr = std::unique_ptr<FILE, int(*)(FILE*)>;
+    const FilePtr file(std::fopen(TEST_DATA_DIRECTORY "/README.md", "r"), std::fclose);
     return file != nullptr;
 }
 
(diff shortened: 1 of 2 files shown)


nlohmann (commented):
I fail to understand the reason for the change. `decltype(&std::fclose)` is supposed to be `int (*)(FILE *)`, so why do you need to make this explicit and add a type alias?

gregmarr:
Explained at https://stackoverflow.com/questions/76867698/what-does-ignoring-attributes-on-template-argument-mean-in-this-context

nlohmann:
Thanks! So to silence GCC's warning about an ignored attribute, we define a unique pointer deleter type without that attribute and call this a fix, because GCC does not realize that we then instantiate this deleter with a function whose argument _has_ an attribute?

gregmarr:
The warning is only on the template declaration, so moving it to the instantiation doesn't trigger that particular warning.  They may expand it or add a new one in the future.

As described in the thread, below would be a better solution for size and performance if it were in the main library, but since it's in the tests, we're not as concerned about size and performance.

```
struct unique_ptr_fclose_deleter {
    void operator()(struct FILE* f) const {
        fclose(f);
    }
};
using FilePtr = std::unique_ptr<FILE, unique_ptr_fclose_deleter>;
```

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks for the clarification! I will merge this once #4701 is merged, because otherwise the CI would fail.

LocutusOfBorg:
thanks!

nlohmann:
Thank you!

PR #4693, opened 2025-03-17
fixes issue 4691

### LINKED ISSUE #4691: gdb-pretty-print not working with python < 3.9, because of removeprefix
### Description

If gdb uses python version < 3.9 the pretty printer do not work because of the missing implementation of 'removeprefix'.

This function is only available since python version 3.9.

I can reproduce this bug by using latest version of CLion 2024.3.4, because they are using python version 3.8.

### Reproduction steps

try to pretty print nlohmann object.
that occurs in an exception because the string obj do not have a function called removeprefix

pretty print should work if python version is < 3.9

### Minimal code example

```Shell

```

### Error messages

```Shell

```

linux, CLion 2024.3.4

### Library version

latest from branch develop 

### Validation

bitFiedler:
fixes the behavior in issue 4691
https://github.com/nlohmann/json/issues/4691

### DIFF
--- a/tools/gdb_pretty_printer/nlohmann-json.py
+++ b/tools/gdb_pretty_printer/nlohmann-json.py
@@ -17,17 +17,19 @@ def json_lookup_function(val):
     if m := ns_pattern.fullmatch(str(val.type.strip_typedefs().name)):
         name = m.group('name')
         if name and name.startswith('basic_json<') and name.endswith('>'):
-            m = ns_pattern.fullmatch(str(val["m_data"]['m_type']))
+            m_data = val['m_data']
+            m_type = m_data['m_type']
+            m = ns_pattern.fullmatch(str(m_type))
             t = m.group('name')
             prefix = 'detail::value_t::'
             if t and t.startswith(prefix):
                 try:
-                    union_val = val['m_data']['m_value'][t.replace(prefix, '', 1)]
+                    union_val = m_data['m_value'][t.replace(prefix, '', 1)]
                     if union_val.type.code == gdb.TYPE_CODE_PTR:
                         return gdb.default_visualizer(union_val.dereference())
                     else:
                         return JsonValuePrinter(union_val)
                 except Exception:
-                    return JsonValuePrinter(val["m_data"]['m_type'])
+                    return JsonValuePrinter(m_type)
 
 gdb.pretty_printers.append(json_lookup_function)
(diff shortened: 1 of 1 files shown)


tools/gdb_pretty_printer/nlohmann-json.py
```diff
@@ -15,18 +17,29 @@ def to_string(self):
...
+            m_data = val["m_data"]
+            m_type = m_data['m_type']
+            m = ns_pattern.fullmatch(str(m_type))
+            t = m.group('name')
+            prefix = 'detail::value_t::'
+            if t and t.startswith(prefix):
+                try:
+                    key_val = None
+                    # python < 3.9 do not have removeprefix
+                    if py_ver.major == 3 and py_ver.minor <= 8 or py_ver.major < 3:
```
nlohmann: I think this check makes the code unnecessarily complex. Why not just implementing the version without `removeprefix`?

bitFiedler: no problem, I reworked my commits

tools/gdb_pretty_printer/nlohmann-json.py
```diff
@@ -15,18 +15,21 @@ def to_string(self):
...
-                  if union_val.type.code == gdb.TYPE_CODE_PTR:
-                      return gdb.default_visualizer(union_val.dereference())
-                  else:
-                      return JsonValuePrinter(union_val)
-              except Exception:
-                  return JsonValuePrinter(val["m_data"]['m_type'])
+        name = m.group('name')
+        if name and name.startswith('basic_json<') and name.endswith('>'):
+            m_data = val["m_data"]
+            m_type = m_data['m_type']
```
nlohmann, line 21: Why did you introduce these variables?

nlohmann (changes requested):
Please check the DCO warning: https://github.com/nlohmann/json/pull/4693/checks?check_run_id=38900328997

tools/gdb_pretty_printer/nlohmann-json.py
gregmarr: Looks to be to eliminate duplicate accesses.  However, `m_type` is only accessed once in the normal case.  The second access is when an exception has been thrown.  `m_data` is accessed twice, or three times if the exceptional case is made to be `m_data["m_type"]`.

While you're here, probably worth changing `"m_data"` to `'m_data'` to match the rest of the function.

bitFiedler: Yes removing duplicates, but if you look in the history some gdb fixes fall in that access. That means if the interface change in the the future the changes should be minimal. I will provide a rebased fix with the single quotes

nlohmann (approved):
Looks good to me.

PR #4712, opened 2025-04-01
patch: Fixed No matching function for call to 'create'

ovaar:
This PR is a part of solving compiling issues for Xcode 16 which are related to https://github.com/nlohmann/json/issues/4713.
- Explicitly set template parameter std::nullptr_t when calling error::create
- Avoid narrow conversion for the error code value by adding static_cast

```
Candidate template ignored: requirement 'is_basic_json_context<long>::value' was not satisfied [with BasicJsonContext = long]
```

- [ ] The changes are described in detail, both the what and why.


gregmarr:
None of these changes make any sense.  I can't see any reason why the compiler would require any of these things to be specified like this.

ovaar:
I agree, I also found it difficult to reason why the compiler doesn't seem to be able to resolve those types.
Maybe it is a compiler bug?

PR #4734, opened 2025-04-11
Remove `<algorithm>` dependency from `from_json.hpp`

vittorioromeo:
As [discussed on Reddit](https://old.reddit.com/r/cpp/comments/1jwl5r1/json_for_modern_c_3120_released/mmjnmz3/), here's a PR taking a first small step towards the improvement of compilation times. Avoiding dependencies on "heavy" standard library header such as `<algorithm>` can help.

Here is the ClangBuildAnalyzer compilation time report:
https://gist.github.com/vittorioromeo/7a89a1e9af8cb89ee217d4e85be83270

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -8,7 +8,6 @@
 
 #pragma once
 
-#include <algorithm> // transform
 #include <array> // array
 #include <forward_list> // forward_list
 #include <iterator> // inserter, front_inserter, end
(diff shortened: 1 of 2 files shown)


nlohmann:
@vittorioromeo Thanks for the PR! Could you please sign off your commit (see https://github.com/nlohmann/json/pull/4734/checks?check_run_id=40403948884).

gregmarr:
Rather than rewrite all of these algorithms which have been in production for many years, and potentially break the use on a user-defined type that specializes `std::front_insert_iterator` or `std::insert_iterator`, I wonder if it would be better to just create a local version of `transform` (and since it's just used for the side effects, not the return, it's really a `copy_transform`, so it could be renamed):

```
template<class InputIt, class OutputIt, class UnaryOp>
void copy_transform(InputIt first1, InputIt last1,
                   OutputIt d_first, UnaryOp unary_op)
{
    for (; first1 != last1; ++d_first, ++first1)
        *d_first = unary_op(*first1);
}
```
based on "Possible Implementation" from https://en.cppreference.com/w/cpp/algorithm/transform

PR #4736, opened 2025-04-11
Provide fallback for missing `char8_t` support

### LINKED ISSUE #4733: Clang 11.0.x compilation error
### Description

Library tests fail to compile using Clang 11.0.x due to

https://github.com/nlohmann/json/blob/7ddea2686f091ea77a3f731ab74c59535db9f684/include/nlohmann/detail/conversions/to_json.hpp#L446-L447

because even in C++20 mode `std::filesystem::path::u8string` does not return `std::u8string` but a `std::string`.

A related issue occurs during conversion in the opposite direction here:

https://github.com/nlohmann/json/blob/7ddea2686f091ea77a3f731ab74c59535db9f684/include/nlohmann/detail/conversions/from_json.hpp#L543-L544

### Reproduction steps

[Godbolt reproducer](https://godbolt.org/z/x3fn7T9r3).

No compilation errors.

### Minimal code example

```Shell

```

### Error messages

```Shell

```

Clang 11.0.x

### Library version

3.12.0

### Validation

sergiud:
Clang 11.0.x with libc++ fails to compile tests in C++20 mode due to incomplete `char8_t` support in `std::filesystem::path`. Use a specific library feature guard instead of performing a generic C++20 language standard check.

Fixes #4733

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -540,7 +540,10 @@ inline void from_json(const BasicJsonType& j, std_fs::path& p)
         JSON_THROW(type_error::create(302, concat("type must be string, but is ", j.type_name()), &j));
     }
     const auto& s = *j.template get_ptr<const typename BasicJsonType::string_t*>();
-#ifdef JSON_HAS_CPP_20
+    // Checking for C++20 standard or later can be insufficient in case the
+    // library support for char8_t is either incomplete or was disabled
+    // altogether. Use the __cpp_lib_char8_t feature test instead.
+#if defined(__cpp_lib_char8_t) && (__cpp_lib_char8_t >= 201907L)
     p = std_fs::path(std::u8string_view(reinterpret_cast<const char8_t*>(s.data()), s.size()));
 #else
     p = std_fs::u8path(s); // accepts UTF-8 encoded std::string in C++17, deprecated in C++20
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -15,7 +15,8 @@
 
 #include <algorithm> // copy
 #include <iterator> // begin, end
-#include <string> // string
+#include <memory> // allocator_traits
+#include <string> //  basic_string, char_traits
 #include <tuple> // tuple, get
 #include <type_traits> // is_same, is_constructible, is_floating_point, is_enum, underlying_type
 #include <utility> // move, forward, declval, pair
@@ -440,15 +441,21 @@ inline void to_json(BasicJsonType& j, const T& t)
 }
 
 #if JSON_HAS_FILESYSTEM || JSON_HAS_EXPERIMENTAL_FILESYSTEM
+#if defined(__cpp_lib_char8_t)
+template<typename BasicJsonType, typename Tr, typename Allocator>
+inline void to_json(BasicJsonType& j, const std::basic_string<char8_t, Tr, Allocator>& s)
+{
+    using OtherAllocator = typename std::allocator_traits<Allocator>::template rebind_alloc<char>;
+    j = std::basic_string<char, std::char_traits<char>, OtherAllocator>(s.begin(), s.end(), s.get_allocator());
+}
+#endif
+
 template<typename BasicJsonType>
 inline void to_json(BasicJsonType& j, const std_fs::path& p)
 {
-#ifdef JSON_HAS_CPP_20
-    const std::u8string s = p.u8string();
-    j = std::string(s.begin(), s.end());
-#else
-    j = p.u8string(); // returns std::string in C++17
-#endif
+    // Returns either a std::string or a std::u8string depending whether library
+    // support for char8_t is enabled.
+    j = p.u8string();
 }
 #endif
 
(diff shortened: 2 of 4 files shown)


include/nlohmann/detail/conversions/to_json.hpp
```diff
@@ -443,12 +443,8 @@ inline void to_json(BasicJsonType& j, const T& t)
 template<typename BasicJsonType>
 inline void to_json(BasicJsonType& j, const std_fs::path& p)
 {
-#ifdef JSON_HAS_CPP_20
-    const std::u8string s = p.u8string();
+    const auto& s = p.u8string();
```
gregmarr: Removing the conditional compilation adds an unnecessary copy to those using it in C++17.

sergiud: Unfortunately, without this change the tests fail because `BasicJsonType::is_string` no longer holds in corresponding `from_json`.

sergiud: Testing against `201811L` seems to solve the issue. However, compilation using Clang 11.0.x starts failing again.

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -540,7 +540,7 @@ inline void from_json(const BasicJsonType& j, std_fs::path& p)
         JSON_THROW(type_error::create(302, concat("type must be string, but is ", j.type_name()), &j));
     }
     const auto& s = *j.template get_ptr<const typename BasicJsonType::string_t*>();
-#ifdef JSON_HAS_CPP_20
+#if defined(__cpp_lib_char8_t) && (__cpp_lib_char8_t >= 201907L)
```
nlohmann, line 546: Please add a comment why checking `__cpp_lib_char8_t` is required and `JSON_HAS_CPP_20` is insufficient.

nlohmann (changes requested):

nlohmann:
Would it make sense to extend the CI to run the Clang tests additionally with `-stdlib=libc++`?

sergiud:
The [`ci_test_standards_clang` job](https://github.com/nlohmann/json/blob/93e957332bddcb244eb479746b4530e8ef2f8bb6/.github/workflows/ubuntu.yml#L207-L208) in the Ubuntu Github workflow seems to be doing exactly that already. The libc++ version used there is likely much more recent and does not expose the regression.

Given GCC 14 (using libstdc++) suffers from a [related issue](https://github.com/nlohmann/json/issues/4733#issuecomment-2798830411), reproducing the setup for the latter is probably also sensible going forward.

nlohmann:
> The [`ci_test_standards_clang` job](https://github.com/nlohmann/json/blob/93e957332bddcb244eb479746b4530e8ef2f8bb6/.github/workflows/ubuntu.yml#L207-L208) in the Ubuntu Github workflow seems to be doing exactly that already. The libc++ version used there is likely much more recent and does not expose the regression.

Yes, these jobs use the latest Clang version. Question is if it makes sense to also do this with (some/all) of the other Clang versions we use in `ci_test_compilers_clang` (3.4 .. 20).

> Given GCC 14 (using libstdc++) suffers from a [related issue](https://github.com/nlohmann/json/issues/4733#issuecomment-2798830411), reproducing the setup for the latter is probably also sensible going forward.

Not sure what additional setup is used there. We already use GCC 14 in the CI. Maybe @jelly can help?

sergiud:
According to https://github.com/nlohmann/json/issues/4733#issuecomment-2799990654 it's [`-fno-char8_t`](https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html#index-fchar8_005ft).

include/nlohmann/detail/conversions/to_json.hpp
gregmarr: @sergiud Oh, that's annoying.  So, libc++ in Clang 11.0.x sets `__cpp_lib_char8_t 201811L` but doesn't actually implement `p.u8string()` returning `std::u8string` until it's set to `201907L`?

That's ugly.

However, this still means that everything using C++17 has to pay for this copy unnecessarily, where they didn't before.  Would the following work?

```
    const auto& s = p.u8string();
#if defined(__cpp_lib_char8_t) && (__cpp_lib_char8_t >= 201907L)
    j = std::string(s.begin(), s.end());
#else
    j = s; // returns std::string in C++17
#endif
```

sergiud: Yes, `__cpp_lib_char8_t` version is somewhat unreliable. Therefore, instead of feature testing against a specific version for `char8_t` library support a better approach is probably to provide a `std::u8string` overload whenever `__cpp_lib_char8_t` is defined.

gregmarr: Looks reasonable.

sergiud:
I've no idea what Codacy's problem is.

nlohmann:
> I've no idea what Codacy's problem is.

I think we can ignore it.

nlohmann:
I suppressed the issue in Codacy.

nlohmann (approved):
Looks good to me.

nlohmann:
@gregmarr Any further comments or is this ready to merge?

gregmarr (approved):

jonesmz:
Hi, my project can't update to json 3.12.0, because we build with char8_t disabled.

Is there appetite for cutting a new release with this fix included, possibly as a 3.12.1 ?

PR #4742, opened 2025-04-14
Fix conversion to std::optional

### LINKED ISSUE #4740: Build issue with std::optional
### Description

Hi,
So great the library now support std::optional!
But I have a build issue with the following simple code.
Do you have any idea?
Thank you.
Vincent

### Reproduction steps

Just compile the code

Compile should be successful

### Minimal code example

```Shell
#include <nlohmann/json.hpp>
#include <iostream>

#if 1
struct toto final
{
  std::optional<std::string> host;
  std::optional<int> port;
  NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(toto, host, port);
};
#else
struct toto final
{
  std::string host;
  int port;
  NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(toto, host, port);
};
#endif

int main()
{
  auto t0 = toto();
  auto j = nlohmann::json(t0);
  std::cout << j.dump(2) << std::endl;
  auto t1 = nlohmann::json::parse(j.dump()).get<toto>();
  return 0;
}
```

### Error messages

```Shell
'nlohmann::json_abi_v3_12_0::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::json_abi_v3_12_0::adl_serializer,std::vector<uint8_t,std::allocator<uint8_t>>,void>::get_to': no matching overloaded function found
```

Visual Studio 2022

### Library version

3.12.0

### Validation

### LINKED ISSUE #4780: `j.get_to(my_struct)` fails to compile for types with `std::optional`
### Description

When creating types that make use of `std::optional`, I am unable to deserialize them using the normal method.

### Reproduction steps

See [this repo](https://github.com/jnickg/json_optional_problem)

## Expected
- Compiles and runs when `std::optional` is used on a member

## Actual
- Does not compile

### Minimal code example

```Shell
See repo linked above, or compile the program below


#include <array>
#include <string>
#include <iostream>
#include <optional>

#include <nlohmann/json.hpp>

struct foo
{
    std::optional<int> _f = 3;
    foo() = default;
};
NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(foo, _f);

struct bar
{
    int _b = 0;
    bar() = default;
};
struct baz
{
    std::optional<bar> _bar = std::nullopt;
    baz() = default;
};

NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(bar, _b);
NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(baz, _bar);

int main()
{
    nlohmann::json j = nlohmann::json::parse(R"({"_b": 37})");
    bar b;
    j.get_to(b);
    std::cout << "b._b: " << b._b << std::endl;

    // Un-comment the code below to break the build.

    baz bz;
    nlohmann::json j2 = nlohmann::json::parse(R"({"_bar": {"_b": 1}})");
    j.get_to(bz);

    std::cout << "bz._bar->_f: " << (bz._bar ? std::to_string(bz._bar->_b) : "null") << std::endl;

    foo f;
    nlohmann::json j3 = nlohmann::json::parse(R"({"_f": 1})");
    j2.get_to(f);
    std::cout << "f._f: " << (f._f ? std::to_string(f._f.value()) : "null") << std::endl;

    return 0;
}
```

### Error messages

```Shell
json.hpp(1825, 17): candidate template ignored: requirement 'detail::has_from_json<nlohmann::basic_json<std::map, std::vector, std::basic_string<char, std::char_traits<char>, std::allocator<char>>, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char>>, void>, std::optional<int>, void>::value' was not satisfied [with ValueType = std::optional<int>]
json.hpp(1838, 17): candidate template ignored: requirement 'detail::is_basic_json<std::optional<int>>::value' was not satisfied [with ValueType = std::optional<int>]
json.hpp(1849, 11): candidate template ignored: could not match 'T[N]' against 'std::optional<int>'
```

clang 19.1.7; Linux 6.12.21-4-MANJARO

### Library version

v3.12.0

### Validation

### LINKED ISSUE #4834: Problems with std::optional
### Description

Note that I am using the "single include file" release of 3.12.0

I'm currently trying to use this library to convert C++ object with std::optional members to/from JSON.  I have been able to make it work, but only after tweaking the json.hpp file slightly.  To get this working, I need the pre-compiler defines/macros to allow the code at lines 4837 to 4847 to be included/compiled.

Issue 1:
It appears as the library is pretty diligent on probing the compiler and automatically setting things like JSON_HAS_CPP_17.  However, this setting is done *after* the initial include of std::optional at line 177.  

Issue 2:
In order for the code mentioned above to be compiled, the macro JSON_USE_IMPLICIT_CONVERSIONS must be *undefined*.  However, the logic on line 2945 *always* defines that macro, and hence the function I need to use is never compiled.  Is it possible that #ifndef on line 4835 was intended to be a #if that should have checked for a value of 1 rather than defined/undefined?

It's quite possible that I am using the library wrongly or don't understand something.  If you see where I am making a mistake, then I apologize and could use a point in the right direction on how to support std::optional translations.

### Reproduction steps

Simply trying something like this exposes the potential bug and gives a compiler error:

std::string s"{\"test\": 7}";
nlohmann:json j = nlohmann::json::parse(s);
std::optional<int> opt;
nlohmann::from_json(json.at("test"), opt);

I would expect this to compile and function

### Minimal code example

```Shell
See reproduction section
```

### Error messages

```Shell

```

Rocky9 with Red Hat 11.5.0-2 version of GCC

### Library version

3.12.0

### Validation

nlohmann:
This MR fixes the conversion from JSON to `std::optional` by making it independent of `JSON_USE_IMPLICIT_CONVERSIONS`.

Fixes #4740

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -13,9 +13,6 @@
 #include <forward_list> // forward_list
 #include <iterator> // inserter, front_inserter, end
 #include <map> // map
-#ifdef JSON_HAS_CPP_17
-    #include <optional> // optional
-#endif
 #include <string> // string
 #include <tuple> // tuple, make_tuple
 #include <type_traits> // is_arithmetic, is_same, is_enum, underlying_type, is_convertible
@@ -31,6 +28,11 @@
 #include <nlohmann/detail/meta/type_traits.hpp>
 #include <nlohmann/detail/string_concat.hpp>
 #include <nlohmann/detail/value_t.hpp>
+
+// include after macro_scope.hpp
+#ifdef JSON_HAS_CPP_17
+    #include <optional> // optional
+#endif
 
 NLOHMANN_JSON_NAMESPACE_BEGIN
 namespace detail
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -47,7 +47,7 @@ inline void from_json(const BasicJsonType& j, typename std::nullptr_t& n)
 }
 
 #ifdef JSON_HAS_CPP_17
-#ifndef JSON_USE_IMPLICIT_CONVERSIONS
+//#if JSON_USE_IMPLICIT_CONVERSIONS == 1
 template<typename BasicJsonType, typename T>
 void from_json(const BasicJsonType& j, std::optional<T>& opt)
 {
@@ -61,7 +61,7 @@ void from_json(const BasicJsonType& j, std::optional<T>& opt)
     }
 }
 
-#endif // JSON_USE_IMPLICIT_CONVERSIONS
+//#endif // JSON_USE_IMPLICIT_CONVERSIONS
 #endif // JSON_HAS_CPP_17
 
 // overloads for basic_json template parameters
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -49,7 +49,6 @@ inline void from_json(const BasicJsonType& j, typename std::nullptr_t& n)
 }
 
 #ifdef JSON_HAS_CPP_17
-//#if JSON_USE_IMPLICIT_CONVERSIONS == 1
 template<typename BasicJsonType, typename T>
 void from_json(const BasicJsonType& j, std::optional<T>& opt)
 {
@@ -62,8 +61,6 @@ void from_json(const BasicJsonType& j, std::optional<T>& opt)
         opt.emplace(j.template get<T>());
     }
 }
-
-//#endif // JSON_USE_IMPLICIT_CONVERSIONS
 #endif // JSON_HAS_CPP_17
 
 // overloads for basic_json template parameters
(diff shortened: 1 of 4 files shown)


vtstmn (changes requested):
OK with code modification and unit test.
Shouldn't the from_json() function for std:optional be inlined like other from_json() functions?

gregmarr:
> Shouldn't the from_json() function for std:optional be inlined like other from_json() functions?

No, that's not necessary.  This is a template function that is not a full specialization, so the `inline` modifier is not required.  It is only required when fully specialized.

nlohmann:
> No, that's not necessary.  This is a template function that is not a full specialization, so the `inline` modifier is not required.  It is only required when fully specialized.

Would you propose reverting this?

gregmarr:
>Would you propose reverting this?

It doesn't hurt, but it is redundant.  You never know when they may add a clang tidy check for it.  Since the PR hasn't been merged yet, it might be worth it.

vtstmn (commented):
OK with all modifications

jozefizso:
Thanks for this change! It helps us a lot as we were patching the code to remove the `JSON_USE_IMPLICIT_CONVERSIONS` to enable support for `std::optional<T>`.

PR #4747, opened 2025-04-16
Check JSON_HAS_THREE_WAY_COMPARSION for existence not value

alexanderbock:
docs/mkdocs/docs/api/macros/index.md describes the preprocessor macros that can be defined to configure nlohmann.  One of these is `JSON_HAS_THREE_WAY_COMPARISON`. However when setting it, the code before in the value_t.hpp was

```cpp
#if JSON_HAS_THREE_WAY_COMPARISON
    #include <compare> // partial_ordering
#endif
```

producing
```
1>A:\openspace\OpenSpace\ext\json\include\nlohmann\detail\value_t.hpp(17,35): error C1017: invalid integer constant expression
```

This PR changes `#if` to `#ifdef`


alexanderbock:
Don't have `make` on my system and not going to bother to install it. If someone else wants to fix this issue, go ahead

gregmarr:
@alexanderbock There is nothing to fix here on the library side.  You are using it incorrectly.  The documentation you referenced says:

> This macro indicates whether the compiler and standard library support 3-way comparison. Possible values are 1 when supported or 0 when unsupported.

So you need to use it like this to enable it: 

    #define JSON_HAS_THREE_WAY_COMPARISON 1

PR #4758, opened 2025-04-24
Fix cppcheck warning

### LINKED ISSUE #4755: Fix cppcheck 1.5.1 warning
In https://github.com/nlohmann/json/pull/4753, a new warning was raised by the new Cppcheck version:

```
include/nlohmann/detail/output/serializer.hpp:835:36: warning: %g in format string (no. 2) requires ‘double’ but the argument type is ‘number_float_t’. [invalidPrintfArgType_float]
        std::ptrdiff_t len = (std::snprintf)(number_buffer.data(), number_buffer.size(), “%.*g”, d, x);
                                   ^
```

We need to check whether to suppress or fix it.

nlohmann:
The serialization of floating-point numbers is handled in two code paths. If `number_float_t` is an IEEE-754 single or double precision number, the Grisu2 algorithm is used. Otherwise, the serialization is done via `std::snprintf`. By updating cppcheck to 1.5.1, it warns that the used specifier `"%.*g"` only works with `double`.

This PR adds a conversion to `double` in the `std::snprintf` call to silence this warning.

Closes #4755

### DIFF
--- a/cmake/requirements/requirements-cppcheck.txt
+++ b/cmake/requirements/requirements-cppcheck.txt
@@ -1,1 +1,1 @@
-cppcheck==1.5.0
+cppcheck==1.5.1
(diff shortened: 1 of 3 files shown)


nlohmann:
Maybe related: https://github.com/nlohmann/json/pull/3929/changes

PR #4760, opened 2025-04-26
Specialize char_traits for std::byte to fix from_msgpack (fixes #4756)

### LINKED ISSUE #4756: Incompatibility of std::char_traits and std::byte in Xcode 16.3
### Description

When using [nlohmann/json](https://github.com/nlohmann/json/tree/v3.12.0), specifically `nlohmann::ordered_json::from_msgpack()` with an argument (buffer) of the type `std::vector<std::byte>` in some tests within the [power-grid-model repo](https://github.com/PowerGridModel/power-grid-model/tree/main), these fail to compile with Xcode 16.3.

In more detail, we use `c++20` and locally, the build fails for Xcode >=16.3. Specifically the problem arises from [this line](https://github.com/PowerGridModel/power-grid-model/blob/fc123b369cc7de95094da4da7f8c132ee8979a6d/tests/native_api_tests/test_api_serialization.cpp#L86), where `nlohmann::ordered_json::from_msgpack()` is used together with `std::vector<std::byte>`. This issue comes from the fact that `std::char_traits` doesn't natively support types other than `char`, `wchar_t`, `char8_t`, `char16_t`, `char32_t` (see [Xcode 16.3 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-16_3-release-notes)), unless some specializations are provided.

Now, in your library, you do provide some specializations in [this place](https://github.com/nlohmann/json/blob/85df7ed59350563c36e39ba999212229ba4bfb1c/include/nlohmann/detail/meta/type_traits.hpp#L191), but there is none compatible with `std::byte`. Furthermore, the use of `std::byte` would make both our and our users' lives easier (which we already supported before the release of Xcode 16.3), so it is desirable that a specialization is included.

For the above reasons, simply using `char` instead is not a solution and, even though we did try to hack this in a [PR](https://github.com/PowerGridModel/power-grid-model/pull/956/files), MSVC really [dislikes it](https://github.com/PowerGridModel/power-grid-model/actions/runs/14493786141/job/40656224096?pr=956). Hence, I would like to request adding the `std::byte` specialization to `std::char_traits`, such that this can easily be used in that way. 

### Reproduction steps

Clone the [power-grid-model repo](https://github.com/PowerGridModel/power-grid-model.git) and try to compile with Xcode 16.3. The [build guide](https://power-grid-model.readthedocs.io/en/stable/advanced_documentation/build-guide.html) may be useful for this. And you would get compilation errors pointing to `nlohmann/json` and `std::char_traits`.

Expected is to be able to compile without any issue. Actual results is a compilation error pointing towards `nlohmann/json`.

### Minimal code example

```Shell

```

### Error messages

```Shell

```

Apple clang version 17.0.0 (clang-1700.0.13.3) and arm64-apple-darwin24.4.0

### Library version

3.12.0

### Validation

XueSongTap:
### Summary

Ensure `from_msgpack` properly handles containers of `std::byte` by mapping `std::byte` to `unsigned char`.

### Changes

- Specialize `char_traits<std::byte>` when `__cpp_lib_byte >= 201603L`.
- Implement `to_int_type`, `to_char_type`, and `eof()` for `std::byte` using `std::char_traits<char>` as a base.
- Guard specialization with `#if defined(__cpp_lib_byte) && __cpp_lib_byte >= 201603L` to maintain compatibility with C++11.

### Reason

Without `char_traits<std::byte>`, using `from_msgpack` on `std::vector<std::byte>` leads to a appleclang compilation failure, because `char_traits<T>` is required by the input adapter.

Fixes #4756

### Testing Done

- Verified that existing tests pass (`ctest --test-dir build -j 10`).
- Added new test cases in `tests/src/unit-msgpack.cpp`

### Checklist

- [x] Changes are described in detail.
- [x] Linked to [[existing issue #4756](https://github.com/nlohmann/json/issues/4756)](https://github.com/nlohmann/json/issues/4756).
- [x] Code coverage remains 100%.
- [x] No breaking changes to the public API.
- [x] Source code is amalgamated with `make amalgamate`.
- [ ] Documentation update not required (behavioral fix, no API change).

### Additional Note on Test Behavior

The newly added unit test verifying `from_msgpack` with `std::byte` containers is only strictly required and triggered on stricter compilers such as:

- Apple Clang 17.0.0 (clang-1700.0.13.3)
- Target: arm64-apple-darwin24.2.0

On most GCC and older Clang versions, this specific issue does not manifest because of more relaxed type deduction and `char_traits` usage, so the corresponding test would pass without triggering a compilation error even without the fix.  
However, maintaining the specialization ensures consistent behavior and future-proof compatibility across all supported compilers.

### DIFF
--- a/include/nlohmann/detail/meta/type_traits.hpp
+++ b/include/nlohmann/detail/meta/type_traits.hpp
@@ -13,7 +13,7 @@
 #include <tuple> // tuple
 #include <type_traits> // false_type, is_constructible, is_integral, is_same, true_type
 #include <utility> // declval
-#include <cstddef>
+#include <cstddef> // byte
 
 #include <nlohmann/detail/iterators/iterator_traits.hpp>
 #include <nlohmann/detail/macro_scope.hpp>
--- a/include/nlohmann/detail/meta/type_traits.hpp
+++ b/include/nlohmann/detail/meta/type_traits.hpp
@@ -13,8 +13,9 @@
 #include <tuple> // tuple
 #include <type_traits> // false_type, is_constructible, is_integral, is_same, true_type
 #include <utility> // declval
-#include <cstddef> // byte
-
+#if defined(__cpp_lib_byte) && __cpp_lib_byte >= 201603L
+    #include <cstddef> // byte
+#endif
 #include <nlohmann/detail/iterators/iterator_traits.hpp>
 #include <nlohmann/detail/macro_scope.hpp>
 #include <nlohmann/detail/meta/call_std/begin.hpp>
--- a/include/nlohmann/detail/meta/type_traits.hpp
+++ b/include/nlohmann/detail/meta/type_traits.hpp
@@ -13,6 +13,7 @@
 #include <tuple> // tuple
 #include <type_traits> // false_type, is_constructible, is_integral, is_same, true_type
 #include <utility> // declval
+#include <cstddef>
 
 #include <nlohmann/detail/iterators/iterator_traits.hpp>
 #include <nlohmann/detail/macro_scope.hpp>
@@ -238,6 +239,30 @@ struct char_traits<signed char> : std::char_traits<char>
         return static_cast<int_type>(std::char_traits<char>::eof());
     }
 };
+
+#if defined(__cpp_lib_byte) && __cpp_lib_byte >= 201603L
+template<>
+struct char_traits<std::byte> : std::char_traits<char>
+{
+    using char_type = std::byte;
+    using int_type = uint64_t;
+
+    static int_type to_int_type(char_type c) noexcept
+    {
+        return static_cast<int_type>(std::to_integer<unsigned char>(c));
+    }
+
+    static char_type to_char_type(int_type i) noexcept
+    {
+        return std::byte(static_cast<unsigned char>(i));
+    }
+
+    static constexpr int_type eof() noexcept
+    {
+        return static_cast<int_type>(std::char_traits<char>::eof());
+    }
+};
+#endif
 
 ///////////////////
 // is_ functions //
--- a/tests/src/unit-msgpack.cpp
+++ b/tests/src/unit-msgpack.cpp
@@ -1914,9 +1914,12 @@ TEST_CASE("MessagePack with std::byte")
         SECTION("empty vector")
         {
             const std::vector<std::byte> empty_data;
-            CHECK_THROWS_WITH_AS(json::from_msgpack(empty_data),
-                                 "[json.exception.parse_error.110] parse error at byte 1: syntax error while parsing MessagePack value: unexpected end of input",
-                                 json::parse_error&);
+            CHECK_THROWS_WITH_AS([&]() { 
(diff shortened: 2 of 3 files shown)


tests/src/unit-msgpack.cpp
```diff
@@ -1880,3 +1880,70 @@ TEST_CASE("MessagePack roundtrips" * doctest::skip())
...
+            std::vector<uint8_t> temp = json::to_msgpack(original);
+            
+            std::vector<std::byte> msgpack_data(temp.size());
+            for (size_t i = 0; i < temp.size(); ++i) {
+                msgpack_data[i] = std::byte(temp[i]);
+            }
+            
+            json direct_result = json::from_msgpack(msgpack_data);
+            
+            // 测试PR中的临时解决方案作为对比
```
nlohmann: Please translate to English.

tests/src/unit-msgpack.cpp
```diff
@@ -1880,3 +1880,70 @@ TEST_CASE("MessagePack roundtrips" * doctest::skip())
...
+            auto const char_start = reinterpret_cast<unsigned char const*>(msgpack_data.data());
+            auto const char_end = char_start + msgpack_data.size();
+            json workaround_result = json::from_msgpack(char_start, char_end);
+            
+            // 验证两种方法产生相同的结果
```
nlohmann: Please translate to English.

include/nlohmann/detail/meta/type_traits.hpp
```diff
@@ -13,6 +13,7 @@
 #include <tuple> // tuple
 #include <type_traits> // false_type, is_constructible, is_integral, is_same, true_type
 #include <utility> // declval
+#include <cstddef>
```
nlohmann: Please add a comment why you included this header.

nlohmann (changes requested):

XueSongTap:
Hi, @nlohmann 

This PR adds a missing comment for #include <cstddef> (to clarify that it is needed for std::byte) and improves the English comments in the unit-msgpack.cpp tests for better clarity.

Could you please help take another look to make sure the added comments look good?

Thanks!

include/nlohmann/detail/meta/type_traits.hpp
```diff
@@ -13,6 +13,7 @@
 #include <tuple> // tuple
 #include <type_traits> // false_type, is_constructible, is_integral, is_same, true_type
 #include <utility> // declval
+#include <cstddef> // byte
```
nlohmann: Since you only need the header for `std::byte`, please only include it in case of

```cpp
#if defined(__cpp_lib_byte) && __cpp_lib_byte >= 201603L
```

nlohmann (changes requested):

nlohmann:
- Please rebase to latest `develop` branch to fix the GCC-related errors.
- Please follow the description at https://github.com/nlohmann/json/pull/4760/checks?check_run_id=41232828348 to sign-off your commits.

nlohmann (approved):
Looks good to me.

PR #4765, opened 2025-04-27
fix: BJData size calculation overflow

thevilledev:
Adds pre-multiplication overflow detection to catch cases where dimension products would exceed size_t max. The previous check only detected when overflow resulted in exactly 0 or SIZE_MAX, missing other cases.

Retains the original post-multiplication check for backward compatibility. Adds tests verifying overflow detection with dimensions (2^32+1)×(2^32), which previously overflowed silently to 2^32.

This prevents custom SAX handlers from receiving incorrect array sizes that could lead to buffer overflows.

### DIFF
--- a/tests/src/unit-bjdata.cpp
+++ b/tests/src/unit-bjdata.cpp
@@ -2815,6 +2815,129 @@ TEST_CASE("BJData")
 #endif
             }
 
+            SECTION("overflow detection in dimension multiplication")
+            {
+                // Simple SAX handler just to monitor if overflow is detected
+                struct SimpleOverflowSaxHandler : public nlohmann::json_sax<json>
+                {
+                    bool overflow_detected = false;
+
+                    // Implement all required virtual methods with minimal implementation
+                    bool null() override
+                    {
+                        return true;
+                    }
+                    bool boolean(bool /*val*/) override
+                    {
+                        return true;
+                    }
+                    bool number_integer(json::number_integer_t /*val*/) override
+                    {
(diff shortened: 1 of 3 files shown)


tests/src/unit-bjdata.cpp
```diff
@@ -2815,6 +2815,126 @@ TEST_CASE("BJData")
...
+                    0x4D,                                           // 'M' (uint64)
+                    0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, // 2^32 (4294967296) as little-endian
+                    0x5D                                            // ']' end dimensions
+                    // No data - we don't need it for this test, we just want to hit the overflow check
+                };
+
+                // Test with overflow dimensions using SAX parser
+                {
+                    SimpleOverflowSaxHandler handler;
+                    auto result = json::sax_parse(bjdata_payload, &handler,
```
nlohmann: This variable is not used which is a warning/error in the used compilers.

thevilledev: Yup, will push a fix soon!

nlohmann: Thanks! There are also some warnings/errors unrelated to your changes which will be fixed with #4763.

nlohmann:
Please rebase to the latest `develop` branch to fix the GCC-related errors.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #4764, opened 2025-04-27
Make library work with C++20 modules

nlohmann:
- Added a test to compile the library as module with the latest Clang and GCC.
- Used the fix from #4759.

Closes #4759.

PR #4768, opened 2025-04-28
Build ci_module_cpp20 with MSVC

nlohmann:
Add a build step for `ci_module_cpp20` for MSVC after merging #4764.


nlohmann:
Fixed by #4952.

PR #4775, opened 2025-05-07
Replace deprecated std::is_trivial in C++26

### LINKED ISSUE #4778: Deprecation warning with gcc 15.1.1: struct std::is_trivial’ is deprecated
### Description

Hi,
starting with GCC 15.1.1 on Arch Linux I'm getting the following deprecation warning:
```
<...>/nlohmann-json-headers/single_include/nlohmann/json.hpp: In static member function ‘static CharType nlohmann::json_abi_v3_11_3::detail::binary_writer<BasicJsonType, CharType>::to_char_type(uint8_t)’:
<...>/nlohmann-json-headers/single_include/nlohmann/json.hpp:16844:28: warning: ‘template<class _Tp> struct std::is_trivial’ is deprecated: use 'is_trivially_default_constructible && is_trivially_copyable' instead [-Wdeprecated-declarations]
16844 |         static_assert(std::is_trivial<CharType>::value, "CharType must be trivial");
      |                            ^~~~~~~~~~
In file included from /usr/include/fmt/base.h:20,
                 from /usr/include/fmt/format.h:41,
                 from foo.cpp:7:
/usr/include/c++/15.1.1/type_traits:898:5: note: declared here
  898 |     is_trivial
      |     ^~~~~~~~~~
```

The troublesome code snippet is the following:
```cpp
    template < typename C = CharType,
               enable_if_t < std::is_signed<C>::value && std::is_unsigned<char>::value > * = nullptr >
    static CharType to_char_type(std::uint8_t x) noexcept
    {
        static_assert(sizeof(std::uint8_t) == sizeof(CharType), "size of CharType must be equal to std::uint8_t");
        static_assert(std::is_trivial<CharType>::value, "CharType must be trivial");
        CharType result;
        std::memcpy(&result, &x, sizeof(x));
        return result;
    }
```

### Reproduction steps

-

-

### Minimal code example

```Shell

```

### Error messages

```Shell

```

Arch Linux (x86_64), GCC 15.1.1

### Library version

3.12.0

### Validation

nmpassthf:
std::is_trivial is deprecated in C++26 ( [P3247: Deprecate the notion of trivial types](https://wg21.link/p3247) ).
GCC 15.1.1 with `-std=c++26` will generate a warning.
Using `std::is_trivially_copyable` && `std::is_trivially_default_constructible` checks for `CharType`  to replace `std::is_trivial`.


Monobr0w:
ahh thx! :-D

nlohmann:
GCC versions < 5 do not support `is_trivially_copyable` and `is_trivially_default_constructible` (see https://stackoverflow.com/questions/25123458/is-trivially-copyable-is-not-a-member-of-std). Either provide an alternative implementation as sketched in SO or keep using `std::is_trivial` for older GCC.

nlohmann (changes requested):
GCC versions < 5 do not support `is_trivially_copyable` and `is_trivially_default_constructible` (see https://stackoverflow.com/questions/25123458/is-trivially-copyable-is-not-a-member-of-std). Either provide an alternative implementation as sketched in SO or keep using `std::is_trivial` for older GCC.

nmpassthf:
As this check is only performed within the `to_char_type()` method, a macro is used to work around missing `std::is_trivially_*` support in GCC 4.x.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1813,8 +1813,18 @@ class binary_writer
                enable_if_t < std::is_signed<C>::value && std::is_unsigned<char>::value > * = nullptr >
     static CharType to_char_type(std::uint8_t x) noexcept
     {
-        static_assert(sizeof(std::uint8_t) == sizeof(CharType), "size of CharType must be equal to std::uint8_t");
+        // Workaround missing "is_trivially_copyable" &&
+        // "is_trivially_default_constructible" in g++ < 5.0.
+        // Fallback to "is_trivial" check.
+        // See <https://github.com/nlohmann/json/issues/4778>.
+#if __GNUG__ && __GNUC__ < 5
```
gregmarr: Feels like there should be some `defined()` checks here.  Also not sure why you are checking `__GNUG__` here since this has to be C++.

nlohmann:
Note the library includes Hedley, so there are already GCC version detection macros available.

nmpassthf:
> Note the library includes Hedley, so there are already GCC version detection macros available.

Because many older compilers do not support the is_trivially_* traits, for example, `clang 3.4.1`,
I think directly checking whether the C++ standard version is greater than 26 is a simpler approach.
I couldn't find a straightforward way to check the C++ version in the Hedley documentation, so I decide to use `__cplusplus` and `_MSVC_LANG` (as referenced from [this](https://stackoverflow.com/questions/52379233/is-has-cxx17-macro-usable-in-custom-project-headers-to-enable-c17-language)).

gregmarr:
The `JSON_HAS_CPP_nn` macros are here: https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/macro_scope.hpp#L35

Since these replacement functions are available in C++11, and we're just looking for library versions that haven't completely implemented C++11, I'm not sure that checking for C++26 is the best way to determine whether to use the two separate ones.  On the other hand, if we prefer `is_trivial` for earlier versions, and only fall back to the others when it isn't available, then I guess it makes sense to check for C++26.  In that case, we should add `JSON_HAS_CPP_26` in the above location and use that.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1813,8 +1813,18 @@ class binary_writer
                enable_if_t < std::is_signed<C>::value && std::is_unsigned<char>::value > * = nullptr >
     static CharType to_char_type(std::uint8_t x) noexcept
     {
-        static_assert(sizeof(std::uint8_t) == sizeof(CharType), "size of CharType must be equal to std::uint8_t");
+        // Workaround missing "is_trivially_copyable" &&
+        // "is_trivially_default_constructible" in multiple older compiler.
+        // Fallback to "is_trivial" check if C++26 not available.
+        // See <https://github.com/nlohmann/json/issues/4778>.
```
gregmarr: Can we word this as something like the below?

The `is_trivial` function is deprecated in C++26.  The replacement is to use `is...` and `is...`.  However, some older libraries that implement `is_trivial` don't implement those functions, and it's not easy to detect all of them, so use `is_trivial` unless we are using a standard where it has been deprecated.

nlohmann: I agree with this and https://github.com/nlohmann/json/pull/4775#issuecomment-2884361627.

nlohmann (approved):
Looks good to me.

gregmarr (approved):

PR #4805, opened 2025-06-01
Use binary_t::value_type

### LINKED ISSUE #4804: `from_cbor` incompatible with `std::vector<std::byte>` as `binary_t`
### Description

Trying to use `json::from_cbor` with a `binary_t` set to `std::vector<std::byte>` will fail because the decoder appears to be hard-coded to `push_back` elements of `std::uint8_t`.

It appears to be related to #4756 where they report essentially the same issue for `from_msgpack`; although that  one was fixed in #4760 by adding a specialisation to `char_traits`. It appears that type trait is not being used in the implementation of `from_cbor`.

### Reproduction steps

Attempt to compile the below minimal example.

Expected to compile.

### Minimal code example

```c++
#include <nlohmann/json.hpp>

using json = nlohmann::basic_json<std::map,        // ObjectType
                                  std::vector,     // ArrayType
                                  std::string,     // StringType
                                  bool,            // BooleanType
                                  std::int64_t,    // NumberIntegerType
                                  std::uint64_t,   // NumberUnsignedType
                                  double,          // NumberFloatType
                                  std::allocator,  // AllocatorType
                                  nlohmann::adl_serializer,  // JSONSerializer
                                  std::vector<std::byte>,    // BinaryType
                                  void                       // CustomBaseClass
                                  >;

int main(){
    std::vector<char> data;
    auto decoded = json::from_cbor(data);
    return 0;
}
```
See also on [Compiler Explorer](https://godbolt.org/z/a3xbG4xaW).

### Error messages

On Clang 20:
```Shell

/opt/compiler-explorer/libs/nlohmann_json/trunk/single_include/nlohmann/json.hpp:12768:20: error: no matching member function for call to 'push_back'
 12768 |             result.push_back(static_cast<std::uint8_t>(current));
```

Clang 20, GCC 15, both on Compiler Explorer (i.e. linux?)

### Library version

trunk

### Validation

nlohmann:
Closes #4804

### DIFF
--- a/tests/src/unit-regression2.cpp
+++ b/tests/src/unit-regression2.cpp
@@ -1114,8 +1114,9 @@ TEST_CASE("regression tests 2")
 #ifdef JSON_HAS_CPP_17
     SECTION("issue #4804: from_cbor incompatible with std::vector<std::byte> as binary_t")
     {
-        const std::vector<char> data;
+        const std::vector<std::uint8_t> data = {0x80};
         const auto decoded = json_4804::from_cbor(data);
+        CHECK((decoded == json_4804::array()));
     }
 #endif
 }
--- a/tests/src/unit-regression2.cpp
+++ b/tests/src/unit-regression2.cpp
@@ -42,6 +42,22 @@ using ordered_json = nlohmann::ordered_json;
     #elif __has_include(<experimental/optional>)
         #include <experimental/optional>
     #endif
+
+    /////////////////////////////////////////////////////////////////////
+    // for #4804
+    /////////////////////////////////////////////////////////////////////
+    using json_4804 = nlohmann::basic_json<std::map,        // ObjectType
+    std::vector,     // ArrayType
+    std::string,     // StringType
+    bool,            // BooleanType
+    std::int64_t,    // NumberIntegerType
+    std::uint64_t,   // NumberUnsignedType
+    double,          // NumberFloatType
+    std::allocator,  // AllocatorType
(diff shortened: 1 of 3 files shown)


gregmarr (approved):

tests/src/unit-regression2.cpp
```diff
@@ -1094,6 +1110,14 @@ TEST_CASE("regression tests 2")
         CHECK(j.type_name() == "invalid");
     }
 #endif
+
+#ifdef JSON_HAS_CPP_17
+    SECTION("issue #4804: from_cbor incompatible with std::vector<std::byte> as binary_t")
+    {
+        const std::vector<char> data;
+        const auto decoded = json_4804::from_cbor(data);
```
burnpanck: Is this test-case even supposed to run successfully? I'm aware that I posted this test-case when raising the issue, but I was expecting it to compile, not necessarily to actually run successfully. This attempts to decode the empty byte-string as CBOR. While this *may* be valid CBOR (though [CBOR playground](https://cbor.me/) rejects it), it's unclear to me how that would be represented within the JSON object model. In particular, the CBOR representation of the empty object is not the same as the empty byte-string. So perhaps `"\xA0"` (the correct CBOR representation of the empty map/object) would be a better test-case.

nlohmann: Good catch. Yes, I indeed just copied your example and made it compile. Very strange to see that some compilers seem to allow parsing from an empty vector.

PR #4814, opened 2025-06-11
[SingleInclude] Missing definition of `iter_impl` and `iteration_proxy` in `json_fwd`

loumalouomega:
Greetings, in https://github.com/KratosMultiphysics/Kratos/ we use this library and the library has helped a lot in our work. The only thing is that in order to properly work we need to modify `single_include/nlohmann/json_fwd.hpp` in order to include `iter_impl` and `iteration_proxy`.

Recently (https://github.com/KratosMultiphysics/Kratos/pull/13461) we updated the lib version, but each time we do an update we need to update the definition of the `single_include/nlohmann/json_fwd.hpp`. I guess if we merge our modification we could skip this issue. This is not a big issue so if doesn't fit the contribution guidelines I can simply close it.


loumalouomega:
Huum, Okay, I will check this

gregmarr:
The `detail` namespace is not part of the public API.  It's for internal use only and may change at any time.  What are you doing that requires access to these types?

loumalouomega:
> The `detail` namespace is not part of the public API. It's for internal use only and may change at any time. What are you doing that requires access to these types?

We are implementing our custom iterators. If this does not fit the API, I will close the PR, for me was just to skip the "patch" each time i update the lib.

gregmarr:
> We are implementing our custom iterators. 

Wouldn't you need the full header for that?

> If this does not fit the API, I will close the PR, for me was just to skip the "patch" each time i update the lib.

Rather than patching, you could probably just use a local header that wraps this one if you really need just the forward declarations.

`my_json_fwd.hpp`
```

NLOHMANN_JSON_NAMESPACE_BEGIN

namespace detail
{
class iter_impl;

template<typename IteratorType>
class iteration_proxy;
}  // namespace detail

NLOHMANN_JSON_NAMESPACE_END
```

gregmarr:
FYI, @nlohmann has the final say here, I'm just a long-time user and part-time contributor.

nlohmann:
@gregmarr is right - we don't want to expose the `detail` namespace. It's not part of the public API and may change at any time without warning.

loumalouomega:
> @gregmarr is right - we don't want to expose the `detail` namespace. It's not part of the public API and may change at any time without warning.

I will close this then

PR #4830, opened 2025-06-24
:bug: add missing header to input_adapters.hpp

ddkilzer:
While the WebKit project was in the process of adopting Swift (clang) modules, a build error occurred because `json.hpp` uses `std::streambuf`, but it doesn't include the `<streambuf>` header.  (A dependent project included `json.hpp`, then WebKit included this project's header, which is how it was found.)

The fix is to add the missing header to `input_adapters.hpp`.

- [X] The changes are described in detail, both the what and why.
- [X] The [Code coverage](https://coveralls.io/github/nlohmann/json) remained at 100%. A test case for every new line of code.
- [X] The source code is amalgamated by running `make amalgamate`.


gregmarr (approved):

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #4835, opened 2025-06-28
knuth up arrow implmentation

zainabhelu:


gregmarr:
Why would this belong in this library?

nlohmann:
I don't see the value of having this in the library. Can you explain the benefit?

skhaz:
This is AI generated crap, I got some non-sense on my repository too.

Issue #4842, opened 2025-07-04
json destructor does not use the provided allocator


veselink1:
### Description

The non-recursive destruction implementation uses a std::vector as a stack to avoid recursive destruction of nested objects.

The vector that is used does not use the provided allocator type.

### Reproduction steps

```
json j = {
  {"object", {
    {"currency", "USD"},
    {"value", 42.99}
  }}
};

// j goes out of scope
```

`basic_json::destroy` is called from `~basic_json` and uses a vector as a stack for the "object".

### Expected vs. actual results

I expected the operation to use the given allocator to allocate the vector of basic_json, but it just uses new/delete.

### Minimal code example

```Shell

```

### Error messages

```Shell

```

### Compiler and operating system

Apple clang version 15.0.0 / macOS

### Library version

v3.12.0

### Validation

- [x] The bug also occurs if the latest version from the [`develop`](https://github.com/nlohmann/json/tree/develop) branch is used.
- [x] I can successfully [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests).

PR #4843, opened 2025-07-04
Use the provided allocator in destroy() (#4842)

veselink1:
Uses the provided allocator to allocate the stack used to avoid recursion in the destroy() implementation used by ~basic_json.

This addresses issue #4842, by changing the std::vector to use the user-provided allocatorl

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -570,7 +570,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
             if (t == value_t::array || t == value_t::object)
             {
                 // flatten the current json_value to a heap-allocated stack
-                std::vector<basic_json> stack;
+                std::vector<basic_json, allocator_type> stack;
 
                 // move the top-level items to stack
                 if (t == value_t::array)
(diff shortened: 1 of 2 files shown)


nlohmann:
Please rebase to the latest `develop` branch to have a green CI.

nlohmann:
Thanks a lot for the fix, @veselink1, and sorry this sat for so long. I merged the latest `develop` into your branch and added a regression test to `unit-allocator.cpp`. The test uses a counting allocator and checks that destroying a nested array or object allocates the temporary stack through it. It fails without your change and passes with it. Once CI is green, this is good to go.

gregmarr:
@nlohmann I wonder if any of the other temporary vectors added in the recent iterative cleanups should also use the provided allocator.

nlohmann:
@gregmarr Good question, thanks! I went through them. Most of the new temporary vectors hold only pointers, iterators and counters: the copy worklist and the stacks for merging (#5547), hashing (#5546), `dump()` (#5285) and the binary readers (#5505). That's the same as the parser's `ref_stack` or the lexer's `token_string`, which have always used `std::allocator`.

The one exception is the deep copy's key scratch space from #5389. It holds `std::pair<key_type, basic_json>` values, so it should use the provided allocator just like the destructor's stack here. That's #5573, which also documents that `AllocatorType` allocates the JSON values while most temporary storage still uses `std::allocator`.

PR #4849, opened 2025-07-20
chore(cmake): natvis building option

niceEli:
Allow a CMake option to allow choosing whether or not to build the .natvis file, there is no issue for this but it caused some issues with me so I fixed it.

The CMake option `JSON_BuildNatvis` being `ON` or `OFF` controls its build, it will still not build if MSVC is not `ON`.

This wont break anyones projects, but if people want to optionally not build the natvis file then they can.


CMakeLists.txt
```diff
@@ -208,7 +209,7 @@ if(JSON_Install)
         FILES ${NLOHMANN_JSON_CMAKE_PROJECT_CONFIG_FILE} ${NLOHMANN_JSON_CMAKE_VERSION_CONFIG_FILE}
         DESTINATION ${NLOHMANN_JSON_CONFIG_INSTALL_DIR}
     )
-    if (NLOHMANN_ADD_NATVIS)
+    if (NLOHMANN_ADD_NATVIS AND JSON_BuildNatvis)
```
gregmarr: Is this part needed?  `NLOHMANN_ADD_NATVIS` will already not be defined if `JSON_BuildNatvis` is false.

niceEli: yea your right its kinda redundant

PR #4851, opened 2025-07-22
Make sure tests can be run offline and by adding new cmake switch

LocutusOfBorg:
As said, I would like to package test_data jsons in Debian...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1883,6 +1883,8 @@ json/tests/src/make_test_data_available.hpp:23: FATAL ERROR: REQUIRE( utils::che
 
 In case you have downloaded the library rather than checked out the code via Git, test `cmake_fetch_content_configure` will fail. Please execute `ctest -LE git_required` to skip these tests. See [issue #2189](https://github.com/nlohmann/json/issues/2189) for more information.
 
+Some tests are requiring network to be properly execute. They are labeled as `git_required`. Please execute `ctest -LE git_required` to skip these tests. See [issue #4851](https://github.com/nlohmann/json/issues/4851) for more information.
+
 Some tests change the installed files and hence make the whole process not reproducible. Please execute `ctest -LE not_reproducible` to skip these tests. See [issue #2324](https://github.com/nlohmann/json/issues/2324) for more information. Furthermore, assertions must be switched off to ensure reproducible builds (see [discussion 4494](https://github.com/nlohmann/json/discussions/4494)).
 
 Note you need to call `cmake -LE "not_reproducible|git_required"` to exclude both labels. See [issue #2596](https://github.com/nlohmann/json/issues/2596) for more information.


LocutusOfBorg:
rebased

cmake/download_test_data.cmake
```diff
@@ -1,14 +1,19 @@
 set(JSON_TEST_DATA_URL     https://github.com/nlohmann/json_test_data)
 set(JSON_TEST_DATA_VERSION 3.1.0)
 
-include(ExternalProject)
+set(JSON_TEST_DATA_DIR /usr/share/json_test_data-${JSON_TEST_DATA_VERSION})
```
nlohmann: Why hard-code this directory?

LocutusOfBorg: Not sure how fast and in an incompatible way they might change... If you don't plan to have multiple versions at the same time, we can just use json_test_data and simplify!

LocutusOfBorg: This PR is a POC work, I'm still waiting for test-data to be accepted in sid (not sure for licensing)

LocutusOfBorg:
json-test-data is now part of sid, and I removed the hardcoded version from the directory. @nlohmann  can you please have another look? thanks

cmake/download_test_data.cmake
```diff
@@ -1,14 +1,19 @@
 set(JSON_TEST_DATA_URL     https://github.com/nlohmann/json_test_data)
 set(JSON_TEST_DATA_VERSION 3.1.0)
 
-include(ExternalProject)
+set(JSON_TEST_DATA_DIR /usr/share/json_test_data)
```
nlohmann: This is still a hardcoded path here.

cmake/download_test_data.cmake
```diff
@@ -1,14 +1,19 @@
...
+
+if(EXISTS ${JSON_TEST_DATA_DIR})
+  set(JSON_TestDataDirectory ${JSON_TEST_DATA_DIR})
```
nlohmann: The idea of `JSON_TestDataDirectory` is to pass the directory from outside.

cmake/download_test_data.cmake
```diff
@@ -1,14 +1,19 @@
...
+endif()
 
 # if variable is set, use test data from given directory rather than downloading them
 if(JSON_TestDataDirectory)
     message(STATUS "Using test data in ${JSON_TestDataDirectory}.")
     add_custom_target(download_test_data)
     file(WRITE ${CMAKE_BINARY_DIR}/include/test_data.hpp "#define TEST_DATA_DIRECTORY \"${JSON_TestDataDirectory}\"\n")
 else()
+    include(ExternalProject)
```
nlohmann: In the latest `develop` this is already included at the top.

tests/CMakeLists.txt
```diff
@@ -192,6 +192,9 @@ if (JSON_Install)
 endif()
 
 add_subdirectory(cmake_add_subdirectory)
-add_subdirectory(cmake_fetch_content)
-add_subdirectory(cmake_fetch_content2)
+if (NOT JSON_No_Internet)
+    add_subdirectory(cmake_fetch_content)
```
nlohmann: I think it makes more sense to add a test label like https://github.com/nlohmann/json/blob/54be9b04f0ec65d0bcfe0da54e7f01ea86fbfc3e/tests/cmake_fetch_content/CMakeLists.txt#L15 to annotate tests that require internet.

nlohmann (changes requested):
I think the idea of the PR can be achieved by adding test labels and leaving the CMake files as is. Then a test directory can be passed via `-DJSON_TEST_DATA_DIR=...` and tests that require internet can be excluded via `ctest -LE ...`.

LocutusOfBorg:
So nothing left to merge :D

nlohmann:
No, I think you still need to annotate the tests that require internet so you can skip those.

LocutusOfBorg:
@nlohmann the tests requiring git are the same requiring internet :)
Maybe git_required should be changed into network_required, because they need git to clone from network, but looks too much to add another label just for the same purpose?

nlohmann:
I'd rather not add an alias label. However, maybe this text from the README could be updated:

> In case you have downloaded the library rather than checked out the code via Git, test cmake_fetch_content_configure will fail. Please execute ctest -LE git_required to skip these tests. See https://github.com/nlohmann/json/issues/2189 for more information.

Thoughts?

nlohmann (changes requested):
Please revert this change. Renaming the label will be breaking other people's workflows.

LocutusOfBorg:
What do you think about this one?

nlohmann (changes requested):
No, please do not add another label. This is all complexity down the road. I think a note to the README that if somebody wants to execute the test suite without internet, then `git_required` is there to help.

(Sorry for being so insistent, but this is a side project, and I have to be careful of adding stuff.)

LocutusOfBorg:
ack done

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

LocutusOfBorg:
thanks to you!

PR #4857, opened 2025-07-26
Give CMake message about the version of the code used only once

vadz:
When consuming the library via `add_subdirectory()` using CMake, the
message about the version of the code being used was given every time
CMake was re-run, even if it happened due to the changes in the super
project completely unrelated to this library.

This was somewhat annoying, so avoid this by remembering the version of
the code used in the cache and not giving the message about it again if
it didn't change since the last run.

-----

Sorry, I don't follow the check list in your guidelines because this change doesn't affect the code, only the build system, so the items in it are mostly inapplicable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,10 +87,14 @@ set(NLOHMANN_JSON_PKGCONFIG_INSTALL_DIR     "${CMAKE_INSTALL_DATADIR}/pkgconfig"
 
 if (JSON_MultipleHeaders)
     set(NLOHMANN_JSON_INCLUDE_BUILD_DIR "${PROJECT_SOURCE_DIR}/include/")
-    message(STATUS "Using the multi-header code from ${NLOHMANN_JSON_INCLUDE_BUILD_DIR}")
+    set(NLOHMANN_JSON_CODE_VERSION_NOW "multi-header")
 else()
     set(NLOHMANN_JSON_INCLUDE_BUILD_DIR "${PROJECT_SOURCE_DIR}/single_include/")
-    message(STATUS "Using the single-header code from ${NLOHMANN_JSON_INCLUDE_BUILD_DIR}")
+    set(NLOHMANN_JSON_CODE_VERSION_NOW "single-header")
+endif()
+if(NOT "${NLOHMANN_JSON_CODE_VERSION_NOW}" STREQUAL "${NLOHMANN_JSON_CODE_VERSION}")
+    set(NLOHMANN_JSON_CODE_VERSION "${NLOHMANN_JSON_CODE_VERSION_NOW}" CACHE INTERNAL "")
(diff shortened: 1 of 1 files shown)


nlohmann:

vadz:
Done. I've also fixed accidental dependency on CMake 3.13 feature which was completely unnecessary.

nlohmann (changes requested):
It seems this change only takes the `JSON_MultipleHeaders` case into account, yet there is a similar output for all other flags (though not always for the default case). As this makes the CMake even more complex, I am hesitant to merge it like this. If we cannot find a denser way to fix the issue, I must say I'd rather keep "my" CMake clean and live with the annoyance of upstream projects.

vadz:
> It seems this change only takes the `JSON_MultipleHeaders` case into account, yet there is a similar output for all other flags (though not always for the default case).

Yes, this is the only message which you get by default, so it was important to fix at least this. I've intentionally tried to keep the changes as small as possible.

> As this makes the CMake even more complex, I am hesitant to merge it like this. If we cannot find a denser way to fix the issue,

We can define a function and use it for all options, which be denser, but probably not less complex.

> I must say I'd rather keep "my" CMake clean and live with the annoyance of upstream projects.

But this is _really_ annoying. How do you deal with this when using the library in your own projects?

nlohmann:
> > As this makes the CMake even more complex, I am hesitant to merge it like this. If we cannot find a denser way to fix the issue,
> 
> We can define a function and use it for all options, which be denser, but probably not less complex.

Please make a proposal - but honestly I don't really look forward to override CMake's `message` function.

> > I must say I'd rather keep "my" CMake clean and live with the annoyance of upstream projects.
> 
> But this is _really_ annoying. How do you deal with this when using the library in your own projects?

I never had an issue with this, and so far, also never got a ticket from outside.

vadz:
> Please make a proposal - but honestly I don't really look forward to override CMake's `message` function.

Well, the goal is not to override it but to avoid using it needlessly. In fact, the _simplest_ fix if you're really looking for one is to just remove all these pretty useless (because they're not communicating the result of any tests but just echo back the options you yourself chose to you for some reason) `message()`s. Or make them `VERBOSE`, maybe. What do you think?

Doing anything more complicated doesn't seem worth it, considering your reaction, so otherwise please just close this. I'll just keep it in my local fork or maybe stop using `add_subdirectory()` and just declare my own `INTERFACE` target.

> I never had an issue with this, and so far, also never got a ticket from outside.

I don't understand, am I doing something wrong or is nobody ever changing their CMakefiles? Because whenever I make a change in any of CMakefiles in my project I get this completely inconsequential message from this library (which is just one of a dozen or so libraries used).

nlohmann:
I frankly do not care about the output when I rerun CMake.

You can try to set [`CMAKE_MESSAGE_LOG_LEVEL`](https://cmake.org/cmake/help/latest/variable/CMAKE_MESSAGE_LOG_LEVEL.html#variable:CMAKE_MESSAGE_LOG_LEVEL) to `NOTICE` before including the library.

PR #4858, opened 2025-07-27
Example: Add student data JSON usage example

aryanrahar:
This PR adds a simple C++ example in `docs/examples/student_data.cpp` showing how to represent structured student data using the nlohmann::json library.

It includes:
- Basic object creation
- Nested key-value pairs
- Realistic field types (string, int, bool, object)

Useful for beginners learning to work with structured JSON objects.

### DIFF
--- a/docs/mkdocs/docs/examples/student_data.cpp
+++ b/docs/mkdocs/docs/examples/student_data.cpp
@@ -1,0 +1,22 @@
+#include <iostream>
+#include <nlohmann/json.hpp>
+
+using json = nlohmann::json;
+
+int main() {
+    // Realistic example: storing student information
+    json student = {
+        {"name", "Aryan"},
+        {"roll_no", 1024},
+        {"branch", "CSE"},
+        {"marks", {
+            {"Math", 85},
+            {"Physics", 90},
+            {"Chemistry", 78}
+        }},
+        {"is_enrolled", true}
+    };
+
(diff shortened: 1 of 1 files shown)


nlohmann:
What is the benefit of adding this example?

aryanrahar:
> What is the benefit of adding this example?

This adds a beginner‑friendly “student record” example that shows all core JSON types (string, int, bool, nested object) in one short snippet.
Most current examples focus on advanced features; this fills a practical create & dump real data gap that new users often ask for.
Happy to tweak or replace with another real‑world object if you prefer just let me know!

nlohmann:
How is it different to the one in https://github.com/nlohmann/json#json-as-a-first-class-data-type?

aryanrahar:
> How is it different to the one in https://github.com/nlohmann/json#json-as-a-first-class-data-type?

The README bit just shows random keys; my snippet wraps those primitives into a real‑world “student” object (name, roll no, nested marks).
It’s a drop in, runnable example under docs/examples, so newcomers can copy run tweak right away.
If that still feels redundant, happy to adjust or close just let me know!

nlohmann:
Thanks for your proposal, but I don't think the example is helpful. If you're willing to contribute, then maybe checking an open issue would be helpful. Or, as a recent problem, you could check out why Clang-Tidy is failing in the latest CI runs.

PR #4860, opened 2025-07-28
Make escape/unescape functions linear and not memory-storming.

puffetto:
Hello,

I am working on a project which does heavy use of JSON pointers (see #4859 ); while digging into the code I saw that the escape/unescape of JSON pointers was not linear, in example escaping "~~~~~~~~~~" requires 10 reallocations and 200 byte copies, escaping ;  "~~~~~~~~~~~~~~~~~~~~" requires 20 reallocations and 800 byte copies.

I rewrote these two functions: one allocation and exactly two cycles on the key for escaping, no allocations and at most two loops for unescaping; I understand they are way less readable in this flavour but also way more efficient,

Existing tests already cover all the new code, I had to add iterators to the unit-alt-string "alt_string" class.

I tried to follow the CONTRIBUTING.md document throughly, but it's my first PR to this project, so forgive any mistake.

A.

### DIFF
--- a/include/nlohmann/detail/string_escape.hpp
+++ b/include/nlohmann/detail/string_escape.hpp
@@ -46,7 +46,7 @@ inline StringType escape(StringType const& s)
             if (ch == CharT('~'))
             {
                 res.push_back(CharT('~'));
-		res.push_back(CharT{'0'});
+                res.push_back(CharT{'0'});
             }
             else if (ch == CharT('/'))
             {
--- a/include/nlohmann/detail/string_escape.hpp
+++ b/include/nlohmann/detail/string_escape.hpp
@@ -15,7 +15,7 @@ namespace detail
 {
 
 /*!
- * @brief Out Of Place string escaping as described in RFC 6901 (Sect. 4)
+ * @brief Returns a copy of a string escaped as described in RFC 6901 (Sect. 4)
  * @param[in] s string to escape
  * @return    escaped string
  *
@@ -45,11 +45,13 @@ inline StringType escape(StringType const& s)
         {
             if (ch == CharT('~'))
             {
-                res.append(StringType{"~0"});
+                res.push_back(CharT('~'));
+		res.push_back(CharT{'0'});
             }
             else if (ch == CharT('/'))
             {
-                res.append(StringType{"~1"});
+                res.push_back(CharT{'~'});
+                res.push_back(CharT{'1'});
             }
             else
             {
@@ -61,7 +63,7 @@ inline StringType escape(StringType const& s)
 }
 
 /*!
- * @brief In Place string unescaping as described in RFC 6901 (Sect. 4)
+ * @brief Unescapes a string as described in RFC 6901 (Sect. 4), in-place
  * @param[in] s string to unescape
  *
  */
--- a/include/nlohmann/detail/string_escape.hpp
+++ b/include/nlohmann/detail/string_escape.hpp
@@ -15,58 +15,79 @@ namespace detail
 {
 
 /*!
-@brief replace all occurrences of a substring by another string
-
-@param[in,out] s  the string to manipulate; changed so that all
-               occurrences of @a f are replaced with @a t
-@param[in]     f  the substring to replace with @a t
-@param[in]     t  the string to replace @a f
-
-@pre The search string @a f must not be empty. **This precondition is
-enforced with an assertion.**
-
-@since version 2.0.0
-*/
-template<typename StringType>
-inline void replace_substring(StringType& s, const StringType& f,
-                              const StringType& t)
-{
-    JSON_ASSERT(!f.empty());
-    for (auto pos = s.find(f);                // find the first occurrence of f
-            pos != StringType::npos;          // make sure f was found
-            s.replace(pos, f.size(), t),      // replace with t, and
-            pos = s.find(f, pos + t.size()))  // find the next occurrence of f
-    {}
-}
-
-/*!
- * @brief string escaping as described in RFC 6901 (Sect. 4)
+ * @brief Out Of Place string escaping as described in RFC 6901 (Sect. 4)
  * @param[in] s string to escape
  * @return    escaped string
  *
- * Note the order of escaping "~" to "~0" and "/" to "~1" is important.
  */
-template<typename StringType>
-inline StringType escape(StringType s)
-{
-    replace_substring(s, StringType{"~"}, StringType{"~0"});
-    replace_substring(s, StringType{"/"}, StringType{"~1"});
-    return s;
-}
+template<typename StringType> // [[nodiscard]]
+inline StringType escape(StringType const & s)
+    {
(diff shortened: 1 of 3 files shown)


nlohmann:
Clang-Tidy fails due to an update. I have not found the time to look into it - probably just some exclusions needed.

nlohmann:
> Clang-Tidy fails due to an update. I have not found the time to look into it - probably just some exclusions needed.

Oh, I see there are indeed some warnings related to the changed code. Please take a look to fix these.

puffetto:
> > Clang-Tidy fails due to an update. I have not found the time to look into it - probably just some exclusions needed.
> 
> Oh, I see there are indeed some warnings related to the changed code. Please take a look to fix these.

0d06a06629aa72788f7bcf96b2b2dbc4b41cbdd3 Should fix it

puffetto:
> [0d06a06](https://github.com/nlohmann/json/commit/0d06a06629aa72788f7bcf96b2b2dbc4b41cbdd3) Should fix it

Ok, it did not, more warnings around... will take a look in the evening.

include/nlohmann/detail/string_escape.hpp
```diff
@@ -15,58 +15,104 @@ namespace detail
...
+    {
+        if (ch == CharT('~') || ch == CharT('/'))
+        {
+            ++esz;
+        }
+    }
+    if (esz == s.size())
+    {
+        res = s;
+    }
```
gregmarr, line 40: Since you commented about this not being as clear, perhaps some use of algorithms would help

```
{
  using CharT = typename StringType::value_type;
  auto results = std::count_if(s.begin(), s.end(), [](CharT ch) { return (ch == CharT('~')) || (ch == CharT('/')); });

  if (results == 0) {
    return s;
  }

  StringType res;
  res.reserve(s.size() + results);

  // for loop here

  return res;
}
```

include/nlohmann/detail/string_escape.hpp
```diff
@@ -15,58 +15,104 @@ namespace detail
...
 {
-    replace_substring(s, StringType{"~1"}, StringType{"/"});
-    replace_substring(s, StringType{"~0"}, StringType{"~"});
+    using CharT = typename StringType::value_type;
+    auto j = s.begin();
+    while (j != s.end() && *j != CharT('~'))
+    {
+        ++j;
+    }
+    auto i = j;
```
gregmarr, line 79: Since you commented about this not being as clear, maybe something like this would help, with some explanatory comments:

```
  using CharT = typename StringType::value_type;
  auto start = s.find(CharT('~');

  if (start == StringType::npos) {
    return;
  }

  for (auto read = s.begin() + start, write = read; read != s.end(); read++, write++) {
    auto ch = *read;

    if (ch == CharT('~')) {
      auto next = read + 1;

      if (next != s.end()) {
        switch (*next) {
          case CharT('0'):
            ch = CharT('~');
            read++;
            break;
          case CharT('1'):
            ch = CharT('/');
            read++;
            break;
        }
      }
    }

    *write = ch;
  } 
```

include/nlohmann/detail/string_escape.hpp
```diff
@@ -15,58 +15,104 @@ namespace detail
...
+        res = s;
+    }
+    else
+    {
+        res.reserve(esz);
+        for (auto const ch : s)   // Yes, this is UTF8-safe
+        {
+            if (ch == CharT('~'))
+            {
+                res.append(StringType{"~0"});
```
gregmarr: This constructs and then destroys a temporary string, two calls to `push_back` would have better performance.

include/nlohmann/detail/string_escape.hpp
```diff
@@ -15,58 +15,104 @@ namespace detail
...
+                res.push_back(ch);
+            }
+        }
+    }
+    return res;
 }
 
 /*!
- * @brief string unescaping as described in RFC 6901 (Sect. 4)
+ * @brief In Place string unescaping as described in RFC 6901 (Sect. 4)
```
gregmarr: in-place

include/nlohmann/detail/string_escape.hpp
```diff
@@ -15,58 +15,104 @@ namespace detail
...
+        }
+        ++i;
+    }
+    s.erase(j, s.end());
+    s.shrink_to_fit();
 }
 
+// Left out, so far we don't use it, so it just lowers test coverage
+// /*!
+// * @brief Out Of Place string unescaping as described in RFC 6901 (Sect. 4)
```
gregmarr, line 107: returns a copy of a string unescaped as described...

nlohmann:
🔴 CI is red due to unrelated issues, see https://github.com/nlohmann/json/issues/4869.

include/nlohmann/detail/string_escape.hpp
puffetto: I did read somewhere that one of the goals was to remove the dependency on ```<algorithm>``` in order to reduce compile times.

include/nlohmann/detail/string_escape.hpp
puffetto: SHould be optimized out by any decent compiler, but I see the point, going to fix it.

include/nlohmann/detail/string_escape.hpp
gregmarr: I can't speak for @nlohmann here, but one person posted a PR to remove one of the 8 uses of `<algorithm>`.  That PR hasn't been accepted and is now marked as stale.  So it isn't necessarily a goal of the project as a whole.  My personal opinion is that precompiled headers and/or using the standard library through modules are the way to go for compiler performance going forward, rather than not using library algorithms.

nlohmann: I'm fine with using `<algorithm>` to have readable code. If we can improve performance, we can do it later, but I think removing `<algorithm>` was just a first shot at improving compilation speed, but not necessarily the most important thing.

nlohmann:

nlohmann:
Hi @puffetto, can you please rebase?

puffetto:
> Hi @puffetto, can you please rebase?

Done, sorry for taking so long, having personal problems these weeks/months.

puffetto:
> ## 🔴 Amalgamation check failed! 🔴
> The source code has not been amalgamated. @puffetto Please read and follow the [Contribution Guidelines](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md#files-to-change).

Sorry I made some noise, I was working on the wrong branch and commits we unintentionally propagated here; I did reset now.

nlohmann:
Thank you very much for this PR, @puffetto, and for pointing out the quadratic behaviour of `escape`/`unescape`. They were rewritten to rebuild the string in a single pass as part of #5443 (717 ms down to 20 ms on the pathological case), which is already on `develop`. I'm closing this PR as superseded. Sorry it took so long, and thanks again for the contribution!

PR #4873, opened 2025-07-29
Fixes #4854 Explicitly handle nullptr in sax_parse

### LINKED ISSUE #4854: `sax_parse` segfaults when given a null handler
### Description

Hey team,

I noticed that the `sax_parse` function family will crash with a segmentation fault if you pass `nullptr` as the SAX handler. 

The cause is a missing null check before the handler is used. You can trigger it easily:

```cpp
// This call crashes instead of throwing an exception
sax_parse(json_string, nullptr);
```

A simple guard clause at the top of the function(s) should fix it:

```cpp
// Add to all relevant sax_parse overloads
if (sax == nullptr) {
    throw json::exception("SAX handler cannot be null.");
}
```

This would make the API safer by throwing a catchable exception instead of crashing.

Thanks!

### Reproduction steps

sax_parse(json_string, nullptr);

expect Exceptions vs crash

### Minimal code example

```Shell

```

### Error messages

```Shell

```

GCC

### Library version

latest

### Validation

nikhilreddydev:
Handle sax_handler nullptr exception and throw an error.
This avoids crashing of program when nullptr is passed.

#4854 

### DIFF
--- a/docs/mkdocs/docs/home/exceptions.md
+++ b/docs/mkdocs/docs/home/exceptions.md
@@ -903,12 +903,10 @@ A JSON Patch operation 'test' failed. The unsuccessful operation is also printed
 
 ### json.exception.other_error.502
 
-when sax_parse operation has failed due to sax handler being nulllptr
+This exception occurs when a null pointer is passed as SAX handler in json.sax_parse() operation.
 
 !!! failure "Example message"
 
-    Executing `nlohmann::json.sax_parse(some_string, nullptrHandle)`:
-
     ```
-    [json.exception.other_error.502] Sax handler cannot be null
+    [json.exception.other_error.502] SAX handler must not be null
     ```
--- a/docs/mkdocs/docs/home/exceptions.md
+++ b/docs/mkdocs/docs/home/exceptions.md
@@ -900,3 +900,15 @@ A JSON Patch operation 'test' failed. The unsuccessful operation is also printed
     ```
     [json.exception.other_error.501] unsuccessful: {"op":"test","path":"/baz","value":"bar"}
     ```
+
+### json.exception.other_error.502
+
+when sax_parse operation has failed due to sax handler being nulllptr
+
+!!! failure "Example message"
+
+    Executing `nlohmann::json.sax_parse(some_string, nullptrHandle)`:
+
+    ```
+    [json.exception.other_error.502] Sax handler cannot be null
+    ```
--- a/docs/mkdocs/docs/home/exceptions.md
+++ b/docs/mkdocs/docs/home/exceptions.md
@@ -903,7 +903,7 @@ A JSON Patch operation 'test' failed. The unsuccessful operation is also printed
 
 ### json.exception.other_error.502
 
-This exception occurs when a null pointer is passed as SAX handler in json.sax_parse() operation.
+This exception occurs when a null pointer is passed as SAX handler in [sax_parse](../api/basic_json/sax_parse.md) function.
 
 !!! failure "Example message"
 
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -338,7 +338,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
           basic_json,
           default_object_comparator_t,
           AllocatorType<std::pair<const StringType,
-          basic_json >>>;
+          basic_json>>>;
 
     /// @brief a type for an array
     /// @sa https://json.nlohmann.me/api/basic_json/array_t/
@@ -1146,7 +1146,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
 
     template<typename JsonRef,
              detail::enable_if_t<detail::conjunction<detail::is_json_ref<JsonRef>,
-                                 std::is_same<typename JsonRef::value_type, basic_json >>::value, int> = 0 >
+                                 std::is_same<typename JsonRef::value_type, basic_json>>::value, int> = 0 >
     basic_json(const JsonRef& ref) : basic_json(ref.moved_or_copied()) {}
 
     /// @brief copy constructor
@@ -4252,7 +4252,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         data(size_type cnt, const basic_json& val)
             : m_type(value_t::array)
         {
-            m_value.array = create < array_t > (cnt, val);
+            m_value.array = create<array_t>(cnt, val);
         }
 
         data() noexcept = default;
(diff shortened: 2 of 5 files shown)


tests/src/unit-regression2.cpp
```diff
@@ -1075,6 +1153,13 @@ TEST_CASE("regression tests 2")
         CHECK(oj["test"].dump() == expected);
     }
 
+    SECTION("issue #4854 sax_parse segfaults when given a null handler")
```
gregmarr: I would think this should be in an existing SAX test file so that we don't need to duplicate the SAX class.

gregmarr (commented):
You have a whole bunch of unrelated formatting changes in the headers.  Have you checked that you're using the right version of the formatting tool?

tests/src/unit-regression2.cpp
nlohmann: Agree.

nlohmann:
Please also extend https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/home/exceptions.md#further-exceptions

nikhilreddydev:
> You have a whole bunch of unrelated formatting changes in the headers. Have you checked that you're using the right version of the formatting tool?

Thanks for the quick review.

I have installed astyle@3.4.13 as per requirements.txt

when I try to format like below from PR source branch
`astyle --options=tools/astyle/.astylerc tests/src/unit-regression2.cpp --verbose` or
`astyle --options=tools/astyle/.astylerc include/nlohmann/json.hpp --verbose` 

It gets below unchanged output

> Artistic Style 3.4.13                             07/30/2025
> Default option file  /Users/nikhilgudi/projects/cpp/json/tools/astyle/.astylerc
>  0 formatted   1 unchanged   0.01 seconds   1,277 lines

@gregmarr could you please help me out if I'm missing something.

nlohmann:
`make amalgamate` should do everything - including installing the required version of the tool.

nikhilreddydev:
> `make amalgamate` should do everything - including installing the required version of the tool.

Just tried it.

Great, I think it is formatting all headers and sources.

But I could not see any changes 😄

nlohmann:
Can you revert the changes to the header, add the line throwing the exception and try running "make amalgamate" again?

nikhilreddydev:
> Can you revert the changes to the header, add the line throwing the exception and try running "make amalgamate" again?

Thank you! This worked.
Next, I will work on extending doc & moving test to another file as reviewed.

include/nlohmann/json.hpp
```diff
@@ -4115,13 +4115,16 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     /// @brief generate SAX events
     /// @sa https://json.nlohmann.me/api/basic_json/sax_parse/
     template <typename InputType, typename SAX>
-    JSON_HEDLEY_NON_NULL(2)
```
gregmarr: Is it useful to keep the non-null annotations?  It is still incorrect to pass nullptr to this function, it just now results in an explicit exception.  It's still better to catch this error early.

nlohmann: I think so - though it has not helped here. It seems the Hedley project is dormant. Would be great to see something similar targeting new compilers.

nikhilreddydev: Hello,
if I understand it correctly, the solution here is to keep both
- Hedley non-null annotations
- explicit exception handling

This has -Werror -Wtautological-pointer-compare error 

```
error: comparison of nonnull parameter 'sax' equal to a null pointer is 'false' on first encounter [-Werror,-Wtautological-pointer-compare]
 4126 |         if (sax == nullptr)
      |             ^~~    ~~~~~~~
note: in instantiation of function template specialization 'nlohmann::basic_json<>::sax_parse<std::string &, SaxCountdown>' requested here
   58 |     nlohmann::json::sax_parse(s, p);
      |                     ^
note: declared 'nonnull' here
 4119 |     JSON_HEDLEY_NON_NULL(2)
```

If suppressing this warning is the next step, may I please get some help on where actually to define this?

gregmarr: Oh, that's ugly.

nikhilreddydev: Yes, I think it is not a good way to suppress that warning.

So, should we keep current change (removed Hedley annotations)?
Please let us know your thoughts @gregmarr  @nlohmann 
Thanks!

docs/mkdocs/docs/home/exceptions.md
```diff
@@ -900,3 +900,15 @@ A JSON Patch operation 'test' failed. The unsuccessful operation is also printed
     ```
     [json.exception.other_error.501] unsuccessful: {"op":"test","path":"/baz","value":"bar"}
     ```
+
+### json.exception.other_error.502
+
+when sax_parse operation has failed due to sax handler being nulllptr
```
nlohmann: Please make a complete sentence out of this. Typo: "nulllptr" -> "null pointer".

include/nlohmann/json.hpp
```diff
@@ -4115,13 +4115,16 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
     static bool sax_parse(InputType&& i, SAX* sax,
                           input_format_t format = input_format_t::json,
                           const bool strict = true,
                           const bool ignore_comments = false,
                           const bool ignore_trailing_commas = false)
     {
+        if (sax == nullptr)
+        {
+            JSON_THROW(other_error::create(502, "Sax handler cannot be null", nullptr));
```
nlohmann:
```suggestion
            JSON_THROW(other_error::create(502, "SAX handler must not be null", nullptr));
```

include/nlohmann/json.hpp
```diff
@@ -4131,13 +4134,16 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
-    JSON_HEDLEY_NON_NULL(3)
     static bool sax_parse(IteratorType first, IteratorType last, SAX* sax,
                           input_format_t format = input_format_t::json,
                           const bool strict = true,
                           const bool ignore_comments = false,
                           const bool ignore_trailing_commas = false)
     {
+        if (sax == nullptr)
+        {
+            JSON_THROW(other_error::create(502, "Sax handler cannot be null", nullptr));
```
nlohmann: See above.

include/nlohmann/json.hpp
```diff
@@ -4151,13 +4157,16 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
-    JSON_HEDLEY_NON_NULL(2)
     static bool sax_parse(detail::span_input_adapter&& i, SAX* sax,
                           input_format_t format = input_format_t::json,
                           const bool strict = true,
                           const bool ignore_comments = false,
                           const bool ignore_trailing_commas = false)
     {
+        if (sax == nullptr)
+        {
+            JSON_THROW(other_error::create(502, "Sax handler cannot be null", nullptr));
```
nlohmann: See above.

tests/src/unit-class_parser.cpp
```diff
@@ -1644,6 +1644,14 @@ TEST_CASE("parser class")
 
         SECTION("SAX parser")
         {
+            SECTION("null sax handler")
+            {
+
+                std::string s = "some_string";
+                SaxCountdown* p = nullptr;
+                nlohmann::json j;
+                CHECK_THROWS_AS(j.sax_parse(s, p), json::other_error);
```
nlohmann: Please use `CHECK_THROWS_WITH_AS` like above to check both the type of the exception and the error message.

docs/mkdocs/docs/home/exceptions.md
```diff
@@ -900,3 +900,15 @@ A JSON Patch operation 'test' failed. The unsuccessful operation is also printed
...
+
+!!! failure "Example message"
+
+    Executing `nlohmann::json.sax_parse(some_string, nullptrHandle)`:
```
nlohmann: Remove this line as the pseudo code is not helping here - the actual error is described above.

nlohmann (changes requested):

include/nlohmann/json.hpp
nlohmann: Please suppress the warning and keep the Hedley annotation. The attribute is a good way to actually annotate the API whereas the runtime check is invisible.

nlohmann:

tests/src/unit-regression2.cpp
nikhilreddydev: This is done https://github.com/nlohmann/json/pull/4873/commits/158c7a5bba3d0cb53c77bf0f330758adc833bc0e

docs/mkdocs/docs/home/exceptions.md
```diff
@@ -900,3 +900,13 @@ A JSON Patch operation 'test' failed. The unsuccessful operation is also printed
     ```
     [json.exception.other_error.501] unsuccessful: {"op":"test","path":"/baz","value":"bar"}
     ```
+
+### json.exception.other_error.502
+
+This exception occurs when a null pointer is passed as SAX handler in json.sax_parse() operation.
```
nlohmann:
```suggestion
This exception occurs when a null pointer is passed as SAX handler in [sax_parse](../api/basic_json/sax_parse.md) function.
```

include/nlohmann/json.hpp
nlohmann: This still needs to be changed.

include/nlohmann/json.hpp
nlohmann: This still needs to be changed.

tests/src/unit-class_parser.cpp
```diff
@@ -1644,6 +1644,14 @@ TEST_CASE("parser class")
 
         SECTION("SAX parser")
         {
+            SECTION("null sax handler")
+            {
+
```
nlohmann: Remove empty line.

nlohmann (changes requested):

tests/src/unit-class_parser.cpp
nikhilreddydev: Removed.

docs/mkdocs/docs/home/exceptions.md
nikhilreddydev: Removed.

tests/src/unit-class_parser.cpp
nikhilreddydev: done.

include/nlohmann/json.hpp
```diff
@@ -4131,13 +4134,16 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
-    JSON_HEDLEY_NON_NULL(3)
     static bool sax_parse(IteratorType first, IteratorType last, SAX* sax,
                           input_format_t format = input_format_t::json,
                           const bool strict = true,
                           const bool ignore_comments = false,
                           const bool ignore_trailing_commas = false)
     {
+        if (sax == nullptr)
+        {
+            JSON_THROW(other_error::create(502, "Sax handler must not be null", nullptr));
```
nlohmann: This string needs to be adjusted. Please also add a test that calls this function.

include/nlohmann/json.hpp
```diff
@@ -4151,13 +4157,16 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
-    JSON_HEDLEY_NON_NULL(2)
     static bool sax_parse(detail::span_input_adapter&& i, SAX* sax,
                           input_format_t format = input_format_t::json,
                           const bool strict = true,
                           const bool ignore_comments = false,
                           const bool ignore_trailing_commas = false)
     {
+        if (sax == nullptr)
+        {
+            JSON_THROW(other_error::create(502, "Sax handler must not be null", nullptr));
```
nlohmann: This string needs to be adjusted. Please also add a test that calls this function.

tests/src/unit-class_parser.cpp
```diff
@@ -1644,6 +1644,14 @@ TEST_CASE("parser class")
 
         SECTION("SAX parser")
         {
+            SECTION("null sax handler")
+            {
+                std::string s = "some_string";
```
nlohmann: Please check the output of the Clang-Tidy job - it suggests that `s` can be const.

nlohmann (changes requested):

include/nlohmann/json.hpp
nikhilreddydev: Fixed in https://github.com/nlohmann/json/pull/4873/commits/2f6155f69e1016407b5d0e4188a29cee7dab7c15

include/nlohmann/json.hpp
nikhilreddydev: Sorry, these string updates were missed by unsaved changes.
Fixed in https://github.com/nlohmann/json/pull/4873/commits/2f6155f69e1016407b5d0e4188a29cee7dab7c15

tests/src/unit-class_parser.cpp
nikhilreddydev: Sure, I can fix it. But I'm facing trouble setting up clang-tidy workflow on local setup

I'm using llvm & ninja. -DJSON_CI=ON
Trying to reproduce the same issues as in CI with below command
`cmake --build build --target ci_clang_tidy`

But the error it shows are very different from CI and there are a lot of warnings as errors, surprisingly if I search through them misc-const-correctness (actual error on CI) is not on local.

May I get some help on some info to setup this, If I'm missing something here.

nlohmann: The CI runs with Clang-Tidy 21. This could be the difference. So maybe just rely on the CI and fix the issues listed in the logs.

nikhilreddydev: Yes, maybe..
I have clang-tidy 20.1.8

Thanks!

tests/src/unit-class_parser.cpp
```diff
@@ -1644,6 +1644,14 @@ TEST_CASE("parser class")
 
         SECTION("SAX parser")
         {
+            SECTION("null sax handler")
+            {
+                const std::string s = "some_string";
+                SaxCountdown* p = nullptr;
+                const json j;
+                CHECK_THROWS_WITH_AS(j.sax_parse(s, p), "[json.exception.other_error.502] SAX handler must not be null", json::other_error&);
```
nlohmann: Still missing: tests for the other two overloads.

include/nlohmann/json.hpp
```diff
@@ -4131,13 +4134,16 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     /// @brief generate SAX events
     /// @sa https://json.nlohmann.me/api/basic_json/sax_parse/
     template<class IteratorType, class SAX>
-    JSON_HEDLEY_NON_NULL(3)
```
nlohmann: Please keep these annotations, but suppress Clang's `Wtautological-pointer-compare` warning via

```cpp
#if defined(__clang__)
    #pragma clang diagnostic push
    #pragma clang diagnostic ignored "-Wtautological-pointer-compare"
#endif
```

and

```cpp
#if defined(__clang__)
    #pragma clang diagnostic pop
#endif
```

nlohmann (changes requested):

include/nlohmann/json.hpp
```diff
@@ -4122,6 +4122,24 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
                           const bool ignore_trailing_commas = false)
     {
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wtautological-pointer-compare"
+#endif
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wnonnull-compare"
+#endif
```
nlohmann:
```suggestion
#if defined(__clang__)
  #pragma clang diagnostic push
  #pragma clang diagnostic ignored "-Wtautological-pointer-compare"
#elif defined(__GNUC__)
  #pragma GCC diagnostic push
  #pragma GCC diagnostic ignored "-Wnonnull-compare"
#endif
```

include/nlohmann/json.hpp
```diff
@@ -4138,6 +4156,24 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
```
nlohmann:
```suggestion
#if defined(__clang__)
  #pragma clang diagnostic push
  #pragma clang diagnostic ignored "-Wtautological-pointer-compare"
#elif defined(__GNUC__)
  #pragma GCC diagnostic push
  #pragma GCC diagnostic ignored "-Wnonnull-compare"
#endif
```

include/nlohmann/json.hpp
```diff
@@ -4158,6 +4194,24 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
```
nlohmann:
```suggestion
#if defined(__clang__)
  #pragma clang diagnostic push
  #pragma clang diagnostic ignored "-Wtautological-pointer-compare"
#elif defined(__GNUC__)
  #pragma GCC diagnostic push
  #pragma GCC diagnostic ignored "-Wnonnull-compare"
#endif
```

include/nlohmann/json.hpp
```diff
@@ -4122,6 +4122,24 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+        if (sax == nullptr)
+        {
+            JSON_THROW(other_error::create(502, "SAX handler must not be null", nullptr));
+        }
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic pop
+#endif
+#if defined (__clang__)
+#pragma clang diagnostic pop
+#endif
```
nlohmann:
```suggestion
#if defined(__clang__)
  #pragma clang diagnostic pop
#elif defined(__GNUC__)
  #pragma GCC diagnostic pop
#endif
```

include/nlohmann/json.hpp
```diff
@@ -4138,6 +4156,24 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
```
nlohmann:
```suggestion
#if defined(__clang__)
  #pragma clang diagnostic pop
#elif defined(__GNUC__)
  #pragma GCC diagnostic pop
#endif
```

include/nlohmann/json.hpp
```diff
@@ -4158,6 +4194,24 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
```
nlohmann:
```suggestion
#if defined(__clang__)
  #pragma clang diagnostic pop
#elif defined(__GNUC__)
  #pragma GCC diagnostic pop
#endif
```

nlohmann (changes requested):
Nearly good to go! Just some little simplifications.

nlohmann (approved):
Looks good to me!

nlohmann:
Thanks a lot! I will merge once the CI is green (don't worry about Cirrus CI).

nikhilreddydev:
> Thanks a lot! I will merge once the CI is green (don't worry about Cirrus CI).

Thanks for your patient reviews!
Looking forward to contribute more. 🙂

PR #4934, opened 2025-09-28
docs: Fix broken package manager links in the integration tab.

DmitriBogdanov:
## Issue

`docs/mkdocs/docs/integration/package_managers.md` renders package manager links & icons as raw code.

This page is linked in the README so it's the first thing users see when looking for package manager docs.

## Cause

Links are wrapped in a `<center>` tag which prevents them from being parsed as markdown.

## Fix

Use `md_in_html` plugin as specified by the Material theme [documentation](https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown/?h=markdown+ht+in#markdown-in-html).

HTML tags with internal md require explicit `markdown` attribute, so it shouldn't affect other pages.

## Additional notes

HTML tag `<center>` is considered [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/center) in favor of CSS like `<p style="text-align: center">`, so it was replaced.

## Before/after

before

after

### DIFF
--- a/docs/mkdocs/docs/integration/package_managers.md
+++ b/docs/mkdocs/docs/integration/package_managers.md
@@ -1,6 +1,6 @@
 # Package Managers
 
-<center>
+<p markdown style="text-align: center">
 ![Homebrew](../images/package_managers/homebrew.svg){: style="height:1em"}&nbsp;[**Homebrew**](#homebrew) `nlohmann-json` &emsp;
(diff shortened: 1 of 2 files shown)


nlohmann:
Thanks! Can you please check https://github.com/nlohmann/json/pull/4934/checks?check_run_id=51431348670 ? Then it should be good to go.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks a lot!

PR #4958, opened 2025-10-22
Enable string_view keys in operator[] by updating is_usable_as_basic_json_key_type (#3912)

### LINKED ISSUE #3912: inconsistency with string/string_view lookups
### Description

`operator string_view()` isn't used, whereas `operator string()` _is_.

### Reproduction steps

https://godbolt.org/z/MYEYG3Gao

compiles

### Minimal code example

```c++
#include <iostream>
#include <string>
#include "nlohmann/json.hpp"

using json = nlohmann::json;

struct Foo {
	operator std::string() const { return "foo"; }
};

struct Bar {
	operator std::string_view() const { return "bar"; }
};

int main() {

	json js {
		{ "foo", "string" },
		{ "bar", "string_view" },
	};

	std::cout << js[std::string{"foo"}] << "\n";
	std::cout << js[Foo{}] << "\n";

	std::cout << js[std::string_view{"bar"}] << "\n";
	std::cout << js[Bar{}] << "\n";
}
```

### Error messages

```Shell
<source>: In function 'int main()':
<source>:26:24: error: no match for 'operator[]' (operand types are 'json' {aka 'nlohmann::json_abi_v3_11_2::basic_json<>'} and 'Bar')
   26 |         std::cout << js[Bar{}] << "\n";
      |                        ^
```
```


linux gcc 12.2

### Library version

trunk

### Validation

### PR DESCRIPTION
**Pull request by khloodelhossiny**
This pull request adds std::string_view support for operator[] to improve usability with modern C++ (C++17 and later). The behavior now matches the existing std::string overloads. Additionally, the exception messages were updated for consistency when operator[] is used with non-object types and a string_view key.

This change addresses and fixes issue #3912.

- [x] An existing issue is referenced (Fixes #3912).
- [x] The Code coverage remains at 100%.. A test case for every new line of code.

### DIFF
--- a/include/nlohmann/detail/meta/type_traits.hpp
+++ b/include/nlohmann/detail/meta/type_traits.hpp
@@ -621,11 +621,14 @@ using is_usable_as_key_type = typename std::conditional <
 template<typename BasicJsonType, typename KeyTypeCVRef, bool RequireTransparentComparator = true,
          bool ExcludeObjectKeyType = RequireTransparentComparator, typename KeyType = uncvref_t<KeyTypeCVRef>>
 using is_usable_as_basic_json_key_type = typename std::conditional <
-    is_usable_as_key_type<typename BasicJsonType::object_comparator_t,
-    typename BasicJsonType::object_t::key_type, KeyTypeCVRef,
-    RequireTransparentComparator, ExcludeObjectKeyType>::value
-    && !is_json_iterator_of<BasicJsonType, KeyType>::value,
-    std::true_type,
+    (is_usable_as_key_type<typename BasicJsonType::object_comparator_t,
+     typename BasicJsonType::object_t::key_type, KeyTypeCVRef,
+     RequireTransparentComparator, ExcludeObjectKeyType>::value
+     && !is_json_iterator_of<BasicJsonType, KeyType>::value)
+#ifdef JSON_HAS_CPP_17
+    || std::is_convertible<KeyType, std::string_view>::value
+#endif
+    , std::true_type,
     std::false_type >::type;
 
 template<typename ObjectType, typename KeyType>
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -4164,11 +4164,14 @@ using is_usable_as_key_type = typename std::conditional <
 template<typename BasicJsonType, typename KeyTypeCVRef, bool RequireTransparentComparator = true,
          bool ExcludeObjectKeyType = RequireTransparentComparator, typename KeyType = uncvref_t<KeyTypeCVRef>>
 using is_usable_as_basic_json_key_type = typename std::conditional <
-    is_usable_as_key_type<typename BasicJsonType::object_comparator_t,
-    typename BasicJsonType::object_t::key_type, KeyTypeCVRef,
-    RequireTransparentComparator, ExcludeObjectKeyType>::value
-    && !is_json_iterator_of<BasicJsonType, KeyType>::value,
-    std::true_type,
+    (is_usable_as_key_type<typename BasicJsonType::object_comparator_t,
+     typename BasicJsonType::object_t::key_type, KeyTypeCVRef,
+     RequireTransparentComparator, ExcludeObjectKeyType>::value
+     && !is_json_iterator_of<BasicJsonType, KeyType>::value)
+#ifdef JSON_HAS_CPP_17
+    || std::is_convertible<KeyType, std::string_view>::value
+#endif
+    , std::true_type,
     std::false_type >::type;
 
 template<typename ObjectType, typename KeyType>
--- a/tests/src/unit-element_access2.cpp
+++ b/tests/src/unit-element_access2.cpp
@@ -1790,3 +1790,55 @@ TEST_CASE_TEMPLATE("element access 2 (additional value() tests)", Json, nlohmann
 #endif
     }
 }
+
+#ifdef JSON_HAS_CPP_17
+TEST_CASE("operator[] with user-defined std::string_view-convertible types")
+{
+    using json = nlohmann::json;
+
+    class TestClass
+    {
+        std::string key_data_ = "foo";
+
+      public:
+        operator std::string_view() const
+        {
+            return key_data_;
+        }
+    };
+
+    struct TestStruct
+    {
+        operator std::string_view() const
+        {
+            return "bar";
+        }
+    };
+
+    json j = {{"foo", "from_class"}, {"bar", "from_struct"}};
+    TestClass foo_obj;
+    TestStruct bar_obj;
+
+    SECTION("read access")
+    {
+        CHECK(j[foo_obj] == "from_class");
+        CHECK(j[TestClass{}] == "from_class");
+        CHECK(j[bar_obj] == "from_struct");
+        CHECK(j[TestStruct{}] == "from_struct");
+    }
+
+    SECTION("write access")
+    {
+        j[TestClass{}] = "updated_class";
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by gregmarr** [comment]
`string_view` support already exists through this function:

```
    /// @brief access specified object element
    /// @sa https://json.nlohmann.me/api/basic_json/operator%5B%5D/
    template<class KeyType, detail::enable_if_t<
                 detail::is_usable_as_basic_json_key_type<basic_json_t, KeyType>::value, int > = 0 >
    reference operator[](KeyType && key)
```
https://github.com/nlohmann/json/pull/3423

As you can see from the godbolt link in the original issue, the issue isn't that `string_view` doesn't work, it's that the compiler doesn't pick up that the `operator string_view()` from the type should allow it to work.

The thing that should really be updated is `is_usable_as_basic_json_key_type`, and then all of the existing functions will work.  I think it currently asks "can this type be used as a key" when this scenario needs it to ask "can this type be converted to a type that can be used as a key".

You can update the test code like this to see the missing piece.
```
	std::cout << js[static_cast<std::string_view>(Bar{})] << "\n";  // Okay
	std::cout << js[Bar{}] << "\n"; // Not okay
```

**Comment by khloodelhossiny** [comment]
@gregmarr 
Yes, I understand that `string_view` is supported through the existing template function, and the issue arises with js[Bar{}] due to ambiguity between the two overloads, causing the compiler to fail. 

My idea is to add specific `string_view` overloads for operator[] to guide the compiler in choosing the correct one, resolving the ambiguity.

**Comment by gregmarr** [comment]
>My idea is to add specific string_view overloads for operator[] to guide the compiler in choosing the correct one, resolving the ambiguity.

Okay, that's not what your PR description says, so it wasn't clear that you knew that.  

I'd be worried about the compiler potentially having a problem selecting between `operator[](KeyType && key)` and `operator[](std::string_view key)` and there also being unnecessary duplication.  I still think that making  `is_usable_as_basic_json_key_type` work with it is a better route here, but @nlohmann may be okay with the new functions.

**Comment by nlohmann** [comment]
I think a cleaner approach would be indeed to fix `is_usable_as_basic_json_key_type` to also work with string views.

**Comment by khloodelhossiny** [comment]
> > My idea is to add specific string_view overloads for operator[] to guide the compiler in choosing the correct one, resolving the ambiguity.
> 
> Okay, that's not what your PR description says, so it wasn't clear that you knew that.
> 
> I'd be worried about the compiler potentially having a problem selecting between `operator[](KeyType && key)` and `operator[](std::string_view key)` and there also being unnecessary duplication. I still think that making `is_usable_as_basic_json_key_type` work with it is a better route here, but @nlohmann may be okay with the new functions.

Thanks a lot for the explanation. I see your point, and I’ll update the code to follow that approach.

**Comment by khloodelhossiny** [comment]
> I think a cleaner approach would be indeed to fix `is_usable_as_basic_json_key_type` to also work with string views.

Thanks for the feedback! Yes, that makes sense — I will update the code.

**Review by nlohmann — CHANGES_REQUESTED** [review]
Please resolve the conflicts.

**Comment by gregmarr** [comment]
I was hoping for something that didn't have to name `string_view` directly, but I'm not sure that's possible.

Another possibility that seems to work and doesn't require any changes to the library is to make your type comparable to `std::string`.

```
struct Bar {
	operator std::string_view() const { return "bar"; }
};
bool operator<(const std::string &lhs, Bar const &rhs) {
    return lhs < static_cast<std::string_view>(rhs);
}
```

It could be a bit more generic, but that might be dangerous in the rest of the program:

```
template<typename T>
bool operator<(const std::string &lhs, T &&rhs) {
    return lhs < static_cast<std::string_view>(rhs);
}
template<typename T>
bool operator<(T&&lhs, const std::string &rhs) {
    return static_cast<std::string_view>(lhs) < rhs;
}
```

**Comment by khloodelhossiny** [comment]
> I was hoping for something that didn't have to name `string_view` directly, but I'm not sure that's possible.
> 
> Another possibility that seems to work and doesn't require any changes to the library is to make your type comparable to `std::string`.
> 
> ```
> struct Bar {
> 	operator std::string_view() const { return "bar"; }
> };
> bool operator<(const std::string &lhs, Bar const &rhs) {
>     return lhs < static_cast<std::string_view>(rhs);
> }
> ```
> 
> It could be a bit more generic, but that might be dangerous in the rest of the program:
> 
> ```
> template<typename T>
> bool operator<(const std::string &lhs, T &&rhs) {
>     return lhs < static_cast<std::string_view>(rhs);
> }
> template<typename T>
> bool operator<(T&&lhs, const std::string &rhs) {
>     return static_cast<std::string_view>(lhs) < rhs;
> }
> ```
I got your point — I understand that you want the library to fully support `std::string_view`.
For example, this code:
```
struct TestStruct {
    operator std::string_view() const { return "key"; }
};
std::cout << j.contains(TestStruct{}) << "\n";
```
I tried to handle that case, but I think for now it might make sense to let users define their own comparison operators if they have custom types.
_At least we’ve fixed the issue that many people usually miss._
I also agree that in the future, we should aim to support full `std::string_view `handling as a proper feature of the library.

khloodelhossiny:
@gregmarr 
Can you help me? I’m having some trouble with the CI checks.
Every time I push, I get these errors:

  - Ubuntu / ci_static_analysis_clang (ci_test_clang)
  - Ubuntu / ci_test_standards_clang (17, libstdcxx)
  - Windows / mingw (x64)

I tried to fix them several times, but after I push new changes, other errors appear.

I believe most of them are related to clang-format.
When I run the formatting command locally, it changes **macros** that should stay _on one line_ into _multiple lines_, which then causes conflicts.

My current steps are:

1.   Edit the two files
2.   Run the tests
3.   Run `make amalgamate`
4.   Push

If I’m missing any step, please let me know — I’m starting to get a bit disappointed :(

At this point, I almost believe that even if I push the develop branch, it will still fail the same checks :)

gregmarr:
@nlohmann will have to help you with the checks.

> I also agree that in the future, we should aim to support full std::string_view handling as a proper feature of the library.

IMO, the library supports `std::string_view` as a proper feature already.  The issue here is that the object isn't directly usable as a key but requires an intermediate conversion to a type that is usable as a key.

nlohmann:
- [ci_static_analysis_clang](https://github.com/nlohmann/json/actions/runs/18982378422/job/54218502063?pr=4958#logs) and [ci_test_standards_clang (17, libstdcxx)](https://github.com/nlohmann/json/actions/runs/18982378422/job/54218502161?pr=4958#logs) complain about a missing new line at the end of a file
- The MinGW job looks like a glitch - I restarted it.
- Please take a look at https://github.com/nlohmann/json/pull/4958/checks?check_run_id=54217855538

khloodelhossiny:
> IMO, the library supports `std::string_view` as a proper feature already. The issue here is that the object isn't directly usable as a key but requires an intermediate conversion to a type that is usable as a key.

Got it, thank you so much for the clarification and guidance!
What I actually meant was that it doesn’t fully support types that can be implicitly converted to std::string_view.
But yes, I’ll try to find the best solution for that.

khloodelhossiny:
@nlohmann Many thanks to you!

khloodelhossiny:
Hi @nlohmann , I've resolved the conflicts and updated the branch. All CI checks should be green now. Let me know if there's anything else needed! 
_Thanks for the feedback_

gregmarr:
Still relevant.  @nlohmann anything still needed here?

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks a lot!

khloodelhossiny:
> Thanks a lot!

_My pleasure!_ Happy this was useful.

PR #4959, opened 2025-10-22
Add null pointer check

lxyjy6:
The `write_characters` method directly calls `str.append(s, length)` without checking if `s` is null. When `s` is null and `length` is non-zero, `std::string::append()` attempts to read from the null pointer, causing the segmentation fault.

### DIFF
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -15802,6 +15802,9 @@ class output_string_adapter : public output_adapter_protocol<CharType>
     JSON_HEDLEY_NON_NULL(2)
     void write_characters(const CharType* s, std::size_t length) override
     {
+        if (s == nullptr) {
+        return;
+        }
         str.append(s, length);
     }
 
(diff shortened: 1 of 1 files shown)


nlohmann:
Can you provide a concrete example where the issue occurs?

lxyjy6:
During fuzz testing, I encountered a case where the first parameter of write_characters was nullptr, and discovered that it caused a crash. Specifically, the following lines of code were executed:
```
std::string str;
nlohmann::detail::output_string_adapter<char, std::string> adapter(str);
adapter.write_characters(nullptr, length);
```
I think it might be appropriate to throw an exception here instead.

nlohmann:
Thanks for your contribution. However, I think there is not an issue:

- `output_string_adapter::write_characters(const CharType* s, std::size_t length)` is part of the `detail` namespace and is not supposed to be called by user code. The function is annotated with `JSON_HEDLEY_NON_NULL(2)` which forbids a null pointer to be passed for `s`. Therefore, it is unsurprising to see an error if `nullptr` is passed anyway.
- In all calls inside the library to this function, `s` is the result of (1) a string literal, (2) `std::string::c_str()`, or (3) `std::vector::data()`. While the (1) and (2) can never be `nullptr`, (3) could be in case the vector is empty. For the vector case, however, we always check the size and never call `write_characters` with the result of `data()`.

Adding a runtime check to the function would unnecessarily slow down the serialization.

If you have an example that does not call the function directly, please provide it. Otherwise, feel free to close the PR.

lxyjy6:
Okay, thank you for your reply.

Issue #2018, opened 2020-03-30
How to create patch for merge_patch input ?


JeremieFiniel:
Hello,

I need to generate a patch like the one from merge_patch example :
```
// the patch
json patch = R"({
            "title": "Hello!",
            "phoneNumber": "+01-123-456-7890",
            "author": {
                "familyName": null
            },  
            "tags": [
                "example"
            ]   
        })"_json;
```

Is there an existing function to generate this ?
I tried diff function but it create a list of actions to apply, so not what I need.

I search somthing working like this:
``` c++
// the source document
json source = R"(
    {
        "baz": "qux",
        "foo": "bar"
    }
)"_json;

// the target document
json target = R"(
    {
        "baz": "boo",
        "hello": [
            "world"
        ]
    }
)"_json;

// create the patch
json patch = json::diff(source, target);
```
Where patch will be
```
{
    "baz": "boo",
    "foo": null,
    "hello": [
        "world"
    ]   
}
```

Thank you for your help.

nickaein:
I don't believe this is currently supported out of the box. The library supports *applying* [JSON Merge Patches](https://github.com/nlohmann/json#json-merge-patch) through [`merge_patch‍()`](https://nlohmann.github.io/json/classnlohmann_1_1basic__json_a7c43ed2a3004c1fa9543913f37b9fca9.html#a7c43ed2a3004c1fa9543913f37b9fca9) function, but there is no counterpart "merge_diff" function to generate one. There are some considerations before implementing a general-purpose one (e.g. how to handle `null` values in source).

nlohmann:
@nickaein is right.

dota17:
Is there a plan to support this case?

nlohmann:
There is no real specification for this. Merge requests are welcome!

PR #876, opened 2017-12-13
JSON Merge Patch (RFC 7396)

nlohmann:
SQLite's json1 extension (https://www.sqlite.org/json1.html) supports JSON Merge Patch (https://tools.ietf.org/html/rfc7396). As the implementation is trivial and we already support JSON Patch, I think this could be a nice extension to the library.

[Describe your pull request here. Please read the text below the line, and make sure you follow the checklist.]

* * *

- [x]  Changes are described in the pull request, or an [existing issue is referenced](https://github.com/nlohmann/json/issues).
- [x]  The test suite [compiles and runs](https://github.com/nlohmann/json/blob/develop/README.md#execute-unit-tests) without error.
- [x]  [Code coverage](https://coveralls.io/github/nlohmann/json) is 100%. Test cases can be added by editing the [test suite](https://github.com/nlohmann/json/tree/develop/test/src).

PR #4965, opened 2025-10-25
support JSON Merge Patch (RFC 7396) diff creation

satelliteprogrammer:
The JSON merge patch document format describes the set of modifications to a resource's content, that more closely mimics the syntax of the resource being modified.

However, and in contrast to JSON Patch (RFC 6902), a JSON Merge Patch cannot express certain modifications, e.g., changing an array element at a specific index, or setting a specific object value to null. The null value in a JSON Merge Patch is used to remove the key from the object.

The diff algorithm is not part of the RFC 7396, but it was tested against all examples provided, plus additional cases on how null values are handled.

JSON Merge Patch PR: https://github.com/nlohmann/json/pull/876
PR discussing the diff: https://github.com/nlohmann/json/issues/2018

If the content is approved, please let me know if/what documentation needs to be updated.

[Describe your pull request here. Please read the text below the line and make sure you follow the checklist.]

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -5254,6 +5254,57 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         }
     }
 
+    /// @brief creates a diff as a JSON Merge Patch
+    JSON_HEDLEY_WARN_UNUSED_RESULT
+    static basic_json merge_diff(const basic_json& source, const basic_json& target)
+    {
+        if (!target.is_object())
+        {
+            return target;
+        }
+
+        basic_json result(value_t::object);
+
+        if (source.is_object())
+        {
+            for (auto it = source.begin(); it != source.end(); ++it)
+            {
+                auto itf = target.find(it.key());
+                if (itf != target.end())
+                {
+                    if (it.value() != itf.value())
+                    {
+                        auto diff = merge_diff(it.value(), itf.value());
+                        if (diff.is_null())
+                        {
+                            JSON_THROW(other_error::create(503, detail::concat("cannot set \"", itf.key(), "\" to null"), &target));
+                        }
+                        result[it.key()] = merge_diff(it.value(), itf.value());
+                    }
+                }
+                else
+                {
+                    result[it.key()] = value_t::null;
+                }
+            }
+        }
+
+        for (auto it = target.begin(); it != target.end(); ++it)
+        {
+            auto itf = source.find(it.key());
+            if (itf == source.end())
+            {
+                if (it.value().is_null())
+                {
+                    JSON_THROW(other_error::create(503, detail::concat("cannot set \"", it.key(), "\" to null"), &target));
+                }
+                result[it.key()] = it.value();
+            }
+        }
+
+        return result;
+    }
+
     /// @}
 };
 
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -5272,14 +5272,22 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
                 auto itf = target.find(it.key());
                 if (itf != target.end())
                 {
-                    if (it.value() != itf.value())
+                    if (!it.value().is_null() && itf.value().is_null())
+                    {
+                        JSON_THROW(other_error::create(503, detail::concat("cannot set \"", it.key(), "\" to null"), &source));
+                    }
+
+                    if (it.value().is_object())
                     {
                         auto diff = merge_diff(it.value(), itf.value());
-                        if (diff.is_null())
+                        if (!diff.empty())
                         {
-                            JSON_THROW(other_error::create(503, detail::concat("cannot set \"", itf.key(), "\" to null"), &target));
+                            result[it.key()] = std::move(diff);
                         }
-                        result[it.key()] = merge_diff(it.value(), itf.value());
+                    }
+                    else if (it.value() != itf.value())
+                    {
+                        result[it.key()] = itf.value();
                     }
                 }
                 else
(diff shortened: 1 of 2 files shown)


nlohmann:
I am not sure if this feature would be widely used, so I'm opening a discussion.

satelliteprogrammer:
We had a (somewhat niche) need for this at work, that's why I've contributed here.

It's essentially a combination of 2 things:

1. we log all our data structures in JSON, this makes it easier to parse the logs and pretty-format large data structures;
2. we have a container that only notifies listeners on data changes.

For very large data structures, reading through a log line to find the one variable that did change is a pain. Not only that, but on some interfaces only a small amount of member variables on the entire structure are actually changing.
Given that we are already serializing in JSON, on those containers that track differences, we thought why not log only those differences using one of the available JSON patch methods. So here we are.

cschreib-ibex:
> [..] we thought why not log only those differences using one of the available JSON patch methods. So here we are.

We had this exact need as well: logging changes in simple JSON structures without all the noise of the formal JSON Patch syntax. Would have been convenient to have this capability built in.

include/nlohmann/json.hpp
```diff
@@ -5254,6 +5254,57 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+                auto itf = target.find(it.key());
+                if (itf != target.end())
+                {
+                    if (it.value() != itf.value())
+                    {
+                        auto diff = merge_diff(it.value(), itf.value());
+                        if (diff.is_null())
+                        {
+                            JSON_THROW(other_error::create(503, detail::concat("cannot set \"", itf.key(), "\" to null"), &target));
+                        }
```
cschreib-ibex: This may be overly constraining. An alternative would be to interpret "field set to `null`" as "field removed".

The RFC isn't very clear on this. While it says:
>This design means that merge patch documents are suitable for describing modifications to JSON documents that primarily use objects for their structure and do not make use of explicit null values.

It also lists examples in appendix where some fields are set to `null` in the source JSON. So they may have meant "do not make use of explicit null values" as "do not attribute specific meaning to explicit null values".

include/nlohmann/json.hpp
```diff
@@ -5254,6 +5254,57 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+                        result[it.key()] = merge_diff(it.value(), itf.value());
```
cschreib-ibex: Inefficient; `merge_diff` was already called on the same inputs above and `diff` could be reused here.

include/nlohmann/json.hpp
```diff
@@ -5254,6 +5254,57 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+        basic_json result(value_t::object);
+
+        if (source.is_object())
+        {
+            for (auto it = source.begin(); it != source.end(); ++it)
+            {
+                auto itf = target.find(it.key());
+                if (itf != target.end())
+                {
+                    if (it.value() != itf.value())
```
cschreib-ibex: Inefficient: this will traverse the whole depth of the JSON structure to check for equality of all sub fields, and we'll do that again in `merge_diff` if going inside the `if`. This check could be removed, calling `merge_diff` unconditionally, then checking the output isn't an empty object.

cschreib-ibex: Though, this will need checking for value types, and only call `merge_diff` if both source and target are objects. 
```c++
if (it.value().is_object() && itf.value().is_object()) 
{
    auto diff = merge_diff(it.value(), itf.value());
    if (!diff.empty()) 
    {
        result[it.key()] = std::move(diff);
    }
} 
else if (it.value() != itf.value()) 
{
    result[it.key()] = itf.value();
}
```

benpeck-eepics:
This is a feature that would be very useful to me. Our usecase is we would like to create a hierarchical data storage system that allows users to override values set in a base layer of json, storing user "overrides" as a merge patch. So being able to easily apply and generate merge patch documents would be essential to such a use case.

I find the merge patch formatting to be more human-readable than the json patch "action list" format. They are more intuitive to interpret for my usecase, where you are essentially looking at a sparse overlay document. You have the context of the changed element's path in the document providing self-documentation as to the intent of the change. A flat list of change actions lacks some of this context when it isn't structured like a typical document.

Thank you for your time, have a nice day.

include/nlohmann/json.hpp
satelliteprogrammer: Thanks! I had figured this one out already, but since the PR wasn't going anywhere I didn't fix it here.

include/nlohmann/json.hpp
satelliteprogrammer: > this will traverse the whole depth of the JSON structure to check for equality of all sub fields

I realised it, but since it would break on the first inequality, I conceded. What nudged me in this direction was that we were already checking if the target is not an object at the start, so it didn't feel right to check again before calling the recursive function.

I've now realised where I was wrong. 
For two non-object identical values, it they are present at the top-level of the JSON, then the patch must contain them, as otherwise it will not apply them.
However, if the identical non-objects are nested within an object, then, even though it would still work, they aren't necessary in the patch file, as the target object will retain its previous values.

I think this difference in behaviour is what requires the extra work within the for-loop. And then you're correct, we need to check the type to make sure we're applying the most efficient comparison.

include/nlohmann/json.hpp
satelliteprogrammer: I've changed the if to explicitly check if we have a transition from !null -> null. That one is impossible to generate, as it would collide with the actual meaning of null in the patch file.

> Null values in the merge patch are given special meaning to indicate the removal of existing values in the target.

However, the end result is the same. We would reach here if the source != target AND target == null. There's no other way for the diff to be null.

satelliteprogrammer:
@cschreib-ibex just in case it's useful for you. You can represent null values iff the source and target represent the same container. In that scenario you don't need null to represent removing existing values.

PR #4526, opened 2024-12-02
optional: use #if instead of #ifndef, fixes #3859

theodelrieu:
Hi, it's been a while.

Just connected to Github and saw the optional-related issue, tests pass with `-DJSON_ImplicitConversions=OFF`.

PR #4970, opened 2025-10-27
MSVC: make #4526 pass (use #if for optional) - fixes  #3859; re-amalgamate

### LINKED ISSUE #3859: .value() with optional default value fails to compile
### Description

We upgraded from 3.10.5 to 3.11.2

Since the Upgrade we can't compile `value()` with optional fallback value anymore

### Reproduction steps

Try to compile value with `std::optional<std::string>{}` as fallback. Is this still possible at all, was I using a non-supported edge-case before?

Compiles vs not

### Minimal code example

Beforehand we said:
```
std::optional<std::string> customer_id;
customer_id = json.value("customer_id", std::optional<std::string>{});
```

This no longer compiles. Even without the assignment and I tried different versions, all fail to compile:

```
json.value("customer_id", std::optional<std::string>{});
json.value<std::optional<std::string>>("customer_id", std::optional<std::string>{});
```
```


### Error messages

_No response_

Ubuntu 22.04, Clang 14

### Library version

3.11.2

### Validation

- (haven't tested this) [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/nlohmann/json/tree/develop) branch is used.
- [X] I can successfully [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests). **All passed**

chelseadzdr:
### Changes made
- pulled pr #4526 and amalgamated 
- merged into develop and fixed conflicts, kept most recent version of from_json.hpp (removed extra JSON_USE_IMPLICIT_CONVERSIONS that broke C++11 builds).
- Regenerated single-header files via the amalgamation target.

### Existing issue is referenced:
- https://github.com/nlohmann/json/issues/3859

### How I tested
- Windows: VS 2019, x64 toolset, CMake
  -  -DJSON_BuildTests=ON
  - Built and ran tests (Debug/Release).

### Amalgamation
- The source code is amalgamated by running `py -3 tools\amalgamate\amalgamate.py`.


gregmarr:
I don't understand what's happening with this PR.  It is not showing the change in the original file, but it is showing it in the amalgamated file, and there's no complaint about that from the checks.  Also strange, there are only 5 checks.

chelseadzdr:
@gregmarr i think most workflows are awaiting approval because it is my first contribution, and the changes in the source files has been made in previous commits, the latest commits just amalgamated the headers, that's why at first it may look like only the amalgamated files are modified

gregmarr:
There are whitespace changes in the amalgamated header, but no changes in any of the other headers, so the base commit was not properly amalgamated (unlikely) or there's a problem with your setup.  The change in `from_json` in the original PR is no longer needed because the `#ifdef` in question was removed as incorrect.

As for the test, it looks like some of the tests require checking `JSON_USE_IMPLICIT_CONVERSIONS` and some don't.

gregmarr:
I think you might need to start over with a single fresh commit that just fixes the test, due to the DCO checks, and the older commits that aren't signed off.

nlohmann:
Thank you very much for picking this up, @chelseadzdr. The `#if` change for #3859 was merged in #5183 and is already on `develop`, so I'm closing this PR as superseded. Thanks again for your help!

PR #4973, opened 2025-10-30
Add a type in the natvis template for detail::json_default_base

### LINKED ISSUE #4972: Natvis file for version 3.12.0 does not contain a type definition for detail::json_default_base
### Description

With version 3.12.0 the underlying type for a nlohmann::json is nlohmann::json_abi_diag_v3_12_0::detail::json_default_base and this type does not have a Type definition in the corresponding natvis template and generated natvis file.

### Reproduction steps

Add the natvis file to your project and try to introspect a nlohmann::json object

Introspection based on natvis file is not used.

### Minimal code example

```Shell

```

### Error messages

```Shell

```

Windows 11, Microsoft Visual Studio Professional 2022 (64-bit) - Current Version 17.14.19

### Library version

3.12.0

### Validation

sakuntalle:
With version 3.12.0 the underlying type for a `nlohmann::json` object has become `nlohmann::json_abi_diag_v3_12_0::detail::json_default_base` and this type does not have a Type definition in the corresponding natvis template and generated natvis file.

This PR updates the Natvis jinja template to add all the types derived from json_default_base. Also the template is rendered to produce the nlohmann_json.natvis

fixes [#4972](https://github.com/nlohmann/json/issues/4972)

### DIFF
--- a/tools/generate_natvis/nlohmann_json.natvis.j2
+++ b/tools/generate_natvis/nlohmann_json.natvis.j2
@@ -36,5 +36,25 @@
         </Expand>
     </Type>
 
+    <Type Name="{{ ns }}::detail::json_default_base">
+        <DisplayString Condition="m_data.m_type == {{ ns }}::detail::value_t::null">null</DisplayString>
+        <DisplayString Condition="m_data.m_type == {{ ns }}::detail::value_t::object">{*(m_data.m_value.object)}</DisplayString>
+        <DisplayString Condition="m_data.m_type == {{ ns }}::detail::value_t::array">{*(m_data.m_value.array)}</DisplayString>
+        <DisplayString Condition="m_data.m_type == {{ ns }}::detail::value_t::string">{*(m_data.m_value.string)}</DisplayString>
+        <DisplayString Condition="m_data.m_type == {{ ns }}::detail::value_t::boolean">{m_data.m_value.boolean}</DisplayString>
+        <DisplayString Condition="m_data.m_type == {{ ns }}::detail::value_t::number_integer">{m_data.m_value.number_integer}</DisplayString>
+        <DisplayString Condition="m_data.m_type == {{ ns }}::detail::value_t::number_unsigned">{m_data.m_value.number_unsigned}</DisplayString>
+        <DisplayString Condition="m_data.m_type == {{ ns }}::detail::value_t::number_float">{m_data.m_value.number_float}</DisplayString>
+        <DisplayString Condition="m_data.m_type == {{ ns }}::detail::value_t::discarded">discarded</DisplayString>
+        <Expand>
+            <ExpandedItem Condition="m_data.m_type == {{ ns }}::detail::value_t::object">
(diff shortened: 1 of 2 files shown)


tools/generate_natvis/nlohmann_json.natvis.j2
```diff
@@ -36,5 +36,25 @@
         </Expand>
     </Type>
 
+    <Type Name="{{ ns }}::detail::json_default_base">
+        <DisplayString Condition="m_data.m_type == {{ ns }}::detail::value_t::null">null</DisplayString>
```
gregmarr, line 40: `m_data` is not a member of `json_default_base`.  Why is this useful?

sakuntalle: Hey Greg!
I understand your concern and I'll have to admit I'm not versed into the library internal data structures.
But what I did notice with the VS version added in the linked issue can be seen in this screenshot
image

So VS sees the nlohmann:json object obj type as a json default_base object which most likely is just a polymorphic type. That's probably the reason why m_data can be found in the object.
And after I've added the natvis type I was suggesting in this PR VS did manage to show properly the content of it as in the screenshot below

image

gregmarr: That's interesting.  What is `json_t`?

sakuntalle: oh yeah, sorry, that's just a typedef to nlohmann::json we use in our codebase

sakuntalle: Hey guys,

I've dig deeper into this situation and I've learned the followings:
In Visual Studio 2022, when inspecting `nlohmann::json` variables, the debugger often shows them as the empty base type `nlohmann::detail::json_default_base`. This happens because VS displays objects according to their static type, not their dynamic type — even though the runtime object is a full `nlohmann::json`. 
Without a visualizer for json_default_base, the debugger cannot interpret the derived JSON contents and only shows raw memory or {...}.

Adding a Natvis for json_default_base ensures that the debugger displays JSON values based on their actual type (object, array, string, etc.) and allows convenient expansion. Although json_default_base is empty in C++, this visualizer is required to bridge the gap between VS’s static-type view and the actual runtime data. 

I've tried a few workarounds to make it take the dynamic type instead but I failed to produce a working natvis file.
Maybe somebody with more Natvis experience would be able to crack this, I dunno :)

gregmarr: That's definitely not their static type, because if that were their static type, you wouldn't be able to do anything with them.  The static type of `obj` is `nlohmann::json`.  I wonder if it's being confused by the inline namespace or the alias, and that's what it's looking up in natvis.  It is finding the base type, though.

gregmarr: This might help track down the issue: https://stackoverflow.com/questions/74713426/debugging-visual-studio-code-natvis-based-custom-view-objects

PR #4989, opened 2025-11-04
Add NLOHMANN_JSON_SERIALIZE_ENUM_STRICT macro that throws type_error.302 on unknown JSON value (Fixes #3992)

### LINKED ISSUE #3992: NLOHMANN_JSON_SERIALIZE_ENUM defaulting to the first enum value is surprising behavior

Ash-Jose:
### Summary
This PR fixes [#3992](https://github.com/nlohmann/json/issues/3992) by introducing a new macro  
`NLOHMANN_JSON_SERIALIZE_ENUM_STRICT`, which throws a `json.exception.type_error.302`  
when deserializing a JSON string that does not match any enumerator.

The existing `NLOHMANN_JSON_SERIALIZE_ENUM` macro remains unchanged to preserve backward compatibility.

---

### Changes
- Added new macro `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` in `macro_scope.hpp` that throws `type_error.302` for unmapped enum strings.  
- Added new test file `unit-serialize_enum_strict.cpp` validating the strict throwing behavior.  
- Regenerated single-header files using `make amalgamate`.  
- Verified all tests using `ctest --output-on-failure`.

---

### Expected test impact
All existing tests for `NLOHMANN_JSON_SERIALIZE_ENUM` continue to pass unchanged.  
New tests for `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` confirm that unknown JSON values now trigger exceptions.  

Other unrelated floating-point mismatches (e.g., `test-regression1_cpp11`) are not affected by this change.

---

### Documentation
A new documentation entry will be added at  
`docs/api/macros/nlohmann_json_serialize_enum_strict.md` describing the stricter deserialization behavior.

---

### Checklist
- [x] Implementation complete  
- [x] New strict macro added (`NLOHMANN_JSON_SERIALIZE_ENUM_STRICT`)  
- [x] New test added and passing  
- [x] Single-header regenerated (`make amalgamate`)  
- [x] Backward compatibility preserved  
- [x] DCO signed  
- [x] Documentation pending maintainer confirmation


nlohmann (changes requested):
Please do not change the existing macros. This would break existing applications. Instead, add a macro `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` with the new behavior.

Ash-Jose:
i have made changes accordingly pls review

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -253,6 +253,48 @@
...
+        auto it = std::find_if(std::begin(m), std::end(m),                                      \
+                               [&j](const std::pair<ENUM_TYPE, BasicJsonType>& ej_pair) -> bool \
+        {                                                                                       \
+            return ej_pair.second == j;                                                         \
+        });                                                                                     \
+        if (it == std::end(m))                                                                  \
+        {                                                                                       \
+            throw ::nlohmann::detail::type_error::create(                                       \
+                302,                                                                            \
+                std::string("invalid value for ") + #ENUM_TYPE + ": " + j.dump(),               \
```
nlohmann: Not tested, but please try

```suggestion
                concat("invalid value for " #ENUM_TYPE ": ", j.dump()),               \
```

nlohmann (changes requested):
- Please add a new line at the end of `unit-serialize_enum_strict.cpp`.
- Please run `make amalgamation` to properly indent the code.

include/nlohmann/detail/macro_scope.hpp
Ash-Jose: ok

Ash-Jose:
is this because of using concat
amalgamation check seems to have passed this time around

nlohmann:
> is this because of using concat amalgamation check seems to have passed this time around

You need to include `#include <nlohmann/detail/string_concat.hpp>`.

Ash-Jose:
included the header

Ash-Jose:
why are tests still failing??

nlohmann:
Maybe some issue with the includes. Please try:

- remove the `string_concat.hpp` include
- revert the code to using `std::string()`... like you did before

Sorry for the confusion.

Ash-Jose:
i have reverted it

Ash-Jose:
should i redo in a different branch??
i dont understand why checks are failing

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -253,6 +253,48 @@
...
+        /* NOLINTNEXTLINE(modernize-avoid-c-arrays) we don't want to depend on <array> */       \
+        static const std::pair<ENUM_TYPE, BasicJsonType> m[] = __VA_ARGS__;                     \
+        auto it = std::find_if(std::begin(m), std::end(m),                                      \
+                               [&j](const std::pair<ENUM_TYPE, BasicJsonType>& ej_pair) -> bool \
+        {                                                                                       \
+            return ej_pair.second == j;                                                         \
+        });                                                                                     \
+        if (it == std::end(m))                                                                  \
+        {                                                                                       \
+            throw ::nlohmann::detail::type_error::create(                                       \
```
nlohmann: This should need the exceptions.hpp header which again includes macro_scope.hpp. I think this is the core of the issue.

Ash-Jose: ill try including the exceptions header

Ash-Jose:
i dont understand why they still fail

nlohmann:
Because there is more to forward-declare. I'm sorry, but it seems to be a more complicated issue to use an exception in the macro header.

gregmarr:
The forward declarations aren't required for the macro definition, they're required where the things are used.  

~I think the issue is that it's using the detail namespace.  I don't think we do that for any of the other user space macros.~
Nope, simple type issue.

include/nlohmann/detail/macro_scope.hpp
gregmarr: Also, this needs to use `JSON_THROW`.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -253,6 +262,48 @@
...
+                    302,                                                                        \
+                    std::string("invalid value for ") + #ENUM_TYPE + ": " + j.dump(),           \
+                    j);                                                                         \
```
gregmarr: This is your problem, it takes a pointer, it should be `&j`.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -17,6 +17,15 @@
...
 #include <nlohmann/detail/abi_macros.hpp>
 
+// forward declaration to avoid circular include with exceptions.hpp
+namespace nlohmann
+{
+namespace detail
+{
+class type_error;
+} // namespace detail
+} // namespace nlohmann
```
gregmarr: This is not needed.

Ash-Jose:
i have made changes accordingly
this was the first version of my change 
although back then i wasnt able to make amalgamate work

Ash-Jose:
its still failing a few tests
is that concerning??

include/nlohmann/detail/macro_scope.hpp
gregmarr: This still needs to be changed, or this macro can't be used by people that disable exceptions.

test.json
```diff
@@ -0,0 +1,7 @@
+{
```
gregmarr: This file should not be in the root of the repo.  Is this just from local testing and got committed accidentally?

gregmarr:
> its still failing a few tests is that concerning??

Yes, one says you need to include `<string>` in `macro_scope.hxx`.

The fix for the other two is here:
https://github.com/nlohmann/json/blob/af524ab666359f09a16bca1a41543a9a3a941251/tests/src/unit-unicode3.cpp#L23-L25

The "code scanning" one seems to be outside of the scope of your PR, since it's failing at

```
  /home/runner/work/json/json/tests/src/unit-comparison.cpp:437:57: warning: ‘bool nlohmann::json_abi_ldvcmp_v3_12_0::basic_json<>::operator<=(const_reference) const is deprecated: Since 3.11.0; use undef JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON [-Wdeprecated-declarations]
    437 |                         CHECK((j_values[i] <= j_values[j]) == !(j_values[j] < j_values[i]));
        |                                                         ^
  /home/runner/work/json/json/tests/thirdparty/doctest/doctest.h:2428:24: note: in definition of macro ‘DOCTEST_ASSERT_IMPLEMENT_1’
   2428 |                     << __VA_ARGS__) DOCTEST_CLANG_SUPPRESS_WARNING_POP
        |                        ^~~~~~~~~~~
  /home/runner/work/json/json/tests/thirdparty/doctest/doctest.h:2970:20: note: in expansion of macro ‘DOCTEST_CHECK’
   2970 | #define CHECK(...) DOCTEST_CHECK(__VA_ARGS__)
        |                    ^~~~~~~~~~~~~
  /home/runner/work/json/json/tests/src/unit-comparison.cpp:437:25: note: in expansion of macro ‘CHECK’
    437 |                         CHECK((j_values[i] <= j_values[j]) == !(j_values[j] < j_values[i]));
        |                         ^~~~~
  In file included from /home/runner/work/json/json/tests/src/unit-comparison.cpp:19:
  /home/runner/work/json/json/include/nlohmann/json.hpp:3775:10: note: declared here
   3775 |     bool operator<=(const_reference rhs) const noexcept
        |          ^~~~~~~~
```

test.json
Ash-Jose: my bad ill take it out

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -11,6 +11,7 @@
 #include <utility> // declval, pair
 #include <nlohmann/detail/meta/detected.hpp>
 #include <nlohmann/thirdparty/hedley/hedley.hpp>
+#include <string>
```
gregmarr: @nlohmann Do you care about ordering of standard vs project includes?

Ash-Jose:
are these fails still related to my pr?

nlohmann:
Yes, the CI is green on the base branch.

include/nlohmann/detail/macro_scope.hpp
nlohmann: Yes, I think in the other headers we do follow this order.

tests/src/unit-serialize_enum_strict.cpp
```diff
@@ -0,0 +1,40 @@
+//     __ _____ _____ _____
+//  __|  |   __|     |   | |  JSON for Modern C++ (supporting code)
+// |  |  |__   |  |  | | | |  version 3.12.0
+// |_____|_____|_____|_|___|  https://github.com/nlohmann/json
+//
+// SPDX-FileCopyrightText: 2013-2025 Niels Lohmann <https://nlohmann.me>
+// SPDX-License-Identifier: MIT
+
+#include <nlohmann/json.hpp>
+#include "../doctest/doctest.h"
```
nlohmann: Please replace with

```cpp
```

and include the `json.hpp` after this.

See https://github.com/nlohmann/json/blob/develop/tests/src/unit-iterators2.cpp#L16C1-L19 for an example.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -253,6 +254,48 @@
...
+        /* NOLINTNEXTLINE(modernize-avoid-c-arrays) we don't want to depend on <array> */       \
+        static const std::pair<ENUM_TYPE, BasicJsonType> m[] = __VA_ARGS__;                     \
+        auto it = std::find_if(std::begin(m), std::end(m),                                      \
+                               [&j](const std::pair<ENUM_TYPE, BasicJsonType>& ej_pair) -> bool \
+        {                                                                                       \
+            return ej_pair.second == j;                                                         \
+        });                                                                                     \
+        if (it == std::end(m))                                                                  \
+        {                                                                                       \
+            throw ::nlohmann::detail::type_error::create(                                       \
```
nlohmann: Please replace `throw` with `JSON_THROW`.

nlohmann (changes requested):

tests/src/unit-serialize_enum_strict.cpp
Ash-Jose: ok thanks

include/nlohmann/detail/macro_scope.hpp
Ash-Jose: will do for sure

Ash-Jose:
ignore this recent push, that was by accident
anyways, when i try to use JSON_THROW
i get error: 'JSON_THROW' was not declared in this scope; did you mean 'JSON_TRY'? 
even though this exists
// allow disabling exceptions
    #define JSON_THROW(exception) throw exception
    #define JSON_TRY try
    #define JSON_INTERNAL_CATCH(exception) catch(exception)
#else
    #include <cstdlib>
    #define JSON_THROW(exception) std::abort()
    #define JSON_TRY if(true)
    #define JSON_CATCH(exception) if(false)
    #define JSON_INTERNAL_CATCH(exception) if(false)
#endif

gregmarr:
Ugh, `JSON_THROW` is undefined by `macro_unscope.hpp`.  @nlohmann Thoughts on letting that stay defined?  Otherwise I guess there would need to be a helper function to do the actual `JSON_THROW`.

include/nlohmann/detail/macro_scope.hpp
gregmarr: `ci_cpplint` enforces system before others, so move `<string>` before `<utility>` since they should be sorted too.

tests/src/unit-serialize_enum_strict.cpp
gregmarr: Also still need to do this:

https://github.com/nlohmann/json/blob/af524ab666359f09a16bca1a41543a9a3a941251/tests/src/unit-unicode3.cpp#L23-L25

Ash-Jose:
```
template <typename BasicJsonType>
inline void throw_enum_error(const BasicJsonType& j, const char* enum_type)
{
    JSON_THROW(::nlohmann::detail::type_error::create(
        302, std::string("invalid value for ") + enum_type + ": " + j.dump(), &j));
}
```
can i define a helper like this??

tests/src/unit-serialize_enum_strict.cpp
Ash-Jose: this exists in unit-unicode3.cpp

Ash-Jose:
i have made the previously suggested changes like the positioning of string header and all
but i am still waiting for reply on what to do about
error: 'JSON_THROW' was not declared in this scope; did you mean 'JSON_TRY'?
sorry i was inactive a couple of days due to college final submissions but i am free again now and would love to solve this issue

gregmarr:
> ```
> template <typename BasicJsonType>
> inline void throw_enum_error(const BasicJsonType& j, const char* enum_type)
> {
>     JSON_THROW(::nlohmann::detail::type_error::create(
>         302, std::string("invalid value for ") + enum_type + ": " + j.dump(), &j));
> }
> ```
> can i define a helper like this??

Seems reasonable to me, @nlohmann  any objections?

nlohmann:
Sure! Sorry for not answering earlier.

Ash-Jose:
are my changes sufficient this time around?
i will make the signoff change to avoid the DCO error
also this pr has become quite messy with all my unecessary commits
if you want i will submit another pr with only the necessary commits to keep things clean

gregmarr:
@nlohmann Looks like the workflows here are awaiting approval.  Not sure why since they ran before.

nlohmann:
@Ash-Jose Can you please update this branch to the latest `develop` branch? This should fix some of the CI issues.

nlohmann:
Thank you very much for this PR, @Ash-Jose. `NLOHMANN_JSON_SERIALIZE_ENUM_STRICT`, which throws on unknown values, was added in #5151 and is already on `develop` (and #3992 is closed). I'm closing this PR as superseded. Thanks again for tackling it!

PR #4323, opened 2024-03-27
Fix serialisation macros no member variables

Yurunsoft:
[Describe your pull request here. Please read the text below the line, and make sure you follow the checklist.]

Just like the code in the test, this code will report an error until it is fixed.

The problem is that serialization macros are not supported in classes without member variables.

```cpp
namespace emptys
{
class empty_intrusive
{
    NLOHMANN_DEFINE_TYPE_INTRUSIVE(empty_intrusive)
};

class empty_intrusive_with_default
{
    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(empty_intrusive_with_default)
};

class empty_only_serialize
{
    NLOHMANN_DEFINE_TYPE_INTRUSIVE_ONLY_SERIALIZE(empty_only_serialize)
};

class empty_non_intrusive
{
};

NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE(empty_non_intrusive)

class empty_intrusive_non_intrusive_only_serialize
{
};

NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_ONLY_SERIALIZE(empty_intrusive_non_intrusive_only_serialize)

class empty_non_intrusive_with_default
{
};

NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT(empty_non_intrusive_with_default)

} // namespace emptys
```

* * *

## Pull request checklist

Read the [Contribution Guidelines](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md) for detailed information.

- [x]  Changes are described in the pull request, or an [existing issue is referenced](https://github.com/nlohmann/json/issues).
- [x]  The test suite [compiles and runs](https://github.com/nlohmann/json/blob/develop/README.md#execute-unit-tests) without error.
- [x]  [Code coverage](https://coveralls.io/github/nlohmann/json) is 100%. Test cases can be added by editing the [test suite](https://github.com/nlohmann/json/tree/develop/test/src).
- [x]  The source code is amalgamated; that is, after making changes to the sources in the `include/nlohmann` directory, run `make amalgamate` to create the single-header files `single_include/nlohmann/json.hpp` and `single_include/nlohmann/json_fwd.hpp`. The whole process is described [here](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md#files-to-change).

## Please don't

[...]

Issue #4041, opened 2023-05-23
NLOHMANN_DEFINE_TYPE_* fails with zero members


dawinaj:
### Description

I found it annoying when working with multiple tiny polymorphic classes, trying to provide identical interface for each of them, that I can't use the macro for a class that has no members to (de)serialize.

### Reproduction steps

Use (probably) any of the `NLOHMANN_DEFINE_TYPE_*` macros with only first argument and no members.
F.e. `NLOHMANN_DEFINE_TYPE_INTRUSIVE(MyClass)`

### Expected vs. actual results

Expected:
```
friend void to_json(nlohmann::json&, const MyClass&) {}
friend void from_json(const nlohmann::json&, MyClass&) {}
```

Actual:
```
friend void to_json(nlohmann::json& nlohmann_json_j, const MyClass& nlohmann_json_t) {
	nlohmann_json_j[] = nlohmann_json_t.;
}
friend void from_json(const nlohmann::json& nlohmann_json_j, MyClass& nlohmann_json_t) {
	nlohmann_json_j.at().get_to(nlohmann_json_t.);
}
```

### Minimal code example

```Shell
class MyClass
{
	NLOHMANN_DEFINE_TYPE_INTRUSIVE(MyClass);
};
```

### Error messages

```Shell
Error	C2661	'nlohmann::json_abi_v3_11_2::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::json_abi_v3_11_2::adl_serializer,std::vector<uint8_t,std::allocator<uint8_t>>>::at': no overloaded function takes 0 arguments
Error	C2661	'nlohmann::json_abi_v3_11_2::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::json_abi_v3_11_2::adl_serializer,std::vector<uint8_t,std::allocator<uint8_t>>>::at': no overloaded function takes 0 arguments
Error	C2661	'nlohmann::json_abi_v3_11_2::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::json_abi_v3_11_2::adl_serializer,std::vector<uint8_t,std::allocator<uint8_t>>>::at': no overloaded function takes 0 arguments
[...]

PR #4994, opened 2025-11-13
Fix NLOHMANN_DEFINE_TYPE_* macros for zero-member types

chelseadzdr:
This pull request solves the issue #4041 using the changes made in the closed PR #4323 

When any of the NLOHMANN_DEFINE_TYPE_* macros are used with no data members, the preprocessor currently expands __VA_ARGS__ to an empty token. This leads to invalid code which breaks compilation for empty types.

This PR adjusts the macro_scope so that the zero-member case is handled explicitly. It finishes the work made in #4323 keeping the new changes like "Generate template functions with NLOHMANN_DEFINE_TYPE macros (#4597)" and to keep the documentations added.

Because develop has moved on since #4323, the changes were reapplied and adjusted manually rather than via a direct merge.

Tests has been included.


include/nlohmann/detail/macro_scope.hpp
```diff
@@ -407,142 +408,233 @@
 #define NLOHMANN_JSON_TO(v1) nlohmann_json_j[#v1] = nlohmann_json_t.v1;
 #define NLOHMANN_JSON_FROM(v1) nlohmann_json_j.at(#v1).get_to(nlohmann_json_t.v1);
 #define NLOHMANN_JSON_FROM_WITH_DEFAULT(v1) nlohmann_json_t.v1 = !nlohmann_json_j.is_null() ? nlohmann_json_j.value(#v1, nlohmann_json_default_obj.v1) : nlohmann_json_default_obj.v1;
-
+#define NLOHMANN_JSON_TO_DEFAULT nlohmann_json_j = nlohmann::json::parse("{}"); (void) nlohmann_json_t;
```
gregmarr: I don't think this should use the namespace or use the parse function, but should just use the `object()` function using the known local object: `nlohmann_json_t::object()`.

chelseadzdr: oh right, since in this context BasicJsonType is the JSON type should i change it to this?
`#define NLOHMANN_JSON_TO_DEFAULT nlohmann_json_j = BasicJsonType::object(); (void) nlohmann_json_t;`

gregmarr: Oh, if that's available, then yes, that would be better!

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -272,7 +272,7 @@
     AllocatorType, JSONSerializer, BinaryType, CustomBaseClass>
 
 // Macros to simplify conversion from/to types
-
+#define NLOHMANN_JSON_EXPAND_ARGS(...) __VA_ARGS__
```
gregmarr, line 274: It's not great to have all those `#ifdef _MSC_VER` variants just for this:

```
MSC:
   friend void to_json(BasicJsonType& nlohmann_json_j, const Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO,  NLOHMANN_JSON_EXPAND_ARGS(__VA_ARGS__))) } 
The rest:
    friend void to_json(BasicJsonType& nlohmann_json_j, const Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO,  ##__VA_ARGS__)) } \
```

Is there no syntax that works for all compilers?

gregmarr: Great, glad it just worked!

chelseadzdr: The CI is failing with `NLOHMANN_DEFINE_TYPE_NON_INTRUSIVE_WITH_DEFAULT`. I tried a more aggressive change to see if it would pass, but the same thing happened. It expands to `NLOHMANN_JSON_PASTE2 `when no arguments are provided.
I’m out of ideas on how to fix this.
Maybe having separate branches for some compilers could solve it?

gregmarr: Maybe it needs to be a different macro for empty classes.  Can it get away with just two, one for intrusive and one for non-intrusive?

```
#define NLOHMANN_DEFINE_EMPTY_TYPE_INTRUSIVE(Type)  \
    template<typename BasicJsonType, nlohmann::detail::enable_if_t<nlohmann::detail::is_basic_json<BasicJsonType>::value, int> = 0> \
    friend void to_json(BasicJsonType& nlohmann_json_j, const Type&) { 
      nlohmann_json_j = BasicJsonType.object();\
    } \
    template<typename BasicJsonType, nlohmann::detail::enable_if_t<nlohmann::detail::is_basic_json<BasicJsonType>::value, int> = 0> \
    friend void from_json(const BasicJsonType&, Type&) { }
#define NLOHMANN_DEFINE_EMPTY_TYPE_NON_INTRUSIVE(Type)  \
    template<typename BasicJsonType, nlohmann::detail::enable_if_t<nlohmann::detail::is_basic_json<BasicJsonType>::value, int> = 0> \
    void to_json(BasicJsonType& nlohmann_json_j, const Type&) { nlohmann_json_j = BasicJsonType::object(); } \
    template<typename BasicJsonType, nlohmann::detail::enable_if_t<nlohmann::detail::is_basic_json<BasicJsonType>::value, int> = 0> \
    void from_json(const BasicJsonType&, Type&) { }

```

chelseadzdr: That would simplify compatibility a lot. If that´s possible then it should be added to the Docs?

gregmarr: Yeah, it would require a new docs entry.  @nlohmann what do you think?

nlohmann: If there is no macro that fits with MSVC and all other compilers, then I guess a new one would be better than having none at all - at least less surprising than one that only works for some compilers. And I guess those few people that actually need a macro for this case will probably find it.

gregmarr:
@chelseadzdr There are still CI failures here.  Do you know what's happening?

PR #5016, opened 2025-11-27
Reference tuple from json

EALePain:
The current std::tuple from_json implementations fail to work with std::tie, aka tuples with reference types. The reason for this is that the template types are maintained all the way to the call to `get<Args>()` resulting in things like `get<int&>()`. The PR modifies this behavior to remove the references from all incoming types if those references are incompatible with the json type being used.

Ultimately this change allows for the result of std::tie to be used with get_to and allows the get method to create std::tuples of references to contained values.

The specific set of changes needed to achieve this are:

1. The addition of the `is_compatible_reference_type` metaprogramming struct that contains ::value = true if the BasicJsonType given has a get_ref instantiation that accepts CompatibleReferenceType (I used get_ptr because SFINAE was not working correctly with get_ref). This allows for overload resolution to decide to use get (and decay the type) or get_ref if a reference type is available for the individual elements of the tuple.
2. The addition of the meta/logic.hpp header adding cxpr_or, cxpr_and, and cxpr_not that allow for c++ 17 fold expressions for `||` and `&&` (with not for semantic consistency) in version older than 17. This is needed for the static assert (which I discuss later)
3. The addition of the `from_json_tuple_get_impl` overload set. In reverse priority order they:
   1. decay all types to values and calls `get<T>` on the json, returning by value
   3. if a type is a compatible reference type, calls get_ref and returns exactly the type given.
   4. if a type is an arethmetic type, call get and return by value. This one specifically exists to allow std::ties containing number_xxx_t to not fail when the type is some other number type.
4. a template using statement `tuple_type` that figures out the return type of a call to `from_json_tuple_get_impl` so that a tuple of exactly the correct types can be constructed from calls to it.
5. modifications to `from_json_tuple_impl_base` implementations adding a priority tag value template argument (to circumvent the `from_json_tuple_get_impl(..., priority_tag<2>)` implementation for calls to get references to members) and to return `tuple_type` and use `from_json_tuple_get_impl`.
6. add a static assert to `from_json_tuple_impl(..., identity_tag<std::tuple<Args...>>, ...)` to explicitly disallow reference types that are not compatible with the json type. This is done because this function _must_ return exactly the type in the identity_tag, but the function it calls will return values when dealing with incompatible references. This creates a dangling reference in the return. If desired, this static assert could be moved into an enable_if on the return to preserve SFINAE, but I could not think of a scenario where SFINAE would allow for recovery from this point during instantiation.
7. The addition of tests for `get_to(std::tie(...))` and `get<std::tuple<const T&,...>>()` and to confirm arithmetic types do not try to pull references even when they are number types.

I would also like to potentially discuss modifying `get_to` to allow for rvalue references in cases like std::tie and mutable range views such as std::span (if it has not already been discussed).

- [x] ~If applicable, an [existing issue](https://github.com/nlohmann/json/issues) is referenced.~

(edited to reflect current implementation as of Nov 29)

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -474,16 +474,16 @@ auto from_json_tuple_get_impl(BasicJsonType&& j, detail::identity_tag<Type> /*un
     return std::forward<BasicJsonType>(j).template get<detail::uncvref_t<Type>>();
 }
 
-template<std::size_t PTagV, typename BasicJsonType, template<typename...> class TupleT, typename... Types>
-using tuple_type = TupleT<decltype(from_json_tuple_get_impl(std::declval<BasicJsonType>(), detail::identity_tag<Types>{}, detail::priority_tag<PTagV>{}))...>;
+template<std::size_t PTagValue, typename BasicJsonType, typename... Types>
+using tuple_type = std::tuple<decltype(from_json_tuple_get_impl(std::declval<BasicJsonType>(), detail::identity_tag<Types>{}, detail::priority_tag<PTagValue>{}))...>;
 
-template<std::size_t PTagV, typename BasicJsonType, typename... Args, std::size_t... Idx>
-tuple_type<PTagV, BasicJsonType, std::tuple, Args...> from_json_tuple_impl_base(BasicJsonType&& j, index_sequence<Idx...> /*unused*/)
+template<std::size_t PTagValue, typename... Args, typename BasicJsonType, std::size_t... Idx>
+tuple_type<PTagValue, BasicJsonType, Args...> from_json_tuple_impl_base(BasicJsonType&& j, index_sequence<Idx...> /*unused*/)
 {
-    return tuple_type<PTagV, BasicJsonType, std::tuple, Args...>(from_json_tuple_get_impl(std::forward<BasicJsonType>(j).at(Idx), detail::identity_tag<Args>{}, detail::priority_tag<PTagV>{})...);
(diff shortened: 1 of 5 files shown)


EALePain:
ultimately I realized copy elation of the whole tuple is probably the more efficient route than elating moves for string, object, and array types allowed with the const reference returning overload.

nlohmann:
🥳 Green CI! I will try to review this in the next week.

EALePain:
Bumping this PR, appears to have gotten lost in the holiday noise.

nlohmann (commented):
The new tuple-returning path allows reference elements, but there's no guard against rvalue JSON arguments. This makes

```cpp
json::array({1,"x"}).get<std::tuple<const number_integer_t&, const string_t&>>()
```

compile and return dangling references once the temporary is destroyed.

EALePain:
In order to truly sand off that sharp edge you'd need to have C++23's explicit object parameter on the get methods etc. because by the time you're in the `from_json` functions you've lost the information needed to deal with the issue without it. Personally I'm of the opinion that its on the user to check for an issue like that, tuples of references already have the sharp edge that they _must_ live shorter lives than what they reference and asking a method to return a tuple of references would implicitly hold that same assumption regarding the tuple and the json.

But ultimately this change is mostly for compatibility with std::tie and the `std::tuple<const T&...>` implementation was a 'happy' accident I happened across with my first pass on an implementation. If you'd prefer not to allow that sharp edge I can remove it.

nlohmann (approved):
Looks good to me.

nlohmann:
@gregmarr Can you please also take a look?

gregmarr:
I don't see anything obvious that jumps out other than the potential that you referenced to have dangling references.

nlohmann:
Thanks a lot!

(Sorry, mis-clicked before and accidentally closed the PR...)

PR #5017, opened 2025-11-30
Fix json::update() with merge_objects==true error

DanielHwangbo:
### Description
This PR fixes a crash when using `update()` on an `ordered_json` object when `JSON_DIAGNOSTICS` is enabled.

### The Issue
When `update()` inserts new elements into an `ordered_json` object, the underlying `std::vector` may reallocate. This invalidates the `m_parent` pointers of existing children, causing `assert_invariant` to fail immediately because the children point to the old memory address.

### The Fix
I have updated the `update()` function to call `set_parents()` after insertion operations when diagnostics are enabled. This ensures that if the container moves in memory, all children are correctly updated to point to the new parent address.

### Reproduction Code
```cpp
#define JSON_DIAGNOSTICS 1
#include "nlohmann/json.hpp"
#include <iostream>

int main() {
    using json = nlohmann::ordered_json;
    json j1 = { {"numbers", {{"one", 1}}} };
    json j2 = { {"numbers", {{"two", 2}}}, {"string", "t"} };

    // This triggers reallocation and assertion failure without the fix
    j1.update(j2, true); 
    std::cout << j1.dump() << std::endl;
}

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -338,7 +338,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
           basic_json,
           default_object_comparator_t,
           AllocatorType<std::pair<const StringType,
-          basic_json>>>;
+          basic_json >>>;
 
     /// @brief a type for an array
     /// @sa https://json.nlohmann.me/api/basic_json/array_t/
@@ -544,7 +544,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         json_value(array_t&& value) : array(create<array_t>(std::move(value))) {}
 
         /// constructor for binary arrays
-        json_value(const typename binary_t::container_type& value) : binary(create<binary_t>(value)) {}
+        json_value(const typename binary_t::container_type & value) : binary(create<binary_t>(value)) {}
 
         /// constructor for rvalue binary arrays
         json_value(typename binary_t::container_type&& value) : binary(create<binary_t>(std::move(value))) {}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by nlohmann** [comment]
Maybe related: #4813

**Comment by DanielHwangbo** [comment]
Hi nlohmann, 

I have fixed the Amalgamation error and would like to request a code review again!

File: include/nlohmann/json.hpp [hunk 1]
```diff
@@ -3478,16 +3478,30 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+#if JSON_DIAGNOSTICS
+                    it2->second.set_parents();
+#endif
                     continue;
                 }
             }
-            m_data.m_value.object->operator[](it.key()) = it.value();
+
+            // Insert/Overwrite logic
+            auto& ref = m_data.m_value.object->operator[](it.key());
```
gregmarr, line 3491: What is the reason for this change?

gregmarr (commented):
You have a LOT of unrelated formatting changes in this file.  Did your editor do an auto-format?

include/nlohmann/json.hpp
```diff
@@ -3478,16 +3478,30 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
                 if (it2 != m_data.m_value.object->end())
                 {
                     it2->second.update(it.value(), true);
+
+                    // Repair the child's subtree.
```
gregmarr, line 3482: I think the comment should go inside the `#if`.

include/nlohmann/json.hpp
```diff
@@ -3478,16 +3478,30 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
                     continue;
                 }
             }
-            m_data.m_value.object->operator[](it.key()) = it.value();
+
+            // Insert/Overwrite logic
+            auto& ref = m_data.m_value.object->operator[](it.key());
+            ref = it.value();
+
+            // We must traverse the entire tree of 'this' and reset everyone's parent
```
gregmarr, line 3494: I think the comment should go inside the `#if`.

gregmarr (commented):
I recommend reverting all of the whitespace changes and the changes at line 3491 to introduce `ref`.  Then you also need to amalgamate and add the file to the commit.  I don't know if it didn't work before, or if you didn't add the file, but there are no changes to the amalgamated header.

gregmarr:
@DanielHwangbo Are you still pursuing this PR?

PR #5018, opened 2025-11-30
Feature: Add constructor support for C++20 Ranges and Views

DanielHwangbo:
### Description
This PR introduces a new constructor to `basic_json` that enables direct initialization from C++20 Ranges and Views (e.g., `std::views::filter`, `std::views::transform`).

### Motivation
Currently, attempting to initialize a JSON object from a C++20 view fails because the library does not natively recognize the view type as a container. Users are currently forced to manually pass iterators, which breaks the ergonomics of modern C++20 pipelines.

Feature: Add constructor support for C++20 Ranges and Views

### Description
This PR introduces a new constructor to `basic_json` that enables direct initialization from C++20 Ranges and Views (e.g., `std::views::filter`, `std::views::transform`).

### Motivation
Currently, attempting to initialize a JSON object from a C++20 view fails because the library does not natively recognize the view type as a container. Users are currently forced to manually pass iterators, which breaks the ergonomics of modern C++20 pipelines.

**Before:**
```
auto view = nums | std::views::filter(...);
// Fails compilation or requires manual iterators:
json j(view.begin(), view.end());
```
```
auto view = nums | std::views::filter(...);
// Direct initialization supported:
json j(view);
````

Implementation Details
I added a templated constructor that delegates to the existing (InputIT, InputIT) constructor.

SFINAE & Ambiguity Resolution: To prevent "ambiguous overload" errors (specifically with std::string and std::vector, which satisfy the Range concept but are already handled by the library), I added specific SFINAE guards. The new constructor is disabled via std::enable_if if:

The type is basic_json (preserves copy/move constructor priority).

The type is a json_ref.

The type is already a compatible_type (e.g., std::string), ensuring existing library behavior is preserved and no ambiguity errors occur during compilation.

Checklist
[x] The changes are described in detail, both the what and why.

[ ] If applicable, an [existing issue](https://github.com/nlohmann/json/issues) is referenced.

[x] The source code is amalgamated by running make amalgamate.

[x] Verified that std::string initialization still works (no ambiguity).

[ ] The [Code coverage](https://coveralls.io/github/nlohmann/json) remained at 100%. (A new test case may be required).

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -1033,6 +1033,51 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         assert_invariant();
     }
 
+    ////////////////////////////////////////////////////////////////////////////
+    // PATCH START: C++20 Range Constructor (Strict Version)
+    ////////////////////////////////////////////////////////////////////////////
+
+    /// @brief create a JSON value from a C++20 Range/View
+    /// @note This allows nlohmann::json j(myView) where myView has .begin()/.end()
+    /// @param[in] r  range to read from
+    template <typename Range,
+              typename std::enable_if<
+                  // 1. It must not be a basic_json object itself
+                  !std::is_same<typename std::decay<Range>::type, basic_json>::value &&
+                  
+                  // 2. It must not be a JSON reference
+                  !detail::is_json_ref<Range>::value &&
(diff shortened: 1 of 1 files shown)


nlohmann (changes requested):
Please amalgamate the source code and make sure the added code is only compiled for C++20.

PR #5049, opened 2026-01-09
Get text after JSON (parse with strict=false)

AbitTheGray:
Hi, I had a need for `parse` where `strict=false`. To be explicit, I need to have a string which starts with JSON but is followed by a text.

```
{ "my_key": "my_value" }That was a cool Json!
["a", "b", "c"]What a nice Array.
[{ "my_key": "my_value" }, { "my_key": "another_value"}]Can we also do an array of objects?
```

Currently the only way I found is to use exceptions to control flow, which I don't like.
```cpp
// https://godbolt.org/z/vveqod138
std::pair<json, std::size_t> parse_some(const std::string& input)
{
    try {
        // If the whole string is pure JSON, this succeeds.
        json j = json::parse(input);
        return {j, input.size()};
    } catch (const json::parse_error& e) {
        // e.byte is the 1-based index of the error location in the input.
        // In the "trailing text" case, that position is the first non-JSON char.
        std::size_t offset = (e.byte > 0) ? (e.byte - 1) : 0;

        // Try parsing up to the error position; if trailing text caused the error,
        // this substring is valid JSON.
        try {
            json j = json::parse(input.substr(0, offset));
            return {j, offset};
        } catch (...) {
            // Not a trailing-text situation; rethrow the original error.
            throw;
        }
    }
}
```
OK, there may be another one where I pre-parse it manually but I don't really see that as an option.
So I've tried implementing `parse_some` that returns not only the object but also the data after it (either as an offset, or span).

In the MR are my proposed changes, with no tests. **Please tell me if you like the idea** and way of doing it.

### DIFF
--- a/include/nlohmann/detail/input/input_adapters.hpp
+++ b/include/nlohmann/detail/input/input_adapters.hpp
@@ -14,6 +14,7 @@
 #include <iterator> // begin, end, iterator_traits, random_access_iterator_tag, distance, next
 #include <memory> // shared_ptr, make_shared, addressof
 #include <numeric> // accumulate
+#include <span> //span
 #include <streambuf> // streambuf
 #include <string> // string, char_traits
 #include <type_traits> // enable_if, is_base_of, is_pointer, is_integral, remove_pointer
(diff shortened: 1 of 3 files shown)


gregmarr:
You can do this with the `operator>>(std::istream& i, basic_json& j)` function.  That parses with `strict` set to false, so it accepts text after the closing `}`.
https://godbolt.org/z/1EjfbeP7c

nlohmann (changes requested):
As @gregmarr pointed out, this functionality should be possible with the existing API. No need to add another function for this.

PR #5050, opened 2026-01-12
Fix -Wtautological-constant-out-of-range-compare in serializer

cacharle:
Fixing this compilation warning:

```
In file included from C:\Users\charles\git\vcpkg\installed\x64-windows\include\nlohmann/json.hpp:60:
C:\Users\charles\git\vcpkg\installed\x64-windows\include\nlohmann/detail/output/serializer.hpp:922:26: warning: result of comparison of constant 400 with expression of type 'const std::uint8_t' (aka 'const unsigned char') is always true [-Wtautological-constant-out-of-range-compare]
  922 |         JSON_ASSERT(byte < utf8d.size());
      |                     ~~~~ ^ ~~~~~~~~~~~~
C:\Users\charles\git\vcpkg\installed\x64-windows\include\nlohmann/detail/macro_scope.hpp:205:35: note: expanded from macro 'JSON_ASSERT'
  205 |     #define JSON_ASSERT(x) assert(x)
      |                                   ^
C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0\ucrt\assert.h:40:17: note: expanded from macro 'assert'
   40 |             (!!(expression)) ||                                                               \
      |                 ^~~~~~~~~~
In file included from C:/Users/charles/git/
```

### DIFF
--- a/include/nlohmann/detail/output/serializer.hpp
+++ b/include/nlohmann/detail/output/serializer.hpp
@@ -919,7 +919,7 @@ class serializer
             }
         };
 
-        JSON_ASSERT(byte < utf8d.size());
+        JSON_ASSERT(static_cast<std::size_t>(byte) < utf8d.size());
         const std::uint8_t type = utf8d[byte];
 
         codep = (state != UTF8_ACCEPT)
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -19754,7 +19754,7 @@ class serializer
             }
         };
 
-        JSON_ASSERT(byte < utf8d.size());
+        JSON_ASSERT(static_cast<std::size_t>(byte) < utf8d.size());
         const std::uint8_t type = utf8d[byte];
 
         codep = (state != UTF8_ACCEPT)


nlohmann:
Interesting - which compiler did you use? I'm asking, because we compile the test suite with a recent Clang 21.1.5 and the following flags

```cmake
set(CLANG_CXXFLAGS
    -Werror
    -Weverything
    -Wno-c++98-compat
    -Wno-c++98-compat-pedantic
    -Wno-deprecated-declarations
    -Wno-extra-semi-stmt
    -Wno-padded
    -Wno-covered-switch-default
    -Wno-unsafe-buffer-usage
    -Wno-missing-noreturn
)
```

and this should have triggered the warning as well.

cacharle:
I'm using clang `19.1.7`

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -919,7 +919,7 @@ class serializer
             }
         };
 
-        JSON_ASSERT(byte < utf8d.size());
+        JSON_ASSERT(static_cast<size_t>(byte) < utf8d.size());
```
nlohmann:
```suggestion
        JSON_ASSERT(static_cast<std::size_t>(byte) < utf8d.size());
```

nlohmann (changes requested):
Please check the proposal and amalgamate the source code (run `make amalgamate`).

include/nlohmann/detail/output/serializer.hpp
gregmarr: This could probably even be a `static_assert`: 
```
   static_assert(std::numeric_limits<decltype(data)>::max() < utf8d.size(), "check that the table is large enough for the passed value");
 ```

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #5090, opened 2026-03-03
fix: treat single-element brace-init as copy/move instead of wrapping in array (#5074)

### LINKED ISSUE #985: Uniform initialization from another json object differs between gcc and clang.
Construction behaves differently with gcc 7.2 and clang 5.

Code to reproduce:
```
auto j1{nlohmann::json::parse("5")};
auto j2 = nlohmann::json::parse("5");
std::cout << j1.is_array() << '\n';
std::cout << j2.is_array() << '\n';
```

gcc prints:
1
0

clang prints:
0
0

Both are nlohmann::json instances however `j1` has an array wrapped around in gcc.

### LINKED ISSUE #1776: pass a json object to a class contructor adds an array around the object
When I pass a json object to a class contructor, an array is added around my object

``` c++
#include <iostream>
#include <nlohmann/json.hpp>


class A {
	public:
		A(const nlohmann::json &by_ref) : _by_ref{by_ref} {
			std::cout << _by_ref.dump(4) << "\n\n";
		}
	private:
		nlohmann::json _by_ref{};
};


class B {
	public:
		B(nlohmann::json by_val) : _by_val{std::move(by_val)} {
			std::cout << _by_val.dump(4) << "\n\n";
		}
	private:
		nlohmann::json _by_val{};
};


int main() {
	nlohmann::json json_object;
	json_object["key1"] = "val1";

	std::cout << json_object.dump(4) << "\n\n";

	A a{json_object};
	B b{json_object};

	return EXIT_SUCCESS;
}
```

Output
``` json
{
    "key1": "val1"
}

[
    {
        "key1": "val1"
    }
]

[
    {
        "key1": "val1"
    }
]
```

I would expect all three outputs like the first one.

I'm using gcc-8.3.0 on gentoo linux and nlohmann_json-3.7.0

### LINKED ISSUE #2026: Different JSON printouts for empty dictionary on Linux and Mac.
- What is the issue you have?

When I create JSON string with this statement `json j = {{}};` and print it, `[null]`is printed on Linux, while `[{}]` is printed on Mac. I think the Linux one is erroneous.

`main.cpp`:
```
#include "json.hpp"
#include <iostream>

using json = nlohmann::json;
using namespace std;

int main()
{
  json j = {{}};
  cout << j << endl;

  return 0;
}
```

Compiled with `g++ -std=c++11 main.cpp`.

The expected behavior is that both printouts should be `[{}]`.

I get `[null]` when I run the same code on Linux.

Linux operating system: Ubuntu 18.04.4 LTS  
Linux compiler: g++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0

Mac operating system: macOS Catalina Version 10.15.3  
Mac compiler: Apple clang version 11.0.0 (clang-1100.0.33.17)

I am using release version 3.7.3.

- If you experience a compilation error: can you [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests)?

No error while compilation.

### LINKED ISSUE #2046: json({}) gives null instead of empty object with GCC and -std=c++17
- What is the issue you have?

Readme says:

```
// ways to express the empty object {}
json empty_object_implicit = json({});
json empty_object_explicit = json::object();
```

But seems when using GCC with `-std=c++17` `json({})` actually gives null. Without `-std=c++17` or with `-std=c++14` or any other version it works. On clang it works with any version.

Seems there is some problem with overload resolution because `basic_json(std::nullptr_t = nullptr)` is called instead of `basic_json(initializer_list_t init, bool type_deduction = true...`. It works after I explicitly set second parameter to `true`.

``` c++
#include "json.hpp"
int main()
{
	nlohmann::json j = nlohmann::json({});
	printf("%s: %s\n", j.type_name(), j.dump().c_str());
	return 0;
}
```

```
g++ -std=c++17 main.cpp
```

```
object: {}
```

```
null: null
```

g++ (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0

version 3.7.0

### LINKED ISSUE #2311: Initialization different between gcc/clang

Initialization seems to differ between clang and gcc (have not yet tried MSVC):

```c++
json j{json::object()};

std::cout << j << std::endl; 
// GCC:   [{}]
// clang: {}
```

I.e., in gcc this creates an array with an object as its first element; in clang it creates an object. 

Running example: https://godbolt.org/z/7r76TE

As above.

Consistent initialization between compilers for the same code.

Different values per compiler.

- Compiler: clang 10, gcc 10.1
- Operating system: Linux
- std: c++17

- [ ] latest release version 3.7.3
- [x] other release - please state the version: 3.6.0, 3.8.0, whatever "trunk" is on godbolt
- [ ] the `develop` branch

### LINKED ISSUE #4488: Brace-Initialization Fails with json::parse and Key Access on Linux
### Description

On Linux, using brace-initialized literals with json::parse or accessing JSON keys with auto results in unexpected behavior or compile-time errors. This issue does not occur on other platforms and requires modifying code to use assignment (=) syntax instead of brace initialization. This is a non-issue for MacOS and iOS. I have not tested Windows.

### Reproduction steps

1.	Create a JSON string, e.g., std::string str = "{\"key\": \"value\"}";
2.	Attempt to parse it using auto parsed{json::parse(str)};
3.	Attempt to access a key using auto key{parsed["key"]};

The code should run on Linux like it does on MacOS and iOS, allowing the use of brace-initialized literals ({}) for json::parse and key access with JSON. It seems to be making an array only on Linux when using {} initialization. 

### Minimal code example

```Shell
#include <nlohmann/json.hpp>
#include <string>

using json = nlohmann::json;

int main() {
    std::string str = "{\"key\": \"value\"}";
    auto parsed{json::parse(str)};   // Fails on Linux
    auto key{parsed["key"]};         // Fails on Linux
    return 0;
}
```

### Error messages

```Shell
On Linux, the code throws exception when using brace-initialization {}, and errors arise specifically in these cases. Changing to assignment (=) syntax for both parsing and key access, like auto parsed = json::parse(str);, works as expected.
```

g++14 with C++20

### Library version

Latest release

### Validation

### LINKED ISSUE #4898: Different results on Linux vs Windows when using json["str"].push_back({json::object})
### Description

**For the following code, I get different results for when running the compiled code on Linux vs Windows.**
```C++
std::ifstream ifs(targetFile);
json original_jf = json::parse(ifs);
json jf_toParse;
jf_toParse["aaa"] = original_jf["aaa"];
json jf;
bool retainElement = false;
for (auto &[arrayElementKey, arrayElementValue] : jf_toParse["aaa"].items())
{
    retainElement = false;
    for (auto &[elementKey, elementValue] : arrayElementValue.items())
    {
        if (elementKey == "bbb" || elementKey == "ccc")
        {
            /// don't have to worry about this being a string. This is error checked prior to this.
            if (elementValue)
            {
                retainElement = true;
            }
        }
    }
    if (retainElement)
    {
        jf["aaa"].push_back({arrayElementValue});
    }
}

std::ofstream outfile (resultFile, std::ios::out | std::ios::binary); 
outfile << std::setw(4) << jf << std::endl;
```

**On Linux, I get the following for the** `resultFile` :
```json
{
    "aaa": [
        [
            {
                "aRandomStr": "randomStr1",
                "bbb": true,
                "randomStrTwo": "randomStr2"
            }
        ],
        [
            {
                "aRandomBool": true,
                "aRandomStr": "randomStr3",
                "bbb": true,
                "randomStrTwo": "randomStr4"
            }
        ],
        [
            {
                "aRandomBool": true,
                "aRandomStr": "randomStr5",
                "bbb": true,
                "randomStrTwo": "randomStr6"
            }
        ],
        [
            {
                "aRandomStr": "randomStr7",
                "ccc": true,
                "randomStrTwo": "randomStr8"
            }
        ],
        [
            {
                "aRandomStr": "randomStr8",
                "ccc": true,
                "randomStrTwo": "randomStr9"
            }
        ]
    ]
}
```

**On Windows, I get the following for the** `resultFile` :
```json
{
    "aaa": [
        {
            "aRandomStr": "randomStr1",
            "bbb": true,
            "randomStrTwo": "randomStr2"
        },
        {
            "aRandomBool": true,
            "aRandomStr": "randomStr3",
            "bbb": true,
            "randomStrTwo": "randomStr4"
        },
        {
            "aRandomBool": true,
            "aRandomStr": "randomStr5",
            "bbb": true,
            "randomStrTwo": "randomStr6"
        },
        {
            "aRandomStr": "randomStr7",
            "ccc": true,
            "randomStrTwo": "randomStr8"
        },
        {
            "aRandomStr": "randomStr8",
            "ccc": true,
            "randomStrTwo": "randomStr9"
        }
    ]
}
```
**I feel like I shouldn't be getting different results for this when the only difference is the operating system and compiler being used.**
**Linux/gcc11 vs Windows/msvc142**

### Reproduction steps

**I put this in the description. I tried to be as specific as possible.**
**For the** `targetFile` **please use a** `targetFile` **that has the following content:**
**Please define** `resultFile` as well.
```json
{
    "aaa": [
        {
            "aRandomStr": "randomStr1",
            "DDD": true,
            "randomStrTwo": "randomStr2"
        },
        {
            "aRandomStr": "randomStr1",
            "bbb": true,
            "randomStrTwo": "randomStr2"
        },
        {
            "aRandomBool": true,
            "aRandomStr": "randomStr3",
            "bbb": true,
            "randomStrTwo": "randomStr4"
        },
        {
            "aRandomBool": true,
            "aRandomStr": "randomStr5",
            "bbb": true,
            "randomStrTwo": "randomStr6"
        },
        {
            "aRandomStr": "randomStr7",
            "ccc": true,
            "randomStrTwo": "randomStr8"
        },
        {
            "aRandomStr": "randomStr8",
            "ccc": true,
            "randomStrTwo": "randomStr9"
        }
    ]
}
```

**I feel like the results I get for Windows/msvc142 are correct and the Linux/gcc11 results via** `resultFile` **are wrong/bugged.**

### Minimal code example

```C++
// You'll still have to define `targetFile` to some file that you filled with the contents I put in the Repoduction Steps.
// You'll have to define your `resultFile` as well.


std::ifstream ifs(targetFile);
json original_jf = json::parse(ifs);
json jf_toParse;
jf_toParse["aaa"] = original_jf["aaa"];
json jf;
bool retainElement = false;
for (auto &[arrayElementKey, arrayElementValue] : jf_toParse["aaa"].items())
{
    retainElement = false;
    for (auto &[elementKey, elementValue] : arrayElementValue.items())
    {
        if (elementKey == "bbb" || elementKey == "ccc")
        {
            /// don't have to worry about this being a string. This is error checked prior to this.
            if (elementValue)
            {
                retainElement = true;
            }
        }
    }
    if (retainElement)
    {
        jf["aaa"].push_back({arrayElementValue});
    }
}

std::ofstream outfile (resultFile, std::ios::out | std::ios::binary); 
outfile << std::setw(4) << jf << std::endl;
```

### Error messages

```Shell
# No error messages. Just different/bad results on Linux vs Windows
```

Various Linux (redhat and suse) systems with gcc11 VERSUS Windows 10/11 with msvc142

### Library version

Noticed bug/issue with 3.9.1. Upgraded to 3.12.0 and the same bug/issue exists. Latest develop branch commit (4106af8) also yielded the same results.

### Validation

### LINKED ISSUE #5074: Copy constructor changes semantics
### Description

As far as I know, this was a known issue on gcc, but clang was OK, but this is now in clang 20. Using {} to construct the object adds an array around an original disctionary. This is IMHO a critical bug and now that gcc and clang are displaying this behavior, it should be fixed.

### Reproduction steps

Just do

`void Object::method(Json json)`

call it with a dictionary, the new json object is now wrapped in an array.

The new object should not be wrapped in an array

### Minimal code example

```Shell
`void Object::method(Json json)`
```

### Error messages

```Shell

```

clang 20, RH8

### Library version

3.12

### Validation

ssam18:
## Summary
When a  value is passed using single-element brace initialization e.g., `json j{someObj}` or via a by-value parameter `void foo(json j)` called as `foo({someObj})`. C++ always prefers the `initializer_list` constructor over the copy/move constructor. This caused the value to be silently wrapped in a single-element array.

This bug was previously compiler-dependent: GCC wrapped the value while Clang did not. As of **Clang 20**, Clang now matches GCC behavior, making this a universal regression.

## Validation

All existing unit tests pass with the fix applied.
Fixes the issue #5074.


gregmarr:
See the first entry in the FAQ.  https://json.nlohmann.me/home/faq/#brace-initialization-yields-arrays
Any change in this behavior would be a regression for all other supported compilers.

nlohmann:
There is currently no way to fix this behavior without breaking existing code.

ssam18:
@nlohmann - Can you please take a look?

nlohmann:
I am not sure about this. Yet another macro adds another moving part to the library...

gregmarr:
It might also need to be encoded in the namespace so you don't accidentally link against a version with the wrong behavior.

docs/mkdocs/docs/home/faq.md
```diff
@@ -38,6 +38,26 @@ for objects.
...
+
+```cpp
+#define JSON_BRACE_INIT_COPY_SEMANTICS 1
+#include <nlohmann/json.hpp>
+
+json obj = {{"key", "value"}};
+json j{obj};   // -> {"key":"value"}  (copy, not array)
+```
+
+Without the macro (default behavior), `json j{obj}` creates `[{"key":"value"}]`. This opt-in macro fixes issue #5074 while preserving backwards compatibility for existing code.
```
gregmarr, line 59: Except that this is not true, the behavior without the macro is compiler dependent.

ssam18: done and tested locally. Please review this PR

docs/mkdocs/docs/home/faq.md
```diff
@@ -38,6 +38,30 @@ for objects.
...
 
+    To explicitly create a single-element array, use `json::array({value})`:
+
+    ```cpp
+    json j = json::array({true});  // [true]
+    ```
+
+**Opt-in copy semantics (since version 3.12.0)**
+
+If you define `JSON_BRACE_INIT_COPY_SEMANTICS` to `1` before including the library, single-element brace initialization is treated as copy/move instead of creating a single-element array:
```
gregmarr, line 49: "single-element brace initialization is treated as copy/move instead of creating a single-element array"

Perhaps something along these lines:
"single-element brace initialization is treated as copy/move instead of being compiler dependent whether it does that or creates a single-element array"

gregmarr:
> I am not sure about this. Yet another macro adds another moving part to the library...

Given how often this is reported as a bug, I think having a better answer than "We know, sorry, don't do that." is a good thing.

I think this is one of those things that should become the only method in 4.0 and be selectable until then.

docs/mkdocs/docs/home/faq.md
ssam18: Good point, updated the wording to make clear it eliminates the compiler-dependent behavior rather than just "not creating an array".

tests/src/unit-regression2.cpp
```diff
@@ -1203,4 +1188,41 @@ TEST_CASE_TEMPLATE("issue #4798 - nlohmann::json::to_msgpack() encode float NaN
...
+{
+    json const j_obj = {{"key", "value"}};
+
+    json const j3 = json::array({j_obj});
+    CHECK(j3.is_array());
+    CHECK(j3.size() == 1);
+    CHECK(j3[0] == j_obj);
+}
+
+#if JSON_BRACE_INIT_COPY_SEMANTICS
```
nlohmann: We don't ensure anywhere that this symbol is actually defined.

docs/mkdocs/docs/home/faq.md
```diff
@@ -38,6 +38,30 @@ for objects.
...
 
+    To explicitly create a single-element array, use `json::array({value})`:
+
+    ```cpp
+    json j = json::array({true});  // [true]
+    ```
+
+**Opt-in copy semantics (since version 3.12.0)**
+
+If you define `JSON_BRACE_INIT_COPY_SEMANTICS` to `1` before including the library, single-element brace initialization is treated as copy/move instead of being compiler-dependent whether it does that or creates a single-element array:
```
nlohmann: Please add a dedicated documentation page for this macro.

nlohmann (changes requested):

ssam18:
@nlohmann - I have addressed your review comment. Please review

ssam18:
@nlohmann - Can we merge this PR?

gregmarr (approved):

gregmarr:
@nlohmann Have all your concerns been addressed?

ssam18:
> @nlohmann Have all your concerns been addressed?

yes, that is true.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #5126, opened 2026-03-31
Add tests for special characters in wide strings

TheUserWW:
## Description
This PR adds a comprehensive test file unit-wstring.cpp for JSON library's wide string support. The file tests three wide string types ( std::wstring , std::u16string , and std::u32string ) JSON parsing capabilities, ensuring the library correctly handles wide character inputs with different encodings.

Validates normal wide string JSON parsing, including multilingual text


nlohmann:
I don't see the benefit of adding some more characters to the existing test - we already test some non-ASCII characters, and this PR is not really adding a "comprehensive" test.

Either we go for a complete UTF coverage or we can keep it as superficial as it is.

PR #5128, opened 2026-04-01
fix: compile error when deserializing into types using ordered_map

ssam18:
When deserializing into a struct that contains an `ordered_map` field (e.g. via `NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT`), GCC fails to compile because `std::transform` + `std::inserter` triggers the element-shifting path in `<algorithm>`, which requires `pair::operator=` deleted when the key is `const`.

Replaced the transform with a simple range-for loop and `emplace`, which constructs in-place and never tries to assign.

### DIFF
--- a/docs/mkdocs/docs/home/faq.md
+++ b/docs/mkdocs/docs/home/faq.md
@@ -20,7 +20,16 @@
 
     yield different results (`#!json [true]` vs. `#!json true`)?
 
-This is a known issue, and -- even worse -- the behavior differs between GCC and Clang. The "culprit" for this is the library's constructor overloads for initializer lists to allow syntax like
+Starting from this version, single-value brace initialization is treated as copy/move instead of wrapping in a single-element array:
+
+```cpp
+json j_orig = {{"key", "value"}};
+
+json j{j_orig};
+j_orig.method();
+```
+
+The root cause was the library's constructor overloads for initializer lists, which allow syntax like
 
 ```cpp
 json array = {1, 2, 3, 4};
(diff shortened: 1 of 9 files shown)


docs/mkdocs/docs/home/faq.md
```diff
@@ -20,7 +20,16 @@
 
     yield different results (`#!json [true]` vs. `#!json true`)?
 
-This is a known issue, and -- even worse -- the behavior differs between GCC and Clang. The "culprit" for this is the library's constructor overloads for initializer lists to allow syntax like
+Starting from this version, single-value brace initialization is treated as copy/move instead of wrapping in a single-element array:
```
nlohmann, line 23: This is a breaking change we cannot accept. Please revert.

nlohmann (changes requested):

include/nlohmann/json.hpp
```diff
@@ -955,6 +955,18 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         }
         else
         {
+            // if there is exactly one element and type deduction is enabled,
```
gregmarr, line 958: This is also that breaking change that we can't accept.

single_include/nlohmann/json.hpp
```diff
@@ -3495,75 +3495,75 @@ NLOHMANN_JSON_NAMESPACE_END
 // |  |  |__   |  |  | | | |  version 3.12.0
 // |_____|_____|_____|_|___|  https://github.com/nlohmann/json
 //
-// SPDX-FileCopyrightText: 2013-2026 Niels Lohmann <https://nlohmann.me>
+// SPDX-FileCopyrightText: 2013-2025 Niels Lohmann <https://nlohmann.me>
 // SPDX-License-Identifier: MIT
 
 #ifndef INCLUDE_NLOHMANN_JSON_FWD_HPP_
-    #define INCLUDE_NLOHMANN_JSON_FWD_HPP_
+#define INCLUDE_NLOHMANN_JSON_FWD_HPP_
```
gregmarr, line 3502: There is something wrong with your amalgamation setup.

gregmarr (commented):
You need to drop commit `2071ba67d9923f4b5349142dc3d12887572ceb61` from this.

docs/mkdocs/docs/home/faq.md
ssam18: I will take a look in a day or two

PR #5151, opened 2026-04-27
Serialize enum

### LINKED ISSUE #3992: NLOHMANN_JSON_SERIALIZE_ENUM defaulting to the first enum value is surprising behavior

nugentcaillin:
[Describe your pull request here. Please read the text below the line and make sure you follow the checklist.]

- added NLOHMANN_SERIALIZE_ENUM_STRICT to throw error on unrecognized enum values, solving [3992](https://github.com/nlohmann/json/issues/3992). 
- Added new unit tests to verify behavior. 
- updated documentation where NLOHMANN_SERIALIZE_ENUM to mention alongside warning that a throwing version exists - i didn't add a separate page for it as it is very close functionally to NLOHMANN_SERIALIZE_ENUM, but can add.

I wasn't able to call JSON_THROW inside detail/macro_scope.hpp, so this PR may have issues when exceptions are disabled - but am unsure if this is an issue if this macro has been added with the purpose of throwing - I'd appreciate some guidance on how to proceed here.


nlohmann (changes requested):
Can you please add a dedicated page for the enum like https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum.md ?

nugentcaillin:
Thanks for taking the time to review! I've added the page and links where seemed appropriate.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -253,6 +253,45 @@
...
+        static_assert(std::is_enum<ENUM_TYPE>::value, #ENUM_TYPE " must be an enum!");          \
+        /* NOLINTNEXTLINE(modernize-avoid-c-arrays) we don't want to depend on <array> */       \
+        static const std::pair<ENUM_TYPE, BasicJsonType> m[] = __VA_ARGS__;                     \
+        auto it = std::find_if(std::begin(m), std::end(m),                                      \
+                               [e](const std::pair<ENUM_TYPE, BasicJsonType>& ej_pair) -> bool  \
+        {                                                                                       \
+            return ej_pair.first == e;                                                          \
+        });                                                                                     \
+        if (it != std::end(m)) j = it->second;                                                  \
+        else throw nlohmann::detail::out_of_range::create(403, "enum value out of range", nullptr);\
```
nlohmann: Please use the `JSON_THROW` macro.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -253,6 +253,45 @@
...
+        static_assert(std::is_enum<ENUM_TYPE>::value, #ENUM_TYPE " must be an enum!");          \
+        /* NOLINTNEXTLINE(modernize-avoid-c-arrays) we don't want to depend on <array> */       \
+        static const std::pair<ENUM_TYPE, BasicJsonType> m[] = __VA_ARGS__;                     \
+        auto it = std::find_if(std::begin(m), std::end(m),                                      \
+                               [&j](const std::pair<ENUM_TYPE, BasicJsonType>& ej_pair) -> bool \
+        {                                                                                       \
+            return ej_pair.second == j;                                                         \
+        });                                                                                     \
+        if (it != std::end(m)) e = it->first;                                                   \
+        else throw nlohmann::detail::out_of_range::create(403, "enum value out of range", nullptr);\
```
nlohmann: Please use the `JSON_THROW` macro.

include/nlohmann/detail/macro_scope.hpp
nlohmann: Please use a new exception code (410 should be free) as 403 has a different semantics. Add it to `docs/mkdocs/docs/home/exceptions.md`.

include/nlohmann/detail/macro_scope.hpp
nlohmann: Please use a new exception code (410 should be free) as 403 has a different semantics. Add it to `docs/mkdocs/docs/home/exceptions.md`.

tests/src/unit-conversions.cpp
```diff
@@ -1657,6 +1657,77 @@ TEST_CASE("JSON to enum mapping")
     }
 }
 
+enum class strict_cards {kreuz, pik, herz, karo};
```
nlohmann: There is no test for the `to_json` part; that is, a situation where an enum value is not mentioned in the mapping, but later used.

tests/src/unit-conversions.cpp
```diff
@@ -1657,6 +1657,77 @@ TEST_CASE("JSON to enum mapping")
...
+
+// NOLINTNEXTLINE(misc-use-internal-linkage,misc-const-correctness,cppcoreguidelines-avoid-c-arrays,hicpp-avoid-c-arrays,modernize-avoid-c-arrays) - false positive
+NLOHMANN_JSON_SERIALIZE_ENUM_STRICT(strict_cards,
+{
+    {strict_cards::kreuz, "kreuz"},
+    {strict_cards::pik, "pik"},
+    {strict_cards::pik, "puk"},  // second entry for cards::puk; will not be used
```
nlohmann:
```suggestion
    {strict_cards::pik, "puk"},  // second entry for cards::pik; will not be used
```

docs/mkdocs/docs/features/enum_conversion.md
```diff
@@ -55,7 +55,8 @@ Just as in [Arbitrary Type Conversions](arbitrary_types.md) above,
 Other Important points:
 
 - When using `get<ENUM_TYPE>()`, undefined JSON values will default to the first pair specified in your map. Select this
-  default pair carefully.
+  default pair carefully. If you desire an exception in this circumstance use [`NLOHMANN_JSON_SERIALIZE_ENUM_STRICT()`](../api/macros/nlohmann_json_serialize_enum_strict.md)
+  which behaves identically except for throwing an exception on unrecognised values.
```
nlohmann:
```suggestion
  which behaves identically except for throwing an exception on unrecognized values.
```

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -253,6 +253,45 @@
...
+    }
+
```
nlohmann, line 310:
```suggestion
```

nlohmann (changes requested):

nugentcaillin:
Thanks for the feedback, I've applied those changes - however for JSON_THROW I couldn't call it directly and had to add a templated function to wrap JSON_THROW.

I would get a compiler error that would be fixed by #defining JSON_THROW in the test file - I'm guessing something to do with the serialize macro creating a template along with JSON_THROW being undef'd and calling JSON_THROW through a wrapper function templated on exception changes the timing on when the JSON_THROW macro is expanded? 

I don't know for sure though but this was the only way I could get the serialization macro to compile without #defining JSON_THROW again wherever it needs to be called. If there's a better way to do this i'd appreciate some guidance.

I also had to void cast exception in this templated function to avoid error if compiled with Werror and Wunused-parameter as JSON_THROW doesn't use exception if exceptions are turned off - not sure if this goes against code style.

maybe a void cast could be added to the JSON_THROW macro when exceptions are turned off to suppress this? - changing JSON_THROW felt beyond the scope of this PR.

docs/mkdocs/docs/home/exceptions.md
```diff
@@ -868,6 +868,16 @@ Key identifiers to be serialized to BSON cannot contain code point U+0000, since
     BSON key cannot contain code point U+0000 (at byte 2)
     ```
 
+### json.exception.out_of_range.410
+
+Undefined json fields cannot be used with JSON_SERIALIZE_ENUM_STRICT
```
nlohmann: Please write the macro in backticks and link the respective documentation page.

docs/mkdocs/docs/api/macros/nlohmann_json_serialize_enum_strict.md
```diff
@@ -0,0 +1,86 @@
...
+```cpp
+#define NLOHMANN_JSON_SERIALIZE_ENUM_STRICT(type, conversion...)
+```
+
+By default, enum values are serialized to JSON as integers. In some cases, this could result in undesired behavior. If
+an enum is modified or re-ordered after data has been serialized to JSON, the later deserialized JSON data may be
+undefined or a different enum value than was originally intended.
+
+`NLOHMANN_JSON_SERIALIZE_ENUM_STRICT` allows to define a user-defined serialization for every enumerator that
+throws an exception on undefined input.
```
nlohmann, line 12: Would it make sense to show this in an example?

nlohmann (commented):
Some small comments, otherwise looks good.

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -253,6 +253,61 @@
...
 
+
+
+/*!
+@brief function to wrap JSON_THROW_MACRO - NLOHMANN_SERIALIZE_ENUM_STRICT has a
+       compilation warning about there being no arguments to JSON_THROW that depend on
+       template arguments otherwise
+*/
+template<typename ExceptionType>
+void templated_json_throw(ExceptionType exception)
```
gregmarr, line 264: FYI, two alternative versions of this function from previous "strict enum" PRs:

https://github.com/nlohmann/json/pull/4612/changes#diff-ed4d9ac7996b56500f709b366672a844db294ffa884bbbcda8c690e6eb8a7711R245-R260
https://github.com/nlohmann/json/pull/4989/changes#diff-0af3903deeaf48f62fcb01acf3c6702a376ece2bb801a80d806c9757f826f41cR32-R41

The first one basically reimplements `JSON_THROW`, the second one uses `JSON_THROW` and is much more focused.

I'm not judging if any of them are better, just pointing out other versions of this same fix for not being able to use `JSON_THROW` in the macros.

nugentcaillin: Thanks for bringing these up! 

I don't think the first would be a good fit as re-implementing `JSON_THROW` ignores the undef and redef of `JSON_THROW` if `JSON_THROW_USER` is defined, and would make the code less maintainable as `JSON_THROW` would have to be updated in an additional place if changes are needed, so I think it's important that `JSON_THROW` is called

https://github.com/nugentcaillin/json/blob/21fe58fa43745a846c6987f772094a6a0a4736c0/include/nlohmann/detail/macro_scope.hpp#L188-L192
https://github.com/Ash-Jose/json_issue/blob/12319044db42395e8709fd58a04e79dad51fceb1/include/nlohmann/detail/conversions/from_json.hpp#L32-L40
```cpp
// override exception macros
#if defined(JSON_THROW_USER)
    #undef JSON_THROW
    #define JSON_THROW JSON_THROW_USER
#endif
```

As for the second, looking at them side by side they seem roughly equivalent with the exception of the previous one being inlined and templated on `BasicJsonType` instead of `ExceptionType`, and hardcoding the specific exception.
https://github.com/nugentcaillin/json/blob/21fe58fa43745a846c6987f772094a6a0a4736c0/include/nlohmann/detail/macro_scope.hpp#L258-L271
```cpp
// current

/*!
@brief function to wrap JSON_THROW_MACRO - NLOHMANN_SERIALIZE_ENUM_STRICT has a
       compilation warning about there being no arguments to JSON_THROW that depend on
       template arguments otherwise
*/
template<typename ExceptionType>
void templated_json_throw(ExceptionType exception)
{
    JSON_THROW(exception);


    /* JSON_THROW(exception) discards exception and aborts - void cast needed to supress
       compilation error if compiled with -Werror and Wunused-parameter */
    (void)exception;
}

// from previous pull

/* helper for strict enum error reporting */
inline void throw_enum_error(const BasicJsonType& j, const char* enum_type)
{
    JSON_THROW(::nlohmann::detail::type_error::create(
                   302,
                   std::string("invalid value for ") + enum_type + ": " + j.dump(),
                   &j));
}
```

I do worry that emulating this would make the change less extensible as hard-coding in the exception would mean that if another strict macro was implemented in the future it would need to implement its own throw function instead of using the current one. 

I also don't know if it would make much sense to pass json object when throwing error in `to_json` as no object would have been created from VA_ARGS, and `dump` cannot be called on nullptr. It does make a lot of sense in `from_json` though - it seems in that PR they decided not to throw in `to_json`.

If that behavior is more desirable I could make a change like this and remove throwing in `to_json` - it does seem against the spirit of a strict serialization macro to not throw though:

```cpp
/*!
@brief function to wrap JSON_THROW_MACRO - NLOHMANN_SERIALIZE_ENUM_STRICT has a
       compilation warning about there being no arguments to JSON_THROW that depend on
       template arguments otherwise
*/
void nlohmann_serialize_enum_throw(const BasicJsonType& j, const char *enum_type)
{
    JSON_THROW(nlohmann::detail::out_of_range::create(
        410,
        std::string("invalid value for") + enum_type + ": " + j.dump(),
        &j));
}
```

I could also leave the templated function for throwing unchanged and change the call site to have a more descriptive error message in `from_json` like the function from the previous pull request and do something like this:
```cpp
inline void to_json(BasicJsonType& j, const ENUM_TYPE& e)
{
    ...
    // add stringification of enum type
    else templated_json_throw<nlohmann::detail::out_of_range>(nlohmann::detail::out_of_range::create(410,"enum value out of range for " + #ENUM_TYPE, nullptr));
}
inline void from_json(BasicJsonType& j, const ENUM_TYPE& e)
{
    ...
    // add stringification of enum type, json dump and pointer
    else templated_json_throw<nlohmann::detail::out_of_range>(nlohmann::detail::out_of_range::create(410,"enum value out of range for " + #ENUM_TYPE + ": " + j.dump(), &j));
}

```
This would take the more descriptive error message from that PR whilst keeping the function available for use for any future strict macros. I've also left the inline out here since it's a templated function and I don't think inline will change how the compiler handles it at all - happy to add it in though.

Please advise which of these solutions, if any would be most appropriate. I'll start on the last one for now as that seems more appropriate but happy to switch.

gregmarr (approved):

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #5160, opened 2026-05-08
test: add missing parse-level tests for ignore_comment edge cases

shub-gour:
## Description
Add missing high-level parse tests for `ignore_comment` edge cases (issue #2330).

The existing test only covers consecutive `//` comments before an empty object.
These new sections cover:
- `//` comment before an array
- `/* */` block comment before an object
- trailing comment after a valid JSON value
- `ignore_comment=false` correctly throws on comment input

## Notes
Low-level lexer comment scanning is already tested in unit-class_lexer.cpp.
These tests cover the missing json::parse() level cases.

### DIFF
--- a/tests/src/unit-regression2.cpp
+++ b/tests/src/unit-regression2.cpp
@@ -781,6 +781,38 @@ TEST_CASE("regression tests 2")
         CHECK(j.dump() == "{}");
     }
 
+    //new test2
+    SECTION("issue #2330 - ignore_comment=true works before an array")
+    {
+        const std::string ss = "// this is a comment\n[1, 2, 3]";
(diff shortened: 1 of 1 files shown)


nlohmann:
These tests already exist.

shub-gour:
Thanks for the feedback! I'll find a genuinely uncovered 
edge case and update the PR.

shub-gour:
Updated the PR — replaced the previous tests with genuinely 
uncovered edge cases for ignore_comment at the json::parse() level.

nlohmann (commented):
Comments are ignored within the lexer, and this is covered by `unit-class_lexer.cpp`. You test the parser - this is not an edge case, but independent features.

PR #5163, opened 2026-05-10
Fix compile error when using nlohmann ordered_map with WITH_DEFAULT macros

### LINKED ISSUE #5122: Compilation error occurs when using nlohmann::ordered_map
### Description

When I use macro 'NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT' to transfer ordered_json to class, the class contains a nlohmann::ordered_map, compile error

### Reproduction steps

```
#include <iostream>
#include <nlohmann/json.hpp>

using namespace std;
using nlohmann::ordered_json;
using nlohmann::ordered_map;

struct J_Data {
    float b = 2;
    ordered_map<string, string> c;
    int a = 1;
    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(J_Data, b, c, a)
};

int main() {
    J_Data data1;
    ordered_json j = data1;
    cout << "j: " << j << endl;

    J_Data data2 = j;
    cout << "j: " << j << endl;
    return 0;
}
```
compiler error:
/usr/include/c++/10/bits/stl_algobase.h:380:18: error: use of deleted function ‘std::pair<const std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >& std::pair<const std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >::operator=(const std::pair<const std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >&)’
  380 |        *__result = *__first;

fix the bug

### Minimal code example

```Shell

```

### Error messages

```Shell

```

ubuntu20

### Library version

gcc version 10.5.0

### Validation

ssam18:
When deserializing into a struct that contains a nlohmann ordered_map field via NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT, the build fails with `pair operator= is deleted` (issue #5122). The macro expands to a ternary whose common type forces copy-assignment of ordered_map, and the inherited std vector copy-assignment uses pair operator= element-wise, which is deleted whenever the key is const. This change gives ordered_map its own copy and move assignment that rebuild via clear plus push_back or transfer the underlying buffer, plus replaces the transform-and-inserter idiom in the map from_json overload with a range-for plus emplace so the same hazard does not surface there. Built and ran the full ctest suite, all 102 targets pass including the new regression test that round-trips a struct with ordered_map<string, string>.

### DIFF
--- a/include/nlohmann/ordered_map.hpp
+++ b/include/nlohmann/ordered_map.hpp
@@ -51,6 +51,39 @@ template <class Key, class T, class IgnoredLess = std::less<Key>,
     ordered_map(std::initializer_list<value_type> init, const Allocator& alloc = Allocator() )
         : Container{init, alloc} {}
 
+    // The inherited std::vector copy/move assignment operators require
+    // value_type to be CopyAssignable / MoveAssignable when the element
+    // copy/move path triggers. value_type here is std::pair<const Key, T>,
+    // whose copy- and move-assignment are deleted because of the const Key.
+    // That breaks any code that assigns an ordered_map, including the
+    // NLOHMANN_JSON_FROM_WITH_DEFAULT macro's ternary which forces
+    // copy-assignment of the field. Provide assignment operators that
+    // rebuild via clear + push_back (copy) or transfer the underlying
+    // vector (move), both of which only need construction of value_type,
+    // never assignment. See nlohmann/json#5122.
+    ordered_map(const ordered_map&) = default;
+    ordered_map(ordered_map&&) noexcept(std::is_nothrow_move_constructible<Container>::value) = default;
+
+    ordered_map& operator=(const ordered_map& other)
+    {
+        if (this != &other)
+        {
+            Container::clear();
+            Container::reserve(other.size());
+            for (const auto& entry : other)
+            {
+                Container::push_back(entry);
+            }
+        }
+        return *this;
+    }
+
+    ordered_map& operator=(ordered_map&& other) noexcept(noexcept(std::declval<Container&>() = std::declval < Container && > ()))
+    {
+        Container::operator=(std::move(static_cast<Container&>(other)));
+        return *this;
+    }
+
     std::pair<iterator, bool> emplace(const key_type& key, T&& t)
     {
         for (auto it = this->begin(); it != this->end(); ++it)
--- a/include/nlohmann/ordered_map.hpp
+++ b/include/nlohmann/ordered_map.hpp
@@ -56,19 +56,12 @@ template <class Key, class T, class IgnoredLess = std::less<Key>,
 
     ordered_map& operator=(const ordered_map& other)
     {
-        if (this != &other)
-        {
-            Container::clear();
-            Container::reserve(other.size());
-            for (const auto& entry : other)
-            {
-                Container::push_back(entry);
-            }
-        }
+        ordered_map tmp(other);
+        Container::operator=(std::move(static_cast<Container&>(tmp)));
         return *this;
     }
 
-    ordered_map& operator=(ordered_map&& other) noexcept(noexcept(std::declval<Container&>() = std::declval < Container && > ()))
+    ordered_map& operator=(ordered_map&& other) noexcept(std::is_nothrow_move_assignable<Container>::value)
     {
         Container::operator=(std::move(static_cast<Container&>(other)));
         return *this;
--- a/include/nlohmann/ordered_map.hpp
+++ b/include/nlohmann/ordered_map.hpp
@@ -56,8 +56,11 @@ template <class Key, class T, class IgnoredLess = std::less<Key>,
 
     ordered_map& operator=(const ordered_map& other)
     {
-        ordered_map tmp(other);
-        Container::operator=(std::move(static_cast<Container&>(tmp)));
+        if (this != &other)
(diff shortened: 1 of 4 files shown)


include/nlohmann/ordered_map.hpp
```diff
@@ -50,6 +50,29 @@ template <class Key, class T, class IgnoredLess = std::less<Key>,
...
+            Container::reserve(other.size());
+            for (const auto& entry : other)
+            {
+                Container::push_back(entry);
+            }
+        }
+        return *this;
+    }
+
+    ordered_map& operator=(ordered_map&& other) noexcept(noexcept(std::declval<Container&>() = std::declval < Container && > ()))
```
nlohmann: Maybe simpler (not tested):


```suggestion
    ordered_map& operator=(ordered_map&& other) noexcept(std::is_nothrow_move_assignable<Container>::value)
```

include/nlohmann/ordered_map.hpp
```diff
@@ -50,6 +50,29 @@ template <class Key, class T, class IgnoredLess = std::less<Key>,
...
+        {
+            Container::clear();
+            Container::reserve(other.size());
+            for (const auto& entry : other)
+            {
+                Container::push_back(entry);
+            }
+        }
+        return *this;
+    }
```
nlohmann, line 65: We lose strong exception safety here. If there is a throw, we already have cleared.

Can we maybe call

```cpp
ordered_map tmp(other);
Container::operator=(std::move(tmp));
```

or

```cpp
Container::operator=(other);
```

here (not tested).

include/nlohmann/ordered_map.hpp
ssam18: yeah, that perfectly make sense. I was doing it in the old way and std::move is the best option.

tests/src/unit-regression2.cpp
```diff
@@ -1240,4 +1240,33 @@ TEST_CASE("regression test #5074 - single-element brace init with JSON_BRACE_INI
...
 
+struct Example_5122
+{
+    float b = 2;
+    nlohmann::ordered_map<std::string, std::string> c{}; // NOLINT(readability-redundant-member-init): needed for GCC -Weffc++
+    int a = 1;
+    NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(Example_5122, b, c, a)
+};
+
+TEST_CASE("regression test #5122 - from_json into types holding nlohmann::ordered_map")
```
nlohmann, line 1251: Can you add a test for self-assignment and move assignment? They seem not to be explicitly covered.

nlohmann (changes requested):

tests/src/unit-regression2.cpp
ssam18: yeah, added test now.

tests/src/unit-regression2.cpp
```diff
@@ -1240,4 +1240,81 @@ TEST_CASE("regression test #5074 - single-element brace init with JSON_BRACE_INI
...
+    auto dst_it = dst.c.begin();
+    for (; src_it != src.c.end(); ++src_it, ++dst_it)
+    {
+        CHECK(dst_it->first == src_it->first);
+        CHECK(dst_it->second == src_it->second);
+    }
+}
+
+DOCTEST_CLANG_SUPPRESS_WARNING_PUSH
+DOCTEST_CLANG_SUPPRESS_WARNING("-Wself-assign-overloaded")
```
nlohmann, line 1279: There is an issue with an older Clang version:

```
[99/162] Building CXX object tests/CMakeFiles/test-regression2_cpp11.dir/src/unit-regression2.cpp.o
/__w/json/json/tests/src/unit-regression2.cpp:1273:1: warning: unknown warning group '-Wself-assign-overloaded', ignored [-Wunknown-pragmas]
DOCTEST_CLANG_SUPPRESS_WARNING("-Wself-assign-overloaded")
^
/__w/json/json/tests/thirdparty/doctest/doctest.h:118:43: note: expanded from macro 'DOCTEST_CLANG_SUPPRESS_WARNING'
#define DOCTEST_CLANG_SUPPRESS_WARNING(w) DOCTEST_PRAGMA_TO_STR(clang diagnostic ignored w)
                                          ^
/__w/json/json/tests/thirdparty/doctest/doctest.h:116:34: note: expanded from macro 'DOCTEST_PRAGMA_TO_STR'
#define DOCTEST_PRAGMA_TO_STR(x) _Pragma(#x)
                                 ^
<scratch space>:60:27: note: expanded from here
 clang diagnostic ignored "-Wself-assign-overloaded"
                          ^
1 warning generated.
```

Maybe you can use the HEDLEY macros to constrain the suppression to those Clang versions that support the warning.

nlohmann (changes requested):

tests/src/unit-regression2.cpp
ssam18: fixed. Let's see.

nlohmann (approved):
Looks good to me.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #5172, opened 2026-05-15
Remove unnecessary if

nlohmann:
`std::copy` on an empty range is a defined no-op. The guard adds a branch and a second call to `.size()` for no benefit.

### DIFF
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -214,10 +214,7 @@ struct external_constructor<value_t::array>
         j.m_data.m_type = value_t::array;
         j.m_data.m_value = value_t::array;
         j.m_data.m_value.array->resize(arr.size());
-        if (arr.size() > 0)
-        {
-            std::copy(std::begin(arr), std::end(arr), j.m_data.m_value.array->begin());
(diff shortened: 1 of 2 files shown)


include/nlohmann/detail/conversions/to_json.hpp
```diff
@@ -214,10 +214,7 @@ struct external_constructor<value_t::array>
         j.m_data.m_type = value_t::array;
         j.m_data.m_value = value_t::array;
         j.m_data.m_value.array->resize(arr.size());
-        if (arr.size() > 0)
-        {
-            std::copy(std::begin(arr), std::end(arr), j.m_data.m_value.array->begin());
-        }
+        std::copy(std::begin(arr), std::end(arr), j.m_data.m_value.array->begin());
```
drcosmin, line 217: This is safe because:
- when arr is empty, both std::begin and std::end return the end() iterator, and std::copy handles the case where both iterators are equal (does nothing)
- also, since j.m_data.m_value.arr has just been resized to arr.size(), there's no risk of buffer overrun

drcosmin (approved):
Looks good to me

nlohmann:
@drcosmin Thanks for the review!

PR #5181, opened 2026-05-17
docs: clarify error_handler_t::ignore UTF-8 behavior

wyf027:
## Summary
Align documentation with actual `error_handler_t::ignore` behavior during `dump()`: invalid UTF-8 sequences are skipped from the output, not copied byte-for-byte.

Updates `dump.md`, `error_handler_t.md`, and `README.md`. Points to #4552 and the proposed `keep` handler in #4555.

## Motivation
Fixes confusion reported in #4552 without changing runtime semantics while #4555 is discussed.

## Test plan
- [x] Documentation only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,7 +363,7 @@ std::cout << j_string << " == " << serialized_string << std::endl;
 
 [`.dump()`](https://json.nlohmann.me/api/basic_json/dump/) returns the originally stored string value.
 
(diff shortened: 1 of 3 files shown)


nlohmann:
I do not like AI-generated documentation. The behavior right now is a bit buggy. It should be fixed or better exemplified. Linking issues is not helpful here.

PR #5182, opened 2026-05-17
docs: clarify ignore handler in FAQ and exceptions pages

wyf027:
## Summary
Align FAQ and exception troubleshooting text with `error_handler_t::ignore` semantics (skip invalid UTF-8 during `dump()`, not byte-for-byte copy).

Complements #5181.

## Motivation
#4552 — reduce confusion from outdated “silently ignore” wording.

## Test plan
- [x] Documentation only.


nlohmann:
I do not like AI-generated documentation. The behavior right now is a bit buggy. It should be fixed or better exemplified. Linking issues is not helpful here.

PR #5185, opened 2026-05-19
Fix BSON conformance issue

nlohmann:
## Summary

Fixes two BSON specification conformance issues in `nlohmann::json::from_bson()`:

1. The 4-byte `int32` `document_size` at the start of a BSON document or array
   is now enforced as a boundary. Documents with `document_size < 5` are
   rejected, and parsing must end exactly at the declared boundary (after
   consuming the trailing `0x00` document terminator).
2. The terminator byte of a length-prefixed BSON string (element type `0x02`)
   is now required to be `0x00`. Previously any non-EOF byte was accepted.

## Changes

- **`include/nlohmann/detail/input/binary_reader.hpp`**
  - `parse_bson_internal()`, `parse_bson_array()`: validate
    `document_size >= 5`; track the document start position; verify parsing
    consumes exactly `document_size` bytes.
  - `get_bson_string()`: require the terminator byte to be `0x00`.
- **`tests/src/unit-bson.cpp`**: regression tests covering the newly-rejected
  inputs.

## Notes

The binary-format readers (`from_bson`, `from_cbor`, `from_msgpack`,
`from_ubjson`, `from_bjdata`) are not designed to validate untrusted input.
An attacker who can submit bytes to `from_bson()` can already cause arbitrary
JSON values to be returned via a spec-conformant payload. These changes close
conformance gaps that should be enforced regardless.

A GitHub Security Advisory will be published referencing this PR.

Reported by @manop55555.

### DIFF
--- a/tests/src/unit-serialization.cpp
+++ b/tests/src/unit-serialization.cpp
@@ -311,14 +311,16 @@ TEST_CASE("dump for basic_json with long double number_float_t")
     SECTION("round-trip dump/parse")
     {
         constexpr std::array<long double, 13> values =
-        {{
-            0.0L, -0.0L, 1.0L, -1.0L,
-            0.5L, -0.5L, 1.5L, -2.25L,
-            1.23e45L, 1.23e-45L,
-            (std::numeric_limits<long double>::min)(),
-            std::numeric_limits<long double>::lowest(),
-            (std::numeric_limits<long double>::max)()
-        }};
+        {
+            {
+                0.0L, -0.0L, 1.0L, -1.0L,
+                0.5L, -0.5L, 1.5L, -2.25L,
+                1.23e45L, 1.23e-45L,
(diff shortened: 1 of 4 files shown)


tests/src/unit-serialization.cpp
```diff
@@ -311,14 +311,16 @@ TEST_CASE("dump for basic_json with long double number_float_t")
...
+        {
+            {
+                0.0L, -0.0L, 1.0L, -1.0L,
+                0.5L, -0.5L, 1.5L, -2.25L,
+                1.23e45L, 1.23e-45L,
+                (std::numeric_limits<long double>::min)(),
+                std::numeric_limits<long double>::lowest(),
+                (std::numeric_limits<long double>::max)()
+            }
+        };
```
nlohmann, line 323: Note: This is an unrelated whitespace change that was done by astyle.

PR #5186, opened 2026-05-20
Fix indentation overflow and correctness bug

nlohmann:
There are two issues in the current `dump` function:

- Overflow (reported by @manop55555): Under large `indent_step` values, `indent_string` growth is insufficient and can lead to buffer over-read / memory safety problems depending on write_characters() behavior. A GitHub Security Advisory will be published referencing this PR.
- In the resize of `indent_string`, a space character instead of the passed `indent_char` was used.

### DIFF
--- a/include/nlohmann/detail/output/serializer.hpp
+++ b/include/nlohmann/detail/output/serializer.hpp
@@ -128,7 +128,8 @@ class serializer
                     const auto new_indent = current_indent + indent_step;
                     if (JSON_HEDLEY_UNLIKELY(indent_string.size() < new_indent))
                     {
-                        indent_string.resize(indent_string.size() * 2, indent_char);
+                        indent_string.resize(std::max(indent_string.size() * 2, static_cast<std::size_t>(new_indent)), indent_char);
+                        JSON_ASSERT(indent_string.size() >= new_indent);
                     }
 
                     // first n-1 elements
@@ -201,7 +202,8 @@ class serializer
                     const auto new_indent = current_indent + indent_step;
                     if (JSON_HEDLEY_UNLIKELY(indent_string.size() < new_indent))
                     {
-                        indent_string.resize(indent_string.size() * 2, indent_char);
+                        indent_string.resize(std::max(indent_string.size() * 2, static_cast<std::size_t>(new_indent)), indent_char);
+                        JSON_ASSERT(indent_string.size() >= new_indent);
                     }
 
                     // first n-1 elements
@@ -262,7 +264,8 @@ class serializer
                     const auto new_indent = current_indent + indent_step;
                     if (JSON_HEDLEY_UNLIKELY(indent_string.size() < new_indent))
                     {
-                        indent_string.resize(indent_string.size() * 2, indent_char);
+                        indent_string.resize(std::max(indent_string.size() * 2, static_cast<std::size_t>(new_indent)), indent_char);
+                        JSON_ASSERT(indent_string.size() >= new_indent);
                     }
 
                     o->write_characters(indent_string.c_str(), new_indent);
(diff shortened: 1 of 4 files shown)


include/nlohmann/detail/output/serializer.hpp
```diff
@@ -128,7 +128,8 @@ class serializer
                     const auto new_indent = current_indent + indent_step;
```
LD-RW: question: this fixes the buffer-growth side (the `* 2` could undershoot a large `indent_step`), but is the accumulation of `new_indent` itself still vulnerable to wraparound? `new_indent = current_indent + indent_step` is unsigned, and the public `dump(const int indent, ...)` lets a caller pass a very large indent. For a deeply nested document the running `current_indent + indent_step` can overflow `unsigned int` and wrap to a small value; the `indent_string.size() < new_indent` guard would then pass on the wrapped-small `new_indent`, and `write_characters(indent_string.c_str(), new_indent)` would emit silently-truncated indentation rather than over-reading. Is that case considered out of scope here, or worth a saturating check / assert on `new_indent` as well?

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -128,7 +128,8 @@ class serializer
...
                     if (JSON_HEDLEY_UNLIKELY(indent_string.size() < new_indent))
                     {
-                        indent_string.resize(indent_string.size() * 2, ' ');
+                        indent_string.resize((std::max)(indent_string.size() * 2, static_cast<std::size_t>(new_indent)), indent_char);
```
LD-RW: suggestion: this resize-and-assert guard is now duplicated verbatim at all three sites (object branch L131, array branch L205, single-element branch L267). The original bug existed in three places precisely because the logic was copy-pasted, and the fix had to be pasted three times too — so the next change to the growth policy has three spots to keep in sync. Extracting a small private helper would collapse them into one and make future divergence impossible:

```cpp
void reserve_indent(std::size_t new_indent)
{
    if (JSON_HEDLEY_UNLIKELY(indent_string.size() < new_indent))
    {
        indent_string.resize((std::max)(indent_string.size() * 2, new_indent), indent_char);
    }
    JSON_ASSERT(indent_string.size() >= new_indent);
}
```

then each site becomes a single `reserve_indent(new_indent);` call before `write_characters`.

LD-RW (commented):
Took a look at the indentation overflow fix — the guarded resize plus honoring `indent_char` both look right. Left two notes inline: a suggestion to extract the now-triplicated resize-and-assert guard into a small `reserve_indent` helper so the three branches can't drift, and a question about whether the unsigned `current_indent + indent_step` accumulation can still wrap on deeply nested input even after this fix. Neither is a blocker — the second is more "is this in scope?" than a defect claim.

include/nlohmann/detail/output/serializer.hpp
nlohmann: Thanks for the suggestion! Done in https://github.com/nlohmann/json/pull/5186/commits/250787cb54652caf0a45b072af89ae12ae86b663.

include/nlohmann/detail/output/serializer.hpp
nlohmann: Good call. Addressed this in https://github.com/nlohmann/json/pull/5186/commits/526c8c50795aa75991997c054ea2ffb2a86cbfb7: Added a `JSON_ASSERT(new_indent >= current_indent)` guard inside the new `reserve_indent` helper to catch the unsigned wraparound in debug builds. Full saturating handling is out of scope here: the wrap only yields truncated indentation (not an over-read) and is unreachable without first allocating ~4 GB for the indent string.

gregmarr:
Looks like this is just tests now.  Do any of these tests duplicate ones that were added in the previous work?

nlohmann:
Confirmed — #5285 (merged the same day you asked) removed `indent_string` and the doubling-resize bug entirely, replacing it with `put_indent()`. Its commit message says explicitly that this removes both bugs #5186 was fixing. #5285 also added its own regression test (`TEST_CASE("indentation is written straight into the write buffer")` in unit-serialization.cpp) covering a wide `indent_step`, a non-space indent char, and multi-buffer-full nesting — the same scenarios this PR's tests target.

Closing this one as superseded by #5285.

PR #5187, opened 2026-05-21
Fix: update() parent pointers not updated after recursive merge with JSON_DIAGNOSTICS

### LINKED ISSUE #4813: json::update() with merge_objects==true may trigger JSON_ASSERT for some objects
### Description

json::update() triggers JSON_ASSERT when certain conditions are met:
- JSON_DIAGNOSTICS turned on (this JSON_ASSERT is inside #if JSON_DIAGNOSTICS block)
- Build type is "Debug"
- ordered_json used (basic_json works fine)
- "merge_objects" argument of update() set to "true"
- passed json objects have some specific structure:
Behavior depends on objects' fields and is very inconsistent. Assertion failure happens with my example, but it won't if you remove any field, change fields' order, or make keys inside "numbers" equal

### Reproduction steps

Build my example and try to run it  
commands:
```
cmake -S . -B ./build
cmake --build ./build
./build/Debug/update_test.exe
```
Output:
```
j1: {"numbers":{"one":1}}
j2: {"numbers":{"two":2},"string":"t"}
Assertion failed: !check_parents || !is_structured() || std::all_of(begin(), end(), [this](const basic_json & j) { return j.m_parent == this; }), file 3rd-party/nlohmann/json.hpp, line 700
```

Expected: No errors
Actual: Assertion failure

### Minimal code example

CMakeLists.txt:
```Shell
cmake_minimum_required(VERSION 3.21)

project (update_test
	LANGUAGES CXX
)
add_executable( ${PROJECT_NAME}
	"main.cpp"
)
target_include_directories(${PROJECT_NAME} PUBLIC
	$<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/3rd-party>	
)
target_compile_definitions(${PROJECT_NAME} PUBLIC 
	JSON_DIAGNOSTICS=1
)
```

main.cpp
Based on update() unit test
[https://github.com/nlohmann/json/blob/develop/tests/src/unit-modifiers.cpp#L728](https://github.com/nlohmann/json/blob/develop/tests/src/unit-modifiers.cpp#L728)
```cpp
#include <iostream>
#include <nlohmann/json.hpp>

int main(int argc, char** argv)
{
	using json = nlohmann::ordered_json;
	json j1 = {
		{"numbers", {{"one", 1}}} };
	json const j2 = {
		{"numbers", {{"two", 2}}},
		{"string", "t"} };
	std::cout << "j1: " << j1.dump() << std::endl;
	std::cout << "j2: " << j2.dump() << std::endl;
	j1.update(j2, true); // assert triggered here
	std::cout << "j1: " << j1.dump() << std::endl;
}
```

### Error messages

```Shell
Assertion failed: !check_parents || !is_structured() || std::all_of(begin(), end(), [this](const basic_json & j) { return j.m_parent == this; }), file 3rd-party/nlohmann/json.hpp, line 700
```

Windows 10, MSVC 2022 or gcc 13.1

### Library version

3.11.3, v3.12.0, develop

### Validation

VasuBhakt:
## Issue
Fixes #4813

## Overview
Added a set_parents() call under `#if JSON_DIAGNOSTICS` in the `update()` iterator overload after a recursive merge.

When update() is called with merge_objects=true, nested objects are recursively merged via `it2->second.update(it.value(), true)`. After this recursive call, the child nodes inside the merged object have stale `m_parent pointers`, causing the parent consistency assertion to fire when `JSON_DIAGNOSTICS=1` is enabled. Calling `set_parents()` after the recursive merge walks the subtree and fixes all parent pointers correctly.

## Changes
- `include/nlohmann/json.hpp` — 3 line fix in `update()` iterator overload
- `single_include/nlohmann/json.hpp` — regenerated via `make amalgamate`

The changes were tested locally which gave the results:

Screenshot 2026-05-21 193726

---

- [x] If applicable, an [existing issue](https://github.com/nlohmann/json/issue/4813) is referenced.

### DIFF
--- a/tests/src/unit-regression2.cpp
+++ b/tests/src/unit-regression2.cpp
@@ -1330,3 +1330,16 @@ TEST_CASE("regression test #5122 - nlohmann::ordered_map move-assignment transfe
 }
 
 DOCTEST_CLANG_SUPPRESS_WARNING_POP
+
+TEST_CASE("regression test #4813 - update() with merge_objects=true triggers JSON_ASSERT with JSON_DIAGNOSTICS")
+{
+    // https://github.com/nlohmann/json/issues/4813
+    nlohmann::ordered_json j1 = {{"numbers", {{"one", 1}}}};
+    nlohmann::ordered_json const j2 = {{"numbers", {{"two", 2}}}, {"string", "t"}};
+    CHECK_NOTHROW(j1.update(j2, true));
+    CHECK(j1["numbers"]["one"] == 1);
+    CHECK(j1["numbers"]["two"] == 2);
+    CHECK(j1["string"] == "t");
+}
+
+
--- a/tests/src/unit-regression2.cpp
+++ b/tests/src/unit-regression2.cpp
@@ -1330,16 +1330,3 @@ TEST_CASE("regression test #5122 - nlohmann::ordered_map move-assignment transfe
 }
 
 DOCTEST_CLANG_SUPPRESS_WARNING_POP
-
-TEST_CASE("regression test #4813 - update() with merge_objects=true triggers JSON_ASSERT with JSON_DIAGNOSTICS")
-{
-    // https://github.com/nlohmann/json/issues/4813
-    nlohmann::ordered_json j1 = {{"numbers", {{"one", 1}}}};
-    nlohmann::ordered_json const j2 = {{"numbers", {{"two", 2}}}, {"string", "t"}};
-    CHECK_NOTHROW(j1.update(j2, true));
-    CHECK(j1["numbers"]["one"] == 1);
-    CHECK(j1["numbers"]["two"] == 2);
-    CHECK(j1["string"] == "t");
-}
-
-
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -3497,6 +3497,9 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
                 if (it2 != m_data.m_value.object->end())
                 {
                     it2->second.update(it.value(), true);
+#if JSON_DIAGNOSTICS
+                    it2->second.set_parents();
+#endif
                     continue;
                 }
             }
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -24000,6 +24000,9 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
                 if (it2 != m_data.m_value.object->end())
                 {
                     it2->second.update(it.value(), true);
+#if JSON_DIAGNOSTICS
+                    it2->second.set_parents();
+#endif
                     continue;
                 }
             }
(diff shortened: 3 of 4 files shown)


nlohmann (changes requested):
Can you please add a regression test (best the example from #4813) to [`unit-regression2.cpp`](https://github.com/nlohmann/json/blob/develop/tests/src/unit-regression2.cpp)?

VasuBhakt:
sure, working on it.

VasuBhakt:
added the regression test @nlohmann, please review it

tests/src/unit-regression2.cpp
```diff
@@ -1330,3 +1330,16 @@ TEST_CASE("regression test #5122 - nlohmann::ordered_map move-assignment transfe
 }
 
 DOCTEST_CLANG_SUPPRESS_WARNING_POP
+
```
nlohmann: Thanks, but I was wrong to propose putting this to unit-regression2.cpp, because it is not compiled with JSON_DIAGNOSTICS. Instead, please move the test to unit-diagnostics.cpp which is designed for this. Sorry for the inconvenience.

The test looks fine otherwise!

VasuBhakt: No worries, fixing it

tests/src/unit-regression2.cpp
```diff
@@ -1330,3 +1330,4 @@ TEST_CASE("regression test #5122 - nlohmann::ordered_map move-assignment transfe
...
```
nlohmann: Is this an unrelated newline?

VasuBhakt: yes my bad it was left while refactoring the test from unit-regression2 to unit-diagnostic

VasuBhakt: should i fix it?

nlohmann: Yes, please. Then this is good to go.

tests/src/unit-regression2.cpp
```diff
@@ -1329,4 +1329,4 @@ TEST_CASE("regression test #5122 - nlohmann::ordered_map move-assignment transfe
     CHECK(src.begin()->first == "after-move");
 }
 
-DOCTEST_CLANG_SUPPRESS_WARNING_POP
+DOCTEST_CLANG_SUPPRESS_WARNING_POP
```
nlohmann: No there is no newline at the end of the file. Just revert everything here.

VasuBhakt: okay so it should stop at 1332 right?

VasuBhakt: and should there be a newline at the end of the unit-diagnostic as well?

nlohmann (approved):
Looks good to me.

tests/src/unit-regression2.cpp
nlohmann: It's fine as is! Thanks a lot! I will merge once the CI is green.

VasuBhakt: Thanks for your patience correcting me on the issues, you are a great mentor!

VasuBhakt:
Thanks a lot @nlohmann! You have been very patient and responsive throughout the review process. Really appreciate the guidance!

nlohmann:
Thanks!

PR #5164, opened 2026-05-11
Fix GCC C++20 modules compilation #5103

hariomphulre:
This PR fixes GCC C++20 modules compilation errors (issue #5103) that occur when `nlohmann/json.hpp` is included in a module preamble after importing other modules.

## What Changed
- **`include/nlohmann/json.hpp`**: Added GCC diagnostic pragmas to suppress false-positive warnings in C++20 module preambles
- **`src/modules/json.cppm`**: Added GCC-specific documentation with workarounds; removed `export namespace std` to avoid symbol conflicts
- **`src/modules/CMakeLists.txt`**: Simplified module structure (removed partition file complexity)

## Why Changed
GCC has limitations handling template instantiations in module preambles when other modules are already imported. The pragmas suppress spurious template redefinition errors without affecting actual warnings. Documentation helps users understand the limitation and provides three workarounds.

## Testing
- ✅ All 10,000+ existing tests pass
- ✅ Module compiles without errors on GCC 15.2.1
- ✅ Zero regressions detected
- ✅ Tested with problematic include order (STL headers before json.hpp)

---

- [x] The changes are described in detail, both the what and why.
- [x] If applicable, an [existing issue](https://github.com/nlohmann/json/issues) is referenced. (#5103)
- [x] The [Code coverage](https://coveralls.io/github/nlohmann/json) remained at 100%. All existing code paths unchanged; only pragmas and comments added.
- [x] If applicable, the [documentation](https://json.nlohmann.me) is updated. (Code comments added to module file)
- [x] The source code is amalgamated by running `make amalgamate`. (Not applicable - only module and header changes)

PR #5137, opened 2026-04-15
Fix missing exports from `json.cppm`

mikomikotaishi:
[Describe your pull request here. Please read the text below the line and make sure you follow the checklist.]

This pull request restores some missing symbols that were present in the `<nlohmann/json.hpp>` header, but weren't exported in `json.cppm`. These include specialisations in namespace `std` (such as `hash<>` and `less<>`), as well as `nlohmann::literals::json_literals::*`. This is necessary as without this those specialisations and symbols are inaccessible from the module.

- [x] The changes are described in detail, both the what and why.
- [ ] If applicable, an [existing issue](https://github.com/nlohmann/json/issues) is referenced.
- [x] The [Code coverage](https://coveralls.io/github/nlohmann/json) remained at 100%. A test case for every new line of code.
- [x] If applicable, the [documentation](https://json.nlohmann.me) is updated.
- [x] The source code is amalgamated by running `make amalgamate`.

PR #5190, opened 2026-05-21
docs: Fix missing newline necessary in docs website

mikomikotaishi:

In my previous PR, #5137, I didn't add a newline at the beginning of the bullet-point lists, which caused the list of exported symbols on the documentation page to not render correctly. This PR adds the needed newlines so that the list renders correctly. This furthermore removes the mention of exporting symbols from `std`, as this is not actually necessary to access these specialisations (and was removed in a previous PR, #5164).

### DIFF
--- a/docs/mkdocs/docs/features/modules.md
+++ b/docs/mkdocs/docs/features/modules.md
@@ -38,9 +38,3 @@ Only the following symbols are exported from `nlohmann.json`:
 - `nlohmann::to_string`
 - `nlohmann::literals::json_literals::operator""_json`
 - `nlohmann::literals::json_literals::operator""_json_pointer`
-
-The following specialisations of `std` symbols are also exported:
-
-- `std::hash`
-- `std::less`
-- `std::swap`
--- a/docs/mkdocs/docs/features/modules.md
+++ b/docs/mkdocs/docs/features/modules.md
@@ -28,6 +28,7 @@ It should be noted that as modules do not export macros, the `nlohmann.json` mod
 
 ## Exported symbols
 Only the following symbols are exported from `nlohmann.json`:
+
 - `nlohmann::adl_serializer`
 - `nlohmann::basic_json`
 - `nlohmann::json`
@@ -39,6 +40,7 @@ Only the following symbols are exported from `nlohmann.json`:
 - `nlohmann::literals::json_literals::operator""_json_pointer`
 
 The following specialisations of `std` symbols are also exported:
+
 - `std::hash`
 - `std::less`
 - `std::swap`


nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

mikomikotaishi:
> Please be aware that the module is experimental and a full test is outstanding, and the exported symbols are subject to change.

By the way, what else would need to be done before this disclaimer could be removed? What would a full test need to consist of for the module to be no longer experimental? (I'm not in a rush for anything, but I would be interested in providing additional improvements so that it can be stabilised.)

Additionally, is there a reason `using json = nlohmann::json;` is used throughout the documentation instead of `using nlohmann::json;` when the latter is more concise and these two do the same thing?

nlohmann:
> By the way, what else would need to be done before this disclaimer could be removed? What would a full test need to consist of for the module to be no longer experimental? (I'm not in a rush for anything, but I would be interested in providing additional improvements so that it can be stabilised.)

I think the best we could do is to provide some mechanism to execute the test suite against the module. Maybe this is possible with some preprocessor magic.

Then we would see if all features actually work at least for some compilers.

Happy to see PRs!

> Additionally, is there a reason using json = nlohmann::json; is used throughout the documentation instead of using nlohmann::json; when the latter is more concise and these two do the same thing?

We have some examples where we use `using json = nlohmann::ordered_json` and then use the same code later on.

mikomikotaishi:
> I think the best we could do is to provide some mechanism to execute the test suite against the module. Maybe this is possible with some preprocessor magic.

Many of the tests I see (the CMake building tests in particular) simply check whether a file compiles after including `<nlohmann/json.hpp>` and constructing a `nlohmann::json`. The [current modules test](https://github.com/nlohmann/json/blob/develop/tests/module_cpp20/main.cpp) does the same thing (and that was how it looked when I first found it); would you like if I made the test a bit more comprehensive to test additional features that the library offers?

I'm not too sure what could be done with preprocessor manipulation, or how much more extensive this would be, but at the very least expanding the current test would provide some greater confidence that things work.

> We have some examples where we use using json = nlohmann::ordered_json and then use the same code later on.

Ah, I see, thanks.

Though, wouldn't it be clearer to just write `using nlohmann::ordered_json;` in those places? Then it would be clear in all places when `basic_json` or `ordered_json` is being used, considering `json` is already a library-defined alias for a template specialisation of `basic_json`.

nlohmann:
Maybe the next step should not necessarily be extending the scope of the tests, but rather the scope of the compilers. It would be good to extend https://json.nlohmann.me/community/quality_assurance/ with a list of compilers for which we know the module work.

PR #5193, opened 2026-05-26
Add null-safe noexcept accessors eval_value/eval_array/eval_object

gitpaladin:
## Summary

Introduces an opt-in header `<nlohmann/eval.hpp>` with three free function templates that complement `basic_json::value()` with semantics designed for untrusted JSON payloads:

| Function | Behaviour |
|---|---|
| `eval_value(j, key\|ptr, default)` | `noexcept`, returns `default` on any non-matching condition (non-object receiver, missing key/path, null resolved value, wrong type, conversion failure). |
| `eval_array(j, key\|ptr)` | `noexcept`, returns a `const&` to a static empty array on any non-matching condition. |
| `eval_object(j, key\|ptr)` | `noexcept`, returns a `const&` to a static empty object on any non-matching condition. |

Follows the design agreed in #5129.

## Why non-member, opt-in header

Per your preference in the discussion ("I definitely prefer this over an extension of the (already bloated) basic_json API"):

- Non-member functions in `namespace nlohmann` so ADL works without explicit qualification (`eval_value(j, ...)` just works).
- Dedicated opt-in header — `basic_json`'s public surface is not extended.
- Uses only the public API (`is_object`, `is_array`, `is_null`, `find`, `end`, `get`, `contains(json_pointer)`, `at(json_pointer)`).
- Pointer overloads guard with `j.contains(ptr)` before `j.at(ptr)`, so they remain correct under `JSON_NOEXCEPTION` too (where `at()` would otherwise abort).
- Static empty array/object fallbacks are returned by const reference via Meyers' singletons — no per-call allocation, thread-safe since C++11.
- A header-private `NLOHMANN_EVAL_TRY`/`NLOHMANN_EVAL_CATCH_ALL` pair mirrors `JSON_TRY`/`JSON_INTERNAL_CATCH`, since those macros are `#undef`'d at the end of `json.hpp` via `macro_unscope.hpp` and are not visible to consumers.

## What is added

- `include/nlohmann/eval.hpp` — 6 overloads (key / json_pointer × value / array / object).
- `tests/src/unit-eval.cpp` — 7 test cases, 79 assertions, picked up automatically by the existing `file(GLOB src/unit-*.cpp)` in `tests/CMakeLists.txt`.
- `docs/mkdocs/docs/api/eval.md` — API page mirroring `basic_json/value.md` style.

## What is not changed

- `basic_json` is untouched — no new members, no signature changes, no ABI impact.
- `single_include/nlohmann/json.hpp` is **not** modified — the new header is intentionally not bundled into the amalgamated single-include (matches Step 4 of the "Next step" plan in #5129). Happy to extend `tools/amalgamate/config_json.json` if you'd prefer it bundled.

## Local test matrix

| Toolchain | Standards | Result |
|---|---|---|
| MSVC cl (VS 2026) | C++17 | 7/7 cases, 79/79 assertions PASS, 0 warnings |
| MSYS2 g++ 15.2 | C++11 / C++17 / C++20 | PASS each, 0 warnings |
| MSYS2 clang++ 21.1 | C++11 / C++17 / C++20 | PASS each, 0 warnings (incl. `-Wshadow -Wconversion -Wsign-conversion -Wold-style-cast`) |
| g++ 15.2 + `-fno-exceptions -DJSON_NOEXCEPTION` | C++17 | smoke test PASS (doctest's own limitations apply to the full suite, same as `test-disabled_exceptions`) |

## Checklist

- [x] Changes described above.
- [x] References discussion #5129.
- [x] New code covered by tests in `tests/src/unit-eval.cpp` (7 cases / 79 assertions).
- [x] API documentation page added under `docs/mkdocs/docs/api/`.
- [ ] `make amalgamate` — not run yet because the new header is intentionally not part of the single-include. If you'd like it bundled, I'll extend the amalgamate config and rerun. Will also rerun should the CI's amalgamation check flag any astyle drift.

Marked as **draft** so you can pre-screen the direction; happy to address feedback (naming, single-include bundling, anything else) before un-drafting.

### DIFF
--- a/docs/mkdocs/docs/api/eval.md
+++ b/docs/mkdocs/docs/api/eval.md
@@ -2,160 +2,234 @@
 
 ```cpp
 #include <nlohmann/eval.hpp>
-```
 
-Null-safe, `noexcept` accessors for retrieving values from a JSON object.
+namespace nlohmann
+{
 
-These free functions are an opt-in alternative to
-[`basic_json::value`](basic_json/value.md). Unlike `value()`, they **never
-throw** -- on any non-matching condition (the receiver is not an object, the
-key/pointer is missing, the resolved value is null, or has the wrong type)
-they silently fall back to the supplied default value (for `eval_value`) or
-to a static empty array/object (for `eval_array` / `eval_object`).
-
-This brings the developer experience of accessing untrusted server-side JSON
-closer to JavaScript's optional chaining (`?.`) and nullish coalescing (`??`).
-
-## API
-
-```cpp
-// (1) -- access by key
+// (1) -- access a JSON object element by key with a default fallback
 template <class BasicJsonType, class ValueType>
 ValueType eval_value(const BasicJsonType& j,
                      const typename BasicJsonType::object_t::key_type& key,
                      const ValueType& default_value) noexcept;
 
-// (2) -- access by JSON Pointer
+// (2) -- access a JSON value by JSON Pointer with a default fallback
 template <class BasicJsonType, class ValueType>
 ValueType eval_value(const BasicJsonType& j,
                      const typename BasicJsonType::json_pointer& ptr,
                      const ValueType& default_value) noexcept;
 
-// (3) -- array access by key
+// (3) -- access a JSON array stored at @a key
 template <class BasicJsonType>
-const BasicJsonType& eval_array(
-    const BasicJsonType& j,
-    const typename BasicJsonType::object_t::key_type& key) noexcept;
+const BasicJsonType& eval_array(const BasicJsonType& j,
+                                const typename BasicJsonType::object_t::key_type& key) noexcept;
 
-// (4) -- array access by JSON Pointer
+// (4) -- access a JSON array resolved by @a ptr
 template <class BasicJsonType>
-const BasicJsonType& eval_array(
-    const BasicJsonType& j,
-    const typename BasicJsonType::json_pointer& ptr) noexcept;
+const BasicJsonType& eval_array(const BasicJsonType& j,
+                                const typename BasicJsonType::json_pointer& ptr) noexcept;
 
-// (5) -- object access by key
+// (5) -- access a JSON object stored at @a key
 template <class BasicJsonType>
-const BasicJsonType& eval_object(
-    const BasicJsonType& j,
-    const typename BasicJsonType::object_t::key_type& key) noexcept;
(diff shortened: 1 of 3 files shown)


docs/mkdocs/docs/api/eval.md
```diff
@@ -0,0 +1,164 @@
+# <small>nlohmann::</small>eval_value, eval_array, eval_object
+
+```cpp
+#include <nlohmann/eval.hpp>
+```
+
+Null-safe, `noexcept` accessors for retrieving values from a JSON object.
```
gregmarr: Should this clarify that this means "null in the JSON data" and not "null C++ objects"?

gregmarr (approved):

nlohmann:
Thank you for putting so much work into this, and for the thorough docs and tests. I also appreciate that you followed up on the discussion in #5129 and kept it out of `basic_json` as I suggested there.

After looking at the finished PR, though, I've decided not to merge it. The main reason is that for most uses, what these functions do already fits in one line of existing API:

```cpp
j.is_object() ? j.value(key, default_value) : default_value
```

Covering null values and failed conversions as well takes only a few more lines:

```cpp
template<class T>
T eval_value(const json& j, const json::object_t::key_type& key, const T& default_value)
{
    if (!j.is_object()) return default_value;
    const auto it = j.find(key);
    if (it == j.end() || it->is_null()) return default_value;
    try { return it->get<T>(); } catch (const json::exception&) { return default_value; }
}
```

`eval_array` and `eval_object` are similar: a `find()` followed by an `is_array()`/`is_object()` check. Adding about 1000 lines of header, tests and docs to the library, and supporting them long term, is a lot for something users can write this easily. It would also be the first header outside `single_include`, which means a second way of distributing the library that I'd rather avoid.

I also checked the code and it seems that getting "never throws" completely right is harder than it looks:
- Returning `default_value` by value copies it, so a `std::string` default can still throw inside a `noexcept` function.
- `eval_array(json::parse(s), "items")` returns a reference into a temporary.
- The JSON pointer overloads return the default for array roots such as `"/0"` on `[10]`.

These are fixable, but they show the functions need design decisions that are easier to make in a small helper for your own project than in a general-purpose library API.

The functions are free functions, so they work just as well in your own code or as a small standalone header. If you publish one, feel free to link it in the discussion so others can find it.

Thanks again for the contribution. I'm sorry this isn't the answer you were hoping for after all the effort.

Issue #4440, opened 2024-08-23
Assert when using std::views::filter and GCC 10


gwjo:
### Description

After upgrading from `3.9.1` to `3.11.3` the code asserts when iterating over a json array that's been filtered using `std::views::filter` due to an uninitialized iterator.

In addition to my failure with 3.11.3 and GCC 10.2, I checked other combinations on Godbolt and observed the following (❌ represents the assert is seen)

|                 | GCC 10.2 | GCC 10.5 | GCC 11.1 | GCC trunk | Clang 16 |
| --------------- | -------- | -------- | -------- | --------- | -------- |
| 3.9.1           | ✅ | ✅ | ✅ | ✅ | ✅ |
| 3.10.5          |  ✅ | ✅ | ✅ | ✅ | ✅ |
| 3.11.1          | ❌ |  ❌ |  ❌ |  ✅ | ✅ |
| trunk           | ❌ |  ❌ |  ❌ |  ✅ | ✅ |

### Reproduction steps

https://godbolt.org/z/WbexYTqnf

### Expected vs. actual results

`std::views::filter` should allow you to filter a json array.

### Minimal code example

```Shell
#include <nlohmann/json.hpp>
#include <iostream>
#include <ranges>

using nlohmann::json;

int main()
{
    auto noOpFilter = std::views::filter([](auto&&){ return true; });
    json j = {1, 2, 3};
    auto filtered = j | noOpFilter;
    std::cout << *filtered.begin() << std::endl;
    return 0;
}
```

### Error messages

```Shell
[...]

PR #5205, opened 2026-06-07
fix: support constructing json from C++20 range views (#4916)

### LINKED ISSUE #4916: Constructing array from C++20 ranges view does not work
### Description

Constructing a `json` object from a C++20 view (e.g. `std::ranges::transform_view`/`std::ranges::filter_view`) does not work.

### Reproduction steps

Constructing a `json` object from a `std::vector` works fine, but constructing from a more basic range such as a vector passed through `std::views::filter` does not work. This holds for a range of `json` object as well as a range of plain `int`s.

Expected: `json` value constructs as array with elements from the view (e.g. below: with contents `[37, 42, 21]`).

Actual: Fails to compile as `is_compatible_array_type` returns false for some reason.

### Minimal code example

```cpp
#include <iostream>
#include <ranges>
#include <vector>

#include <nlohmann/json.hpp>

int main() {
    std::vector<int> nums{1, 2, 37, 42, 21};
    auto filteredNums = nums | std::views::filter([](int i) { return i > 10; });
    nlohmann::json j(filteredNums);
    std::cout << j << std::endl;
}
```

[Repro on Compiler Explorer](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:1,endLineNumber:13,positionColumn:1,positionLineNumber:13,selectionStartColumn:1,selectionStartLineNumber:13,startColumn:1,startLineNumber:13),source:'%23include+%3Ciostream%3E%0A%23include+%3Cranges%3E%0A%23include+%3Cvector%3E%0A%0A%23include+%3Cnlohmann/json.hpp%3E%0A%0Aint+main()+%7B%0A++++std::vector%3Cint%3E+nums%7B1,+2,+37,+42+,+21%7D%3B%0A++++auto+filteredNums+%3D+nums+%7C+std::views::filter(%5B%5D(int+i)+%7B+return+i+%3E+10%3B+%7D)%3B%0A++++nlohmann::json+j(filteredNums)%3B%0A++++std::cout+%3C%3C+j+%3C%3C+std::endl%3B%0A%7D%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:39.2695652173913,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:clang_trunk,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!((name:nlohmann_json,ver:trunk)),options:'-std%3Dc%2B%2B20+-Wall+-Wextra+-Wconversion+-Wpedantic+-Wdeprecated',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+clang+(trunk)+(Editor+%231)',t:'0')),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+(trunk)',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'0'),l:'5',n:'0',o:'Output+of+x86-64+clang+(trunk)+(Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:60.73043478260869,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

### Error messages

```
<source>:10:20: error: no matching constructor for initialization of 'nlohmann::json' (aka 'basic_json<>')
   10 |     nlohmann::json j(filteredNums);
      |                    ^ ~~~~~~~~~~~~
[...]
/opt/compiler-explorer/libs/nlohmann_json/trunk/single_include/nlohmann/json.hpp:20942:5: note: candidate template ignored: requirement 'nlohmann::detail::is_compatible_type<nlohmann::basic_json<std::map, std::vector, std::basic_string<char, std::char_traits<char>, std::allocator<char>>, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char>>, void>, std::ranges::filter_view<std::ranges::ref_view<std::vector<int, std::allocator<int>>>, (lambda at <source>:9:51)>>::value' was not satisfied [with CompatibleType = typename __invoke_result<const std::ranges::views::_Filter &, std::vector<int, std::allocator<int>> &, const (lambda at <source>:9:51) &>::type &, U = detail::uncvref_t<typename __invoke_result<const std::ranges::views::_Filter &, std::vector<int, std::allocator<int>> &, const (lambda at <source>:9:51) &>::type &>]
 20942 |     basic_json(CompatibleType && val) noexcept(noexcept( // NOLINT(bugprone-forwarding-reference-overload,bugprone-exception-escape)
       |     ^
[...]
```

Clang 22 / GCC 16

### Library version

3.11.3, 3.12.0

### Validation

federicosfriso05-dotcom:
## Description

Fixes #4916.

Constructing a `nlohmann::json` object from a C++20 range view (e.g. `std::views::filter`, `std::views::transform`) failed to compile because `is_compatible_array_type` did not recognize range views as compatible types.

### What changed

- `include/nlohmann/detail/meta/type_traits.hpp`: added a new partial specialization of `is_compatible_array_type_impl` for C++20 range views, guarded by `#ifdef JSON_HAS_CPP_20`. The specialization matches types satisfying `std::ranges::view` while excluding character sequences (e.g. `std::string_view`).
- `include/nlohmann/detail/conversions/to_json.hpp`: added a new `construct` overload for C++20 range views inside `external_constructor<value_t::array>`, also guarded by `#ifdef JSON_HAS_CPP_20`.
- The specialization uses `std::ranges::view` rather than just `std::ranges::range` to avoid ambiguity with standard containers like `std::string` and `std::string_view`, which already satisfy `std::ranges::range` but are handled by existing specializations. Character sequences are additionally excluded by checking that the range's value type is not char or wchar_t.

### How to test

```cpp
std::vector<int> nums{1, 2, 37, 42, 21};
auto filtered = nums | std::views::filter([](int i) { return i > 10; });
nlohmann::json j(filtered);
assert(j == nlohmann::json({37, 42, 21}));
```

- [ ] If applicable, the documentation is updated.

### DIFF
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -177,8 +177,11 @@ struct external_constructor<value_t::array>
     }
 
     template < typename BasicJsonType, typename CompatibleArrayType,
-               enable_if_t < !std::is_same<CompatibleArrayType, typename BasicJsonType::array_t>::value,
-                             int > = 0 >
+               enable_if_t < !std::is_same<CompatibleArrayType, typename BasicJsonType::array_t>::value
+#ifdef JSON_HAS_CPP_20
+                             && !std::ranges::view<CompatibleArrayType>
+#endif
+                             , int > = 0 >
     static void construct(BasicJsonType& j, const CompatibleArrayType& arr)
     {
         using std::begin;
@@ -218,6 +221,24 @@ struct external_constructor<value_t::array>
         j.set_parents();
         j.assert_invariant();
     }
+
+#ifdef JSON_HAS_CPP_20
+    template<typename BasicJsonType, typename CompatibleArrayType,
+             enable_if_t<std::ranges::view<std::remove_cv_t<CompatibleArrayType>>, int> = 0>
+    static void construct(BasicJsonType& j, const CompatibleArrayType& arr)
+    {
+        auto view = arr;
+        j.m_data.m_value.destroy(j.m_data.m_type);
+        j.m_data.m_type = value_t::array;
+        j.m_data.m_value = value_t::array;
+        for (const auto& x : view)
+        {
+            j.m_data.m_value.array->push_back(x);
+            j.set_parent(j.m_data.m_value.array->back());
+        }
+        j.assert_invariant();
+    }
+#endif
 };
 
 template<>
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -178,7 +178,7 @@ struct external_constructor<value_t::array>
 
     template < typename BasicJsonType, typename CompatibleArrayType,
                enable_if_t < !std::is_same<CompatibleArrayType, typename BasicJsonType::array_t>::value
-#ifdef JSON_HAS_CPP_20
+#if JSON_HAS_RANGES && !defined(__MINGW32__)
                              && !std::ranges::view<CompatibleArrayType>
 #endif
                              , int > = 0 >
@@ -222,7 +222,9 @@ struct external_constructor<value_t::array>
         j.assert_invariant();
     }
 
-#ifdef JSON_HAS_CPP_20
+    // std::ranges does not work properly on MinGW due to incomplete C++20 support
+    // see https://github.com/nlohmann/json/issues/4916
+#if JSON_HAS_RANGES && !defined(__MINGW32__)
     template<typename BasicJsonType, typename CompatibleArrayType,
              enable_if_t<std::ranges::view<std::remove_cv_t<CompatibleArrayType>>, int> = 0>
     static void construct(BasicJsonType& j, const CompatibleArrayType& arr)
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -232,7 +232,7 @@ struct external_constructor<value_t::array>
         j.m_data.m_value.destroy(j.m_data.m_type);
         j.m_data.m_type = value_t::array;
         j.m_data.m_value = value_t::array;
-        for (auto& x : std::forward<CompatibleArrayType>(arr))
+        for (auto&& x : std::forward<CompatibleArrayType>(arr))
         {
             j.m_data.m_value.array->push_back(x);
             j.set_parent(j.m_data.m_value.array->back());
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -178,7 +178,7 @@ struct external_constructor<value_t::array>
 
     template < typename BasicJsonType, typename CompatibleArrayType,
                enable_if_t < !std::is_same<CompatibleArrayType, typename BasicJsonType::array_t>::value
-#if JSON_HAS_RANGES && !defined(__MINGW32__)
+#if JSON_HAS_RANGES && !defined(__MINGW32__) && (!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE >= 12)
                              && !std::ranges::view<CompatibleArrayType>
 #endif
                              , int > = 0 >
@@ -224,7 +224,7 @@ struct external_constructor<value_t::array>
 
     // std::ranges does not work properly on MinGW due to incomplete C++20 support
     // see https://github.com/nlohmann/json/issues/4916
-#if JSON_HAS_RANGES && !defined(__MINGW32__)
+#if JSON_HAS_RANGES && !defined(__MINGW32__) && (!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE >= 12)
     template<typename BasicJsonType, typename CompatibleArrayType,
              enable_if_t<std::ranges::view<std::remove_cv_t<CompatibleArrayType>>, int> = 0>
     static void construct(BasicJsonType& j, const CompatibleArrayType& arr)
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -232,7 +232,7 @@ struct external_constructor<value_t::array>
         j.m_data.m_value.destroy(j.m_data.m_type);
         j.m_data.m_type = value_t::array;
         j.m_data.m_value = value_t::array;
-        for (auto& x : arr)
+        for (auto& x : std::forward<CompatibleArrayType>(arr))
         {
             j.m_data.m_value.array->push_back(x);
             j.set_parent(j.m_data.m_value.array->back());
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -226,14 +226,13 @@ struct external_constructor<value_t::array>
     // see https://github.com/nlohmann/json/issues/4916
 #if JSON_HAS_RANGES && !defined(__MINGW32__)
     template<typename BasicJsonType, typename CompatibleArrayType,
-             enable_if_t<is_compatible_range_view<std::remove_cv_t<CompatibleArrayType>>::value, int> = 0>
-    static void construct(BasicJsonType& j, const CompatibleArrayType& arr)
+             enable_if_t<is_compatible_range_view<std::remove_cvref_t<CompatibleArrayType>>::value, int> = 0>
+    static void construct(BasicJsonType& j, CompatibleArrayType && arr)
     {
-        auto view = arr;
         j.m_data.m_value.destroy(j.m_data.m_type);
         j.m_data.m_type = value_t::array;
         j.m_data.m_value = value_t::array;
-        for (const auto& x : view)
+        for (auto& x : arr)
         {
             j.m_data.m_value.array->push_back(x);
             j.set_parent(j.m_data.m_value.array->back());
@@ -378,12 +377,28 @@ template < typename BasicJsonType, typename CompatibleArrayType,
                          !is_compatible_object_type<BasicJsonType, CompatibleArrayType>::value&&
                          !is_compatible_string_type<BasicJsonType, CompatibleArrayType>::value&&
                          !std::is_same<typename BasicJsonType::binary_t, CompatibleArrayType>::value&&
-                         !is_basic_json<CompatibleArrayType>::value,
+                         !is_basic_json<CompatibleArrayType>::value
+#if JSON_HAS_RANGES && !defined(__MINGW32__)
+    && !is_compatible_range_view<CompatibleArrayType>::value
+#endif
+                         ,
                          int > = 0 >
 inline void to_json(BasicJsonType& j, const CompatibleArrayType& arr)
 {
     external_constructor<value_t::array>::construct(j, arr);
 }
+
+#if JSON_HAS_RANGES && !defined(__MINGW32__)
+template < typename BasicJsonType, typename T,
+           enable_if_t < is_compatible_range_view<std::remove_cvref_t<T>>::value
+                         && !is_compatible_string_type<BasicJsonType, std::remove_cvref_t<T>>::value
+                         && !is_compatible_object_type<BasicJsonType, std::remove_cvref_t<T>>::value
+                         && !is_basic_json<std::remove_cvref_t<T>>::value, int > = 0 >
+inline void to_json(BasicJsonType& j, T && arr)
+{
+    external_constructor<value_t::array>::construct(j, std::forward<T>(arr));
+}
+#endif
 
 template<typename BasicJsonType>
 inline void to_json(BasicJsonType& j, const typename BasicJsonType::binary_t& bin)
--- a/include/nlohmann/detail/conversions/to_json.hpp
+++ b/include/nlohmann/detail/conversions/to_json.hpp
@@ -178,8 +178,8 @@ struct external_constructor<value_t::array>
 
     template < typename BasicJsonType, typename CompatibleArrayType,
                enable_if_t < !std::is_same<CompatibleArrayType, typename BasicJsonType::array_t>::value
-#if JSON_HAS_RANGES && !defined(__MINGW32__) && (!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE >= 12)
-                             && !std::ranges::view<CompatibleArrayType>
+#if JSON_HAS_RANGES && !defined(__MINGW32__)
+                             && !is_compatible_range_view<CompatibleArrayType>::value
 #endif
                              , int > = 0 >
     static void construct(BasicJsonType& j, const CompatibleArrayType& arr)
@@ -224,9 +224,9 @@ struct external_constructor<value_t::array>
 
     // std::ranges does not work properly on MinGW due to incomplete C++20 support
     // see https://github.com/nlohmann/json/issues/4916
-#if JSON_HAS_RANGES && !defined(__MINGW32__) && (!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE >= 12)
+#if JSON_HAS_RANGES && !defined(__MINGW32__)
     template<typename BasicJsonType, typename CompatibleArrayType,
-             enable_if_t<std::ranges::view<std::remove_cv_t<CompatibleArrayType>>, int> = 0>
+             enable_if_t<is_compatible_range_view<std::remove_cv_t<CompatibleArrayType>>::value, int> = 0>
     static void construct(BasicJsonType& j, const CompatibleArrayType& arr)
     {
         auto view = arr;
--- a/include/nlohmann/detail/meta/type_traits.hpp
+++ b/include/nlohmann/detail/meta/type_traits.hpp
@@ -470,7 +470,7 @@ struct is_compatible_array_type_impl <
 
 // std::ranges does not work properly on MinGW due to incomplete C++20 support
 // see https://github.com/nlohmann/json/issues/4916
-#if JSON_HAS_RANGES && !defined(__MINGW32__)
+#if JSON_HAS_RANGES && !defined(__MINGW32__) && (!defined(_GLIBCXX_RELEASE) || _GLIBCXX_RELEASE >= 12)
 template<typename BasicJsonType, typename CompatibleArrayType>
 struct is_compatible_array_type_impl <
     BasicJsonType, CompatibleArrayType,
--- a/include/nlohmann/detail/meta/type_traits.hpp
+++ b/include/nlohmann/detail/meta/type_traits.hpp
@@ -468,7 +468,9 @@ struct is_compatible_array_type_impl <
         range_value_t<CompatibleArrayType>>::value;
 };
 
-#ifdef JSON_HAS_CPP_20
+// std::ranges does not work properly on MinGW due to incomplete C++20 support
+// see https://github.com/nlohmann/json/issues/4916
+#if JSON_HAS_RANGES && !defined(__MINGW32__)
 template<typename BasicJsonType, typename CompatibleArrayType>
 struct is_compatible_array_type_impl <
     BasicJsonType, CompatibleArrayType,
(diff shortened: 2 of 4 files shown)


federicosfriso05-dotcom:
Hi, I noticed the Windows/MinGW builds are failing. After investigating the CI logs, the issue seems to be related to an incomplete C++20 ranges implementation in MinGW GCC 12.2.0, where `std::ranges::ref_view` and `std::ranges::filter_view` do not work correctly on that toolchain. The same code compiles and runs correctly on GCC/Linux and Clang/Mac.
Should the test be guarded against MinGW with something like `#ifndef __MINGW32__`, or is there a different approach you'd recommend for handling this compiler limitation?
Thanks!

nlohmann:
IIRC we also excluded some other older compilers from other C++20 features. So I guess and ifdef with some specific versions and comments would be sufficient.

federicosfriso05-dotcom:
Updated. Replaced `JSON_HAS_CPP_20` with `JSON_HAS_RANGES && !defined(__MINGW32__)` as suggested, following the existing pattern in macro_scope.hpp

federicosfriso05-dotcom:
I noticed there's still a failure on GCC 11 in `unit-iterators2.cpp` — `iteration_proxy`, an internal nlohmann type, appears to satisfy `std::ranges::view`, causing a circular constraint evaluation error. I'm not sure how to exclude internal types cleanly without knowing the codebase better. Any guidance on the right approach would be appreciated

federicosfriso05-dotcom:
Hi, just checking in after the CI failures. I understand the issue with `iteration_proxy` satisfying `std::ranges::view` — happy to implement any fix you suggest for handling internal types correctly.

nlohmann:
Now that #5161 has merged, a few things:

1. Please rebase on develop: it now conflicts (we touched the same `JSON_HAS_RANGES`/type_traits.hpp areas). After rebasing, your #4916 regression test will sit next to the #4440 one #5161 added; please reconcile so we don't duplicate.

2. Guard the new `#include <ranges>` in unit-regression2.cpp: it's currently unconditional and breaks every pre-C++20 build (GCC 5, Clang 3.x: ranges: No such file or directory). Wrap it the way #5161 does: `#if JSON_HAS_RANGES`.

3. Keep the `!defined(__MINGW32__)` guards: develop's ranges detection still doesn't special-case MinGW, so they're needed.

4. The remaining real blocker is GCC 11. The failure is on `std::ranges::ref_view<basic_json>` - a view over `basic_json` itself - which trips libstdc++ 11's self-referential swappable/movable constraint (atomic constraint depends on itself). [#5161](https://github.com/nlohmann/json/issues/5161) doesn't cover this (it only disables ranges through GCC 10). Simplest fix: gate the view specialization off on libstdc++ 11, e.g. add `_GLIBCXX_RELEASE >= 12` to the `#if`. Excluding `basic_json`-valued views by `range_value_t` would be cleaner conceptually, but may not help here since GCC recurses while merely evaluating `std::ranges::view<ref_view<basic_json>>` — so the version guard is the safe bet.

federicosfriso05-dotcom:
I've rebased and applied all 4 fixes you suggested. Given the `ci_test_compilers_gcc (12)` failure, I guess that the `_GLIBCXX_RELEASE >= 12` does not fully solve the problem. Is there a more specific way to exclude `basic_json`-valued views specifically?
Then, looking at the library code, the library guard (`!defined(__MINGW32__)`) correctly excludes the feature, but `tests/src/unit-regression2.cpp` isn't guarded the same way, so it still tries to instantiate `json const j(filteredNums)` on MinGW.
Happy to follow whatever you suggest.

nlohmann:
Thanks for rebasing! I asked Claude Code to dig into this locally with g++-12 + libstdc++, which reproduces the failure, and I want to save you some cycles because a couple of the natural fixes don't actually work.

The red across the whole gcc/clang matrix isn't your new test — it's `tests/src/unit-iterators2.cpp`, which pipes the library's own `j.items()` (an `iteration_proxy`) and `json` through `std::views::transform`/`reverse`. Your new `is_compatible_array_type_impl` specialization then gets evaluated for views whose value type is `iteration_proxy`, and `is_constructible<BasicJsonType, range_value_t<view>>` re-enters the compatibility machinery → `satisfaction of atomic constraint … depends on itself`. It's a hard error, so it can't be SFINAE'd away after the fact, and it's why `_GLIBCXX_RELEASE >= 12` doesn't help (gcc 12–15 all hit it) — please drop that guard.

Two things I checked with Claude:
- Just excluding `basic_json` value types isn't enough — the poison value type here is `iteration_proxy` (a `detail` type), not `basic_json`. I tried `!is_basic_json<range_value_t>` and it still recurses.
- You can't sidestep the trait either: on baseline, `is_compatible_array_type<filter_view<…>>` is already `false`, so the trait change is genuinely needed.

The crux is that any specialization keyed on `std::ranges::view` / `range_value_t` will be evaluated for the library's internal `iteration_proxy` during `is_constructible<json, …>` and that evaluation itself recurses. So the fix has to (a) avoid the self-referential `is_constructible<json, range_value_t>` in the constraint, and (b) exclude json's own `detail` range types up front with a plain (non-ranges) predicate, before any ranges concept is touched. And the test guard must mirror the feature guard exactly (the MinGW exclusion too), or MinGW stays red.

This one's genuinely fiddly template-wise. The key is to build `unit-iterators2.cpp` with g++-12/libstdc++ locally, since the C++20 standards jobs alone don't surface it.

federicosfriso05-dotcom:
Here's what changed after the last commit:

`type_traits.hpp`:
- Added `is_iteration_proxy_type<T>` to identify nlohmann's internal proxy types before touching any `std::ranges` concept
- Added `is_compatible_range_view<T>` with a two-level `SafeToCheck` guard to prevent circular constraints on GCC 12: excludes `iteration_proxy`/`iteration_proxy_value`, views wrapping them (e.g. `owning_view<iteration_proxy<...>>`), and views wrapping `basic_json` (e.g. `ref_view<json>`)
- Made the iterator-based `is_compatible_array_type_impl` specialization explicitly exclude views to avoid ambiguity with types like `std::string_view` that satisfy both iterator traits and `std::ranges::view`

`to_json.hpp`:
- Added a dedicated `construct()` overload for range views that iterates and pushes each element into the json array

I replaced `_GLIBCXX_RELEASE >= 12` with `JSON_HAS_RANGES && !defined(__MINGW32__)` guard due to incomplete C++20 ranges support on MinGW.

After building `unit-iterators2.cpp` with GCC11, GCC12, Clang 18, MinGW and Apple Clang locally (and then the whole project), all CI checks are green.

Is there something I have to change? Happy to follow any your advice.

include/nlohmann/detail/conversions/to_json.hpp
```diff
@@ -218,6 +221,26 @@ struct external_constructor<value_t::array>
...
     }
+
+    // std::ranges does not work properly on MinGW due to incomplete C++20 support
+    // see https://github.com/nlohmann/json/issues/4916
+#if JSON_HAS_RANGES && !defined(__MINGW32__)
+    template<typename BasicJsonType, typename CompatibleArrayType,
+             enable_if_t<is_compatible_range_view<std::remove_cv_t<CompatibleArrayType>>::value, int> = 0>
+    static void construct(BasicJsonType& j, const CompatibleArrayType& arr)
+    {
+        auto view = arr;
```
nlohmann: This requires copy-constructing the view, but this fails to compile for the most idiomatic C++20 usage: piping directly from a temporary/prvalue container.

MWE:

```cpp
#include <nlohmann/json.hpp>
#include <ranges>
#include <vector>

using json = nlohmann::json;

int main()
{
    // Note: no intermediate variable for the vector -- this is a prvalue.
    json j(std::vector<int>{1, 2, 37, 42, 21}
           | std::views::filter([](int i) { return i > 10; }));
}
```

include/nlohmann/detail/meta/type_traits.hpp
```diff
@@ -461,13 +496,37 @@ struct is_compatible_array_type_impl <
...
         range_value_t<CompatibleArrayType>>::value;
 };
 
+#if JSON_HAS_RANGES && !defined(__MINGW32__)
+template<typename BasicJsonType, typename CompatibleArrayType>
+struct is_compatible_array_type_impl <
+    BasicJsonType, CompatibleArrayType,
+    enable_if_t < is_compatible_range_view<CompatibleArrayType>::value
+    && !std::is_same<detected_t<range_value_t, CompatibleArrayType>, char>::value
+    && !std::is_same<detected_t<range_value_t, CompatibleArrayType>, wchar_t>::value >>
```
nlohmann, line 533: This specialization unconditionally sets `value = true` for any `std::ranges::view` whose element type isn't `char`/`wchar_t`, without checking `is_constructible<BasicJsonType, range_value_t<CompatibleArrayType>>::value` the way the sibling iterator-based specialization does.

MWE:

```cpp
#include <nlohmann/json.hpp>
#include <ranges>
#include <thread>
#include <vector>

using json = nlohmann::json;

int main()
{
    std::vector<std::thread> threads;
    threads.emplace_back([] {});

    auto view = threads | std::views::filter([](const std::thread&) { return true; });
    json j(view); // fails deep inside construct()/push_back(), not at this line
}
```

whereas this is immediately ruled out with "no matching constructor" error

```cpp
#include <nlohmann/json.hpp>
#include <thread>
#include <vector>

using json = nlohmann::json;

int main()
{
    std::vector<std::thread> threads;
    threads.emplace_back([] {});
    json j(threads); // clean "no matching constructor" error
}
```

nlohmann (changes requested):

federicosfriso05-dotcom:
Before pushing the new edits: should I also add a test case for the prvalue/owning_view scenario (e.g. `json j(std::vector<int>{1,2,3} | std::views::filter(...))`)? If I should, it would need a guard for GCC 11 and older versions because there is no owning_view in listdc++ 11. What do you suggest to me?

nlohmann:
> Before pushing the new edits: should I also add a test case for the prvalue/owning_view scenario (e.g. `json j(std::vector<int>{1,2,3} | std::views::filter(...))`)? If I should, it would need a guard for GCC 11 and older versions because there is no owning_view in listdc++ 11. What do you suggest to me?

Good catch - this is actually the case I flagged in the review comment on line 232: `construct()` requires copy-constructing the view, and `owning_view` (which you get from piping a prvalue) is move-only, so it can't compile today. That needs an implementation fix (take the view by value/move rather than copy) before adding the test, not a version-guarded test around the current behavior. Once that's fixed, yes please add the prvalue test — and separately, the line-520 comment about the missing is_constructible check on the view specialization is still open too, worth resolving in the same pass.

federicosfriso05-dotcom:
Thanks for the clarification. 
I've already implemented the forwarding reference fix for `construct()`. However, I noticed that the prvalue test is still failing to compile on GCC 11 because `owning_view` doesn't exist in libstdc++ 11. The CI runs C++20 tests on GCC 11 (`JSON_HAS_RANGES` is enabled there). How should I handle this? After this last clarification I'll push the edits.

nlohmann:
I think then you have to exclude the test with something like:

```cpp
#if JSON_HAS_RANGES && !defined(__MINGW32__) && !(defined(__GLIBCXX__) && _GLIBCXX_RELEASE == 11)
SECTION("issue #4916 - constructing array from prvalue C++20 ranges view (owning_view)")
{
    ...
}
#endif
```

with a short comment noting `owning_view` isn't available in libstdc++ 11. Could you double check whether `== 11` is precise, or whether the gap is wider (e.g. use `< 12` instead)?

federicosfriso05-dotcom:
Both requested changes have been implemented.
- Fix 1 (`type_traits.hpp`): The view specialization of `is_compatible_array_type_impl now checks is_constructible<BasicJsonType, std::ranges::range_value_t<CompatibleArrayType>>::value` instead of unconditionally returning true, consistent with the iterator-based specialization. Used `std::ranges::range_value_t` instead of the plain nlohmann alias because Apple Clang's C++20 iterators don't always define classic `iterator_traits`, which caused the plain alias to return the wrong type.

- Fix 2 (`to_json.hpp`): `construct()` and `to_json()` now take the view by forwarding reference (T&&) instead of copying it, so move-only views like owning_view (produced by piping a prvalue) are handled correctly. `std::forward` is also applied in the range-for loop to satisfy clang-tidy.

A prvalue test has also been added (`json j(std::vector<int>{...} | std::views::filter(...))`), guarded with `!(defined(__GLIBCXX__) && _GLIBCXX_RELEASE < 12)` since `owning_view` is not available in libstdc++ < 12.

include/nlohmann/detail/conversions/to_json.hpp
```diff
@@ -218,6 +221,25 @@ struct external_constructor<value_t::array>
...
+    // see https://github.com/nlohmann/json/issues/4916
+#if JSON_HAS_RANGES && !defined(__MINGW32__)
+    template<typename BasicJsonType, typename CompatibleArrayType,
+             enable_if_t<is_compatible_range_view<std::remove_cvref_t<CompatibleArrayType>>::value, int> = 0>
+    static void construct(BasicJsonType& j, CompatibleArrayType && arr)
+    {
+        j.m_data.m_value.destroy(j.m_data.m_type);
+        j.m_data.m_type = value_t::array;
+        j.m_data.m_value = value_t::array;
+        for (auto& x : std::forward<CompatibleArrayType>(arr))
```
nlohmann: This only works for views whose `operator*` returns an lvalue reference (e.g. `std::views::filter` over a container, which is the only case the tests cover). It fails to compile for views yielding prvalues — notably `std::views::transform` with a by-value-returning callable, which the PR description itself calls out as a supported case:

```cpp
std::vector<int> nums{1, 2, 3};
auto t = nums | std::views::transform([](int i){ return i * 2; });
json j(t);  // fails: non-const lvalue reference cannot bind to a temporary
```

Changing this to `auto &&` should work:

```suggestion
        for (auto&& x : std::forward<CompatibleArrayType>(arr))
```

Please add the example above as test case.

nlohmann (changes requested):
Thanks! I only have one comment.

nlohmann:
Regarding the CI issues: the documentation issue is fixed on develop. The GCC issue is something you need to look into.

federicosfriso05-dotcom:
Firstly, I've just added `noexcept` to the prvalue test since CI adopts GCC 16 with `-Werror=noexcept`. Then, the CI is still failing on GCC 26 because `std::optional` is considered as `std::range::view`, so `to_json(T&&)` catches all `std::range::view`, including `std::optional<std::string>`. The result is a failing test case at line 1792 of `unit-conversions.cpp`, which checks if `json(std::optional<std::string>)` produces `"string"`, but it produces `[string]`. How do you wanna manage this case? Is a specific guard like `!is_optional<T>` in `is_compatible_range_view` or in `to_json(T&&)` enough?

nlohmann:
`!is_optional<T>` is the right instinct, but two implementation notes:

**1. You can't reuse `detail::is_std_optional`.** It's defined in `json.hpp` (around line 100), but `to_json.hpp` is `#include`d before that point (line 47 vs. line 100), so it doesn't exist yet where you'd need it. You'll need a small standalone trait in `type_traits.hpp`, next to `is_iteration_proxy_type` (which this PR already added there for the same kind of reason). Don't add a fresh `#include <optional>` there, though — `is_std_optional` was deliberately kept out of `type_traits.hpp` to avoid disturbing the C++20 module's global-module-fragment include order (see the comment right above its definition in `json.hpp`). You don't actually need to: `to_json.hpp` already includes `<optional>` (line 13) before it includes `type_traits.hpp` (line 29), so `std::optional` is visible by the time this header's content is processed. Just give the new trait a different name than `is_std_optional` so it's clear these are two independent definitions.

**2. Put the guard inside `is_compatible_range_view` itself, not at each call site.** That trait already has a `SafeToCheck` gate excluding `iteration_proxy`/`basic_json`-valued ranges for the GCC-12 circular-constraint issue — folding the optional exclusion in there (rather than adding `!is_optional<T>` separately to `to_json(T&&)`'s `enable_if`) keeps it centralized, so nothing else feeding off that trait needs to remember to add it too.

```cpp
// in type_traits.hpp, near is_iteration_proxy_type
#ifdef JSON_HAS_CPP_17
template<typename T> struct is_range_view_optional_type : std::false_type {};
template<typename T> struct is_range_view_optional_type<std::optional<T>> : std::true_type {};
#else
template<typename T> struct is_range_view_optional_type : std::false_type {};
#endif

template < typename T, bool SafeToCheck =
           !is_iteration_proxy_type<T>::value &&
           !is_iteration_proxy_type<detected_t<range_value_t, T>>::value &&
           !is_basic_json<detected_t<range_value_t, T>>::value &&
           !is_range_view_optional_type<T>::value >
struct is_compatible_range_view : std::false_type {};
```

I applied this locally (plus the `auto&&` fix from earlier in the thread) and re-ran your filter-view test, the prvalue/`owning_view` test, a plain `std::optional<std::string>` conversion, and (since I wanted to make sure this doesn't regress it) a *named lvalue* of a move-only view (`auto ov = vec | views::filter(pred); json j(ov);`) — all four compile and behave correctly with ASan/UBSan on.

I also looked at whether there's a way to avoid a per-type exclusion list altogether (e.g. switching the generic overload from `T&&` to `const T&`, or to by-value) — both looked structurally cleaner at first but each has a real problem: `const T&` makes `const_cast`-ing to get non-const `begin()`/`end()` undefined behavior for your prvalue `owning_view` test specifically (binding a prvalue to a `const&` parameter materializes a genuinely `const`-qualified temporary), and by-value breaks compiling a *named* lvalue of a move-only view since it forces a copy. Your original `T&&` design is actually correct on that front — the only real gap is the overload-hijacking for `optional`, so a targeted exclusion is the right scope for this fix, not a bug in the overall design.

Also worth double-checking against your `ci_module_cpp20` job specifically once pushed, since I can't test the actual C++20 module build locally and the include-order reasoning above is inferred from the header structure rather than verified against that job directly.

nlohmann:
(Sorry for the AI response. I started to reproduce locally, then ran into a road block, asked Claude and then, after trying several "fixes", it concluded your approach is the right one. Wanted to add the reasoning here so it's not in vain.)

federicosfriso05-dotcom:
I tried adding `is_range_view_optional_type` to `type_traits.hpp` as suggested, but it breaks on all compilers (GCC 11, 12, Clang 18) with `'optional' is not a member of 'std'`. The issue is that `type_traits.hpp` is also included via `iteration_proxy.hpp` through paths that don't go through `to_json.hpp`, so `<optional>` is not yet visible at definition time. How should I handle this?

nlohmann:
It seems the CI fails, because `std::optional` is used without including the respective header.

federicosfriso05-dotcom:
All CI checks pass except for `ci_test_compilers_gcc_old` (4.8, 4.9, 5, 6), which fail at the "Install g++" step before any compilation. Is there a way to check them?

nlohmann:
> All CI checks pass except for `ci_test_compilers_gcc_old` (4.8, 4.9, 5, 6), which fail at the "Install g++" step before any compilation. Is there a way to check them?

I restarted the job. I recently moved the old GCC versions from a homebrew Docker image to another image (#5263). If the installation remains flaky, I may revert this.

federicosfriso05-dotcom:
All CI checks are green. Is there anything else needed?

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks for your effort and patience! I'll merge this once the CI is through.

Issue #288, opened 2016-08-04
Silently get numbers into smaller types


vladimirgamalian:
This code:
```cpp
	json j = { { "foo", 65536 } };
	int16_t v = j["foo"].get<int16_t>();
	std::cout << v << "\n";
```
prints
```
0
```
Probably, it would be better, if `get` thrown some exception (maybe `std::out_of_range`).

whackashoe:
If you are grabbing smaller widths then you must know that the numbers will be small.

vladimirgamalian:
@whackashoe yes, but the same rule can be apply to get number from a string, nevertheless we have an exception in that case. And, in most cases, json comes from untrusted sources, so we should check all the values from it. For numbers it can be:

```
		json j = { { "foo", 65536 } };
		json v = j["foo"];
		if (!v.is_number_integer())
			throw;
		int64_t a = v;
		if ((a < std::numeric_limits<int16_t>::min()) || (a > std::numeric_limits<int16_t>::max()))
			throw;
		int16_t b = static_cast<int16_t>(a);
```

whackashoe:
If the data you are getting is untrusted, then you must pepper with checks anyways. May as well just read it into the largest width size number you have then check if you can cast downwards from there. Something expected to be a number but is actually "0234hnnnng111111" is I guess exceptional - it isn't a number at all- whereas rest of C++ historically thinks overflow isn't exceptional. I don't understand why you'd cast untrusted data to small size and assume a library will throw an exception if that doesn't fit.

vladimirgamalian:
@whackashoe I understood the idea. Maybe my experience with another libs, with separate methods (`get` for strict extraction and `to` for best try conversion) misled me.

whackashoe:
It's just my opinion on it, very subjective :dancers:

nlohmann:
[RFC 7159, Section 6](https://tools.ietf.org/html/rfc7159#section-6) states:

> Since software that implements IEEE 754-2008 binary64 (double precision) numbers [IEEE754] is generally available and widely used, good interoperability can be achieved by implementations that expect no more precision or range than these provide, in the sense that implementations will approximate JSON numbers within the expected precision.

and

> Note that when such software is used, numbers that are integers and are in the range [-(2^53)+1, (2^53)-1] are interoperable in the sense that implementations will agree exactly on their numeric values.

[...]

PR #5207, opened 2026-06-09
Fix UBSAN errors when converting types using from_json

### LINKED ISSUE #5206: from_json - Type conversions without range clamping cause UBSan warnings
### Description

https://github.com/InternationalColorConsortium/iccDEV/issues/1230

from_json.hpp has type casting without checking/clamping ranges. Casting from float to int is the most dangerous of these, and triggers UBSan warnings when the float value is outside of the integer range.
```from_json.hpp:356:47: runtime error: 1.11111e+21 is outside the range of representable values of type 'int'```   
(original report appears to be header version 3.11.3 on Linux)

### Reproduction steps

https://github.com/InternationalColorConsortium/iccDEV/issues/1230

Any from_json code that converts from float to int can cause this.
Converting larger int to smaller ints can have the same problem.

```from_json.hpp:356:47: runtime error: 1.11111e+21 is outside the range of representable values of type 'int'```
Expect no UBSan warnings from the json library code.

### Minimal code example
```Shell
#include <cstdio>
#include <cstdint>
#include <cmath>
#include <limits>
#include "json.hpp"
using json = nlohmann::json;

int main(int argc, const char * argv[])
{
    const json::number_float_t float_reference{ std::numeric_limits<json::number_float_t>::max() };
    json j(float_reference);

    int n_int = j.get<int>();
    if (n_int != std::numeric_limits<int>::max())
        fprintf(stderr,"ERROR - int from float failed! Got %d instead of %d\n", n_int, std::numeric_limits<int>::max());

    uint64_t n_int64 = j.get<uint64_t>();
    if (n_int64 != std::numeric_limits<uint64_t>::max())
        fprintf(stderr,"ERROR - uint64 from float failed! Got %llu instead of %llu\n", n_int64, std::numeric_limits<uint64_t>::max());
    
    // Inf and NaN fail with undefined behavior
    // integer overrange also fails with bit aliasing behavior (might as well be random)
    // See the updated unit-conversions.cpp for the fill list of problems and tests
    return 0;
}
```

### Error messages

```Shell
/Volumes/Development/json_ubsan_range_errors/json.hpp:5344:47: runtime error: 1.79769e+308 is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Volumes/Development/json_ubsan_range_errors/json.hpp:5344:47 in 
ERROR - int from float failed! Got -2147483648 instead of 2147483647
/Volumes/Development/json_ubsan_range_errors/json.hpp:5007:47: runtime error: 1.79769e+308 is outside the range of representable values of type 'unsigned long long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Volumes/Development/json_ubsan_range_errors/json.hpp:5007:47 in 
ERROR - uint64 from float failed! Got 9223372036854775808 instead of 18446744073709551615
```

Clang, MacOS
GCC, Linux - but the bug is not compiler or platform specific

### Library version
3.12.0

### Validation

ChrisCoxArt:
Fixes  #5206

from_json.hpp
- Clamp ranges when converting from internal to external types.
- Flush NaN and Inf to 0 when converting float to int, preserve when converting to another float.
- Use compile time tests to remove the clamping when not necessary.

unit-conversions.cpp
- Test range problems (overflow and underflow) for integer and float types.
- Test +-Infinity and NaN conversions.

### Checklist
    I'm not sure about this - this change doesn't really fit the current number_handling.md documentation.
- [x] Make sure all ctests are run and passing.

### Note
DCO does not appear to know about digital signatures - it should be removed or replaced with smarter tests.

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -105,11 +105,13 @@ void get_arithmetic_value(const BasicJsonType& j, ArithmeticType& val)
         case value_t::number_float:
         {
             auto temp = *j.template get_ptr<const typename BasicJsonType::number_float_t*>();
-            if (!std::isfinite(temp))   // flush NaN and Inf to zero
+            // flush NaN and Inf to zero before converting to a type that cannot represent them
+            bool valueIsFinite = std::isfinite(temp);
+            if (!std::numeric_limits<ArithmeticType>::has_infinity && !valueIsFinite)
                 val = 0;
-            else if (temp < std::numeric_limits<ArithmeticType>::lowest())
+            else if (valueIsFinite && temp < std::numeric_limits<ArithmeticType>::lowest())
                 val = std::numeric_limits<ArithmeticType>::lowest();
-            else if (temp > std::numeric_limits<ArithmeticType>::max())
+            else if (valueIsFinite && temp > std::numeric_limits<ArithmeticType>::max())
                 val = std::numeric_limits<ArithmeticType>::max();
             else
                 val = static_cast<ArithmeticType>(temp);
@@ -458,11 +460,13 @@ inline void from_json(const BasicJsonType& j, ArithmeticType& val)
         case value_t::number_float:
         {
             auto temp = *j.template get_ptr<const typename BasicJsonType::number_float_t*>();
-            if (!std::isfinite(temp))   // flush NaN and Inf to zero
+            // flush NaN and Inf to zero before converting to a type that cannot represent them
+            bool valueIsFinite = std::isfinite(temp);
+            if (!std::numeric_limits<ArithmeticType>::has_infinity && !valueIsFinite)
                 val = 0;
-            else if (temp < std::numeric_limits<ArithmeticType>::lowest())
+            else if (valueIsFinite && temp < std::numeric_limits<ArithmeticType>::lowest())
                 val = std::numeric_limits<ArithmeticType>::lowest();
-            else if (temp > std::numeric_limits<ArithmeticType>::max())
+            else if (valueIsFinite && temp > std::numeric_limits<ArithmeticType>::max())
                 val = std::numeric_limits<ArithmeticType>::max();
             else
                 val = static_cast<ArithmeticType>(temp);
@@ -470,8 +474,7 @@ inline void from_json(const BasicJsonType& j, ArithmeticType& val)
         }
         case value_t::boolean:
         {
-            // comparison forces the value to be 0 or 1, which is in range for all types
-            auto temp = (0 != (*j.template get_ptr<const typename BasicJsonType::boolean_t*>()));
+            auto temp = *j.template get_ptr<const typename BasicJsonType::boolean_t*>();
             val = static_cast<ArithmeticType>(temp);
             break;
         }
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -19,6 +19,8 @@
 #include <unordered_map> // unordered_map
 #include <utility> // pair, declval
 #include <valarray> // valarray
+#include <cmath> // isfinite
+#include <limits> // max, lowest
 
 #include <nlohmann/detail/exceptions.hpp>
 #include <nlohmann/detail/macro_scope.hpp>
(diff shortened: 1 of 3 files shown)


gregmarr:
There was a previous discussion of this about 10 years ago [in #288](https://github.com/nlohmann/json/issues/288#issuecomment-237757373).  I don't know if his position has changed since then, but at the time it was that if you are extracting to a smaller data type than what is being used internally to hold the data value, then it is your responsibility to perform the range checking, not the library's, and thus people that extract it to the proper type don't pay the cost of range checking.

ChrisCoxArt:
Again:
You literally can't do the range checking after the fact, unless you extract everything to the exact type used inside the library, then do all the work yourself... at which point, why is the library creating errors and more work for everyone instead of doing the right thing? Yes, it is 100% the library's responsibility to not cause undefined behavior errors.

gregmarr:
> You literally can't do the range checking after the fact, unless you extract everything to the exact type used inside the library, then do all the work yourself...

Yes, that is exactly how one does range checking when one is going to fit an unknown value from a larger datatype into a smaller datatype.

> at which point, why is the library creating errors and more work for everyone instead of doing the right thing? 

The library could limit the extraction function to just the internal types, but it does more as a convenience for the user.  If the user cares about fitting larger values into smaller types, then it is the user's responsibility to do the checking to make sure it fits.

In general, the library follows the "don't pay for what you don't use" philosophy.  If you know that you don't need to do range checking, then you can call the function that doesn't do range checking.  If you do need to do range checking because you absolutely need a smaller datatype than what is used internally, do the range checking yourself before putting the data in the smaller datatype.

> Yes, it is 100% the library's responsibility to not cause undefined behavior errors.

It is not causing them.  If you are 100% certain that the value will fit in the smaller datatype, then you can use the function that allows you to extract it into the smaller datatype.  If you are not 100% certain, extract it as the native datatype and do the range checking yourself.

I am not the maintainer, just a long time user, and in general that has been the philosophy of the library.  As the maintainer, @nlohmann is definitely able to make changes, but if those changes can cause behavior differences, then it is less likely that they will be accepted.

ChrisCoxArt:
Wow. Your (almost Lewis Carroll like) "logic" belongs in a museum somewhere.

Yes, the nlohmann json code is 100% causing the undefined behavior errors because of type conversions without range testing/clamping. Fixing that requires changes to the library code, or re-implementing chunks of the library code by every user of the library. Yes, fixing the errors in the library will cause behavior changes - because it will no longer cause errors, or return random unexpected values.

This is JSON we're talking about - you can never be 100% certain of the data range unless you just wrote that data and only immediately consume your own internal data. As soon as storage is involved -- data can be changed, and ranges are undefined. As soon as third party data is involved -- ranges are undefined. As soon as end-users are involved -- ranges are undefined.   If you assume that the data is always in the range you created, then you have created attack vectors and vulnerabilities in your code -- which is why tools like UBSan exist to expose those errors and vulnerabilities.

Pushing responsibility for the errors in the library onto the users of the library is irresponsible, at best.

I certainly hope that the maintainer has a better grasp on the world than you seem to be conveying.

gregmarr:
The fix is very simple, extract using the internal datatypes, and then you don't have to worry about introducing undefined behavior using the library functions.  What you might want for range checking behavior isn't necessarily the same as someone else wants for range checking behavior, and it can vary from element to element.

>This is JSON we're talking about - you can never be 100% certain of the data range unless you just wrote that data and only immediately consume your own internal data. 

And yet there are users of this library that are 100% certain of the data that they are reading, or they ALREADY do their own range checking, and thus for performance reasons absolutely do not want the extra range checking.

ChrisCoxArt:
No, pushing the fix onto users of the library means re-implementing parts of the library to do the right thing. Why should the user have to re-implement parts of the library? Why can't the library do the right thing in the first place?  Why ship a library with known UB errors?  Can users trust the library code if maintainers think that UB shouldn't be fixed and should require third party code to replace parts of the library?

Your "logic" is beyond broken, and you are defending bad code and bad practices.

gregmarr:
Please define "the right thing" for every possible set of circumstances experienced by every user of this library.

ChrisCoxArt:
You can start with "don't generate errors from UBSan", and "don't return random unexpected values".

Please stop defending broken code, and start working on how to fix it.

nlohmann:
I stand with #288. The library has a well-defined and documented behavior of storing data. When you decide to ask for conversions like `get<int>()`, it is not the library's duty to check if this conversion makes sense - this is also [documented](https://json.nlohmann.me/api/basic_json/get/).

Also, it's [documented](https://json.nlohmann.me/features/types/number_handling/) how the library is handling number types. There is no surprise, and we have functions to query the stored type. This means there is some caution and work needed by clients when numbers are read from the library - but all this is possible by calling public library functions.

Finally, I would like to mention that the the UBSAN errors mentioned here are not caused by the library, but by client code making the wrong assumptions. It's basically an example of casting `double` to `int` via a `static_cast` and complaining about cases where this conversion does not make sense.

ChrisCoxArt:
I strongly disagree. The library is causing the UBsan errors, because the library exposes functionality that has undefined behavior and fails to take precautions against that undefined behavior.
This is 100% a bug in the json library code.
This is 100% a security vulnerability in the library code that affects most clients (returning random values based on byte aliasing when returning smaller types).
Expecting all clients to re-implement library functionality just to avoid UB and security problems is extremely irresponsible.

PR #5211, opened 2026-06-17
Fix integer comparison bug

### LINKED ISSUE #5210: Mixed comparison bug: number_unsigned (INT64_MAX+1~UINT64_MAX) vs number_integer
### Description

Comparisons between `number_unsigned` and `number_integer` can return incorrect results when the unsigned value is greater than `INT64_MAX`.

I consider this a bug because the library returns mathematically incorrect results for `==` and `<` on valid JSON number values. The issue is reproducible through multiple construction paths (`json` direct construction, object assignment, array insertion, initializer lists, nested access, and `json::parse(...)`), so it appears to be in the core comparison implementation rather than in a specific API entry point.

The problematic code appears to be in `include/nlohmann/json.hpp` inside `JSON_IMPLEMENT_OPERATOR`, especially these mixed signed/unsigned branches:

```cpp
else if (lhs_type == value_t::number_unsigned && rhs_type == value_t::number_integer)
{
    return static_cast<number_integer_t>(lhs.m_data.m_value.number_unsigned) op rhs.m_data.m_value.number_integer;
}
else if (lhs_type == value_t::number_integer && rhs_type == value_t::number_unsigned)
{
    return lhs.m_data.m_value.number_integer op static_cast<number_integer_t>(rhs.m_data.m_value.number_unsigned);
}
```

When `number_unsigned` is in `[INT64_MAX + 1, UINT64_MAX]`, converting it to `number_integer_t` changes the value into the negative signed range and breaks comparison semantics.

Relevant code:
- `include/nlohmann/json.hpp`
- mixed comparison branches in `JSON_IMPLEMENT_OPERATOR`

### Reproduction steps

1. Create one `json` value from a `uint64_t` greater than `INT64_MAX`.
2. Create another `json` value from `-1`.
3. Compare them using `==` and `<`.
4. Observe that the results are mathematically incorrect.

Expected:

- `json(uint64_t_value) == json(-1)` should always be `false`
- `json(-1) < json(uint64_t_value)` should always be `true`

Actual:

- For values above `INT64_MAX`, `json(-1) < json(uint64_t_value)` becomes `false`
- For `UINT64_MAX`, `json(UINT64_MAX) == json(-1)` becomes `true`

Examples observed:

- `json(uint64_t(9223372036854775808ULL)) == json(-1)` -> `false`
- `json(-1) < json(uint64_t(9223372036854775808ULL))` -> `false` (should be `true`)
- `json(uint64_t(18446744073709551615ULL)) == json(-1)` -> `true` (should be `false`)
- `json(-1) < json(uint64_t(18446744073709551615ULL))` -> `false` (should be `true`)

### Minimal code example

```Shell
#include <cstdint>
#include <iostream>
#include <limits>
#include <nlohmann/json.hpp>

using json = nlohmann::json;

int main()
{
    const json a = static_cast<std::uint64_t>((std::numeric_limits<std::int64_t>::max)()) + 1ULL;
    const json b = (std::numeric_limits<std::uint64_t>::max)();
    const json neg = -1;

    std::cout << std::boolalpha;
    std::cout << "a == -1: " << (a == neg) << '\n';
    std::cout << "-1 < a : " << (neg < a) << '\n';
    std::cout << "b == -1: " << (b == neg) << '\n';
    std::cout << "-1 < b : " << (neg < b) << '\n';
}


Observed output:


a == -1: false
-1 < a : false
b == -1: true
-1 < b : false
```

### Error messages

```Shell
There is no compilation error or exception. This is a runtime correctness issue: comparison operators return incorrect boolean results.
```

- Operating system: Linux - Compiler: GCC / g++ - Language standard: C++17

### Library version

 - Version observed in this repository checkout: `3.12.0` - Source: GitHub repository checkout

### Validation

ljcjclljc:
This PR provides a minimal implementation for #5210.
Fiex #5210 
## Background

When comparing `number_unsigned` and `number_integer`, the current implementation converts the unsigned value to `number_integer_t` before applying the operator.

This is problematic for values above `INT64_MAX`, because converting such `uint64_t` values to `number_integer_t` changes them into the negative signed range. As a result, valid mixed signed/unsigned comparisons can return mathematically incorrect results.

## Problem

The affected logic is in the mixed comparison branches of `JSON_IMPLEMENT_OPERATOR` in `include/nlohmann/json.hpp`.

Before this change, comparisons such as these could fail:

- `json(-1) < json(18446744073709551615ULL)` returned `false`
- `json(18446744073709551615ULL) == json(-1)` returned `true`

This is not limited to one construction path. The same incorrect behavior can be observed after direct construction, object assignment, array insertion, initializer-list construction, nested access, and `json::parse()`.

## Solution

This PR keeps the existing macro shape and applies the smallest possible fix in the mixed `number_unsigned` / `number_integer` comparison branches:

- If the signed operand is negative, return the mathematically correct ordering result directly.
- If the signed operand is non-negative, cast it to `number_unsigned_t` and compare in the unsigned domain.

This avoids the unsafe `uint64_t -> int64_t` conversion while preserving the existing comparison structure.

## Scope

This PR is intentionally minimal:

- no refactor of the comparison macro
- no change to the public API
- no unrelated behavior changes

It only fixes the signed/unsigned mixed comparison bug described in #5210.

- [x] If applicable, an `https://github.com/nlohmann/json/issues` is referenced.
- [x] The `https://coveralls.io/github/nlohmann/json` remained at 100%. A test case for every new line of code.
- [x] If applicable, the `https://json.nlohmann.me` is updated.


include/nlohmann/json.hpp
```diff
@@ -3686,14 +3690,22 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     {                                                                                                    \
         return lhs.m_data.m_value.number_float op static_cast<number_float_t>(rhs.m_data.m_value.number_unsigned);     \
     }                                                                                                    \
-    else if (lhs_type == value_t::number_unsigned && rhs_type == value_t::number_integer)                \
+   else if (lhs_type == value_t::number_unsigned && rhs_type == value_t::number_integer)                \
```
gregmarr: You lost a space here.

include/nlohmann/json.hpp
```diff
@@ -3686,14 +3690,22 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
     {                                                                                                    \
-        return static_cast<number_integer_t>(lhs.m_data.m_value.number_unsigned) op rhs.m_data.m_value.number_integer; \
+        if (rhs.m_data.m_value.number_integer < 0)                                                       \
+        {                                                                                                \
+            return number_integer_t(1) op number_integer_t(0);                                           \
+        }                                                                                                \
+        return lhs.m_data.m_value.number_unsigned op static_cast<number_unsigned_t>(rhs.m_data.m_value.number_integer); \
```
gregmarr: This is a good change, aligns more with the C++ promotion rules.

gregmarr (commented):
This seems reasonable at a low level.  There may be some cases that should be documented where `a < b` and `json(a) < json(b)` will provide different results.  Those are the same cases where the compiler should have been warning you not to do that `a < b`, so that's probably okay, but good to call it out explicitly.

tests/src/uint64_compare_test.cpp
```diff
@@ -0,0 +1,19 @@
...
+    const json a = static_cast<std::uint64_t>((std::numeric_limits<std::int64_t>::max)()) + 1ULL;
+    const json b = (std::numeric_limits<std::uint64_t>::max)();
+    const json neg = -1;
+
+    std::cout << std::boolalpha;
+    std::cout << "a == -1: " << (a == neg) << '\n';
+    std::cout << "-1 < a : " << (neg < a) << '\n';
+    std::cout << "b == -1: " << (b == neg) << '\n';
+    std::cout << "-1 < b : " << (neg < b) << '\n';
+}
```
nlohmann: Please move the test to `unit-comparison.cpp` and add a section there like

```cpp
SECTION("signed/unsigned mixed comparison above INT64_MAX")
```

and add `CHECK` macros for the tests.

nlohmann (changes requested):

include/nlohmann/detail/conversions/from_json.hpp
gregmarr, line 562: Should this be part of this PR, or separate?  I am leaning towards separate.

include/nlohmann/detail/conversions/to_json.hpp
gregmarr, line 450: Should this be part of this PR, or separate?  I am leaning towards separate.

include/nlohmann/json.hpp
```diff
@@ -3686,14 +3690,22 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
     }                                                                                                    \
-    else if (lhs_type == value_t::number_integer && rhs_type == value_t::number_unsigned)                \
+    else if (lhs_type == value_t::number_integer && rhs_type == value_t::number_unsigned) /* NOLINT(readability/braces) */ \
```
gregmarr: Rather than disabling the lint warning, why not just add the parens and make it clear what the precedence is?

include/nlohmann/json.hpp
```diff
@@ -3630,6 +3630,10 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
 
     // note parentheses around operands are necessary; see
     // https://github.com/nlohmann/json/issues/1530
+    // Mixed signed/unsigned integer comparisons check for negative signed
+    // values before casting. This avoids wraparound when converting a negative
+    // integer to an unsigned type and preserves the rule that any negative
+    // integer is smaller than any unsigned integer.
```
gregmarr: Since we probably don't want to put a comment inside the macro explaining the actual op, should there be one here?

```
  // This is done by checking for the signed value being less than zero.
  // If it is, the comparison is performed using `0` for the signed value
  // and `1` for the unsigned value.  Since the negative signed value will
  // always be less than the unsigned value, any two fixed values that 
  // have the same relationship will produce the right result.
  // If the signed value is not less than zero, it is cast to unsigned
  // before doing the comparison.
```

I'm wondering if `-1` and `1` would be clearer than `0` and `1`.  That way you can explicitly see the negative sign.

include/nlohmann/json.hpp
```diff
@@ -3686,14 +3690,22 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     {                                                                                                    \
         return lhs.m_data.m_value.number_float op static_cast<number_float_t>(rhs.m_data.m_value.number_unsigned);     \
     }                                                                                                    \
-    else if (lhs_type == value_t::number_unsigned && rhs_type == value_t::number_integer)                \
+    else if (lhs_type == value_t::number_unsigned && rhs_type == value_t::number_integer) /* NOLINT(readability/braces) */ \
```
gregmarr: Rather than disabling the lint warning, why not just add the parens and make it clear what the precedence is?

gregmarr (commented):
IMO, this needs a lot more documentation, both in the main documentation section, and somewhere that will show up in the changes section of the readme as a potentially breaking change, with the appropriate rationale, as described in the linked issue.

include/nlohmann/detail/conversions/from_json.hpp
nlohmann: Yes, please separate this from the original PR scope.

include/nlohmann/detail/conversions/to_json.hpp
nlohmann: Yes, please separate this from the original PR scope.

include/nlohmann/json.hpp
nlohmann: Yes, please fix this.

include/nlohmann/json.hpp
nlohmann: Yes, please fix this.

nlohmann (commented):
Some comments. Will review later.

tests/src/uint64_compare_test.cpp
```diff
@@ -0,0 +1,19 @@
+#include <cstdint>
```
nlohmann: This file can now be removed given we have `tests/src/unit-comparison.cpp`.

include/nlohmann/detail/conversions/from_json.hpp
nlohmann: This change is already on `develop` branch. Please rebase.

include/nlohmann/detail/conversions/to_json.hpp
nlohmann: This change is already on `develop` branch. Please rebase.

tests/src/unit-comparison.cpp
```diff
@@ -255,6 +257,23 @@ TEST_CASE("lexicographical comparison operators")
             {f_, f_, f_, f_, f_, f_, f_, f_, f_, f_, f_, f_, f_, f_, f_, f_, f_, f_, f_, f_, f_, f_}, // 21
         };
 
+        SECTION("signed/unsigned mixed comparison above INT64_MAX")
+        {
+            const json above_int64_max = static_cast<std::uint64_t>((std::numeric_limits<std::int64_t>::max)()) + 1ULL;
+            const json max_uint64 = (std::numeric_limits<std::uint64_t>::max)();
+            const json negative_one = -1;
+
+            CHECK_FALSE(above_int64_max == negative_one);
```
nlohmann, line 268: The tests only check `==` and `<` but not `<=`, `>`, `>=`, `!=`, or `operator<=>`, and don't cover the non‑negative‑signed + large‑unsigned path. Please add these.

nlohmann (changes requested):

nlohmann (approved):
Looks good to me.

nlohmann:
For the release notes, I would add the following to the changelog:

> Fixed incorrect results when comparing a `number_unsigned` value greater than `INT64_MAX` against a `number_integer` value. Mixed signed/unsigned comparisons previously cast the unsigned operand to the signed type, wrapping large values into the negative range (e.g. `json(-1) < json(18446744073709551615ULL`) returned `false`, and `json(18446744073709551615ULL) == json(-1)` returned `true`). Comparisons are now performed in a domain that preserves the mathematical result ([#5210](https://github.com/nlohmann/json/issues/5210), [#5211](https://github.com/nlohmann/json/pull/5211)).

Furthermore, I would add somewhere notable at the top:

> Comparison operators (`==`, `!=`, `<`, `<=`, `>`, `>=`, `<=>`) now return the mathematically correct result when a `number_unsigned` value above `INT64_MAX` is compared with a negative `number_integer`. Values ≤ `INT64_MAX` are unaffected. Code that relied on the previous (incorrect) ordering — e.g. basic_json values used as keys in ordered containers — may observe different ordering or equality after upgrading.

@gregmarr @ljcjclljc Are you OK with this?

@gregmarr Anything you would like to add or discuss for this PR - https://github.com/nlohmann/json/pull/5211#discussion_r3506918142 is still open.

ljcjclljc:
I fully agree with the wording you drafted for the changelog and top highlight notice. It clearly explains the root cause, wrong examples, new correct comparison logic, and reminds users about possible ordering changes after upgrade. No further revisions or discussions needed from my side for PR #5211.

nlohmann:
Thanks!

gregmarr (commented):
LGTM

Issue #4996, opened 2025-11-14
Tests don't build with VS 2026


t-b:
### Description

no text

### Reproduction steps

```
cmake -S . -B build -G "Visual Studio 18 2026" -A x64
cmake --build build
```

### Expected vs. actual results

no text

### Minimal code example

```Shell
no text
```

### Error messages

```Shell
E:\projekte\json>cmake -S . -B build -G "Visual Studio 18 2026" -A x64
-- Selecting Windows SDK version 10.0.26100.0 to target Windows 10.0.26200.
-- The CXX compiler identification is MSVC 19.50.35717.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/18/Community/VC/Tools/MSVC/14.50.35717/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Using the multi-header code from E:/projekte/json/include/
-- Operating system: Windows-10.0.26200; MSYS_NT-10.0-26200 DESKTOP-A4CHLC0 3.6.5-22c95533.x86_64 2025-10-10 12:02 UTC x86_64 Msys
-- Compiler: Microsoft (R) C/C++-Optimierungscompiler Version 19.50.35717 für x64; Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.; ; Syntax: cl [ Option... ] Dateiname... [ /link Linkeroption... ]; MSVC_VERSION=1950; MSVC_TOOLSET_VERSION=145
-- C++ standard library: Microsoft C++ Standard Library (MSVC STL), _MSVC_STL_VERSION=145
-- Testing standards: 11 14 17 20 23
-- Looking for C++ include sys/types.h
-- Looking for C++ include sys/types.h - found
-- Looking for C++ include stdint.h
-- Looking for C++ include stdint.h - found
-- Looking for C++ include stddef.h
-- Looking for C++ include stddef.h - found
-- Check size of size_t
-- Check size of size_t - done
-- Configuring done (5.1s)
-- Generating done (0.4s)
-- Build files have been written to: E:/projekte/json/build



  unit-conversions.cpp
E:\projekte\json\tests\src\unit-conversions.cpp(1667,36): error C2672: "nlohmann::json_abi_v3_12_0::basic_json<std::map
[...]

PR #5215, opened 2026-06-21
Minor: unique_ptr template resolution workaround for MSVC

### LINKED ISSUE #5002: VS2026 Insiders, C2678 With C++23 Modules
### Description

When using the Module build of  nlohmann::json I get the following error:
`binary '!=': no operator found which takes a left-hand operand of type 'nullptr' (or there is no acceptable conversion)`
This occurs at line 20512 of the current dev branch of json.hpp.

```'
module;
#include <cstdint>
#include <string>
//#include <sstream> // TJT: Note: Without this include, VS2026 will not complile the toJson() 

export module Dnn.Modules.Softmax:Config;

import Dnn.ModuleConfig;
import nlohmann.json;

namespace Mila::Dnn
{
    using json = nlohmann::json;
// Later

json toJson() const
{
    json j;
    j["name"] = name_;
    //j["precision"] = static_cast<int>( precision_ );
    //j["axis"] = axis_;

    return j;
}`
```

With the included <sstream> (Another similar module had this include and compiled fine AND tested fine ) there is no compilation error.

I'll look into this more but I am busy for the rest of the day

### Reproduction steps

not having the include 
`#include <sstream>`
triggers the compilation error c2678

The include essentially is providing a needed dependency that is missing in the json.hpp module based code.

### Minimal code example

```Shell

```

### Error messages

```Shell

```

Win 11, VS2026

### Library version

Dev branch, Module build

### Validation

drcosmin:
Fixes #5002 

Very similar to #4996, template resolution is the current MSVC compiler is broken and, until MS fixes it, workarounds are the only viable fixes.

In this case, we can rely on unique_ptr's implicit conversion to bool to achieve the same thing

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -400,7 +400,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         };
         std::unique_ptr<T, decltype(deleter)> obj(AllocatorTraits::allocate(alloc, 1), deleter);
         AllocatorTraits::construct(alloc, obj.get(), std::forward<Args>(args)...);
-        JSON_ASSERT(obj != nullptr);
+        JSON_ASSERT(obj);
         return obj.release();
     }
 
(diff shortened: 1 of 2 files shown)


nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #5216, opened 2026-06-27
Make MSVC Release build build binaries with default Release optimization.

risa2000:
MSVC `Release` build of library tests disabled all optimizations by using `/Od` compiler flag. This produced the binaries that despite having all debug helpers and infos disabled, still produced "debug friendly" and totally unoptimized code.

The patch changes two things:
 * removes forced /Od flag in MSVC Release builds
 * adds /wd4702 to disable `warning C4702: unreachable code` in MSVC Release build, which would otherwise prevent the tests to run on CI which enforces "treat warnings as errors" policy.


nlohmann (approved):
Looks good to me.

nlohmann:
Will merge once the pipeline is green (I had to restart an unrelated job).

nlohmann:
Thanks!

PR #5218, opened 2026-06-28
Overwork project infrastructure

### LINKED ISSUE #4354: (official) example for parsing "custom data source" (implementing input iterator) does not compile
### Description

The example in [the current section "Custom data source"](https://github.com/nlohmann/json/blob/8c391e04fe4195d8be862c97f38cfe10e2a3472e/README.md#custom-data-source) does not compile with [the current header](https://github.com/nlohmann/json/blob/8c391e04fe4195d8be862c97f38cfe10e2a3472e/single_include/nlohmann/json.hpp). I don't know if the example or the library is erroneous (I suspect the former). But even if it is just an example I would expect it to compile - or to have an explanation of the necessary steps to make it work.

<sup>Current refers to the [current `develop`](https://github.com/nlohmann/json/tree/8c391e04fe4195d8be862c97f38cfe10e2a3472e) branch.</sup>

### Reproduction steps

To reproduce:

1. Copy the code into a source file.
2. Add the missing `#include` and `using namespace`.
3. Optionally add dummy definition for `MyContainer::advance()` and `MyContainer::get_current()` (does not make any difference at this stage)
4. Try to compile.

See [result on Compiler Explorer with clang 15.0.0](https://godbolt.org/z/PKTGj9Mox).

I would expect the code to compile (without linking) without errors.

Note: It compiles (not linking) if this call to the parser is used instead:

```diff
-    json j = json::parse(c);
+    json j = json::parse(begin(c), end(c));
```

But the example is about using a container.

### Minimal code example

```Shell
++
#include <nlohmann/json.hpp>

using namespace nlohmann;

struct MyContainer {
  void advance();
  const char& get_current();
};

struct MyIterator {
    using difference_type = std::ptrdiff_t;
    using value_type = char;
    using pointer = const char*;
    using reference = const char&;
    using iterator_category = std::input_iterator_tag;

    MyIterator& operator++() {
        target->advance();
        return *this;
    }

    bool operator!=(const MyIterator& rhs) const {
        return rhs.target != target;
    }

    reference operator*() const {
        return target->get_current();
    }

    MyContainer* target = nullptr;
};

MyIterator begin(MyContainer& tgt) {
    return MyIterator{&tgt};
}

MyIterator end(const MyContainer&) {
    return {};
}

void foo() {
    MyContainer c;
    json j = json::parse(c);
}
```

### Error messages

```Shell
In file included from <source>:1:
/app/raw.githubusercontent.com/nlohmann/json/8c391e04fe4195d8be862c97f38cfe10e2a3472e/single_include/nlohmann/json.hpp:23324:16: error: no matching function for call to 'input_adapter'
        parser(detail::input_adapter(std::forward<InputType>(i)), cb, allow_exceptions, ignore_comments).parse(true, result);
               ^~~~~~~~~~~~~~~~~~~~~
<source>:43:20: note: in instantiation of function template specialization 'nlohmann::basic_json<>::parse<MyContainer &>' requested here
    json j = json::parse(c);
                   ^
/app/raw.githubusercontent.com/nlohmann/json/8c391e04fe4195d8be862c97f38cfe10e2a3472e/single_include/nlohmann/json.hpp:6563:27: note: candidate function not viable: no known conversion from 'MyContainer' to 'std::FILE *' (aka '_IO_FILE *') for 1st argument
inline file_input_adapter input_adapter(std::FILE* file)
                          ^
/app/raw.githubusercontent.com/nlohmann/json/8c391e04fe4195d8be862c97f38cfe10e2a3472e/single_include/nlohmann/json.hpp:6568:29: note: candidate function not viable: no known conversion from 'MyContainer' to 'std::istream &' (aka 'basic_istream<char> &') for 1st argument
inline input_stream_adapter input_adapter(std::istream& stream)
                            ^
/app/raw.githubusercontent.com/nlohmann/json/8c391e04fe4195d8be862c97f38cfe10e2a3472e/single_include/nlohmann/json.hpp:6573:29: note: candidate function not viable: no known conversion from 'MyContainer' to 'std::istream' (aka 'basic_istream<char>') for 1st argument
inline input_stream_adapter input_adapter(std::istream&& stream)
                            ^
/app/raw.githubusercontent.com/nlohmann/json/8c391e04fe4195d8be862c97f38cfe10e2a3472e/single_include/nlohmann/json.hpp:6556:109: note: candidate template ignored: substitution failure [with ContainerType = MyContainer]: no type named 'adapter_type' in 'nlohmann::detail::container_input_adapter_factory_impl::container_input_adapter_factory<MyContainer>'
typename container_input_adapter_factory_impl::container_input_adapter_factory<ContainerType>::adapter_type input_adapter(const ContainerType& container)
                                                                                               ~~~~~~~~~~~~ ^
/app/raw.githubusercontent.com/nlohmann/json/8c391e04fe4195d8be862c97f38cfe10e2a3472e/single_include/nlohmann/json.hpp:6589:32: note: candidate template ignored: requirement 'std::is_pointer<MyContainer>::value' was not satisfied [with CharT = MyContainer]
contiguous_bytes_input_adapter input_adapter(CharT b)
                               ^
/app/raw.githubusercontent.com/nlohmann/json/8c391e04fe4195d8be862c97f38cfe10e2a3472e/single_include/nlohmann/json.hpp:6597:6: note: candidate template ignored: could not match 'T[N]' against 'MyContainer'
auto input_adapter(T (&array)[N]) -> decltype(input_adapter(array, array + N)) // NOLINT(cppcoreguidelines-avoid-c-arrays,hicpp-avoid-c-arrays,modernize-avoid-c-arrays)
     ^
/app/raw.githubusercontent.com/nlohmann/json/8c391e04fe4195d8be862c97f38cfe10e2a3472e/single_include/nlohmann/json.hpp:6522:69: note: candidate function template not viable: requires 2 arguments, but 1 was provided
typename iterator_input_adapter_factory<IteratorType>::adapter_type input_adapter(IteratorType first, IteratorType last)
                                                                    ^
1 error generated.
Compiler returned: 1
```

x86-64 clang 15.0.0

### Library version

8c391e04fe4195d8be862c97f38cfe10e2a3472e

### Validation

### LINKED ISSUE #5012: error_handler_t::ignore documentation is incorrect
### Description

```
ignore invalid UTF-8 sequences; all bytes are copied to the output unchanged
```

`all bytes` should be `all valid bytes`, as the former implies invalid UTF-8 characters are copied.

### Reproduction steps

Reading causes potential misunderstanding.

`all bytes` should be `all valid bytes`

### Minimal code example

```Shell

```

### Error messages

```Shell

```

eyes

### Library version

latest

### Validation

nlohmann:
Used Claude Code to review the project infrastructure. Here are the fixes:

- Avoid compiling CMake in every Cirrus CI run.
- Add `persist-credentials: false` to all read-only jobs.
- Pinned uses hash where possible.
- CMake cleanup.
- Add missing C++26 detection for tests.
- Update CI compiler matrix in documentation.
- Improve coverage for module smoke test.
- Build the module smoke test with `import std;` so it compiles under GCC, Clang, and MSVC.
- Fix a `std::optional` conversion ambiguity exposed by GCC 16.

It also fixes some documentation issues:

- Fix the "Custom data source" README example so it compiles (closes #4354).
- Clarify that error_handler_t::ignore drops invalid UTF-8 bytes (closes #5012).
- Update outdated master links to develop.
- Replace a dead Google+ link.
- Fix a wrong variable name in a conversion example.
- Fix typos.


tests/module_cpp20/main.cpp
```diff
@@ -8,7 +8,50 @@
 
 import nlohmann.json;
 
+#include <sstream>
```
gregmarr: If this is using modules, shouldn't it use `import` instead of `include`?

cmake/download_test_data.cmake
```diff
@@ -32,13 +32,17 @@ endif()
...
 # determine the operating system (for debug and support purposes)
 find_program(UNAME_COMMAND uname)
-find_program(VER_COMMAND ver)
 find_program(LSB_RELEASE_COMMAND lsb_release)
 find_program(SW_VERS_COMMAND sw_vers)
 set(OS_VERSION_STRINGS "${CMAKE_SYSTEM}")
-if (VER_COMMAND)
-    execute_process(COMMAND ${VER_COMMAND} OUTPUT_VARIABLE VER_COMMAND_RESULT OUTPUT_STRIP_TRAILING_WHITESPACE)
-    set(OS_VERSION_STRINGS "${OS_VERSION_STRINGS}; ${VER_COMMAND_RESULT}")
+if (CMAKE_HOST_WIN32)
```
gregmarr, line 38: Should this have an else branch that uses `ver` for other OSes?

tests/module_cpp20/main.cpp
nlohmann: Probably. Interestingly though, this file fails to compile with GCC, but is OK with MSVC and Clang. I'll check.

cmake/download_test_data.cmake
nlohmann: No, `ver` was only used in Windows.

tests/module_cpp20/main.cpp
nlohmann: So `import std;` fixes GCC but breaks Clang, because on the silkeh/clang image, the default libstdc++ ships no std module (`libstdc++.modules.json` is absent; only `libc++.modules.json` exists), and the libc++ path doesn't configure cleanly without extra setup. It would also require:

- C++23 (the test is currently C++20), and
- CMake's experimental import std gate — CMAKE_EXPERIMENTAL_CXX_IMPORT_STD, a version-specific UUID that CMake bumps across releases (the value for 4.0 is already rejected by 4.3). Because CI pulls the latest CMake, a release that changes the UUID would fail the build until the hard-coded value is updated.

What do you think? What should we do here?

gregmarr: Maybe for now use `import` for GCC and `include` for everything else?

nlohmann: Done.

gregmarr (approved):

PR #5233, opened 2026-07-03
Read binary strings/blobs in bulk chunks with a memcpy fast path

nlohmann:
## Summary

The binary reader (`binary_reader::get_string` / `get_binary`) read CBOR/MessagePack/BSON/UBJSON strings and byte arrays **one byte at a time** via `get()` + `push_back()`. Worse, the iterator input adapter's `get_elements()` fallback (`input_adapters.hpp`) was itself a per-byte loop, so even contiguous inputs (`const char*`, `std::vector`/`std::string` iterators) never benefited from a block copy — `get_number`'s multi-byte reads didn't either.

## Changes

1. **`iterator_input_adapter::get_elements()` contiguous fast path.** When the underlying iterator refers to a contiguous range, the whole requested range is copied with a single `std::memcpy`. Contiguity is detected via `std::is_pointer` on all standards and, under C++20, additionally via `std::contiguous_iterator` (so `std::vector`/`std::string` iterators qualify too). Non-contiguous iterators keep the element-by-element loop.

2. **`get_string()` / `get_binary()` read in bounded chunks.** Both now delegate to a shared `get_bytes()` that pulls data through `get_elements()` in capped chunks instead of byte by byte. Capping the chunk size preserves the deliberate *"don't `reserve(len)` for an untrusted length"* DoS protection (peak allocation stays bounded regardless of the announced length) while turning the inner loop into block copies. The `min(chunk_size, len)` computation is width-safe, so a narrow length type (e.g. MessagePack ext-8's `uint8_t`) cannot truncate `chunk_size` to zero.

## Benchmark

Payload: a 2 MiB text string + 2 MiB binary blob + 2000×1 KiB strings, re-encoded to CBOR/MessagePack and decoded 200× (Apple clang, `-O2`):

| Std | Case | before | after | speedup |
|-----|------|-------:|------:|--------:|
| C++20 | `from_cbor(vector)` | 20.1 ms | 1.0 ms | ~20× |
| C++20 | `from_cbor(pointer)` | 18.9 ms | 1.0 ms | ~19× |
| C++17 | `from_cbor(pointer)` | 18.7 ms | 1.0 ms | ~19× (memcpy path) |
| C++17 | `from_cbor(vector)` | 18.7 ms | 4.0 ms | ~4.6× (tight loop, no memcpy pre-C++20) |

`from_msgpack` shows the same pattern.

## Correctness

- Behavior is unchanged: truncated input still throws `parse_error.110` at the same byte offset (verified for lengths straddling the chunk boundary).
- All binary-format unit tests pass (`test-cbor`, `test-msgpack`, `test-ubjson`, `test-bson`, `test-bjdata`, `test-binary_formats` — several million assertions, 0 failures) under C++11.
- Added a regression test (`CBOR large strings and binaries (chunked reader)`) covering string/binary roundtrips and truncation at lengths of 0, 1, 4095, 4096, 4097, 8192, and 100000 bytes, for both vector and pointer inputs.
- `single_include` regenerated; `make check-amalgamation` passes.


include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -2941,18 +2930,60 @@ class binary_reader
...
+            const std::size_t old_size = result.size();
+            result.resize(old_size + wanted);
+            const std::size_t read = ia.get_elements(&result[old_size], wanted);
+            chars_read += read;
+            if (JSON_HEDLEY_UNLIKELY(read < wanted))
+            {
+                // premature end of input: shrink to what was actually read and
+                // report the failure at the first missing byte
+                result.resize(old_size + read);
+                ++chars_read;
```
gregmarr, line 2984: Why are you increasing `chars_read` here?  It was already increased by `read` above.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -2941,18 +2930,60 @@ class binary_reader
...
+                current = char_traits<char_type>::eof();
+                return unexpect_eof(format, context);
+            }
+            len -= static_cast<NumberType>(wanted);
```
gregmarr: Add an assert that `read == wanted` to catch `get_elements` returning too large?

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -177,9 +177,44 @@ class iterator_input_adapter
...
+
+  private:
+    // whether IteratorType refers to a contiguous range and therefore supports
+    // a std::memcpy fast path (pointers always do; in C++20 we can also detect
+    // library iterators such as those of std::vector and std::string)
+#if defined(__cpp_lib_concepts) && defined(JSON_HAS_CPP_20)
+    static constexpr bool iterator_is_contiguous = std::is_pointer<IteratorType>::value || std::contiguous_iterator<IteratorType>;
+#else
+    static constexpr bool iterator_is_contiguous = std::is_pointer<IteratorType>::value;
+#endif
```
gregmarr, line 223: Less duplication:

```
    static constexpr bool iterator_is_contiguous =
#if defined(__cpp_lib_concepts) && defined(JSON_HAS_CPP_20)
        std::contiguous_iterator<IteratorType> ||
#endif
        std::is_pointer<IteratorType>::value;
```

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -177,9 +177,44 @@ class iterator_input_adapter
...
+    // contiguous fast path: bulk copy the remaining range with std::memcpy
+    template<class T>
+    std::size_t get_elements_impl(T* dest, std::size_t count, std::true_type /*contiguous*/)
+    {
+        const std::size_t wanted = count * sizeof(T);
+        const std::size_t available = static_cast<std::size_t>(std::distance(current, end)) * sizeof(char_type);
+        const std::size_t copied = (std::min)(wanted, available);
+        if (JSON_HEDLEY_LIKELY(copied != 0))
+        {
+            std::memcpy(dest, &(*current), copied);
```
gregmarr: Why do you need `&(*current)` rather than just `current`?

include/nlohmann/detail/input/binary_reader.hpp
nlohmann: (Claude Code) This mirrors the position accounting in `get_to()` for a partial number read: `chars_read` is advanced by the bytes actually read, then incremented by one more to point at the *first missing* byte, which is where error 110 should be reported. The original per-byte `get_string`/`get_binary` did the same implicitly (the failing `get()` on EOF also bumped `chars_read`). The binary EOF-offset tests (e.g. the MessagePack "unexpected end" cases) still pass unchanged, so the reported byte offsets are identical to before.

include/nlohmann/detail/input/binary_reader.hpp
nlohmann: (Claude Code) Good idea — added `JSON_ASSERT(read == wanted)` after the short-read branch to document/enforce that `get_elements()` never returns more than requested.

include/nlohmann/detail/input/input_adapters.hpp
nlohmann: (Claude Code) Done, adopted verbatim — thanks.

include/nlohmann/detail/input/input_adapters.hpp
nlohmann: (Claude Code) Because in C++20 the fast path also accepts non-pointer contiguous iterators (e.g. `std::vector`/`std::string` iterators via `std::contiguous_iterator`), and for those `current` is a class type that is not implicitly convertible to `const void*`, so `memcpy(dest, current, ...)` would not compile. `&*current` yields the raw element address for both raw pointers and those wrapper iterators (safe here since `copied != 0` guarantees `current != end`). I dropped the redundant parentheses and added a comment.

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -177,9 +177,44 @@
...
```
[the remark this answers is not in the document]
nlohmann: (Claude Code) The copy is bounds-checked: `copied = std::min(wanted, available)` where `available` is the number of bytes remaining in the input range, so `memcpy` never reads past the source or writes past `dest` (which was sized by the caller). This is the same pattern as the existing `std::memcpy` calls in `binary_writer.hpp` and `to_chars.hpp`, so I believe this Flawfinder alert can be dismissed as a false positive.

nlohmann:
(Claude Code) Addressed the Flawfinder alerts:

- The three **`buffer/read`** alerts were false positives triggered by a local variable named `read` (Flawfinder matches it as the POSIX `read()` syscall). Renamed it to `bytes_read` — pure rename, identical semantics.
- The single **`buffer/memcpy`** alert is intrinsic to the fast path. The copy is bounds-checked (`copied = min(wanted, available)`), and it matches the existing unsuppressed `std::memcpy` uses elsewhere in the library (`binary_writer.hpp`, `to_chars.hpp`), so it should be dismissable as a false positive.

Also fixed the GCC `-Werror=arith-conversion` build failure (narrowing in the chunk-length subtraction for narrow `NumberType`s like UBJSON int16) and formatted the new test. All binary-format unit suites pass locally.

include/nlohmann/detail/input/binary_reader.hpp
gregmarr: Thanks, that makes sense.

include/nlohmann/detail/input/input_adapters.hpp
gregmarr: Ah, I didn't realize that `current` is an iterator.  It wasn't in any of the modified code and I didn't take the time to go looking for it.  That makes sense.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -2970,18 +2970,24 @@ class binary_reader
                                        : chunk_size;
             const std::size_t old_size = result.size();
             result.resize(old_size + wanted);
-            const std::size_t read = ia.get_elements(&result[old_size], wanted);
-            chars_read += read;
-            if (JSON_HEDLEY_UNLIKELY(read < wanted))
+            // the destination now has room for `wanted` bytes at offset
+            // old_size, which is what get_elements() is allowed to write
+            JSON_ASSERT(result.size() >= old_size + wanted);
```
gregmarr: This should be an exact match.  It is not allowed for `resize()` to make it larger than specified.  Only `capacity()` maybe has that freedom.

gregmarr (approved):

include/nlohmann/detail/input/binary_reader.hpp
nlohmann: (Claude Code) Good catch, fixed to `==` — you're right that `resize()` guarantees exact size, so `>=` was needlessly weak.

nlohmann: Good point! Fixed in e36340da398eefce17fec5acf173dad659bd6040.

nlohmann:
(Claude Code) Merged latest `develop` to resolve the conflict in `include/nlohmann/detail/input/input_adapters.hpp` against #5234 (lazy lexer diagnostics), which touched the same `iterator_input_adapter` class. The two features are orthogonal — the lexer uses the new `get_consumed_count()`/`copy_consumed_range()` helpers, this PR uses `get_elements()` — and both are preserved. Verified locally: all binary-format suites (CBOR/MessagePack/UBJSON/BSON/BJData) plus the lexer/parser/diagnostic-positions suites pass with the merge in place, and `make check-amalgamation` is clean.

PR #5234, opened 2026-07-04
Reconstruct lexer diagnostics lazily for seekable input

nlohmann:
## Summary

The lexer maintained a per-character `token_string` diagnostic buffer on 100% of successful parses, but that buffer is only read on error.

`lexer::get()` pushed every scanned character into `token_string` (and `unget()` popped it), yet the buffer is consumed exclusively by `get_token_string()` when rendering the *"last read"* fragment of a `parse_error`. On well-formed input this per-byte copy is pure overhead that is always discarded.

This PR implements the "bigger win" option from the todo: reconstruct the offending token lazily on error from a saved start offset for random-access/contiguous input, keeping the eager per-byte copy only for streaming adapters that cannot seek back.

## Approach

- **Seekable adapters** — random-access, single-byte iterators (backing `std::string`, `const char*`, `std::vector<char>`, arrays): the eager copy is skipped. `reset()` records only the token's start offset; on error the token is rebuilt from the input via `iterator_input_adapter::copy_consumed_range()`.
- **Streaming adapters** — `FILE*`, `std::istream`, wide-string, and user-defined adapters: unchanged eager-copy behavior.
- The strategy is selected at compile time by `input_adapter_supports_seek` (an overload-based detection that is tolerant of adapters lacking the flag, so user-defined adapters keep working).

## Correctness

Error messages are **byte-for-byte identical** across all adapter types. Verified against the existing exhaustive parser/diagnostic tests plus a new regression test (`last-read diagnostics are identical across input adapters`) that cross-checks `std::string`, `const char*`, `std::vector<char>`, `std::list` (bidirectional), `std::istringstream` (streaming), and `std::u16string`/`std::u32string` (wide) — covering control-character escaping, number overflow, BOM, and whitespace/structural accumulation.

## Benchmark

Microbenchmark parsing a 4 MB mixed JSON document (objects, arrays, strings, numbers) from a `std::string`, best-of-many medians:

| | throughput |
|---|---|
| before | ~149 MB/s |
| after  | ~160 MB/s |

≈ **+8%** parse throughput (theoretical ceiling from removing the buffer entirely is ~+11%).

### DIFF
--- a/include/nlohmann/detail/input/input_adapters.hpp
+++ b/include/nlohmann/detail/input/input_adapters.hpp
@@ -161,8 +161,18 @@ class iterator_input_adapter
   public:
     using char_type = typename std::iterator_traits<IteratorType>::value_type;
 
+    // Whether the lexer may reconstruct already-consumed input on demand (for
+    // diagnostics) instead of copying every scanned character eagerly. This is
+    // only sound for multi-pass, randomly-addressable byte input: the iterator
+    // must be random-access (so the consumed prefix can be revisited in O(1))
+    // and each element must map 1:1 to an input byte (wide inputs are wrapped
+    // in wide_string_input_adapter, which does not expose this).
+    static constexpr bool supports_seek =
+        std::is_same<typename std::iterator_traits<IteratorType>::iterator_category, std::random_access_iterator_tag>::value
+        && sizeof(char_type) == 1;
+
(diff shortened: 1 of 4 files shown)


gregmarr (approved):

nlohmann:
@gregmarr Thanks for the reviews!

PR #5235, opened 2026-07-04
Replace snprintf with a branch-free writer for \uXXXX escapes

nlohmann:
## Summary
- `serializer::dump_escaped` called `std::snprintf(..., "\u%04x", ...)` once per escaped code point (both for the single-BMP-codepoint case and the surrogate-pair case). `snprintf` re-parses the format string and drags in locale/`printf` machinery on every call, which is much heavier than the fixed 6-/12-byte output requires. This is hot for any string with control characters, and for **all** non-ASCII text when `ensure_ascii` is set (e.g. serializing CJK/emoji-heavy data).
- Replaced it with `write_u_escape`, a small branch-free helper that writes `\uXXXX` directly into `string_buffer` via a `nibble_to_hex` lookup table, operating on a buffer position so it can assert sufficient headroom internally. Mirrors the already-hand-rolled `dump_integer` fast path in the same file.
- Added a `JSON_ASSERT` inside `write_u_escape` to guard the "must have at least 6 bytes of headroom" precondition, since the surrounding code already tracks/guarantees this.

## Test plan
- [x] Added a new `"string escape with ensure_ascii"` section in `tests/src/unit-convenience.cpp` covering: BMP escapes across all nibble positions (U+0080, U+00FF, U+07FF, U+4F60, U+ABCD, U+FFFD), the surrogate-pair path for astral code points (U+10000, U+1F600, U+10FFFF), and verbatim pass-through when `ensure_ascii` is disabled.
- [x] Verified output is byte-identical to the old `snprintf`-based implementation across all escapable BMP code points plus astral cases.
- [x] Re-amalgamated `single_include/nlohmann/json.hpp` with the pinned astyle version.
- [x] Ran the updated unit test standalone against both `include/` and `single_include/`; all assertions pass.

### DIFF
--- a/include/nlohmann/detail/output/serializer.hpp
+++ b/include/nlohmann/detail/output/serializer.hpp
@@ -465,18 +465,12 @@ class serializer
                             {
                                 if (codepoint <= 0xFFFF)
                                 {
-                                    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg,hicpp-vararg)
-                                    static_cast<void>((std::snprintf)(string_buffer.data() + bytes, 7, "\\u%04x",
-                                                                      static_cast<std::uint16_t>(codepoint)));
-                                    bytes += 6;
+                                    bytes = write_u_escape(bytes, static_cast<std::uint16_t>(codepoint));
                                 }
                                 else
                                 {
-                                    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg,hicpp-vararg)
-                                    static_cast<void>((std::snprintf)(string_buffer.data() + bytes, 13, "\\u%04x\\u%04x",
-                                                                      static_cast<std::uint16_t>(0xD7C0u + (codepoint >> 10u)),
-                                                                      static_cast<std::uint16_t>(0xDC00u + (codepoint & 0x3FFu))));
-                                    bytes += 12;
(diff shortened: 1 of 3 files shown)


include/nlohmann/detail/output/serializer.hpp
```diff
@@ -465,18 +465,12 @@ class serializer
                             {
                                 if (codepoint <= 0xFFFF)
                                 {
-                                    // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg,hicpp-vararg)
-                                    static_cast<void>((std::snprintf)(string_buffer.data() + bytes, 7, "\\u%04x",
-                                                                      static_cast<std::uint16_t>(codepoint)));
-                                    bytes += 6;
+                                    bytes = write_u_escape(bytes, static_cast<std::uint16_t>(codepoint));
```
gregmarr: Should there be a comment about why this is `=` instead of `+=`?  Alternatively, return the number of bytes written.  Here and two lines below.  This looks like a bug at first glance.  Could also make it clearer by taking a reference and advancing it instead of taking a value and returning a value.

nlohmann: Good point! Fixed in https://github.com/nlohmann/json/pull/5235/commits/feeab08982c4e20192785e356b3156cd3b43c9b5.

PR #5237, opened 2026-07-05
Switch number parsing to `std::from_chars` / extended locale `strto*_l` if available

### LINKED ISSUE #5198: TOCTOU race between lexer construction and locale changes causes float truncation
### Description

There appears to be a time-of-check-to-time-of-use (TOCTOU) race condition in float parsing that causes silent data loss when another thread changes the locale during JSON deserialization.

The lexer queries `localeconv()->decimal_point` at construction time, replacing `.` in the token buffer with that character. Later, `strtod` parses the buffer using the *current* locale. If another thread changes the locale between these two points, the locale's current decimal point won't match what's in the buffer, causing `strtod` to stop parsing at the wrong character.

**Check:** `lexer` constructor (`lexer.hpp:127`) queries `get_decimal_point()` which calls `localeconv()`.

**Use:** `scan_number_done` (`lexer.hpp:1293`) calls `strtod()` which respects the *current* locale.

Any `setlocale()` call from another thread between construction and parsing can cause misalignment.

Additionally, changing locale in parallel with locale-dependent functions like `strtod()` is undefined behavior. Since JSON itself does not allow for decimal separators other than `.`, it is quite surprising that `json::parse()` exhibits the same safety issue and should be document if it cannot be avoided.

### Reproduction steps

1. Compile `main.cpp` (included below) with `-std=c++20`. Alternatively, clone https://github.com/jgreitemann/json-locale-bug-repro and run the commands below.
2. Run the binary (it switches locale between `C` and `de_DE` on the main thread while a worker thread parses JSON)
3. Observe truncation or assertion failure

```bash
# Debug builds trigger an assertion
make debug && ./build/repro-debug

# Release builds silently truncate
make release && ./build/repro
```

#### Expected behavior

`99.123456` parsed correctly, with 10M+ iterations completing without error.

##### Actual behavior (release)

Fractional part is silently truncated: `99.123456` becomes `99.0`.

I've also observed the reproduction to crash with `SIGSEGV` occasionally. I suspect that's another symptom of the  undefined behavior due to using `strtod` while the locale is being set.

###### Sample output (release)

```
Locale: C
TRUNCATION at iteration 68: expected 99.123456000, got 99.000000000
```
or
```
Locale: C
fish: Job 1, './build/repro' terminated by signal SIGSEGV (Address boundary error)
```

##### Actual behavior (debug)

`JSON_ASSERT(endptr == token_buffer.data() + token_buffer.size())` at `lexer.hpp:1296` fires since the token buffer is not fully consumed.

###### Sample output (debug)

```
Locale: C
Assertion failed: (endptr == token_buffer.data() + token_buffer.size()), function scan_number, file lexer.hpp, line 1296.
```

### Minimal code example

```Shell
// main.cpp
#include <cstdio>
#include <cstdlib>
#include <thread>

#include <nlohmann/json.hpp>

using json = nlohmann::json;

const char *kJson = R"({"val": 99.123456})";
static constexpr double kExpected = 99.123456;
static constexpr int kMaxSwitches = 10'000'000;

void worker(std::stop_token stop) {
  for (int i = 0; !stop.stop_requested(); ++i) {
    json j = json::parse(kJson);
    double val = j.value("val", 0.0);
    if (val != kExpected) {
      fprintf(stderr, "TRUNCATION at iteration %d: expected %.9f, got %.9f\n",
              i, kExpected, val);
      std::exit(1);
    }
  }
}

int main() {
  printf("Locale: %s\n", setlocale(LC_NUMERIC, nullptr));

  {
    int switches = 0;
    bool german = false;

    std::jthread w(worker);

    for (; switches < kMaxSwitches; ++switches) {
      setlocale(LC_NUMERIC, german ? "de_DE.UTF-8" : "C");
      german = !german;
    }
  }

  printf("OK: %d iterations without truncation\n", kMaxSwitches);
}
```

- Apple clang 21.0.0 / macOS 26.5
- GCC 14.1.0 (glibc 2.31.0) / Linux 7.0.0
- VS 2022, MSVC 19.44.35227 / Windows 11

### Library version

v3.11.3 (commit d10879bca, latest develop as of this report)

### Validation

jgreitemann:
Fixes #5198. The issue describes a TOCTOU bug when changes to the global C locale race against the use of `localeconv`/`strto*` by the lexer. Even discounting the TOCTOU issue, using locale-dependent functions while switching locale concurrently is undefined behavior.

Implements a two-tiered approach to avoid locale dependent number parsing:
* When available, the C++17 `std::from_chars` provides an inherently locale-independent API.
* Otherwise, use extended locale APIs (`strto*_l` on POSIX, `_strto*_l` on Windows) and pass them an instance of the "C" locale. This isolates them from the effects of the global locale.
* Otherwise, fall back to the locale-dependent `strto*` functions. The bug is not fixed in this case.

All approaches are encapsulated in terms of a `nlohmann::detail::from_chars` API that is more or less identical to that of `std::from_chars`.

Caveats / Availability:
- `std::from_chars` support is detected based on the `__cpp_lib_to_chars` feature-test macro. Implementation support for `std::from_chars` has not been great. For example, libc++ only added support for `float` and `double` in Clang 20 and to this day does not support `long double`. Hence, libc++ does not set `__cpp_lib_to_chars` and we won't be using `std::from_chars`, even for `float` and `double`. If a future version of libc++ adds support for `long double`, the PR would start using it, though.
- Standard library implementations diverge in their handling of floating-point under-/overflow. [P4168](https://isocpp.org/files/papers/P4168R0.html) does a good job in describing their disagreement and proposes the behavior that would actually be more useful as it allows distinguishing under-/overflow towards ±0 vs. ±∞. This PR implements a shim around `std::from_chars` which inspects the result in case of `result_out_of_range` and ensures behavior consistent with what P4168 proposes (and which libc++ and MS STL already implement), even on libstdc++ (which currently leaves the value unset). If P4168 is accepted as a defect report and implementations converge in the future, this shim should no longer be exercised (though it would have to stay in to still support older toolchains).
- Extended locale functions, i.e. non-standard variants which accept an additional locale argument, cannot be assumed universally available on POSIX platforms and live outside of the POSIX standard. That being said, I'm not aware of any mainstream platform that doesn't provide the extended locale functions. This PR uses a trait class template with a SFINAE-based detection of the availability of `strtof_l`. On Windows, `_strtof_l` and friends have been available since at least Visual Studio 2005, so I think we don't need to bother with SFINAE in that case.
- AFAICT, musl doesn't do locale, in the sense that all locales are essentially identical to the "C" locale. Curiously, the extended locale functions still exist for compatibility with software expecting them but they simply ignore the locale argument.

A last-resort mitigation (in case extended locale APIs are not available) would potentially see us using `uselocale` (POSIX) and `_configthreadlocale` (Windows) to set (and unset) a thread-local locale. This approach comes with degraded performance, even for users who do not need to mitigate the issue because they don't change locale. In practice, the absence of the two other mitigations should be quite rare, so this approach is _not_ implemented by this PR (see discussion in issue). An open question is whether this is still a sufficiently important issue to add a note to the documentation.


nlohmann (changes requested):
Some observations (thanks to Claude Code):

1. The `std::from_chars-tier` P4168 shim overflows a plain `int` when re-parsing the exponent, causing an assertion abort (debug builds) or a silently wrong result (release builds). E.g., `"1e-99999999999999999999"` incorrectly parses to `inf` instead of the correct `0.0`.

2. The POSIX extended_locale_traits SFINAE probe only tests `strtof_l`, but the selected specialization unconditionally uses `strtoll_l`/`strtoull_l`/`isspace_l` too, which musl doesn't declare.

nlohmann:
Hi @jgreitemann, thanks for your patch! I had a look at the CI failures:

1. **Widespread linker failures ("undefined reference to `pthread_create`")** — this is the single biggest cluster (gcc, gcc_old, clang 3.9+, icpc, noexceptions, and more). The new regression test for #5198 spawns a `std::thread`, but `tests/CMakeLists.txt` never links against `Threads::Threads`/`pthread`. It happened to link on some toolchains only because their driver auto-adds `-pthread`. Fix: add `find_package(Threads REQUIRED)` + link `Threads::Threads` for the test targets.
2. **`#if defined(__has_include) && __has_include(<version>)` breaks old preprocessors** (from_chars.hpp:13, :29) — GCC 4.8 errors with `missing binary operator before token "("`, and MSVC (AppVeyor VS2015 jobs) warns C4067 `unexpected tokens following preprocessor directive` (promoted to error under `/WX`). The rest of the codebase avoids this by nesting the checks instead of combining them on one line, e.g. `macro_scope.hpp:61-62` (`#ifdef __has_include` / `#if __has_include(<version>)`). Same fix should apply here.
3. **`-Wfloat-equal -Werror` (gcc)** at from_chars.hpp:116 and :418 — the exact-equality checks (`value == orig_value`, `value == T{}`/`infinity()`) are intentional, but need the same `#pragma GCC diagnostic push` / `ignored "-Wfloat-equal"` / `pop` wrapping already used in `to_chars.hpp:1078-1090`, `json.hpp:3761/3865`, and `binary_writer.hpp:1779`.
4. **clang-tidy `misc-const-correctness`** at from_chars.hpp:260 (`c_locale` can be `const`) and :485 (pointee of `ptr` can be `const`) — straightforward `const` additions.
5. **MinGW build fails**: `_strtoll_l`/`_strtoull_l`/`_strtof_l`/`_strtold_l` are "not declared in this scope". The comment "on Windows, `_strtof_l` et al. are always available" doesn't hold for MinGW's CRT (only real MSVC). Might need a MinGW-specific carve-out that falls back to the locale-dependent path, similar to the musl case.
6. **MSVC VS2015 jobs (AppVeyor)**: `error C2131: expression did not evaluate to a constant` at from_chars.hpp:379/387/395/403/411 — the `from_chars_traits<T>` specializations initialize `static constexpr` function-pointer members (`&extended_locale_traits<>::strtoll` etc.) at class-definition time, which VS2015's constexpr support apparently can't handle. Worth checking if VS2015 is still a hard requirement for this project's CI matrix; if so this may need a non-constexpr indirection for that toolset.

jgreitemann:
Hello @nlohmann,

thanks for the initial comments. I've addressed all but these two points:

1. > The POSIX extended_locale_traits SFINAE probe only tests strtof_l, but the selected specialization unconditionally uses strtoll_l/strtoull_l/isspace_l too, which musl doesn't declare.

   I didn't realize musl only provided the floating-point functions for compatibility. Since those are only wrappers for their localeless counterparts (since musl doesn't do locale), I guess I will extend the SFINAE probe to check for _all_ of the `*_l` functions I'm using.

2. > MinGW build fails: _strtoll_l/_strtoull_l/_strtof_l/_strtold_l are "not declared in this scope". The comment "on Windows, _strtof_l et al. are always available" doesn't hold for MinGW's CRT (only real MSVC). Might need a MinGW-specific carve-out that falls back to the locale-dependent path, similar to the musl case.

   I'm not sure that to expect to work on MinGW and I have little experience with it. I'm not sure
   - if it is a bug in MinGW that the symbol for `_strtof_l` is not exported, or
   - if it instead provides `strtof_l` and the real issue is that the `#ifdef _WIN32` "bites" on MinGW when I should instead be checking for the MSVC CRT (would that be `#ifdef _MSC_VER`?), or
   - if there's no support for either version.
   
   I suppose it doesn't really matter which one is the case — I might just as well avoid making any assumptions about the platform and add another `extended_locale_traits<>` specialization that's SFINAE-predicated on `_strto*_l` without any preprocessor logic? Hope that wouldn't tank compile times too much. What do you think?

jgreitemann:
Sorry about putting this on the back burner. I still plan on finishing it once I have some time to spare.

nlohmann:
Thank you very much for this thorough PR, @jgreitemann. Number parsing now goes through `std::from_chars` when the standard library provides it. That landed in #5283, which is already on `develop`, and your PR now conflicts with it heavily. I'm closing it as superseded. If you think something in it is still missing on `develop`, such as the extended-locale `strto*_l` fallback, please open an issue or a smaller follow-up PR against the current code. Thanks again for all the effort!

Issue #4320, opened 2024-03-22
Behavioral change of serializers in 3.11.3. Name lookup related.


TobiSchluter:
### Description

This is a behavioral change in nlohmann::json 3.11.3 compared to previous versions. Old code of mine started behaving differently.

Output of the test program changed from:
nlohmann::json 3.11.2 (and older versions)
```
{"x":1.0,"y":2.0,"z":3.0}
{"x":1.0,"y":2.0,"z":3.0}
```

To:
nlohmann::json 3.11.3
```
[1.0,2.0,3.0]
{"x":1.0,"y":2.0,"z":3.0}
```

The issue appears to be related to the lookup of to_json converters. Notice that the old behavior can be triggered by adding the namespace prefix to the `to_json` call in `N::to_json`.

### Reproduction steps

Run the code reproduced below. Results will differ between 3.11.2 and 3.11.3.

It needs the Eigen library, version 3.4.0. What appears to happen is that the custom vector type is recognized as "array compatible" and then output as such.

### Expected vs. actual results

The output should be as in older versions of nlohmann::json:
```
{"x":1.0,"y":2.0,"z":3.0}
{"x":1.0,"y":2.0,"z":3.0}
```
instead we get
```
[1.0,2.0,3.0]
{"x":1.0,"y":2.0,"z":3.0}
```
i.e. my serializer is bypassed.

### Minimal code example

```C++
#include <Eigen/Dense>
#include <iostream>
#include <nlohmann/json.hpp>

// Types for data exchange.
namespace N {
    struct Vector {
        double x, y, z;
    };
}

namespace L {
    using Vec3d = Eigen::Vector3d;
}

namespace N {
    // I don't recall why my code goes through this intermediate type, but this is likely related to the change in behavior.
    class Vec3d : public L::Vec3d {
    public:
        explicit Vec3d(const L::Vec3d& v) noexcept
            : L::Vec3d(v)
        {}
        Vec3d(const Vector& v) noexcept
            : L::Vec3d(v.x, v.y, v.z)
        {}
        using L::Vec3d::Vec3d;
    };
    inline Vec3d vToEigen(const Vector& v) noexcept { return Vec3d(v); }
}

namespace Eigen {
    inline void to_json(nlohmann::json& j, const L::Vec3d& v) noexcept
    {
        j = { { "x", v.x() },{ "y", v.y() },{ "z", v.z() } };
    }
}

namespace N {
[...]

PR #5239, opened 2026-07-06
Fix allocation failure during JSON destruction

### LINKED ISSUE #5135: basic_json destructor allocates memory, violating noexcept semantics
### Description

The destructor of `basic_json` invokes `json_value::destroy`, which may allocate a large `std::vector` to flatten out its object/array hierarchy in order to avoid recursion.  In the event of an allocation failure, the vector will throw an uncatchable `std::bad_alloc`, terminating the program.

(I noticed this while debugging a *related* issue, where destroying an invalidated JSON object caused a four-exabyte allocation.)

### Reproduction steps

- Create a complex JSON document by nesting a few million arrays/objects in an arbitrary structure.
- Artificially exhaust the application's available memory.
- Destroy your JSON document.

Allocating a bunch of memory in a destructor is *weird* and widely understood to be unsafe.

Normal recursive destruction would be suitable for my use-case.  If that is unacceptable, there are methods (such as forming an ad-hoc linked list) that could be used to avoid recursion without allocating.  If those methods are unacceptable, a non-throwing allocator could be used with a fallback to recursive destruction in the event that no memory is available.

The latter approach is implemented by PR #4654.

### Minimal code example

```Shell

```

### Error messages

```Shell

```

Win11/MSVC

### Library version

3.12.0

### Validation

michaelsam94:
This PR fixes #5135 by preventing allocation failures from the temporary destruction stack from escaping `basic_json`'s `noexcept` destructor.

The existing iterative destruction path is kept unchanged for the normal case. It is now wrapped with the library's `JSON_TRY` / `JSON_CATCH` macros; if growing the heap-allocated stack throws, destruction falls back to the regular array/object cleanup below instead of allowing the exception to cross the destructor boundary and call `std::terminate`.

A regression test in `unit-regression2.cpp` injects a one-shot global allocation failure while destroying a structured JSON value, verifying that the destructor consumes the failing stack allocation path and returns normally.

Local verification:

```sh
/usr/bin/c++ -std=c++11 -Iinclude -Itests/src -Itests/thirdparty/doctest tests/src/unit.cpp tests/src/unit-regression2.cpp -o build-local/unit-regression2 && ./build-local/unit-regression2 --test-case='*5135*'
./build-local/unit-regression2
/usr/bin/c++ -std=c++11 -Iinclude -Itests/src -Itests/thirdparty/doctest tests/src/unit.cpp tests/src/unit-allocator.cpp -o build-local/unit-allocator && ./build-local/unit-allocator
printf '#include <nlohmann/json.hpp>\nint main(){ nlohmann::json j = nlohmann::json::array({1,2,3}); return j.size() == 3 ? 0 : 1; }\n' | /usr/bin/c++ -std=c++11 -Isingle_include -x c++ - -o build-local/single-smoke && ./build-local/single-smoke
printf '#include <nlohmann/json.hpp>\nint main(){ nlohmann::json j = nlohmann::json::array({1,2,3}); return j.size() == 3 ? 0 : 1; }\n' | /usr/bin/c++ -std=c++11 -fno-exceptions -DJSON_NOEXCEPTION -Iinclude -x c++ - -o build-local/noexceptions-smoke && ./build-local/noexceptions-smoke
git diff --check
```

I could not run the full CMake suite locally because `cmake` is not available on this machine's `PATH`.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -598,50 +598,58 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
             }
             if (t == value_t::array || t == value_t::object)
             {
-                // flatten the current json_value to a heap-allocated stack
-                std::vector<basic_json> stack;
+                JSON_TRY
+                {
+                    // flatten the current json_value to a heap-allocated stack
+                    std::vector<basic_json> stack;
 
-                // move the top-level items to stack
-                if (t == value_t::array)
-                {
-                    stack.reserve(array->size());
-                    std::move(array->begin(), array->end(), std::back_inserter(stack));
-                }
-                else
-                {
-                    stack.reserve(object->size());
-                    for (auto&& it : *object)
+                    // move the top-level items to stack
+                    if (t == value_t::array)
                     {
-                        stack.push_back(std::move(it.second));
+                        stack.reserve(array->size());
+                        std::move(array->begin(), array->end(), std::back_inserter(stack));
                     }
-                }
-
-                while (!stack.empty())
-                {
-                    // move the last item to a local variable to be processed
-                    basic_json current_item(std::move(stack.back()));
-                    stack.pop_back();
-
-                    // if current_item is array/object, move
-                    // its children to the stack to be processed later
-                    if (current_item.is_array())
+                    else
                     {
-                        std::move(current_item.m_data.m_value.array->begin(), current_item.m_data.m_value.array->end(), std::back_inserter(stack));
-
-                        current_item.m_data.m_value.array->clear();
-                    }
-                    else if (current_item.is_object())
-                    {
-                        for (auto&& it : *current_item.m_data.m_value.object)
+                        stack.reserve(object->size());
+                        for (auto&& it : *object)
                         {
                             stack.push_back(std::move(it.second));
                         }
-
-                        current_item.m_data.m_value.object->clear();
                     }
 
-                    // it's now safe that current_item gets destructed
-                    // since it doesn't have any children
+                    while (!stack.empty())
+                    {
+                        // move the last item to a local variable to be processed
+                        basic_json current_item(std::move(stack.back()));
+                        stack.pop_back();
+
+                        // if current_item is array/object, move
+                        // its children to the stack to be processed later
+                        if (current_item.is_array())
+                        {
+                            std::move(current_item.m_data.m_value.array->begin(), current_item.m_data.m_value.array->end(), std::back_inserter(stack));
+
+                            current_item.m_data.m_value.array->clear();
+                        }
+                        else if (current_item.is_object())
+                        {
+                            for (auto&& it : *current_item.m_data.m_value.object)
+                            {
+                                stack.push_back(std::move(it.second));
+                            }
+
+                            current_item.m_data.m_value.object->clear();
+                        }
+
+                        // it's now safe that current_item gets destructed
+                        // since it doesn't have any children
+                    }
+                }
+                JSON_CATCH(...) // LCOV_EXCL_LINE
+                {
+                    // If the heap-allocated stack cannot grow, fall back to
+                    // recursive destruction rather than escaping a noexcept dtor.
                 }
             }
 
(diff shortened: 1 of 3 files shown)


nlohmann:
I just merged a test which caused a conflict. Sorry, please rebase.

michaelsam94:
Rebased onto current develop (acf076a67) and resolved the test conflict by keeping the newly merged #4320 regression and placing the #5135 regression after it.\n\nRe-ran local checks:\n\n- unit-regression2: 9 test cases, 128 assertions passed\n- unit-allocator: 3 test cases, 17 assertions passed\n- single-header smoke compile\n- no-exceptions smoke compile\n- git diff --check

nlohmann:
Thanks for the fix — the `JSON_TRY`/`JSON_CATCH` wrapper does address the crash from #5135 in the common case. A review turned up a few things worth a look before merging:

**Correctness**

1. **New dangling-pointer risk in `external_constructor<>::construct()`** ([`to_json.hpp`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/conversions/to_json.hpp), e.g. `external_constructor<value_t::string>::construct`). Every overload does `destroy(old_type); m_type = new_type; m_value = new_value;`. Before this PR, if `destroy()` threw (OOM), execution never reached the `m_type` reassignment, so `j` stayed consistent (leaked, but not corrupted). Now `destroy()` never throws, so `m_type` is always updated — and if the *second* allocation (for the new value) then also fails under the same sustained memory pressure, `j` ends up with `m_type` and `m_value` referring to different types, i.e. a dangling pointer. Reproduction sketch: `j` holds a large/deeply-nested array, then `j = someString;` under persistent OOM — `destroy()` swallows the failure and frees the old array, `m_type` becomes `string`, then the string allocation also fails, leaving `m_value.string` pointing at freed memory while `m_type == string`.

2. **Fix is a no-op under `JSON_NOEXCEPTION` alone.** `JSON_TRY`/`JSON_CATCH` only expand to real `try`/`catch` when exceptions are compiler-enabled *and* `JSON_NOEXCEPTION` is undefined; otherwise they expand to `if(true)`/`if(false)`. A build with `-DJSON_NOEXCEPTION` but without `-fno-exceptions` (a supported, documented combination) still gets a genuine `std::bad_alloc` from `std::vector`, but the new catch compiles away — so it still escapes the `noexcept` destructor and calls `std::terminate()`, reproducing #5135 exactly in that configuration. The new regression test is itself guarded with `!defined(JSON_NOEXCEPTION)`, so this gap isn't covered.

3. **`JSON_CATCH(...)` swallows more than allocation failures.** It's a blanket catch, so a custom `AllocatorType` that throws something other than `std::bad_alloc` gets silently discarded and funneled into the same fallback path, even though the comment describes only the "heap-allocated stack cannot grow" case.

**Design tradeoff worth a second look**

4. The fallback (swallow, then let the container's normal allocator-based destroy run) falls through to ordinary recursive destruction for whatever wasn't flattened yet. Since every nested level's `destroy()` independently hits the same catch, under a *sustained* (not just transient) low-memory condition, each level fails its own flatten attempt too — so potentially the entire remaining tree ends up destroyed via plain recursion, which is exactly the stack-overflow-prone pattern the iterative flatten-to-heap-stack algorithm was originally introduced to avoid, and it kicks in specifically under memory pressure. Might be worth a comment (or a depth-based test) documenting that this residual risk was considered and accepted.

**Minor**

- The new regression test only injects a single one-shot failure on a 2–3 level deep structure, so it doesn't exercise any of the scenarios above.
- `tests/src/unit-regression2.cpp`: the global `operator new`/`delete` override is active for the whole `unit-regression2` binary rather than scoped to just the new `TEST_CASE`.
- `const bool allocation_failure_was_injected = !fail_next_global_allocation; ... CHECK(allocation_failure_was_injected);` could just be `CHECK_FALSE(fail_next_global_allocation);`.

*(Comment drafted with the help of Claude Code.)*

nlohmann (changes requested):
Fix conflicts and check out https://github.com/nlohmann/json/pull/5239#issuecomment-5346217269

Issue #5246, opened 2026-07-08
std::optional<T> direct construction from JSON null throws instead of yielding std::nullopt


nlohmann:
## Description

Constructing or assigning `std::optional<T>` directly from a JSON `null` value throws `type_error` instead of yielding `std::nullopt`. Only the explicit `.get<std::optional<T>>()` form works correctly.

## Reproduction

```cpp
#include <nlohmann/json.hpp>
#include <optional>

int main() {
    nlohmann::json j_null;
    
    // Works correctly
    auto opt1 = j_null.get<std::optional<std::string>>();  // → std::nullopt ✓
    
    // Throws type_error 302
    std::optional<std::string> opt2 = j_null;  // ✗ throws
    std::optional<std::string> opt3(j_null);   // ✗ throws (direct-init also fails)
}
```

Error: `[json.exception.type_error.302] type must be string, but is null`

## Expected behavior

Direct construction/assignment should yield `std::nullopt`, consistent with `.get<std::optional<T>>()`.

## Environment

- Compiler: GCC 16.1.0, Apple Clang 21.0.0
- C++ standard: C++17 and later
- Library version: `develop` branch

## Why this cannot be cleanly fixed

This is a limitation at the C++ language level, not a library bug:

1. `std::optional<T>` has its own converting constructor that accepts any type constructible from `T`
2. `T` (e.g., `std::string`) is constructible from `basic_json` via `basic_json::operator T()`
3. When both constructors are viable, std::optional's own constructor wins overload resolution
4. There is no SFINAE trick to distinguish "being called from inside std::optional's own constructor" from a direct call

A proper fix would require restricting or redesigning `basic_json::operator ValueType()` (json.hpp:1946-1971), which is relied upon for all ordinary implicit conversions (e.g., `std::string s = json_obj;`). Such a change would be a breaking change to the core conversion API and belongs in a major-version redesign.

## Workaround

Use `.get<std::optional<T>>()` or `.get_to()` for `std::optional<T>` conversions:

```cpp
auto opt = j_null.get<std::optional<std::string>>();  // → std::nullopt ✓
j_null.get_to(opt);                                     // → std::nullopt ✓
```

## Related

- #4864: Original `std::optional` support issue (now fixed on `develop`)
- #3453: Type-strictness redesign for 4.0 (where a permanent fix could live)

---

*Note: This issue was filed by Claude Code to track a limitation separate from #4864.*

hiSandog:
[...]

Issue #4864, opened 2025-07-28
C++17 std::optional feature not enabled


Stevan1996:
### Description

I tried to use `std::optional` with the library. I am using CMake, and, in my CMakeLists.txt, I set the language standard to C++23 with `set(CMAKE_CXX_STANDARD 23)`.

When trying to read a JSON field as `std::optional<std::string>>`, I get a compilation error that no matching overload could be found.
By skimming over the source code, I could not find any checks for `JSON_HAS_CPP_17` in the implementation of the `get` method.

### Reproduction steps

Reading a JSON field as `std::optional<T>` with `json.get<std::optional<T>>`.

### Expected vs. actual results

*Expected*
The `get` method should be able to return `std::optional<T>`.

*Actual*
Compiler cannot find a matching overload when using `std::optional<T>` as template parameter of `get`.

### Minimal code example

```Shell
nlohmann::json json;
json.get<std::optional<std::string>>();
```

### Error messages

```Shell
error C2672: 'nlohmann::json_abi_v3_12_0::basic_json<std::map,std::vector,std::string,bool,int64_t,uint64_t,double,std::allocator,nlohmann::json_abi_v3_12_0::adl_serializer,std::vector<uint8_t,std::allocator<uint8_t>>,void>::get': no matching overloaded function found
```

### Compiler and operating system

MSVC 19.44.35213.0 & Windows 11

### Library version

3.12.0

### Validation

- [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/nlohmann/json/tree/develop) branch is used.
- [x] #4865

gregmarr:
Are you using `JSON_USE_IMPLICIT_CONVERSIONS` set as `false`?  If so, you can't use 3.12.0, you need to use the latest `develop` branch until a new release is made.

>By skimming over the source code, I could not find any checks for JSON_HAS_CPP_17 in the implementation of the get method.

That's correct, it isn't part of `get`, it's done through a `from_json` function.

Stevan1996:
I did not set `JSON_USE_IMPLICIT_CONVERSIONS`. According to the documentation, the default value should be `true`.
I explicitly set the variable with `set(JSON_ImplicitConversions ON)` in the CMakeLists.txt, and it still does not work.

gregmarr:
It works in general: https://www.godbolt.org/z/1qhe8PY55

Can you try to compile that simple file in your environment and share the full error?  There may be clues in that output.

Stevan1996:
[...]

Issue #3453, opened 2022-04-24
Define roadmap for 4.0.0


nlohmann:
### Description

As discussed in https://github.com/nlohmann/json/pull/3444#discussion_r857072525, we currently have no single place to collect all things deemed to wait until the next breaking change.

The goal for this issue is to have a document in the repository to sketch what will come with 4.0.0.

### Reproduction steps

-

### Expected vs. actual results

-

### Compiler and operating system

-

### Library version

-

### Validation

- [ ] The bug also occurs if the latest version from the [`develop`](https://github.com/nlohmann/json/tree/develop) branch is used.
- [ ] I can successfully [compile and run the unit tests](https://github.com/nlohmann/json#execute-unit-tests).

nlohmann:
Related discussion: https://github.com/nlohmann/json/discussions/3473

amanbadone:
Hey, It seems like maybe I can help you out with developing the roadmap for v4.0.0. Please send the necessary details or some instructions you want me to follow as it would be my first contribution to the project and maybe you can also help me out with this.

nlohmann:
Thanks for offering help here, but I think this is an issue that I have to work on myself. There are plenty of other issues where any help is greatly appreciated.

amanbadone:
Actually, I am a newbie in open source. I have a basic understanding of C++ maybe anyone can specifically suggest to me how can I help to contribute

nlohmann:
Then maybe you can contribute best by checking out the documentation - there is surely a lot to improve, and especially beginners can make a difference there. See https://json.nlohmann.me/ for a starting point.

ricardscandit:
Hey @nlohmann what about the 3.x roadmap? Will there be more 3. releases? It's been more than a year since 3.12.0 but I see many commits landing to the main branch frequently.

Cheers!

nlohmann:
Yes, there will be more 3.x releases.

PR #5247, opened 2026-07-08
Document std::optional<T> direct construction limitation

nlohmann:
## Summary

Documents and tests a known limitation with `std::optional<T>` direct construction/assignment from JSON `null` values.

When constructing `std::optional<T>` directly from a JSON `null`, the library throws `type_error 302` instead of yielding `std::nullopt`. Only the explicit `get<std::optional<T>>()` form works correctly.

**Related to:** #4864 (original `std::optional` support issue, which is fixed on `develop` pending release)

**New tracking issue:** #5246 (documents this direct-init/copy-init limitation separately)

## Changes

1. **Test regression** (`tests/src/unit-conversions.cpp`):
   - Added `CHECK_THROWS_AS_WITH` assertions in the `null` section of `TEST_CASE("std::optional")`
   - Verifies both exception type and error message for `std::optional<std::string>(j_null)` and `std::optional<int>(j_null)`
   - Includes detailed comment explaining the C++ language-level root cause

2. **Documentation** (`docs/mkdocs/docs/features/conversions.md`):
   - Added warning callout in the "Implicit conversions" section
   - Clearly states the limitation and provides the correct workaround (`get<std::optional<T>>()` / `get_to()`)

## Why this limitation exists

This is a C++ language-level constraint, not a library bug:
- `std::optional<T>` has its own converting constructor that accepts any type constructible from `T`
- `T` is constructible from `basic_json` via `basic_json::operator T()`
- When both constructors are viable, std::optional's own constructor wins overload resolution
- There is no SFINAE path to distinguish "called from inside std::optional's own constructor" from "direct call"

A proper fix would require restricting/redesigning `basic_json::operator ValueType()`, which would be a breaking change to the core conversion API and belongs in a 4.0 major-version redesign (see #3453).

## Workaround

Users should use explicit conversion:
```cpp
auto opt = j_null.get<std::optional<std::string>>();  // ✓ std::nullopt
j_null.get_to(opt);                                     // ✓ std::nullopt
```

---

🤖 Generated with Claude Code

### DIFF
--- a/docs/mkdocs/docs/features/conversions.md
+++ b/docs/mkdocs/docs/features/conversions.md
@@ -66,6 +66,24 @@ which forces the explicit `get` form and can catch unintended conversions at com
     floating-point value as an integer truncates it, and narrowing conversions may overflow. See
     [number conversion](types/number_handling.md#number-conversion) for details and how to guard against it.
 
+!!! warning "std::optional direct construction from JSON null throws"
+
+    Constructing or assigning `std::optional<T>` directly from a JSON value does not correctly produce
+    `std::nullopt` for a JSON `null`:
+
+    ```cpp
+    json j_null;
+    std::optional<std::string> opt = j_null;  // ❌ throws type_error 302
+    ```
+
+    This is due to C++ language rules: `std::optional<T>` has its own converting constructor that is chosen over
+    `basic_json::operator T()` when both are viable. Use `get<std::optional<T>>()` or `get_to()` instead:
+
+    ```cpp
+    auto opt = j_null.get<std::optional<std::string>>();  // ✅ std::nullopt
+    j_null.get_to(opt);                                     // ✅ std::nullopt
+    ```
+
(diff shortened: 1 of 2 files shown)


gregmarr:
Does this work as desired when implicit conversions are disabled?

nlohmann:
> Does this work as desired when implicit conversions are disabled?

Yes, same behavior either way. `std::optional<T>(j)` throws on `null` regardless of `JSON_USE_IMPLICIT_CONVERSIONS`, since `std::optional<T>`'s own constructor only requires `is_constructible<T, basic_json>`, not `is_convertible`, so marking `operator T()` explicit doesn't stop it from being picked.

Your question also led me to notice something else worth fixing here: `TEST_CASE("std::optional")` was gated behind `#if !JSON_USE_IMPLICIT_CONVERSIONS`, so it never actually ran in the default build.

Pushed a fix, let's see.

PR #5262, opened 2026-07-09
Fix std::hash contract violation for numeric types (#5256)

### LINKED ISSUE #5256: Int and uint compare equal but hashes do not
### Description

On the implementation of `std::hash`, [the C++ specification](https://en.cppreference.com/cpp/utility/hash) states:

> If `k1 == k2` is `true`, `h(k1) == h(k2)` is also `true`

Yet this is not the case for `json` values of the same numeric value but of signed and unsigned storage types. I can see this behaviour is documented [here](https://json.nlohmann.me/api/basic_json/std_hash/), but it violates the `std::hash` contract. While this behaviour is great for distinguishing between the other types, it feels unnatural for it to distinguish between signed and unsigned types.

### Reproduction steps

```C++
using namespace nlohmann;

int main()
{
    const json a{42};
    const json b{42u};

    assert(a == b);
    assert(std::hash<json>{}(a) == std::hash<json>{}(b));
}
```

I would expect `assert(std::hash<json>{}(a) == std::hash<json>{}(b));` to succeed because `a == b` is `true`, but it actually fails.

I believe the `hash<basic_json>` implementation should treat signed and unsigned integers the same. This is especially important because the storage type selected by the parser is not specified (currently, see #5255 ).

In all other aspects, signed and unsigned integers are treated the same: there is no distinction made in the JSON standard and they are silently cast between when calling `.at()` or `.get<>()`.

### Minimal code example

```C++
// I'd like to be able to write code like this

using namespace nlohmann;

int main()
{
    const json j = R"({"a": 42})"_json;
    const std::unordered_map<json, std::string> m = {{42, "hello"}};
    std::cout << m.at(j.at("a")) << '\n'; // throws std::out_of_range
}
```

### Error messages

```Shell

```

GCC 12.2.0, Linux

### Library version

3.12.0

### Validation

nlohmann:
## Summary

Fixes #5256: The C++ `std::hash` contract requires that if two values compare equal, they must hash to the same value. This fix addresses the violation where `json(42)`, `json(42u)`, and `json(42.0)` all compare equal but hashed differently.

### Core Fix: `include/nlohmann/detail/hash.hpp`
- Added `is_exactly_representable_as_float<BasicJsonType>()` helper to safely check if an integer value round-trips losslessly through the float type
- Modified hash function to:
  - Normalize unsigned integers to signed via `static_cast<number_integer_t>()`, exactly matching the `operator==` behavior
  - Bridge values to float domain only when exactly representable (zero precision loss)
  - Use a shared numeric type tag so all numeric types that are equal hash identically

### Tests: `tests/src/unit-hash.cpp`
- Updated expected distinct hash count from 21 to 19 (json(0), json(0U), json(0.0) now collide)
- Added explicit `std::hash` contract verification tests for both `json` and `ordered_json`

### Documentation
- `docs/mkdocs/docs/api/basic_json/std_hash.md`: Updated to describe the new numeric hash unification behavior and edge cases
- `docs/mkdocs/docs/examples/std_hash.cpp`: Added `hash(0.0)` to show all three forms colliding
- `docs/mkdocs/docs/examples/std_hash.output`: Regenerated output showing equal hashes

### Build
- `single_include/nlohmann/json.hpp`: Regenerated via `tools/amalgamate` to stay in sync

## Testing

✅ All 108 existing tests pass
✅ Hash tests specifically verify the std::hash contract
✅ No regressions detected

## Notes

- This is a **breaking change** in observable behavior (hash values will differ from previous versions), but not in API/source compatibility
- The fix is fully rigorous for signed/unsigned integers at any magnitude
- For integer/float comparisons, there's a documented edge case at extreme magnitudes (beyond float exact range) due to float precision limits, mirroring limitations already in `operator==`


include/nlohmann/detail/hash.hpp
```diff
@@ -90,20 +146,47 @@ std::size_t hash(const BasicJsonType& j)
...
+                return combine(numeric_type, h);
+            }
         }
 
         case BasicJsonType::value_t::number_unsigned:
         {
-            const auto h = std::hash<number_unsigned_t> {}(j.template get<number_unsigned_t>());
-            return combine(type, h);
+            const auto v = j.template get<number_unsigned_t>();
+            // Normalize to signed (matching operator== behavior for U-vs-I comparison)
```
gregmarr, line 173: This will no longer match if #5211 is merged.

docs/mkdocs/docs/examples/std_hash.output
```diff
@@ -1,8 +1,9 @@
 hash(null) = 2654435769
 hash(false) = 2654436030
-hash(0) = 2654436095
-hash(0U) = 2654436156
-hash("") = 6142509191626859748
+hash(0) = 2654436221
+hash(0U) = 2654436221
+hash(0.0) = 2654436221
+hash("") = 11160318156688833227
```
gregmarr, line 6: Why did the string hashes change?

include/nlohmann/detail/hash.hpp
nlohmann: Good catch, and thank you for flagging this. You're right — this normalization currently mirrors today's `operator==` U-vs-I wraparound cast specifically to stay contract-consistent with it, so it would need to be revisited once #5211 lands (which replaces the wraparound cast with a mathematically correct comparison).

I'll wait for #5211 to merge first, then rebase this hash normalization on top of it. The forward-compatible fix will check `value_in_range_of<number_integer_t>(v)` before normalizing, falling back to hashing in the pure `number_unsigned_t` domain when a value doesn't fit — avoiding any wraparound-based collision regardless of which comparison semantics are in effect.

docs/mkdocs/docs/examples/std_hash.output
nlohmann: This isn't caused by the code change — verified by comparing against the original file: `hash(null)`, `hash(false)`, `hash({})`, and `hash([])` (none of which touch the modified code paths) are byte-identical to the previously committed values. Re-running the exact same compiled binary repeatedly also gives identical `hash("")` results, which rules out per-process hash randomization.

The `hash("")` / `hash({"hello": "world"})` values differ only because I regenerated this file locally (Apple Clang/libc++/ARM64), which is a different toolchain than whatever originally produced the checked-in values. `std::hash<std::string>` legitimately differs across standard library implementations — this is exactly the platform-dependence the docs already call out ("Note the output is platform-dependent").

Happy to regenerate on a Linux/GCC toolchain instead if you'd prefer the diff minimized to just the numeric lines that actually changed — let me know.

include/nlohmann/detail/hash.hpp
nlohmann: (Sorry for the auto-generated message. Indeed let's wait for #5211 to land first. Can you check the open question to you there please?)

docs/mkdocs/docs/examples/std_hash.output
nlohmann: (Sorry for the auto-generated message - I need to learn how to restrict Claude here...)

include/nlohmann/detail/hash.hpp
```diff
@@ -90,20 +153,43 @@ std::size_t hash(const BasicJsonType& j)
...
+            const auto h = std::hash<number_integer_t> {}(v_as_signed);
+            return combine(numeric_type, h);
         }
 
         case BasicJsonType::value_t::number_float:
         {
             const auto h = std::hash<number_float_t> {}(j.template get<number_float_t>());
-            return combine(type, h);
+            const auto numeric_type = static_cast<std::size_t>(BasicJsonType::value_t::number_float);
+            return combine(numeric_type, h);
```
Jezza2: This change seems to be redundant - `type` is already `number_float`.

nlohmann: You're right, thanks — `type` already equals `number_float`'s enum value within that case, so the extra `numeric_type` local was redundant there. Removed in 10d18792b, just reusing `type` directly.

(The `numeric_type` variable is still needed in the `number_integer`/`number_unsigned` cases, since `type` there is `number_integer`/`number_unsigned` respectively, not `number_float` — that's the whole point of the shared tag for those two.)

nlohmann:
CI is green here, but that is misleading: the new tests only exercise `0` and `42`. I merged this branch onto current `develop` and probed the boundaries. The contract is still violated, and there is undefined behavior on the way.

### 1. UB — negating `INT64_MIN`

`hash.hpp` does `static_cast<number_unsigned_t>(-val)`. The negation happens in the *signed* domain, so `val == INT64_MIN` is signed overflow. The comment two lines above (`"Using unsigned arithmetic to avoid UB with negating INT_MIN"`) describes what the code was meant to do, not what it does.

Confirmed with UBSan against the merged tree:

```
hash.hpp:75:84: runtime error: negation of -9223372036854775808 cannot be represented in
type 'typename basic_json<>::number_integer_t' (aka 'long long')
```

Reachable via `json(std::numeric_limits<std::int64_t>::min())` and via `json(1ULL << 63)` (the unsigned path casts to `INT64_MIN` first). `ci_test_clang_sanitizer` passes only because nothing in the test suite hashes those values.

### 2. Threshold is off by a power of two

`max_exact = static_cast<number_integer_t>(1) << (float_digits - 1)` is 2^52, but the comment directly above it says "use 2^digits as the threshold", i.e. 2^53. Integers in [2^52, 2^53] *are* exactly representable as `double`, yet they are excluded from the float bridge, so they keep hashing in the integer domain while the equal `double` hashes in the float domain.

### 3. The contract is still violated — measured

All pairs below have `a == b` true on `develop` + this PR:

| pair | `==` | hashes equal |
|---|---|---|
| `json(42)` vs `json(42u)` / `json(42.0)` | yes | yes |
| `json(2^52)` vs `json(2^52 as double)` | yes | **no** |
| `json(2^52+1)` vs `json((double)(2^52+1))` | yes | **no** |
| `json(2^53)` vs `json(2^53 as double)` | yes | **no** |
| `json(-2^52)` vs `json(-2^52 as double)` | yes | **no** |
| `json(UINT64_MAX)` vs `json((double)UINT64_MAX)` | yes | **no** |
| `json(2^63 unsigned)` vs `json((double)2^63)` | yes | **no** |

2^52 is about 4.5e15, which is not exotic — nanosecond timestamps land past it. The `is_exactly_representable_as_float` escape hatch *is* the violation: whenever it returns `false`, an integer that `operator==` considers equal to some float hashes differently.

### 4. The unsigned-to-signed premise is now stale

The comment "Normalize to signed (matching `operator==` behavior for U-vs-I comparison)" stopped being true when #5211 landed on `develop` (two days after this branch's last push): mixed comparison is now value-correct, so `json(-1) != json(UINT64_MAX)`. The hash still maps `UINT64_MAX` to `-1`, so those two now collide — legal, but gratuitous, and it folds the whole upper half of the `uint64` range onto negative `int64` values. This branch needs a rebase regardless.

### Suggestion: drop the special-casing entirely

`operator==` compares integer against float by casting the integer to `number_float_t`. So hashing *every* numeric value in the float domain is exactly consistent with equality — no helper, no threshold, no UB, no MSVC pragma, no `<limits>`/`<cmath>`:

```cpp
case BasicJsonType::value_t::number_integer:
case BasicJsonType::value_t::number_unsigned:
case BasicJsonType::value_t::number_float:
{
    // hash all numeric values in the number_float domain under a shared
    // type tag, mirroring how operator== compares them across types
    const auto numeric_type = static_cast<std::size_t>(BasicJsonType::value_t::number_float);
    number_float_t v{};
    if (j.type() == BasicJsonType::value_t::number_integer)
    {
        v = static_cast<number_float_t>(j.template get<number_integer_t>());
    }
    else if (j.type() == BasicJsonType::value_t::number_unsigned)
    {
        v = static_cast<number_float_t>(j.template get<number_unsigned_t>());
    }
    else
    {
        v = j.template get<number_float_t>();
    }
    const auto h = std::hash<number_float_t> {}(v);
    return combine(numeric_type, h);
}
```

I built this and re-ran the probe: every row in the table above passes, both "legal collision" rows stop colliding, UBSan is clean, and the set in `unit-hash.cpp` still yields exactly 19 distinct hashes — so the test changes in this PR stay valid as written. It is about 55 lines shorter than the current diff.

The trade-off is real but small: distinct integers at or above 2^53 that round to the same `double` collide. Below 2^53 the mapping is injective, so ordinary integer keys are unaffected — and this collision is *forced* by the lossy cross-type comparison in `operator==`. The only way to keep full integer hash precision would be to make integer/float equality exact, which is a separate and larger change to `JSON_IMPLEMENT_OPERATOR`.

### Minor

- `std_hash.md` says the edge case starts "beyond ~2^53", but the code uses 2^52. With the change above the edge-case paragraph can be replaced by a short note about large-integer collisions.
- `std_hash.output`: `hash("")` and `hash({"hello": "world"})` changed although no string or object hashing code changed — the file was regenerated on libc++ rather than libstdc++. I reproduced the committed values byte-for-byte on macOS, which confirms it. `std_hash.test` is excluded from `check_output_portable`, so CI will not catch it, but `make check_output` locally will. The `hash(0)`/`hash(0U)`/`hash(0.0)` values are the same under both standard libraries (both map `0.0` to hash `0`), so only those three lines need to change; the previous string values should be kept.
- Test coverage: nothing in the new tests goes above 42. Worth adding `INT64_MIN`, `UINT64_MAX`, `1ULL << 63`, and the 2^52/2^53 boundaries — those are what turned all of this up.
- Release note: every numeric hash value changes. Not an API or ABI break, but anyone persisting hashes across versions is affected.

*— reviewed and written by Claude Code on behalf of @nlohmann*

nlohmann:
I dug into this vs. #5447, which fixes the same underlying bug (duplicate issue #5400 vs #5256 here), and found a correctness problem with this PR's `is_exactly_representable_as_float` strategy — confirmed with a compiled repro against both PR branches.

**1. `number_unsigned` normalization bug:** the `number_unsigned` branch does `static_cast<number_integer_t>(v)` before checking representability and before the float cast. For any `uint64_t` value above `INT64_MAX`, that cast wraps to a negative garbage value, and the hash ends up computed from that wrapped value instead of the real one:

```cpp
uint64_t big = std::numeric_limits<uint64_t>::max();
json a = big;
json b = static_cast<double>(big);   // a == b is true

hash(a) != hash(b)   // contract violated on this branch
```

**2. A deeper issue beyond that fix:** even correcting (1), the whole "hash non-exactly-representable integers in their native domain" idea is incompatible with how `operator==` compares across types. `operator==` between an integer/unsigned and a float always compares `static_cast<double>(v) == float_value`, with no regard for whether that cast was lossless. So for *any* integer `v`, `json(static_cast<double>(v))` compares equal to it via `operator==` — whether or not `v` round-trips through `double`. That means routing "not exactly representable" integers through integer-domain hashing (as this PR does) breaks the contract for those values too, independent of the unsigned/signed wraparound:

```cpp
int64_t v = (1LL << 60) + 1;   // NOT exactly representable as double
json a = v;
json b = static_cast<double>(v);   // a == b is true

hash(a) != hash(b)   // contract violated here as well
```

Fixing this fully would mean always hashing `number_integer`/`number_unsigned` by converting to `number_float_t` (matching how `operator==` actually compares them across types) — which removes any benefit of the `is_exactly_representable_as_float` machinery and effectively converges on the simpler approach in #5447.

Given that, I think we should pursue #5447 instead of this PR. Closing this in favor of that one.

PR #5265, opened 2026-07-10
Add iterator+sentinel tests and docs for binary deserializers

nlohmann:
## Summary

This PR extends C++20 ranges support (iterator+sentinel pairs) to the binary format deserializers, matching the JSON text parsers. Users can now read binary files directly via `std::istreambuf_iterator<char>` + a sentinel without pre-buffering to a vector.

- Adds `istreambuf_sentinel` helper to test utilities for stream EOF detection
- Implements tests for all 5 binary formats reading directly from test data files
- Updates documentation for `from_cbor`, `from_msgpack`, `from_ubjson`, `from_bjdata`, and `from_bson` with overload (3) descriptions
- All tests pass with existing test suite data

## Test Plan

✅ CBOR: 9 test cases pass
✅ MessagePack: 4 test cases pass  
✅ UBJSON: 5 test cases pass
✅ BJData: 25 test cases pass
✅ BSON: 8 test cases pass (includes temp file + cleanup test)

🤖 Generated with Claude Code

### DIFF
--- a/include/nlohmann/detail/input/input_adapters.hpp
+++ b/include/nlohmann/detail/input/input_adapters.hpp
@@ -155,7 +155,9 @@ class input_stream_adapter
 
 // General-purpose iterator-based adapter. It might not be as fast as
 // theoretically possible for some containers, but it is extremely versatile.
-template<typename IteratorType>
+// SentinelType defaults to IteratorType for backward compatibility, but may
+// be a different type (e.g., a C++20 sentinel or counted_iterator).
+template<typename IteratorType, typename SentinelType = IteratorType>
 class iterator_input_adapter
 {
   public:
@@ -169,9 +171,10 @@ class iterator_input_adapter
     // in wide_string_input_adapter, which does not expose this).
     static constexpr bool supports_seek =
         std::is_same<typename std::iterator_traits<IteratorType>::iterator_category, std::random_access_iterator_tag>::value
+        && std::is_same<IteratorType, SentinelType>::value
         && sizeof(char_type) == 1;
 
-    iterator_input_adapter(IteratorType first, IteratorType last)
+    iterator_input_adapter(IteratorType first, SentinelType last)
         : begin(first), current(std::move(first)), end(std::move(last))
     {}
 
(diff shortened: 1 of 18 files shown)


include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -476,25 +505,38 @@ struct is_iterator_of_multibyte
...
+    return factory_type::create(first, last);
+}
+
+// General purpose iterator-based input (iterator+sentinel pair with different types)
+// Only enable for types that can be compared with !=
+template < typename IteratorType, typename SentinelType,
+           typename = typename std::enable_if <
+               !std::is_same<IteratorType, SentinelType>::value &&
+               can_compare_ne<IteratorType, SentinelType>::value >::type >
+typename iterator_input_adapter_factory<IteratorType, SentinelType>::adapter_type input_adapter(IteratorType first, SentinelType last)
```
gregmarr, line 541: Does this actually need to be a separate function?  Can you just do this?

```
template < typename IteratorType, typename SentinelType = IteratorType,
           typename = typename std::enable_if <
               can_compare_ne<IteratorType, SentinelType>::value >::type >
typename iterator_input_adapter_factory<IteratorType, SentinelType>::adapter_type input_adapter(IteratorType first, SentinelType last)
```

nlohmann: Good point! I'll check!

tests/src/unit-bson.cpp
```diff
@@ -1208,6 +1208,25 @@ TEST_CASE("BSON numerical data")
...
+        std::ofstream file("test_bson_sentinel.bson", std::ios::binary);
+        file.write(reinterpret_cast<const char*>(bson_bytes.data()), static_cast<std::streamsize>(bson_bytes.size()));
+    }
+
+    std::ifstream file("test_bson_sentinel.bson", std::ios::binary);
+    std::istreambuf_iterator<char> first(file);
+    const json parsed = json::from_bson(first, utils::istreambuf_sentinel{});
+    CHECK(parsed == j);
+
+    static_cast<void>(std::remove("test_bson_sentinel.bson"));
```
nlohmann: We should parse a BSON file from the test suite instead of writing/deleting a temp file.

include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -216,12 +219,14 @@ class iterator_input_adapter
   private:
     // whether IteratorType refers to a contiguous range and therefore supports
     // a std::memcpy fast path (pointers always do; in C++20 we can also detect
-    // library iterators such as those of std::vector and std::string)
+    // library iterators such as those of std::vector and std::string).
+    // The fast path also requires SentinelType == IteratorType so std::distance works.
```
gregmarr, line 223: Should this also support `std::counted_iterator` for the `IteratorType` and `std::default_sentinel_t` for the `SentinelType`?  https://en.cppreference.com/cpp/iterator/counted_iterator

That would require that `current.count` be used instead of `std::distance` for that type, which would probably require a member function with SFINAE.

nlohmann: Sorry, saw this too late - opened #5268.

gregmarr: No worries, I wrote it after you merged.  :)

PR #5268, opened 2026-07-11
Extend memcpy fast path to sized sentinels (e.g. std::counted_iterator)

nlohmann:
## Summary

Follow-up to #5265's review discussion: https://github.com/nlohmann/json/pull/5265#discussion_r3564237584 (@gregmarr asked whether the fast path should support `std::counted_iterator` + `std::default_sentinel_t`).

`std::counted_iterator` over a contiguous iterator already satisfies `std::contiguous_iterator`, and `std::default_sentinel_t` is a `std::sized_sentinel_for` it — verified this compiles and behaves correctly (`std::ranges::distance` returns the right count, `&*it` gives a valid contiguous address). This generalizes beyond just `counted_iterator` to any C++20 sized-sentinel pair, without needing a custom count-based member function as originally suggested.

- `iterator_input_adapter::iterator_is_contiguous` now also allows `std::sized_sentinel_for<SentinelType, IteratorType>` under C++20 (previously required `IteratorType == SentinelType`).
- The memcpy fast path computes the available element count with `std::ranges::distance` under C++20 (works for both same-type and sized-sentinel pairs), falling back to `std::distance` pre-C++20 where `SentinelType` is always `IteratorType`.
- Added a C++20-only test exercising `json::parse`/`accept` with `std::counted_iterator` + `std::default_sentinel_t`.
- Documented `std::default_sentinel_t` + `std::counted_iterator` as a `SentinelType` example across `parse.md`, `accept.md`, `sax_parse.md`, and the five `from_*.md` pages, replacing an earlier ambiguously-worded bullet.

Before this combination correctly fell back to the byte-by-byte path (functionally correct, just not optimized); this lets it reach the `memcpy` fast path.

## Test plan

- [x] New C++20-gated test case (`test-user_defined_input_cpp20`) covering `json::parse`/`accept` with `std::counted_iterator` + `std::default_sentinel_t`
- [x] Full local test suite: 109/109 passing across C++11/14/17/20/23
- [x] Verified `iterator_is_contiguous` evaluates true for the new combination and false-path (pre-C++20, byte-by-byte) still compiles/works unchanged
- [x] `make check-amalgamation` clean

🤖 Generated with Claude Code

### DIFF
--- a/docs/mkdocs/docs/api/basic_json/accept.md
+++ b/docs/mkdocs/docs/api/basic_json/accept.md
@@ -49,7 +49,7 @@ Unlike the [`parse()`](parse.md) function, this function neither throws an excep
 :   defaults to `IteratorType`; may be a different type comparable to `IteratorType` via `operator!=`, for instance.
 
     - a custom sentinel type for C++20 ranges
-    - `std::counted_iterator` with a different sentinel type
+    - `std::default_sentinel_t`, when `IteratorType` is `std::counted_iterator`
 
 ## Parameters
 
--- a/docs/mkdocs/docs/api/basic_json/from_bjdata.md
+++ b/docs/mkdocs/docs/api/basic_json/from_bjdata.md
@@ -36,8 +36,10 @@ The exact mapping and its limitations are described on a [dedicated page](../../
 :   a compatible iterator type
 
 `SentinelType`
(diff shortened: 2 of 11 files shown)


gregmarr:
Excellent that there is already a `distance` function that handles this.

gregmarr (approved):

PR #5142, opened 2026-04-20
Fix #4041: Support zero-member types in NLOHMANN_DEFINE_TYPE_* macros

suchetindrakanty:
Fixes #4041

This PR addresses an issue where NLOHMANN_DEFINE_TYPE_* macros fail when used with zero members.

### Problem
When __VA_ARGS__ is empty, macro expansion generates invalid code, causing compilation errors.

### Solution
- Wrapped macro expansions with __VA_OPT__ to ensure code is only generated when arguments are present
- Prevents invalid expansions for zero-member types
- Maintains existing behavior for non-empty cases

### Testing
- Added test cases for intrusive and non-intrusive macros
- Covered WITH_DEFAULT and ONLY_SERIALIZE variants
- Verified empty structs serialize to {}
- Verified deserialization works correctly

PR #5099, opened 2026-03-05
Fix serialization macros with no member variables

GhostVaibhav:
This PR builds upon the previous PR opened for the same issue and updates the relevant files in the relevant places for the same behaviour intended as in the original PR. This is the link to the original PR - https://github.com/nlohmann/json/pull/4323

- [x] The changes are described in detail, both the what and why.
- [x] If applicable, an [existing issue](https://github.com/nlohmann/json/issues) is referenced.
- [x] The [Code coverage](https://coveralls.io/github/nlohmann/json) remained at 100%. A test case for every new line of code.
- [x] If applicable, the [documentation](https://json.nlohmann.me) is updated.
- [x] The source code is amalgamated by running `make amalgamate`.

Read the [Contribution Guidelines](https://github.com/nlohmann/json/blob/develop/.github/CONTRIBUTING.md) for detailed information.

PR #5272, opened 2026-07-11
Support zero-member types in NLOHMANN_DEFINE_TYPE_* macros (#4041)

### LINKED ISSUE #4041: NLOHMANN_DEFINE_TYPE_* fails with zero members

nlohmann:
## Summary

Fixes #4041: `NLOHMANN_DEFINE_TYPE_INTRUSIVE(Type)` and its 11 sibling macros (`_WITH_DEFAULT`, `_ONLY_SERIALIZE`, `NON_INTRUSIVE` equivalents, and the `DERIVED_TYPE` family) produced broken generated code for types with no member variables to serialize.

Three prior PRs attempted this (#5142, #5099, #4994); all are stale and fail CI. This is a fresh implementation, not based on any of them:

- #5142's `__VA_OPT__` approach breaks every pre-C++20 build (`error C3861: '__VA_OPT__': identifier not found` on MSVC's C++11 test project) because it isn't gated behind a C++20 check.
- More fundamentally: invoking a variadic macro so that `__VA_ARGS__` is empty is only standard-conforming since C++20 — GCC/Clang reject it under `-pedantic` pre-C++20 **regardless of what the macro body does** (verified directly against real GCC 16 / Clang with this repo's actual `-pedantic -Werror -Wvariadic-macros` flags). So a `__VA_OPT__`-only fix can, at best, only ever cover C++20+.
- #4994's `##__VA_ARGS__` GNU-extension attempt doesn't avoid this either (also verified directly) and only special-cases GCC.

Instead, this PR makes all 12 macros purely variadic (no fixed leading parameter) and dispatches on argument count, using a sentinel-padded extension of the existing `NLOHMANN_JSON_GET_MACRO` positional-dispatch idiom already used by `NLOHMANN_JSON_PASTE1..64` in this file. `NLOHMANN_JSON_TYPE_TAG(...)` yields the token `EMPTY` for a lone `Type` and `MEMBERS` otherwise; `NLOHMANN_JSON_CAT` pastes that onto the macro's prefix to pick the body. The derived-type macros drop their leading `Type` and reuse the same tag, which keeps the dispatch inside `NLOHMANN_JSON_GET_MACRO`'s 64 positional slots and so preserves the full 63-member limit for them too. This gives full C++11 through C++26 support with **no feature-test gate at all**.

`WITH_NAMES` variants are intentionally out of scope, matching issue #4041 and all three prior PRs.

## Changes

- `include/nlohmann/detail/macro_scope.hpp`: new `NLOHMANN_JSON_CAT`/`NLOHMANN_JSON_TYPE_TAG`/`NLOHMANN_JSON_DERIVED_TYPE_TAG` dispatch helpers; each of the 12 macros split into a `_MEMBERS` body (byte-identical to the previous public macro body) and a new `_EMPTY` body, with the public macro name becoming a 1-line dispatcher. The four `_WITH_DEFAULT_EMPTY` bodies are one-line aliases of their plain counterparts, since with no members there is nothing to default.
- `single_include/nlohmann/json.hpp`: re-amalgamated (`make amalgamate`, astyle 3.4.13).
- `tests/src/unit-udt_macro.cpp`: 12 new zero-member test cases (all 6 base macro variants + all 6 derived variants, the latter using a real base class to confirm base-class delegation still works with zero own members).
- `tests/src/unit-udt_macro.cpp`: a max-arity regression test (`max_members`, `max_members_derived`) pinning the documented 63-member limit for both the plain and the derived macros.
- `docs/mkdocs/docs/features/arbitrary_types.md`: new "Zero-member types" note with an example.
- `docs/mkdocs/docs/api/macros/nlohmann_define_type_intrusive.md`, `..._non_intrusive.md`, `nlohmann_define_derived_type.md`: the `member` parameter is documented as allowing an empty list, with a note on what the zero-member case generates.

## API compatibility

**No breaking changes.** The change is purely additive: argument lists that compiled before produce byte-identical generated code (verified by comparing preprocessed token streams), and the documented 63-member limit is unchanged for both the plain and the derived macros. The only new behaviour is that a previously ill-formed invocation — a macro with no member arguments — now compiles.

## Test plan

- [x] Existing `unit-udt_macro.cpp` suite passes unmodified — zero regressions from the macro rename/dispatch refactor; the 12 `_MEMBERS` bodies are byte-identical to the previous public macro bodies.
- [x] New zero-member cases (12 sections, both `nlohmann::json` and `nlohmann::ordered_json`) pass.
- [x] Verified on real `g++-16` and `clang++` at `-std=c++11/14/17/20` with `-pedantic -Werror -Wvariadic-macros` (matching `cmake/gcc_flags.cmake`).
- [x] `make check-amalgamation` clean.
- [x] Spot-checked an unrelated test file (`unit-udt.cpp`) still compiles/passes to rule out macro-name collisions from the new helpers.
- [x] Verified the 63-member limit holds for all 6 derived macros as well as the 6 plain ones, and that 64 members is still rejected.

---
This PR description was written by Claude Code.

### DIFF
--- a/include/nlohmann/detail/macro_scope.hpp
+++ b/include/nlohmann/detail/macro_scope.hpp
@@ -587,15 +587,57 @@ void templated_json_throw(ExceptionType exception)
 
 #define NLOHMANN_JSON_BASIC_TYPE_TEMPLATE(...) template<typename BasicJsonType, nlohmann::detail::enable_if_t<nlohmann::detail::is_basic_json<BasicJsonType>::value, int> = 0> __VA_ARGS__
 
+// Tags used to dispatch the NLOHMANN_DEFINE_TYPE_*/NLOHMANN_DEFINE_DERIVED_TYPE_*
+// macros below between a zero-member and a one-or-more-member implementation
+// (issue #4041, e.g. NLOHMANN_DEFINE_TYPE_INTRUSIVE(Type) with no further
+// arguments). NLOHMANN_JSON_TYPE_TAG(...) expands to the token 1 when
+// __VA_ARGS__ is a single token (Type alone) and to N for two or more (Type,
+// member...); NLOHMANN_JSON_DERIVED_TYPE_TAG(...) is the same idea shifted by
+// one slot for the two-token Type,BaseType prefix used by the derived-type
+// macros. Both reuse the existing 64-slot NLOHMANN_JSON_GET_MACRO dispatch
+// with one extra trailing sentinel token appended so its own trailing "..."
+// is never left completely empty at the lowest supported argument count --
+// invoking a variadic macro so that "..." matches nothing is only granted
+// unconditionally by the standard since C++20, and pre-C++20 compilers may
+// reject it under -pedantic regardless of what the macro body does.
+#define NLOHMANN_JSON_CAT_(x, y) x ## y
+#define NLOHMANN_JSON_CAT(x, y) NLOHMANN_JSON_CAT_(x, y)
+
+#define NLOHMANN_JSON_TYPE_TAG(...) NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_GET_MACRO(__VA_ARGS__, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, 1, \
+        NLOHMANN_JSON_TYPE_TAG_SENTINEL))
+
+#define NLOHMANN_JSON_DERIVED_TYPE_TAG(...) NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_GET_MACRO(__VA_ARGS__, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, 2, N, \
+        NLOHMANN_JSON_DERIVED_TYPE_TAG_SENTINEL))
+
 /*!
 @brief macro
 @def NLOHMANN_DEFINE_TYPE_INTRUSIVE
 @since version 3.9.0
 @sa https://json.nlohmann.me/api/macros/nlohmann_define_type_intrusive/
 */
-#define NLOHMANN_DEFINE_TYPE_INTRUSIVE(Type, ...)  \
+#define NLOHMANN_JSON_DEFINE_TYPE_INTRUSIVE_N(Type, ...)  \
     NLOHMANN_JSON_BASIC_TYPE_TEMPLATE(friend void to_json(BasicJsonType& nlohmann_json_j, const Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_TO, __VA_ARGS__)) }) \
     NLOHMANN_JSON_BASIC_TYPE_TEMPLATE(friend void from_json(const BasicJsonType& nlohmann_json_j, Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(NLOHMANN_JSON_FROM, __VA_ARGS__)) })
+
+#define NLOHMANN_JSON_DEFINE_TYPE_INTRUSIVE_1(Type)  \
+    NLOHMANN_JSON_BASIC_TYPE_TEMPLATE(friend void to_json(BasicJsonType& nlohmann_json_j, const Type&) { nlohmann_json_j = BasicJsonType::object(); }) \
+    NLOHMANN_JSON_BASIC_TYPE_TEMPLATE(friend void from_json(const BasicJsonType&, Type&) { })
+
+#define NLOHMANN_DEFINE_TYPE_INTRUSIVE(...) NLOHMANN_JSON_CAT(NLOHMANN_JSON_DEFINE_TYPE_INTRUSIVE_, NLOHMANN_JSON_TYPE_TAG(__VA_ARGS__))(__VA_ARGS__)
 
 #define NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_NAMES(Type, ...)  \
     NLOHMANN_JSON_BASIC_TYPE_TEMPLATE(friend void to_json(BasicJsonType& nlohmann_json_j, const Type& nlohmann_json_t) { NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_DOUBLE_PASTE(NLOHMANN_JSON_TO_WITH_NAME, __VA_ARGS__)) }) \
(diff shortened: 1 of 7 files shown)


gregmarr (approved):

include/nlohmann/detail/macro_scope.hpp
```diff
@@ -587,16 +587,59 @@ void templated_json_throw(ExceptionType exception)
...
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, N, N, N, N, N, N, N, \
+        N, N, N, 1, \
+        NLOHMANN_JSON_TYPE_TAG_SENTINEL))
+
+#define NLOHMANN_JSON_DERIVED_TYPE_TAG(...) NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_GET_MACRO(__VA_ARGS__, \
```
nlohmann: **Blocking: this dispatch drops one member slot — all six `NLOHMANN_DEFINE_DERIVED_TYPE_*` macros regress from 63 to 62 members.**

`NLOHMANN_JSON_GET_MACRO` has 64 positional slots with `NAME` at position 65. `NLOHMANN_JSON_TYPE_TAG` dispatches on `1 + members`, so it resolves correctly up to 63 members — exactly matching `NLOHMANN_JSON_PASTE64`'s limit. `NLOHMANN_JSON_DERIVED_TYPE_TAG` dispatches on `2 + members`, so it runs out one slot early: at 63 members, position 65 lands on the *last member name* instead of a sentinel, and `NLOHMANN_JSON_CAT` builds an undefined identifier such as `NLOHMANN_JSON_DEFINE_DERIVED_TYPE_INTRUSIVE_m63`.

Verified against g++-16 and clang at `-std=c++11 -pedantic -Werror -Wvariadic-macros`:

| case | `develop` | this PR |
|---|---|---|
| `NLOHMANN_DEFINE_TYPE_INTRUSIVE` @ 63 members | OK | OK |
| `NLOHMANN_DEFINE_DERIVED_TYPE_INTRUSIVE` @ 62 members | OK | OK |
| `NLOHMANN_DEFINE_DERIVED_TYPE_INTRUSIVE` @ 63 members | **OK** | **FAIL** |

All six derived variants regress identically (`_INTRUSIVE`, `_WITH_DEFAULT`, `_ONLY_SERIALIZE` × intrusive/non-intrusive). The diagnostic is also unhelpful — g++ emits `error: unknown type name 'm1'` repeated 63 times, with nothing pointing at an argument-count limit.

This contradicts the shipped documentation: `docs/mkdocs/docs/api/macros/nlohmann_define_derived_type.md:60` states "up to 63 members can be given as a comma-separated list". It is therefore a public API break, and the PR description has no breaking-change section declaring it.

### Suggested fix

Strip the leading `Type` and delegate to `NLOHMANN_JSON_TYPE_TAG`, so the dispatch operates on `BaseType + members` and fits within the 64 slots:

```c
#define NLOHMANN_JSON_DERIVED_TYPE_TAG_(Type, ...) NLOHMANN_JSON_TYPE_TAG(__VA_ARGS__)
#define NLOHMANN_JSON_DERIVED_TYPE_TAG(...) NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_DERIVED_TYPE_TAG_(__VA_ARGS__))
```

then rename the six `NLOHMANN_JSON_DEFINE_DERIVED_TYPE_*_2(Type, BaseType)` bodies to `_1(Type, BaseType)` and re-amalgamate. This also removes the 16-line sentinel table above.

I applied and tested this locally: derived @ 63 members compiles again, and all 12 zero-member cases still pass at C++11/14/17/20 under `-pedantic -Werror -Wvariadic-macros` on both g++-16 and clang. One tradeoff — for the misuse case `NLOHMANN_DEFINE_DERIVED_TYPE_INTRUSIVE(D)` with no `BaseType`, the fix emits an extra `-Wc++20-extensions` warning ahead of the hard error, where the current form errors without it; both still fail to compile. MSVC's traditional preprocessor was not covered locally and needs CI confirmation, though the `NLOHMANN_JSON_EXPAND` wrapper is the same idiom used throughout this file.

Worth adding a regression test at the arity boundary too — the existing suite stops well short of the maximum, which is exactly why this slipped through.

nlohmann (commented):
> [!CAUTION]
> **Blocking — not ready to merge as-is.** GitHub does not allow a formal `Request changes` review on your own PR, so this is filed as a review comment; treat it as blocking.

The approach is sound — pure argument-count dispatch reusing the existing `NLOHMANN_JSON_GET_MACRO` idiom is the right call, and the reasoning for rejecting `__VA_OPT__` and `##__VA_ARGS__` holds up. Zero-member coverage is thorough (all 12 macros × `json`/`ordered_json`), the amalgamation is in sync, and the new `NLOHMANN_JSON_CAT` helpers don't collide with anything.

One blocking issue though: `NLOHMANN_JSON_DERIVED_TYPE_TAG` silently reduces the maximum member count for all six `NLOHMANN_DEFINE_DERIVED_TYPE_*` macros from 63 to 62, which is a public API break that also contradicts the shipped docs. Details and a tested fix inline.

Two smaller points, non-blocking:

- **Zero-member `from_json` accepts any JSON value.** It is a no-op, so `json(42).get<Zero>()`, `"str"`, `[1,2]` and `null` all succeed, whereas the 1+-member version throws `type_error.304` on a non-object. Defensible — there is nothing to read — but worth stating explicitly in the docs rather than leaving implicit.
- **API docs not updated.** `docs/mkdocs/docs/features/arbitrary_types.md` got the new note, but the three `docs/mkdocs/docs/api/macros/*.md` pages, where the parameter contract is actually specified, say nothing about zero members.

---
*This review was written by Claude Code.*

include/nlohmann/detail/macro_scope.hpp
nlohmann: Addressed in 0b28c63.

`NLOHMANN_JSON_DERIVED_TYPE_TAG` now drops the leading `Type` and defers to `NLOHMANN_JSON_TYPE_TAG`, so the tag is computed from `BaseType` plus the member list and fits within `NLOHMANN_JSON_GET_MACRO`'s 64 positional slots. The six `NLOHMANN_JSON_DEFINE_DERIVED_TYPE_*_2` bodies became `_1`, and the derived tag's sentinel table is gone (net -18 lines in `macro_scope.hpp`). The comment above now records why the derived tag delegates instead of shifting the table, so the cap does not get reintroduced.

Also added a regression test at the documented maximum — `max_members` (63 members, intrusive) and `max_members_derived` (63 own members over a base) — checked for key count and round-trip across `json` and `ordered_json`. It produces 20 compile errors against the previous dispatch, so it is a real guard rather than a passing-by-construction test.

Verified locally on clang and g++-16 at `-pedantic -Werror -Wvariadic-macros`:

| check | result |
|---|---|
| derived @ 63 members | OK (was FAIL) |
| base @ 63, derived @ 62 | OK, unchanged |
| base/derived @ 64 members | still rejected, as documented |
| zero-member, all 12 macros, runtime | OK at C++11/14/17/20/23 |
| `unit-udt_macro` full suite | 50/50 cases, `-Iinclude` and `-Isingle_include`, C++11/17/20 |
| `unit-udt.cpp` (macro-name collision check) | 11/11 |
| `make check-amalgamation` | clean |

MSVC's traditional preprocessor was not covered locally and is left to CI, though the `NLOHMANN_JSON_EXPAND` wrapper is the same idiom used throughout this file.

No documentation change was needed: the "up to 63 members" contract in `docs/mkdocs/docs/api/macros/nlohmann_define_derived_type.md` holds again. And since the PR no longer narrows the derived macros' capacity, there is no public API break left to declare.

---
*This reply was written by Claude Code.*

PR #5273, opened 2026-07-14
Fix out-of-bounds read in dump() for large indentation widths

Angadi56:
When pretty-printing, the serializer keeps one reusable `indent_string` buffer and widens it on demand before writing a run of indentation. The width it needs is `current_indent + indent_step`, which grows with nesting depth and with the indent argument given to `dump()`. The buffer only doubled once behind an `if`, so when the required width was more than twice the current size the buffer stayed too short and the following `write_characters(indent_string.c_str(), new_indent)` ran off its end, copying whatever heap bytes followed into the output. It reproduces under AddressSanitizer with `json j = {{"a", 1}}; j.dump(1100);`, which reads 1100 bytes out of a 1040-byte allocation. Looping the doubling until the buffer actually holds `new_indent` closes it while keeping the same amortized growth. I applied it to the three pretty-print branches (object, array, binary) that share the pattern, and added a serialization test that checks the emitted indentation is the expected run of spaces at a width past the first doubling.

### DIFF
--- a/include/nlohmann/detail/output/serializer.hpp
+++ b/include/nlohmann/detail/output/serializer.hpp
@@ -126,9 +126,9 @@ class serializer
 
                     // variable to hold indentation for recursive calls
                     const auto new_indent = current_indent + indent_step;
-                    while (JSON_HEDLEY_UNLIKELY(indent_string.size() < new_indent))
+                    if (JSON_HEDLEY_UNLIKELY(indent_string.size() < new_indent))
                     {
-                        indent_string.resize(indent_string.size() * 2, ' ');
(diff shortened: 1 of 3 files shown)


include/nlohmann/detail/output/serializer.hpp
```diff
@@ -126,7 +126,7 @@ class serializer
 
                     // variable to hold indentation for recursive calls
                     const auto new_indent = current_indent + indent_step;
-                    if (JSON_HEDLEY_UNLIKELY(indent_string.size() < new_indent))
+                    while (JSON_HEDLEY_UNLIKELY(indent_string.size() < new_indent))
```
gregmarr: Maybe instead go in a single step instead of growing over multiple calls.
```
indent_string.resize((std::max)(indent_string.size() * 2, new_indent));
```

PR #5274, opened 2026-07-15
reject CBOR array/map length equal to the indefinite-length marker

Angadi56:
CBOR arrays and maps can prefix their element count with a full eight-byte integer, and the reader tells a definite-length container from an indefinite one by comparing that count against `detail::unknown_size()`, which is `SIZE_MAX`. Nothing stops a definite count from being exactly `SIZE_MAX`, so on a 64-bit build a stream that declares `0xFFFFFFFFFFFFFFFF` elements narrows straight onto the sentinel and `get_cbor_array`/`get_cbor_object` read the container as indefinite rather than rejecting it. The effect is that bytes like `9B FF FF FF FF FF FF FF FF 01 02 FF` come back as `[1, 2]`, read up to the next `0xFF` break, even though the neighboring count `0xFFFFFFFFFFFFFFFE` is correctly turned away as excessive. I noticed the gap while checking how the eight-byte length path lines up with the indefinite marker. The fix adds a small `get_cbor_container_size` helper that narrows the length and rejects the one value that aliases the sentinel; such a count is larger than any container's `max_size()` anyway, so no representable input changes behavior. The four-byte and eight-byte array and map cases route through it, and the indefinite forms stay as they were.

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -1114,10 +1114,11 @@ class binary_reader
     /*!
     @brief narrow a definite CBOR array/map length to std::size_t
 
-    detail::unknown_size() is reserved to mark an indefinite-length container,
-    so a definite length that equals it would be read as indefinite instead of
-    as the (impossible) size it claims. Reject that value here; it also exceeds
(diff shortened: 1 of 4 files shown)


nlohmann (approved):
Looks good to me.

PR #5276, opened 2026-07-16
reject unpaired UTF-16 surrogates in wide-string input

Angadi56:
The wide-string input path in `wide_string_input_helper::fill_buffer` converts UTF-16 code units to UTF-8 one unit at a time. When it reaches a value in the surrogate range it assumes a high surrogate and reads the next unit to build a pair, without checking that the first unit is actually a high surrogate or that the second is a low surrogate. So a lone low surrogate, or a high surrogate followed by an ordinary character, is silently combined into a different code point, and the trailing unit is consumed even when it is the closing quote, which throws the rest of the parse off. I noticed it while comparing this path against the `\uXXXX` surrogate handling in the lexer, which already rejects these same malformed sequences with a parse error. The fix forms a pair only when a high surrogate is followed by a low surrogate and otherwise passes the offending unit through unchanged, so the UTF-8 decoder reports it the same way the escape path does. Valid input and the existing lone-high-surrogate-at-end case are unchanged.

Separately, this PR also fixes a pre-existing bug in the test suite itself: the `"invalid std::u16string"` section in `tests/src/unit-wstring.cpp` was gated on `wstring_is_utf16()` instead of `u16string_is_utf16()`. That is independent of the surrogate fix, but correcting it was necessary so the new `std::u16string` cases run under the intended guard.

### DIFF
--- a/tests/src/unit-wstring.cpp
+++ b/tests/src/unit-wstring.cpp
@@ -54,14 +54,26 @@ TEST_CASE("wide strings")
             json _;
             CHECK_THROWS_AS(_ = json::parse(w), json::parse_error&);
 
+            // the exact message depends on the width of wchar_t: a 16-bit
+            // wchar_t passes the lone surrogate to the UTF-8 decoder unchanged
+            // (rejected as a single ill-formed byte at column 2), while a
+            // 32-bit wchar_t first encodes it as an ill-formed three-byte
+            // sequence (rejected one byte later, at column 3)
+            const char* const error_low_surrogate = sizeof(wchar_t) == 2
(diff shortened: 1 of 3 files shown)


tests/src/unit-wstring.cpp
```diff
@@ -53,6 +53,15 @@ TEST_CASE("wide strings")
             std::wstring const w = L"\"\xDBFF";
             json _;
             CHECK_THROWS_AS(_ = json::parse(w), json::parse_error&);
+
+            // a lone low surrogate cannot start a pair
+            CHECK_THROWS_AS(_ = json::parse(std::wstring{L'"', static_cast<wchar_t>(0xDC00), L'"'}), json::parse_error&);
```
nlohmann: Please use the macro variant that also checks the exception message (CHECK_THROWS_AS_WITH?).

PR #5283, opened 2026-07-20
Speed up parsing of contiguous input (numbers, strings, UTF-8)

nlohmann:
## What & why

The parser reads input one character at a time through the input adapter and converts numbers with `strtod`/`strtoull`. For the common case of **contiguous byte input** (`std::string`, `std::vector<char>`, string literals, `const char*` ranges), the per-character `get()`/`add()` overhead and the locale/errno-heavy number conversion dominate. This PR adds fast paths for that case, borrowing ideas from the simdjson / fast_float ecosystem, while keeping the library **header-only** with a **C++11 fallback** on every path. Streaming inputs (files, `std::istream`, wide strings, user adapters) are untouched and keep the byte-at-a-time scanner. Every fast path defers to the byte-path scanner on anything malformed, so parse errors keep the same message and the same `byte` offset; the one deliberate exception is the reported column after a newline, described under *Public API impact* below.

## Stack

```
develop
└─ claude/parser-performance-research-4hkdf8    (this PR)
   └─ claude/dump-function-performance-hcj2bl   (#5285)
```

### The changes (each independent and reversible)

1. **Number conversion** — replace `strtoull`/`strtoll`/`strtod` over the validated digit buffer with dedicated parsers: hand-rolled integer parsing with overflow→float fallback (matching the old semantics exactly), and Clinger's exact fast path for `double`, bit-identical to `strtod` on its subset and declining otherwise. `float`/`long double` keep `strtof`/`strtold`.
2. **SWAR string scanning** — scan 8 bytes at a time with a `uint64_t` word test to bulk-append ordinary characters. Pure C++11, no intrinsics.
3. **UTF-8 bulk validation** — fold well-formed multibyte sequences into the bulk run (validated with the same ranges as the byte-path switch), so non-ASCII text no longer drops to the slow per-byte path.
4. **Contiguous-container pointer adapter** — route `std::string` & friends through a pointer-based adapter so 1–3 apply to `json::parse(std::string)` in **C++11–17** (previously only `const char*`/literals, and C++20 contiguous iterators, qualified). The pointer keeps the container's element type, so `char_type` — and all behavior, including binary formats from `std::vector` — is unchanged. Detection requires the container's own `value_type` to be the single-byte element `data()` points at, so a type whose `size()` counts something else keeps the iterator-based adapter instead of being read as `[data(), data() + size())` bytes.
5. **Contiguous number fast path** — parse the whole number token straight from the buffer (no per-byte `get()`/`add()`), materializing `token_buffer` in one copy with the locale decimal point substituted, then reusing the shared `convert_number()` tail.
6. **Optional simdutf backend** (`JSON_USE_SIMDUTF`, off by default) — since simdutf ships a `.cpp` and uses runtime dispatch, it is **not vendored**; when defined, bulk UTF-8 validation routes through `simdutf::validate_utf8` (the user supplies/links simdutf), otherwise the C++11 scalar validator is used. The library stays header-only either way. simdutf requires C++17 — its header rejects older standards with an `#error` — so the backend is gated on `JSON_HAS_CPP_17` and the macro is inert below that, which keeps it safe to set project-wide in a mixed-standard build.
7. **New byte-level headers** — the value conversion and string scanning helpers live in two new focused headers, `detail/input/number_parse.hpp` (302 lines) and `detail/input/string_scan.hpp` (241 lines), as free functions operating on raw bytes, so `lexer.hpp` keeps only the state machine and the adapter-facing glue. `lexer.hpp` itself grows (1760 → 2025 lines) with the bulk scanners and the number fast path; the only code moved out of it is the ~30-line `strtoull`/`strtoll`/`strtod` block replaced in item 1.
8. **Sized sentinels** — the bulk paths apply to any iterator/sentinel pair whose remaining length is computable in O(1), not just same-type pairs, so `std::counted_iterator` + `std::default_sentinel_t` reaches them too (per @gregmarr's review, matching #5268).

### Measured gains

Cumulative, `json::parse` / `accept(std::string)`, C++11, g++ 13 `-O3`, vs the v3.12.0 baseline (representative synthetic data):

| dataset | `parse` | `accept` |
|---|---|---|
| floats | 1.6× | 1.8× |
| ints | 1.2× | 1.3× |
| strings (twitter-like) | 1.3× | 1.7× |
| long ASCII strings | **8.5×** | **18×** |
| dense CJK | 2.4× | 2.3× |
| mixed international | 2.1× | 2.2× |

`accept()`/SAX/validation see the largest gains (not allocation-bound); full-DOM `parse()` gains on object-heavy input are capped by DOM allocation, which is a separate concern. With `JSON_USE_SIMDUTF`, dense-UTF-8 validation has further headroom (~4.7 GB/s ceiling measured with an SSE validator; ~7× the scalar path).

### Public API impact

**No breaking changes to the documented public API.** `parse()`, `accept()`, `sax_parse()`, and the `from_*()` signatures are unchanged.

Two internal `nlohmann::detail` changes are worth recording:

- `detail::input_adapter(container)` now returns a pointer-based adapter for contiguous byte containers, so `detail::string_input_adapter_type` changes from `iterator_input_adapter<std::string::iterator>` to `iterator_input_adapter<const char*>`. Source-compatible (these types are normally deduced), but the mangled names change, which matters to anyone shipping a precompiled wrapper that names them.
- `iterator_input_adapter` gains `supports_bulk_scan` / `bulk_data()` / `bulk_remaining()` / `bulk_skip()`, and `supports_seek` is relaxed from "same-type iterator/sentinel" to "distance computable in O(1)".

#### One deliberate behavior change: error column after a newline

`scan_number()` reads the character that terminates a number and then ungets it, so the reported column is the one reached after the number's last character. When that terminator is a newline, `get()` has already cleared `chars_read_current_line` and `unget()` could only restore `lines_read`, leaving the column at 0 — a number followed by a newline reported a different position than the same number followed by a space.

`unget()` now remembers the column the newline was read at, so **both** terminators report the same position:

```cpp
json::parse("[01\n]");             // was: line 1, column 0    now: line 1, column 3
json::parse("[01 ]");              //      line 1, column 3         line 1, column 3  (unchanged)

json::parse("null -2.5e3\nfalse"); // was: line 1, column 0    now: line 1, column 11
json::parse("null -2.5e3 false");  //      line 1, column 11        line 1, column 11 (unchanged)
```

Only the column changes — message text, `byte` offset, and accept/reject are identical, and the contiguous and streaming routes agree. 22 of 481k differential documents are affected, all of this shape.

### Verification

Differential testing against the merge base across all routes (`std::string` / `std::istream` / `std::vector` / `accept`), C++11 and C++17, clang/libc++ and g++/libstdc++, in the `C` and `de_DE`/`fr_FR` locales:

- 481k mixed documents: identical values, accept/reject, `byte` offsets, and message text; **zero disagreement between the contiguous and streaming routes** across 298,333 error documents.
- 900k hard floating-point cases (subnormals, 15–25 significant digits, the Clinger boundary at |scale| = 22/23, significands near 2^53, over/underflow edges): bit-identical to `strtod`, both with `std::from_chars` active under libstdc++ and under `-ffast-math`.
- `JSON_DIAGNOSTIC_POSITIONS` start/end offsets: identical over 60k documents.
- ASan + UBSan clean over the full corpus and over counted-iterator inputs whose count ends before the underlying buffer does.
- Millions of differential-fuzz documents (arbitrary bytes; overlong/surrogate/truncated UTF-8; overflow-range integers, long digit strings, `%.17g` doubles; escapes) parse and accept/reject identically; the `JSON_USE_SIMDUTF` build matches too.
- Full test suite: **109/109**, warning-clean on g++ and clang, C++11/17/20, in both build configurations; clang-tidy clean on the new headers; `make check-amalgamation` clean.
- With `JSON_TestSimdutf=ON`, the whole suite (162 tests, built at C++11 and C++17) passes against the simdutf backend, and its results are byte-identical to the scalar validator over the differential corpus.

### Notes for reviewers / remaining work

- Dedicated `tests/` cases cover the number fast path (including an exhaustive grammar-parity sweep over the number alphabet), the string and UTF-8 bulk scanners (exhaustive contiguous-vs-streaming parity, special bytes at every offset of the SWAR stride, and the boundaries of every accepted UTF-8 range), contiguous-container detection, the counted-iterator bulk paths (including counts that end before the underlying buffer), and contiguous-vs-streaming error-position parity.
- `JSON_TestSimdutf` (off by default) builds the unit tests against simdutf, fetched with `FetchContent`; `ci_test_simdutf` runs the suite in that configuration and is part of the `ci_cmake_options` matrix. It warns and falls back to the scalar validator when no tested standard can reach simdutf. The **library** target and the installed CMake package are deliberately untouched: making `nlohmann_json` link `simdutf::simdutf` would add a `find_dependency(simdutf)` to the exported package, which is a separate decision.
- The `JSON_USE_SIMDUTF` doc page says "Added in version 3.13.0", matching the other pages documenting unreleased features — adjust if this lands in a different release.

---

- [ ] If applicable, an existing issue is referenced. _(none)_
- [ ] The code coverage remained at 100%. A test case for every new line of code. _(targeted tests now cover the number fast path, the string/UTF-8 bulk scanners, contiguous-container detection, the counted-iterator paths, and error-position parity; a full coverage run has not been done)_
- [x] If applicable, the documentation is updated. _(`JSON_USE_SIMDUTF` macro page, nav entry, and supported-macros overview)_

---


include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -231,6 +231,33 @@ class iterator_input_adapter
         std::is_same<IteratorType, SentinelType>::value && std::is_pointer<IteratorType>::value;
 #endif
 
+  public:
+    // Whether the remaining input is a single contiguous block of 1-byte
+    // elements that the lexer can inspect directly (used for the SWAR string
+    // fast path). Restricted to same-type iterator/sentinel pairs so that plain
+    // std::distance/std::advance are well-defined in all standards.
```
gregmarr: No support for counted iterators?

wilkolbrzym-coder:
# Quick review notes – PR #5283

- **No critical bugs found** – the implementation is correct. The pointer arithmetic in `scan_number_bulk_contiguous()` is safe because `bulk_data()` is only called after the current character has been consumed.

- **SWAR string scanner & UTF‑8 validation** – correctly implemented; the UTF‑8 range checks exactly match the byte‑by‑byte scanner, so error messages and positions stay identical.

- **Number parsers** – handle overflow correctly and fall back to float when needed, preserving legacy behaviour.

- **Simdutf integration** – opt‑in and safe; on validation failure it falls back to the scalar path.

---

**What's missing (should be addressed before merge):**

1. **Dedicated unit tests** for the bulk string/UTF‑8 scanning and for the container adapter (e.g., `std::vector<uint8_t>`). Currently covered only by fuzzing – explicit test cases would lock in the behaviour.

2. **CMake option for `JSON_USE_SIMDUTF`** – not included yet; users must manually add the header and link the library.

3. **Support for counted iterators** – the adapter uses `std::distance(current, end)` and assumes iterators of the same type. Niche case, worth noting but not a blocker.

include/nlohmann/detail/input/input_adapters.hpp
wilkolbrzym-coder: Right – counted iterators aren't supported here. But the fast path only applies to containers with `.data()`/`.size()` (true contiguous storage). Everything else falls back to the generic byte‑at‑a‑time scanner, so correctness is preserved. Supporting counted iterators would add complexity for a niche case.

gregmarr: https://github.com/nlohmann/json/pull/5268 recently added counted iterator support for a similar case.

include/nlohmann/detail/input/lexer.hpp
```diff
@@ -1279,45 +1346,78 @@ class lexer : public lexer_base<BasicJsonType>
...
+    The digit sequence in [first, last) has been validated by the caller, so a
+    dedicated parser can avoid the locale/errno overhead of std::strtoull.
 
-            if (errno != ERANGE)
+    @return the token type on success; token_type::uninitialized if @a
+            number_type is not an integer type or the value does not fit, in
+            which case the caller falls back to the floating-point conversion
+            (matching the previous std::strtoull/std::strtoll behavior)
+    */
+    token_type convert_integer(token_type number_type, const char* first, const char* last)
```
gregmarr, line 1378: `convert_integer()` was inserted between `convert_number()` and its documentation.

gregmarr: Resolved

docs/mkdocs/docs/api/macros/json_use_simdutf.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Examples
+
+??? example
+
+    The code below enables the simdutf backend for UTF-8 validation.
+
+    ```cpp
+    #define JSON_USE_SIMDUTF 1
+    #include <simdutf.h>
```
nlohmann:
```suggestion
```

In the actual code, `<simdutf.h>` is already being included. So we should not make it look as if it is required to be included before using the library.

include/nlohmann/detail/input/string_scan.hpp
```diff
@@ -0,0 +1,237 @@
...
+#if defined(JSON_USE_SIMDUTF)
+    const std::size_t run = find_string_delimiter(data, n);
+    if (run != 0 && simdutf::validate_utf8(reinterpret_cast<const char*>(data), run))
+    {
+        return run;
+    }
+    return scalar_string_bulk_run(data, n);
+#else
+    return scalar_string_bulk_run(data, n);
+#endif
```
nlohmann: This looks redundant.

include/nlohmann/detail/input/input_adapters.hpp
nlohmann: Done.

gregmarr (approved):

PR #5284, opened 2026-07-20
reject negative UBJSON/BJData string length

Angadi56:
UBJSON and BJData write string and object-key lengths with signed integer markers (i, I, l, L). `get_ubjson_string` reads that marker into a signed value and hands it straight to `get_string`, whose read loop runs only while the count is positive, so a negative length reads nothing and quietly produces an empty string. A corrupt or crafted document such as `S i FF` then decodes as `""` and the bytes that were meant to follow get reinterpreted as the next value, so the parser accepts malformed input and returns a different document rather than reporting an error. I found it while lining this path up against `get_ubjson_size_value`, which already rejects a negative optimized-container count with `parse_error.113`. The fix applies the same non-negative check to the four signed string-length markers before the bytes are read, so a negative length now fails in both UBJSON and BJData. A zero length still yields an empty string and the unsigned markers were never affected.

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -1847,6 +1847,29 @@ class binary_reader
     }
 
     /*!
+    @brief reject a negative UBJSON/BJData string length
+
+    String and key lengths are written with signed integer markers (i, I, l,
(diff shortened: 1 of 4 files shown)


nlohmann (approved):
Looks good to me.

Issue #2217, opened 2020-06-24
Getting Qt types to work


pylvain:
following https://github.com/nlohmann/json/issues/2216
I try to make Qt types work with the lib
I did this : 
```C++
using qtjson = nlohmann::basic_json<QMap,QList,QString>;
```
It compiles
 but this doesn't
```C++
qtjson j;
```
The compiler yells at me things like
```
json.hpp:16302: error: wrong number of template arguments (4, should be 2)
json.hpp:16978: error: wrong number of template arguments (4, should be 2)
json.hpp:17027: error: ‘const union nlohmann::basic_json<QMap, QList, QString>::json_value’ has no member named ‘object’
```

I'm a bit lost, do I have do implement functions ? Maybe did i forget to specify a type somewhere ?
Thanks in advance !

pylvain:
I made that to convert between Qt Types and standard types https://gist.github.com/pylvain/d8127d4a3da206fcd49dd05463a4cc07, should anyone need it.
The lib takes 400ms to parse a 4MB json file. Converting back and fourth between QJsonType and json::value_type takes 800ms. It's not awful, i will use that, that's good enough for me.

Just out of curiosity, has anyone any idea on how to improve performances ?
(I will close after that)

PR #5285, opened 2026-07-20
Speed up dump(), and keep it from overflowing the stack

### LINKED ISSUE #5413: dump() performs avoidable per-call heap allocations (eager 512-byte indent_string + shared_ptr output adapter) — ~17% overhead on the dump-many-small-values pattern
### Description

Every `dump()` call constructs a fresh `serializer` and output adapter, and two of the resulting heap allocations are avoidable for the extremely common case of serializing many small values (logging, per-record/message serialization, `dump()` in a loop). These are fixed per-call costs, so they dominate when the output itself is small.

**1. The 512-byte `indent_string` is allocated even for compact output.** The serializer constructor unconditionally runs:

```cpp
// include/nlohmann/detail/output/serializer.hpp  (constructor)
, indent_string(512, indent_char)   // <-- 512-byte std::string, heap-allocated every dump()
```

`indent_string` ([member at `serializer.hpp:1014`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/output/serializer.hpp)) is only ever read inside the `pretty_print` branches. For a **compact** `dump()` (the default, `indent = -1`) it is never used, yet the 512-byte buffer is allocated and freed on every call. (`number_buffer` and `string_buffer` are `std::array` — on the stack — so only `indent_string` heap-allocates.)

**2. The output adapter is heap-allocated via `make_shared` every call.** `dump()` builds `detail::output_adapter<char, string_t>(result)`, whose constructor does ([`output_adapters.hpp:135`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/output/output_adapters.hpp)):

```cpp
: oa(std::make_shared<output_string_adapter<CharType, StringType>>(s)) {}
```

and `output_adapter_t = std::shared_ptr<output_adapter_protocol<CharType>>` ([`:45`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/output/output_adapters.hpp)). So each `dump()` heap-allocates the adapter + control block, and the serializer then makes a virtual call per write through the `shared_ptr`.

### Measurement

`dump()` (compact, default) of a small 3-field object `{"level":"info","msg":"hello world","id":12345}`, g++ 13.3.0 `-O2`, 500k iterations, global `operator new` counting:

| version | time | allocations |
|---|---|---|
| baseline | 313.0 ns/call | 4.00 |
| lazy `indent_string` (only) | 259.5 ns/call | 3.00 |

Making `indent_string` lazy alone is **~17% faster and removes one of four allocations**. Removing the `shared_ptr` adapter would remove a second.

### Minimal benchmark

```cpp
#include <new>
#include <cstdlib>
#include <cstdio>
static long allocs=0; static bool count=false;
void* operator new(std::size_t n){ if(count)++allocs; void*p=std::malloc(n); if(!p)throw std::bad_alloc(); return p; }
void operator delete(void*p)noexcept{std::free(p);}
void operator delete(void*p,std::size_t)noexcept{std::free(p);}
#include <nlohmann/json.hpp>
using json = nlohmann::json;
int main(){
    json j = { {"level","info"}, {"msg","hello world"}, {"id",12345} };
    count=true; allocs=0;
    for(long i=0;i<500000;i++){ auto s=j.dump(); (void)s; }
    std::printf("%.2f allocs/call\n", (double)allocs/500000);   // 4.00 on develop
}
```

### Suggested direction

**`indent_string` (clean, high-value):** allocate it only when pretty-printing. Note the current resize logic in the three pretty branches (`indent_string.resize(indent_string.size() * 2, ' ')`) assumes a 512-byte preallocation and hard-codes a space fill — which is exactly the code **PR #5186** rewrites (it also fixes the `indent_char`-vs-space bug and the large-indent overflow). Because a lazy `indent_string` must change that resize logic, this is best folded into #5186: initialize `indent_string` empty and grow/fill it on first pretty use. (Verified: a lazy fill using `indent_char` produces byte-identical output to today except in the beyond-512 tab-indent case, which is the #5186 bug being corrected.)

**`shared_ptr` adapter (larger, overlaps #5285):** the per-call `make_shared` and the virtual-dispatch-per-write could both be removed by templating the serializer on the concrete adapter type instead of holding a `shared_ptr<output_adapter_protocol>`. This overlaps the devirtualization goal of the open dump-performance PR #5285 (which buffers writes to cut virtual-call *count* but, per its description, still flushes through the adapter) — so it should be coordinated with that PR rather than done independently.

### Behavior preservation

Output is unchanged (the `indent_string` change is byte-identical except where it corrects the #5186 fill-char bug). Both changes are internal to `detail`.

### Library version

nlohmann:
## What & why

`dump()` had two hot spots that mirror exactly what #5283 fixed on the parse side, so this PR reuses that PR's SWAR/UTF-8 primitives (`detail/input/string_scan.hpp`) to fix the serializer. **Output is byte-for-byte unchanged** on every input; only the path taken to produce it is faster.

## Stack

```
develop
└─ claude/parser-performance-research-4hkdf8    (#5283)  ← base of this PR
   └─ claude/dump-function-performance-hcj2bl   (this PR)
```

> **Note on the base branch:** this is stacked on `claude/parser-performance-research-4hkdf8` (#5283), because it reuses that PR's `string_scan.hpp` scanner and validator. The diff here is only the serializer + tests; once #5283 lands on `develop`, this can be retargeted to `develop` unchanged.

### The changes (each independent and reversible)

1. **SWAR bulk string escaping (`ensure_ascii == false`)** — `dump_escaped()` ran the UTF-8 DFA over *every* byte of every string and key, even for ordinary text with nothing to escape (the serializer twin of the parser's per-character `get()`). At a character boundary it now bulk-copies the longest run needing no escaping via `string_bulk_run()` — the same scanner + UTF-8 bulk validator the lexer's contiguous path uses — and drops to the byte-at-a-time DFA only for the first byte that needs individual handling (a quote, backslash, control char, or ill-formed/truncated UTF-8). For `ensure_ascii == false`, `string_bulk_run()`'s stop set is an exact match for the bytes `dump_escaped` copies verbatim, so escaping and error handling (including strict-mode error 316 position and message) are identical. Picks up `JSON_USE_SIMDUTF` for free.

2. **Internal write buffer (devirtualization)** — every structural character (`{`, `"`, `,`, …) previously went straight to the output adapter through a virtual call. All writes now route through `put_char`/`put_chars` into a 1 KiB buffer flushed in bulk; the public `dump()` flushes once the top-level value is serialized (the recursive worker is split out as `dump_internal`). Runs larger than the buffer are written straight through, so large payloads are not copied twice. This is the dominant cost for object/array-heavy values.

3. **`ensure_ascii == true` fast path** — added `find_ascii_copyable_run()` (a SWAR scan stopping at `"`, `\`, `< 0x20`, `0x7F`, and `>= 0x80`) so runs of printable ASCII are bulk-copied even when non-ASCII must be `\u`-escaped. `0x7F`/DEL is a deliberate stop (it is escaped under `ensure_ascii`), which is why `find_string_special()` could not be reused directly here.

4. **Bounded descent, so `dump()` cannot overflow the stack (#5387)** — serializing a container serialized its elements, so `dump()` descended into one call per nesting level and a value nested deeply enough terminated the process with `SIGSEGV`, no exception to catch. Parsing such a value works (the parser is iterative) and so does destroying one (#1436); serializing never got the same treatment. The descent is now bounded: the first 128 levels are written by exactly the code that always wrote them, and only below that does a new `dump_iteratively` write out what is left, keeping the containers it has entered on an explicit stack.

    Writing *every* value that way instead — the obvious fix — measured **2% to 20% slower**, 20% on object-heavy documents, which would have undone much of what this PR is for. Keeping the descent for everything that can afford it costs one comparison per container: between **-1.4% and +1.2%** across compact and pretty output of number, integer, string, object-heavy, wide-object and deeply nested documents.

    Both writers emit the separator *in front of* every element but the first, rather than after every element but the last; that puts exactly one between each pair and none at the end.

5. **`ensure_ascii` folded into the escaper** — `dump_escaped()` took it as a runtime flag and tested it inside the loop, once per character run, although it cannot change while a string is written. It is now a template parameter dispatched once per string, which folds the choice of scanner and lets each of the two be inlined into a loop of its own. This is the hottest loop in the serializer — it runs over every string *and every object key*.

6. **Binary bytes no longer go through `dump_integer`** — a byte is always in `[0, 255]`, so it needs neither digit counting nor 64-bit arithmetic; `dump_byte` writes the at most three digits straight into the write buffer. Any byte type that is not a plain unsigned byte is still left to `dump_integer`, whose representation of it may differ.

    Gains from 5 and 6 (medians of 9 interleaved runs, clang -O3, against the commit before them):

    | workload | |
    |---|---|
    | binary values | **-33.8%** |
    | dense CJK, `ensure_ascii=true` | **-20.6%** |
    | small value dumped in a loop | -21.4% |
    | deeply nested, pretty | -17.9% |
    | key-heavy objects | -17.8% |
    | dense CJK, `ensure_ascii=false` | -11.8% |
    | object-heavy, compact / pretty | -9.3% / -9.5% |
    | wide objects | -2.3% |
    | integer and float arrays | unchanged |
    | **arrays of plain ASCII strings** | **+3.5% to +4.2%** |

    The last row is the one shape that loses, consistently across two independent benchmarks. It is a small cost against the rest, but it is a real one.

    **Tried and dropped:** leaving the write and string buffers uninitialized instead of zeroing 1.5 KB per `dump()` call. Worth -30% on small values, but two nearly identical string workloads moved 18% apart *in opposite directions*, so the measurement did not support the change — and it would have traded away safety margin that cannot be checked here (MemorySanitizer is unavailable on arm64 macOS).

### Measured gains

`json::dump()`, C++17, g++ 13 `-O3`, vs the pre-change serializer (representative synthetic data):

| dataset | speedup |
|---|---|
| long ASCII strings, `ensure_ascii=false` | 4.2× |
| long ASCII strings, `ensure_ascii=true` | 4.1× |
| twitter-like objects | 2.7× |
| pretty-printed objects | 2.4× |
| dense CJK | 1.8× (further headroom with `JSON_USE_SIMDUTF`) |
| integer arrays | 1.1× |

### Verification

- **Differential** (used for items 4-6): every check below is byte-for-byte identical against the pre-change serializer, and now also covers **every one of the 256 byte values**, alone and together, in both binary layouts.
- **Differential for the bounded descent:** ~980k lines of output compared against the pre-change serializer and found identical — 700 randomized values plus curated shapes, each dumped in every combination of six indents (including one wider than the write buffer), both indent characters, both `ensure_ascii` settings and all three error handlers; plus **every nesting depth from 1 to 300** in array, object and mixed shapes at three indent settings, so the depth at which the two writers hand over is covered from every side. New tests cover a 100,000-deep value (which crashes on the base), the depths around the bound, pretty-printing across it, and empty containers reached below it.
- **Differential:** dump output is byte-for-byte identical to the pre-change implementation across ~20k randomized byte strings plus curated edge cases (all escapes, control chars, `0x7F`, valid multibyte, surrogates, overlong, truncated sequences), for compact/pretty and object/array output, both `ensure_ascii` settings, and all three error handlers, in C++11/17/20 at `-O2`/`-O3`.
- New serialization tests cover the write-buffer flush boundaries (>1 KiB strings, 2000-element arrays, 1100-deep nesting), escape and `0x7F` handling, multibyte under both settings, and invalid-UTF-8 handling.
- Warning-clean under clang `-Weverything` and the gcc pedantic flag set; clang-tidy clean on the changed headers; `make check-amalgamation` clean.

## How other libraries handle this

Worth knowing where this sits, because the trade-off taken here is deliberately not the common one.

**Most libraries reject deeply nested input rather than support it** — a hard depth limit at parse time, which then protects everything downstream:

| library | limit |
|---|---|
| [Boost.JSON](https://www.boost.org/doc/libs/1_83_0/libs/json/doc/html/json/ref/boost__json__parse_options.html) | `parse_options::max_depth` = 32 |
| [serde_json](https://github.com/serde-rs/json/issues/162) | 128, then `RecursionLimitExceeded` |
| [Jackson](https://github.com/FasterXML/jackson-core/pull/943) | `maxNestingDepth` 1000 (500 in 3.0) |
| [simdjson](https://simdjson.org/api/0.9.0/md_doc_basics.html) | 1024, refuses to parse deeper |
| .NET `System.Text.Json` | 65 |

Jackson's limit exists because of [CVE-2025-52999](https://www.herodevs.com/blog-posts/cve-2025-52999-denial-of-service-via-stack-overflow-in-jackson-core), and [CVE-2026-29062](https://github.com/advisories/GHSA-6v53-7c9g-w56r) covers a parser that bypassed it; PostgreSQL [patched its OAuth JSON parser](https://www.postgresql.org/message-id/CAJ7c6TO_MwMnpuw3%2Bdub%2BGif7c2tkS90YjHbyR%2BtmzpSP_ooXw%40mail.gmail.com) the same way. Only a few libraries are iterative throughout instead: [yyjson](https://github.com/ibireme/yyjson) advertises "unlimited JSON nesting levels" as a headline feature, and [miniserde](https://docs.rs/miniserde) gives its value type a non-recursive `Drop`. A third approach is to grow the stack on demand, as [serde_stacker](https://github.com/dtolnay/serde-stacker) does.

**This library sits in the worst spot of the three.** Its parser is iterative, so it accepts input nested arbitrarily deeply — and then hands it to operations that cannot cope. [lovasoa/bad_json_parsers](https://github.com/lovasoa/bad_json_parsers), which measures exactly this, buckets it under *"unlimited → segfault"*. A library with a parse limit never builds such a value in the first place.

**The closest C++ peer has the same split and has fixed none of it.** RapidJSON's `kParseIterativeFlag` makes only *parsing* iterative, and is opt-in because the recursive parser is faster; its destructor is still recursive ([#2217](https://github.com/Tencent/rapidjson/issues/2217), open, asking for exactly what #1436 did here in 2019), `CopyFrom` still overflows on deep values, and `Value::Accept()` — its serializer — still recurses.

Two things that supports about the approach taken here: RapidJSON keeps recursion as the *default because it is faster*, which matches the measurement that writing every value iteratively costs up to 20%; and a bound of 128 sits in the same range as everyone else's hard limits — except that nothing is rejected at it, so this library stays strictly more permissive than all of them while no longer crashing.

### Public API impact

**No breaking changes to the documented public API.** `dump()`, `operator<<`, the `error_handler_t` values, and every serialization option behave exactly as before, and the output is byte-for-byte identical on every input.

Two internal signatures change. They live in `nlohmann::detail` and are not part of the documented API, so this only matters to code that reaches into that namespace:

- `detail::serializer::dump()`'s `indent_step` and `current_indent` parameters change from `unsigned int` to `std::size_t`. Callers convert implicitly, so this is source-compatible, but the mangled name changes — the same caveat class as the adapter type change documented in #5283.
- `dump_escaped()` becomes a template on `ensure_ascii`, and the recursive worker is split out as `dump_internal()`.

One behavior change, and it is strictly permissive: `dump()` on a sufficiently deeply nested value previously terminated the process with `SIGSEGV` (#5387), and now serializes it successfully. Nothing that used to serialize stops working, no new exception is introduced, and no depth is rejected.

---

- [ ] If applicable, an existing issue is referenced. _(none; follow-up to #5283)_
- [x] The code coverage remained at 100%. A test case for every new line of code.
- [ ] If applicable, the documentation is updated. _(no public API or option changes)_

---

### DIFF
--- a/include/nlohmann/detail/input/string_scan.hpp
+++ b/include/nlohmann/detail/input/string_scan.hpp
@@ -87,6 +87,58 @@ inline std::size_t find_string_special(const unsigned char* data, std::size_t n)
     return n;
 }
 
+// classify a byte as one the serializer must NOT copy verbatim when
+// ensure_ascii is requested: the closing quote, an escape, a control character
+// (< 0x20), DEL (0x7F), or any non-ASCII byte (>= 0x80). Everything else -
+// printable ASCII except '"' and '\\' - is emitted unchanged. Note this differs
+// from is_string_special() only in that 0x7F is also a stop (it is escaped as
+// \u007f under ensure_ascii).
+inline bool is_ascii_copyable(unsigned char c) noexcept
+{
+    return c >= 0x20u && c < 0x7Fu && c != '\"' && c != '\\';
+}
+
+// return the index of the first byte in [data, data+n) that is NOT
+// is_ascii_copyable(), or n if every byte can be copied verbatim; scans 8 bytes
+// at a time. Used by the serializer's ensure_ascii fast path.
+inline std::size_t find_ascii_copyable_run(const unsigned char* data, std::size_t n) noexcept
+{
+    constexpr std::uint64_t ones = 0x0101010101010101ull;
+    constexpr std::uint64_t high = 0x8080808080808080ull;
+    std::size_t i = 0;
+    for (; i + 8 <= n; i += 8)
+    {
+        std::uint64_t v = 0;
+        std::memcpy(&v, data + i, sizeof(v));
+        const std::uint64_t q = v ^ 0x2222222222222222ull; // '"'  (0x22)
+        const std::uint64_t b = v ^ 0x5C5C5C5C5C5C5C5Cull; // '\\' (0x5C)
+        const std::uint64_t d = v ^ 0x7F7F7F7F7F7F7F7Full; // DEL  (0x7F)
+        const std::uint64_t stop = ((q - ones) & ~q & high)   // == '"'
+                                   | ((b - ones) & ~b & high)  // == '\\'
+                                   | ((d - ones) & ~d & high)  // == 0x7F
+                                   | ((v - 0x2020202020202020ull) & ~v & high) // < 0x20
+                                   | (v & high);               // >= 0x80
+        if (stop != 0)
+        {
+            for (std::size_t j = 0; j < 8; ++j)
+            {
+                if (!is_ascii_copyable(data[i + j]))
+                {
+                    return i + j;
+                }
+            }
+        }
+    }
+    for (; i < n; ++i)
+    {
+        if (!is_ascii_copyable(data[i]))
+        {
+            return i;
+        }
+    }
+    return n;
+}
+
 // Validate one UTF-8 sequence at the front of [data, data+avail). Returns its
 // length (2..4) only when the bytes form a *well-formed* sequence using exactly
 // the same ranges as scan_string()'s per-byte switch, so the bulk path accepts
--- a/include/nlohmann/detail/input/string_scan.hpp
+++ b/include/nlohmann/detail/input/string_scan.hpp
@@ -93,6 +93,58 @@ inline std::size_t find_string_special(const unsigned char* data, std::size_t n)
     return n;
 }
 
+// classify a byte as one the serializer must NOT copy verbatim when
+// ensure_ascii is requested: the closing quote, an escape, a control character
+// (< 0x20), DEL (0x7F), or any non-ASCII byte (>= 0x80). Everything else -
+// printable ASCII except '"' and '\\' - is emitted unchanged. Note this differs
+// from is_string_special() only in that 0x7F is also a stop (it is escaped as
+// \u007f under ensure_ascii).
+inline bool is_ascii_copyable(unsigned char c) noexcept
+{
+    return c >= 0x20u && c < 0x7Fu && c != '\"' && c != '\\';
+}
+
+// return the index of the first byte in [data, data+n) that is NOT
+// is_ascii_copyable(), or n if every byte can be copied verbatim; scans 8 bytes
+// at a time. Used by the serializer's ensure_ascii fast path.
+inline std::size_t find_ascii_copyable_run(const unsigned char* data, std::size_t n) noexcept
+{
+    constexpr std::uint64_t ones = 0x0101010101010101ull;
+    constexpr std::uint64_t high = 0x8080808080808080ull;
+    std::size_t i = 0;
+    for (; i + 8 <= n; i += 8)
+    {
+        std::uint64_t v = 0;
+        std::memcpy(&v, data + i, sizeof(v));
+        const std::uint64_t q = v ^ 0x2222222222222222ull; // '"'  (0x22)
+        const std::uint64_t b = v ^ 0x5C5C5C5C5C5C5C5Cull; // '\\' (0x5C)
+        const std::uint64_t d = v ^ 0x7F7F7F7F7F7F7F7Full; // DEL  (0x7F)
+        const std::uint64_t stop = ((q - ones) & ~q & high)   // == '"'
+                                   | ((b - ones) & ~b & high)  // == '\\'
+                                   | ((d - ones) & ~d & high)  // == 0x7F
+                                   | ((v - 0x2020202020202020ull) & ~v & high) // < 0x20
+                                   | (v & high);               // >= 0x80
+        if (stop != 0)
+        {
+            for (std::size_t j = 0; j < 8; ++j)
+            {
+                if (!is_ascii_copyable(data[i + j]))
+                {
+                    return i + j;
+                }
+            }
+        }
+    }
+    for (; i < n; ++i)
+    {
+        if (!is_ascii_copyable(data[i]))
+        {
+            return i;
+        }
+    }
+    return n;
+}
+
 // Validate one UTF-8 sequence at the front of [data, data+avail). Returns its
 // length (2..4) only when the bytes form a *well-formed* sequence using exactly
 // the same ranges as scan_string()'s per-byte switch, so the bulk path accepts
--- a/include/nlohmann/detail/input/string_scan.hpp
+++ b/include/nlohmann/detail/input/string_scan.hpp
@@ -101,7 +101,7 @@ inline std::size_t find_string_special(const unsigned char* data, std::size_t n)
 // \u007f under ensure_ascii).
 inline bool is_ascii_copyable(unsigned char c) noexcept
 {
-    return c >= 0x20u && c < 0x7Fu && c != '\"' && c != '\\';
+    return c >= 0x20u && c < 0x7Fu && c != '"' && c != '\\';
 }
 
 // return the index of the first byte in [data, data+n) that is NOT
@@ -126,13 +126,7 @@ inline std::size_t find_ascii_copyable_run(const unsigned char* data, std::size_
                                    | (v & high);               // >= 0x80
         if (stop != 0)
         {
-            for (std::size_t j = 0; j < 8; ++j)
-            {
-                if (!is_ascii_copyable(data[i + j]))
-                {
-                    return i + j;
-                }
-            }
+            break;
         }
     }
     for (; i < n; ++i)
--- a/include/nlohmann/detail/output/serializer.hpp
+++ b/include/nlohmann/detail/output/serializer.hpp
@@ -16,6 +16,7 @@
 #include <cstddef> // size_t, ptrdiff_t
 #include <cstdint> // uint8_t
 #include <cstdio> // snprintf
+#include <cstring> // memcpy
 #include <limits> // numeric_limits
 #include <string> // string, char_traits
 #include <iomanip> // setfill, setw
@@ -111,19 +112,38 @@ class serializer
               const unsigned int indent_step,
               const unsigned int current_indent = 0)
     {
+        dump_internal(val, pretty_print, ensure_ascii, indent_step, current_indent);
+        flush();
+    }
+
+  JSON_PRIVATE_UNLESS_TESTED:
+    /*!
+    @brief recursive worker for @ref dump
+
+    Identical in behavior to the historical @ref dump, but writes into the
+    serializer's internal @ref write_buffer instead of issuing a virtual call
+    per token. The public @ref dump wraps this and flushes the buffer once the
+    top-level value has been serialized.
+    */
+    void dump_internal(const BasicJsonType& val,
+                       const bool pretty_print,
+                       const bool ensure_ascii,
+                       const unsigned int indent_step,
+                       const unsigned int current_indent = 0)
+    {
         switch (val.m_data.m_type)
         {
             case value_t::object:
             {
                 if (val.m_data.m_value.object->empty())
                 {
-                    o->write_characters("{}", 2);
+                    put_chars("{}", 2);
                     return;
                 }
 
                 if (pretty_print)
                 {
-                    o->write_characters("{\n", 2);
+                    put_chars("{\n", 2);
 
                     // variable to hold indentation for recursive calls
                     const auto new_indent = current_indent + indent_step;
@@ -136,51 +156,51 @@ class serializer
                     auto i = val.m_data.m_value.object->cbegin();
                     for (std::size_t cnt = 0; cnt < val.m_data.m_value.object->size() - 1; ++cnt, ++i)
                     {
-                        o->write_characters(indent_string.c_str(), new_indent);
-                        o->write_character('\"');
+                        put_chars(indent_string.c_str(), new_indent);
+                        put_char('\"');
                         dump_escaped(i->first, ensure_ascii);
-                        o->write_characters("\": ", 3);
-                        dump(i->second, true, ensure_ascii, indent_step, new_indent);
-                        o->write_characters(",\n", 2);
+                        put_chars("\": ", 3);
+                        dump_internal(i->second, true, ensure_ascii, indent_step, new_indent);
+                        put_chars(",\n", 2);
                     }
 
                     // last element
                     JSON_ASSERT(i != val.m_data.m_value.object->cend());
                     JSON_ASSERT(std::next(i) == val.m_data.m_value.object->cend());
-                    o->write_characters(indent_string.c_str(), new_indent);
-                    o->write_character('\"');
+                    put_chars(indent_string.c_str(), new_indent);
+                    put_char('\"');
                     dump_escaped(i->first, ensure_ascii);
-                    o->write_characters("\": ", 3);
-                    dump(i->second, true, ensure_ascii, indent_step, new_indent);
+                    put_chars("\": ", 3);
+                    dump_internal(i->second, true, ensure_ascii, indent_step, new_indent);
 
-                    o->write_character('\n');
-                    o->write_characters(indent_string.c_str(), current_indent);
-                    o->write_character('}');
+                    put_char('\n');
+                    put_chars(indent_string.c_str(), current_indent);
+                    put_char('}');
                 }
                 else
                 {
-                    o->write_character('{');
+                    put_char('{');
 
                     // first n-1 elements
                     auto i = val.m_data.m_value.object->cbegin();
                     for (std::size_t cnt = 0; cnt < val.m_data.m_value.object->size() - 1; ++cnt, ++i)
                     {
-                        o->write_character('\"');
+                        put_char('\"');
                         dump_escaped(i->first, ensure_ascii);
-                        o->write_characters("\":", 2);
-                        dump(i->second, false, ensure_ascii, indent_step, current_indent);
-                        o->write_character(',');
+                        put_chars("\":", 2);
+                        dump_internal(i->second, false, ensure_ascii, indent_step, current_indent);
+                        put_char(',');
                     }
 
                     // last element
                     JSON_ASSERT(i != val.m_data.m_value.object->cend());
                     JSON_ASSERT(std::next(i) == val.m_data.m_value.object->cend());
-                    o->write_character('\"');
+                    put_char('\"');
                     dump_escaped(i->first, ensure_ascii);
-                    o->write_characters("\":", 2);
-                    dump(i->second, false, ensure_ascii, indent_step, current_indent);
+                    put_chars("\":", 2);
+                    dump_internal(i->second, false, ensure_ascii, indent_step, current_indent);
 
-                    o->write_character('}');
+                    put_char('}');
                 }
 
                 return;
@@ -190,13 +210,13 @@ class serializer
             {
                 if (val.m_data.m_value.array->empty())
                 {
-                    o->write_characters("[]", 2);
+                    put_chars("[]", 2);
                     return;
                 }
 
                 if (pretty_print)
                 {
-                    o->write_characters("[\n", 2);
+                    put_chars("[\n", 2);
 
                     // variable to hold indentation for recursive calls
                     const auto new_indent = current_indent + indent_step;
@@ -209,37 +229,37 @@ class serializer
                     for (auto i = val.m_data.m_value.array->cbegin();
                             i != val.m_data.m_value.array->cend() - 1; ++i)
                     {
-                        o->write_characters(indent_string.c_str(), new_indent);
-                        dump(*i, true, ensure_ascii, indent_step, new_indent);
-                        o->write_characters(",\n", 2);
+                        put_chars(indent_string.c_str(), new_indent);
+                        dump_internal(*i, true, ensure_ascii, indent_step, new_indent);
+                        put_chars(",\n", 2);
                     }
 
                     // last element
                     JSON_ASSERT(!val.m_data.m_value.array->empty());
-                    o->write_characters(indent_string.c_str(), new_indent);
-                    dump(val.m_data.m_value.array->back(), true, ensure_ascii, indent_step, new_indent);
+                    put_chars(indent_string.c_str(), new_indent);
+                    dump_internal(val.m_data.m_value.array->back(), true, ensure_ascii, indent_step, new_indent);
 
-                    o->write_character('\n');
-                    o->write_characters(indent_string.c_str(), current_indent);
-                    o->write_character(']');
+                    put_char('\n');
+                    put_chars(indent_string.c_str(), current_indent);
+                    put_char(']');
                 }
                 else
                 {
-                    o->write_character('[');
+                    put_char('[');
 
                     // first n-1 elements
                     for (auto i = val.m_data.m_value.array->cbegin();
                             i != val.m_data.m_value.array->cend() - 1; ++i)
                     {
-                        dump(*i, false, ensure_ascii, indent_step, current_indent);
-                        o->write_character(',');
+                        dump_internal(*i, false, ensure_ascii, indent_step, current_indent);
+                        put_char(',');
                     }
 
                     // last element
                     JSON_ASSERT(!val.m_data.m_value.array->empty());
-                    dump(val.m_data.m_value.array->back(), false, ensure_ascii, indent_step, current_indent);
+                    dump_internal(val.m_data.m_value.array->back(), false, ensure_ascii, indent_step, current_indent);
 
-                    o->write_character(']');
+                    put_char(']');
                 }
 
                 return;
@@ -247,9 +267,9 @@ class serializer
 
             case value_t::string:
             {
-                o->write_character('\"');
+                put_char('\"');
                 dump_escaped(*val.m_data.m_value.string, ensure_ascii);
-                o->write_character('\"');
+                put_char('\"');
                 return;
             }
 
@@ -257,7 +277,7 @@ class serializer
             {
                 if (pretty_print)
                 {
-                    o->write_characters("{\n", 2);
+                    put_chars("{\n", 2);
 
                     // variable to hold indentation for recursive calls
                     const auto new_indent = current_indent + indent_step;
@@ -266,9 +286,9 @@ class serializer
                         indent_string.resize(indent_string.size() * 2, ' ');
                     }
 
-                    o->write_characters(indent_string.c_str(), new_indent);
+                    put_chars(indent_string.c_str(), new_indent);
 
-                    o->write_characters("\"bytes\": [", 10);
+                    put_chars("\"bytes\": [", 10);
 
                     if (!val.m_data.m_value.binary->empty())
                     {
@@ -276,30 +296,30 @@ class serializer
                                 i != val.m_data.m_value.binary->cend() - 1; ++i)
                         {
                             dump_integer(*i);
-                            o->write_characters(", ", 2);
+                            put_chars(", ", 2);
                         }
                         dump_integer(val.m_data.m_value.binary->back());
                     }
 
-                    o->write_characters("],\n", 3);
-                    o->write_characters(indent_string.c_str(), new_indent);
+                    put_chars("],\n", 3);
+                    put_chars(indent_string.c_str(), new_indent);
 
-                    o->write_characters("\"subtype\": ", 11);
+                    put_chars("\"subtype\": ", 11);
                     if (val.m_data.m_value.binary->has_subtype())
                     {
                         dump_integer(val.m_data.m_value.binary->subtype());
                     }
                     else
                     {
-                        o->write_characters("null", 4);
+                        put_chars("null", 4);
                     }
-                    o->write_character('\n');
-                    o->write_characters(indent_string.c_str(), current_indent);
-                    o->write_character('}');
+                    put_char('\n');
+                    put_chars(indent_string.c_str(), current_indent);
+                    put_char('}');
                 }
                 else
                 {
-                    o->write_characters("{\"bytes\":[", 10);
+                    put_chars("{\"bytes\":[", 10);
 
                     if (!val.m_data.m_value.binary->empty())
                     {
@@ -307,20 +327,20 @@ class serializer
                                 i != val.m_data.m_value.binary->cend() - 1; ++i)
                         {
                             dump_integer(*i);
-                            o->write_character(',');
+                            put_char(',');
                         }
                         dump_integer(val.m_data.m_value.binary->back());
                     }
 
-                    o->write_characters("],\"subtype\":", 12);
+                    put_chars("],\"subtype\":", 12);
                     if (val.m_data.m_value.binary->has_subtype())
                     {
                         dump_integer(val.m_data.m_value.binary->subtype());
-                        o->write_character('}');
+                        put_char('}');
                     }
                     else
                     {
-                        o->write_characters("null}", 5);
+                        put_chars("null}", 5);
                     }
                 }
                 return;
@@ -330,11 +350,11 @@ class serializer
             {
                 if (val.m_data.m_value.boolean)
                 {
-                    o->write_characters("true", 4);
+                    put_chars("true", 4);
                 }
                 else
                 {
-                    o->write_characters("false", 5);
+                    put_chars("false", 5);
                 }
                 return;
             }
@@ -359,13 +379,13 @@ class serializer
 
             case value_t::discarded:
             {
-                o->write_characters("<discarded>", 11);
+                put_chars("<discarded>", 11);
                 return;
             }
 
             case value_t::null:
             {
-                o->write_characters("null", 4);
+                put_chars("null", 4);
                 return;
             }
 
@@ -401,28 +421,35 @@ class serializer
 
         for (std::size_t i = 0; i < s.size(); ++i)
         {
-            // Fast path: when not escaping non-ASCII characters and sitting on a
-            // character boundary (state == UTF8_ACCEPT), bulk-copy the longest
-            // run of bytes that need no escaping. string_bulk_run() (shared with
-            // the lexer's contiguous scanner) stops exactly at the first byte
-            // that dump_escaped would handle individually - a quote, a backslash,
-            // a control character (< 0x20), or an ill-formed/truncated UTF-8
-            // sequence - so that byte is left to the byte-at-a-time path below,
-            // keeping error handling and diagnostics unchanged.
-            if (!ensure_ascii && state == UTF8_ACCEPT)
+            // Fast path: at a character boundary (state == UTF8_ACCEPT),
+            // bulk-copy the longest run of bytes that need no escaping using a
+            // SWAR scanner shared with the lexer's contiguous path. The scanner
+            // stops exactly at the first byte dump_escaped would handle
+            // individually, so that byte is left to the byte-at-a-time path
+            // below, keeping escaping output and error diagnostics unchanged.
+            //
+            // - ensure_ascii == false: string_bulk_run() copies ordinary bytes
+            //   and complete well-formed UTF-8, stopping at a quote, backslash,
+            //   control character (< 0x20), or ill-formed/truncated sequence.
+            // - ensure_ascii == true: only printable ASCII may be copied
+            //   verbatim; find_ascii_copyable_run() additionally stops at 0x7F
+            //   and every non-ASCII byte (>= 0x80), which must be \u-escaped.
+            if (state == UTF8_ACCEPT)
             {
                 const auto* const data = reinterpret_cast<const unsigned char*>(s.data());
-                const std::size_t run = string_bulk_run(data + i, s.size() - i);
+                const std::size_t run = ensure_ascii
+                                        ? find_ascii_copyable_run(data + i, s.size() - i)
+                                        : string_bulk_run(data + i, s.size() - i);
                 if (run != 0)
                 {
                     // emit any bytes still pending in string_buffer first to
                     // preserve output order, then write the run directly
                     if (bytes != 0)
                     {
-                        o->write_characters(string_buffer.data(), bytes);
+                        put_chars(string_buffer.data(), bytes);
                         bytes = 0;
                     }
-                    o->write_characters(s.data() + i, run);
+                    put_chars(s.data() + i, run);
                     bytes_after_last_accept = 0;
                     undumped_chars = 0;
                     i += run;
@@ -521,7 +548,7 @@ class serializer
                     // written ("\uxxxx\uxxxx\0") for one code point
                     if (string_buffer.size() - bytes < 13)
                     {
-                        o->write_characters(string_buffer.data(), bytes);
+                        put_chars(string_buffer.data(), bytes);
                         bytes = 0;
                     }
 
@@ -580,7 +607,7 @@ class serializer
                                 // written ("\uxxxx\uxxxx\0") for one code point
                                 if (string_buffer.size() - bytes < 13)
                                 {
-                                    o->write_characters(string_buffer.data(), bytes);
+                                    put_chars(string_buffer.data(), bytes);
                                     bytes = 0;
                                 }
 
@@ -619,7 +646,7 @@ class serializer
             // write buffer
             if (bytes > 0)
             {
-                o->write_characters(string_buffer.data(), bytes);
+                put_chars(string_buffer.data(), bytes);
             }
         }
         else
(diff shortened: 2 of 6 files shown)


wilkolbrzym-coder:
## Quick review notes

- **Critical bug** in binary pretty‑print: `indent_string.resize(indent_string.size() * 2, ' ')` doubles the string every time a binary value is serialized, causing exponential memory growth. Should be a conditional resize to `new_indent` instead.

- **Missing check**: the fast path uses `string_bulk_run`, but its implementation isn’t shown – please verify it correctly stops at incomplete/ill‑formed UTF‑8 to keep error handling (`replace`/`ignore`) intact.

- **Nit**: replace magic `1024` buffer size with a named constant for readability.

Otherwise, the SWAR scanner and buffering logic look solid. 👍

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -109,20 +111,39 @@ class serializer
...
+                       const unsigned int current_indent = 0)
     {
         switch (val.m_data.m_type)
         {
             case value_t::object:
             {
                 if (val.m_data.m_value.object->empty())
                 {
-                    o->write_characters("{}", 2);
+                    put_chars("{}", 2);
```
gregmarr: Seems like this is asking for the number to get out of sync with the length of the string literal.  Could this be something like the following?
```
template<int N> void put_literal(const char (&str)[N])
{
  for (int i = 0; i < N - 1; i++)
  {
    put_char(str[i]);
  }
}
```

include/nlohmann/detail/output/serializer.hpp
gregmarr: not guaranteed to be big enough (handled by #5186).

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -135,51 +156,51 @@ class serializer
                     auto i = val.m_data.m_value.object->cbegin();
                     for (std::size_t cnt = 0; cnt < val.m_data.m_value.object->size() - 1; ++cnt, ++i)
                     {
-                        o->write_characters(indent_string.c_str(), new_indent);
-                        o->write_character('\"');
+                        put_chars(indent_string.c_str(), new_indent);
```
gregmarr: Simplify and split from putting the literals?
```
put_indent(new_indent);
```
Then `put_indent` can handle resizing `indent_string`, and you don't need to convert it to a `char *`.

nlohmann:
Thanks for your feedback!

1. Real, but pre-existing: already fixed by #5186
2. Not a bug: verified correct in `string_scan.hpp`
3. Yes, style nit.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -628,6 +688,66 @@ class serializer
...
+
+    /*!
+    @brief append @a length characters to the write buffer
+
+    Runs that do not fit the buffer are written straight through the output
+    adapter (after flushing what is pending), so large string/number payloads
+    are not copied an extra time.
+    */
+    JSON_HEDLEY_NON_NULL(2)
+    void put_chars(const char* s, std::size_t length)
```
gregmarr: There are multiple separate things being handled by this function.
1. Write the characters of a string literal.
2. Write a number of indent characters.
3. Write the first `len` characters from a `std::array<char, N>`.
4. Write an arbitrary number of characters from an arbitrary position in a `string_t`.

I'm not sure that each of these should be the same function.

Number 1 is asking for the number to get out of sync with the length of the string literal. It could be something like this:
```
template<int N> void put_literal(const char (&str)[N])
{
  const int length = N - 1;
  if (JSON_HEDLEY_UNLIKELY(write_buffer_pos + length >= write_buffer.size()))
  {
    flush();
  }
  std::memcpy(write_buffer.data() + write_buffer_pos, str, length);
  write_buffer_pos += length;
}
```
maybe with some `enable_if` to keep `N` under `1024` so you don't need to worry about it overflowing.

Number 2 could be:
```
void put_indent(unsigned int indent)
{
        if (JSON_HEDLEY_UNLIKELY(write_buffer_pos + indent > write_buffer.size()))
        {
            flush();
        }
        if (JSON_HEDLEY_UNLIKELY(indent > write_buffer.size()))
        {
          // the buffer is clear, fill it with the indent character
          std::memset(write_buffer.data(), indent_char, write_buffer.size());
          // keep writing it out until all the indent is processed.
          while (JSON_HEDLEY_UNLIKELY(indent > write_buffer.size()))
          {
            flush();
            indent -= write_buffer.size();
          }
          // set the write position to cover the rest of the indent
          write_buffer_pos = indent;

          return;
        }

        std::memset(write_buffer.data() + write_buffer_pos, indent_char, indent);
        write_buffer_pos += indent;
  }
```
which also eliminates the need for `indent_string` and resizing it.

Number 3 and 4 could be safer than just a `char *` and a count, not sure if they could easily share a "safer" implementation though.

include/nlohmann/detail/output/serializer.hpp
nlohmann: Right, and it turned out to be worth chasing: `indent_string` is gone as of a1e9147.

`put_indent()` now memsets the indentation straight into the write buffer, filling and flushing it as needed, so there is no auxiliary string to size, grow, or keep in sync with the deepest nesting level reached. That removes both halves of #5186 structurally rather than patching the growth rule — the doubling that isn't enough when `indent_step` more than doubles the string, and the `' '` literal in the resize instead of `indent_char`.

Both were reproducible against `develop`:

- `json{{"a", 1}}.dump(2000)` — `indent_string` starts at 512, doubles once to 1024, then 2000 bytes are read from it. Right length, garbage content: a heap over-read.
- `json{{"a", 1}}.dump(600, '\t')` — the part past 512 comes back as spaces, not tabs.

Both now pass, and there are tests for them (plus one for nesting whose accumulated indentation spans several buffer-fulls). `next_indent()` keeps #5186's assertion that the unsigned indentation accumulation hasn't wrapped on deep nesting, since that one is independent of how the characters get written.

@nlohmann — this does mean #5186 will conflict here and its `reserve_indent()` has nothing left to reserve. Worth deciding which lands first, especially with an advisory pointing at it.

*(Written by Claude Code.)*

include/nlohmann/detail/output/serializer.hpp
nlohmann: Agreed — split into separate entry points in a1e9147, close to what you sketched.

**1. String literals** → `put_literal()`, taking the literal by reference and deducing the length from the array bound, as you had it. All 27 call sites converted. For what it's worth I checked every hand-written count before replacing it and none of them was actually wrong today, but the class of bug is gone.

One deviation: instead of an `enable_if` to bound `N`, the length is checked against the buffer with a `static_assert`, which lets the oversized-run branch disappear entirely rather than becoming unreachable — this repo wants 100% line coverage, so an uncoverable branch is its own small problem.

```cpp
template<std::size_t N>
void put_literal(const char (&s)[N])
{
    static_assert(N >= 2, "put_literal expects a non-empty string literal");
    static_assert(N - 1 < write_buffer_size, "string literal must fit into the write buffer");
    ...
}
```

**2. Indentation** → `put_indent()`, and this removes `indent_string` and the resizing exactly as you predicted. Details in the other thread, since it also settles #5186.

Your sketch needs one adjustment though: in the wider-than-the-buffer loop, `write_buffer_pos` is already `0` from the first `flush()` and never set back to `write_buffer.size()`, so each `flush()` inside the `while` writes zero characters — the buffer gets memset but nothing comes out, and the indentation ends up truncated. I used a fill-and-flush loop instead, which also collapses the two cases into one:

```cpp
void put_indent(unsigned int indent)
{
    while (indent > 0)
    {
        if (JSON_HEDLEY_UNLIKELY(write_buffer_pos == write_buffer.size()))
        {
            flush();
        }

        const std::size_t chunk = (std::min)(static_cast<std::size_t>(indent),
                                             write_buffer.size() - write_buffer_pos);
        std::memset(write_buffer.data() + write_buffer_pos, indent_char, chunk);
        write_buffer_pos += chunk;
        indent -= static_cast<unsigned int>(chunk);
    }
}
```

**3. `std::array` + length** → `put_buffer(const std::array<char, N>&, std::size_t)`, so the length is checked against the buffer's own bound instead of travelling next to a bare pointer. Covers the `string_buffer` and `number_buffer` writes.

**4. Arbitrary range of a `string_t`** → left on `put_chars()`. Together with `to_chars()`'s output those are the only two callers now, and both are genuinely a pointer into the middle of something plus a length; I didn't find a safer shape that didn't just move the arithmetic. Open to ideas.

*(Written by Claude Code.)*

gregmarr: >write_buffer_pos is already 0 from the first flush() and never set back to write_buffer.size(), so each flush() inside the while writes zero characters

Yep, needed a `write_buffer_pos = write_buffer.size();` before the `flush()`.

> I used a fill-and-flush loop instead, which also collapses the two cases into one:

I was trying to avoid the `memset()` on every iteration. If you have an indent of 4x the buffer size, you'll fill the whole buffer 4 times with the indent character, which is essentially a no-op on 3 of them because you know the state of the buffer.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -628,6 +688,126 @@ class serializer
...
+    void put_char(char c)
+    {
+        if (JSON_HEDLEY_UNLIKELY(write_buffer_pos == write_buffer.size()))
+        {
+            flush();
+        }
+        write_buffer[write_buffer_pos++] = c;
+    }
+
+    /*!
```
gregmarr: This comment block got detached from `put_chars`.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -842,7 +1022,7 @@ class serializer
         auto* begin = number_buffer.data();
         auto* end = ::nlohmann::detail::to_chars(begin, begin + number_buffer.size(), x);
 
-        o->write_characters(begin, static_cast<size_t>(end - begin));
+        put_chars(begin, static_cast<size_t>(end - begin));
```
gregmarr: update to `put_buffer`?

include/nlohmann/detail/output/serializer.hpp
nlohmann: Both fair — fixed in b4fdf76.

You're right that re-filling the buffer is wasted work: flushing doesn't disturb what the buffer holds, so it only needs writing once. It now fills the room that's left, and if anything remains, flushes, fills the buffer once, and re-flushes that same content:

```cpp
void put_indent(unsigned int indent)
{
    // closing braces at the outermost level ask for no indentation at all
    if (indent == 0)
    {
        return;
    }

    const std::size_t capacity = write_buffer.size();

    const std::size_t head = (std::min)(static_cast<std::size_t>(indent), capacity - write_buffer_pos);
    std::memset(write_buffer.data() + write_buffer_pos, indent_char, head);
    write_buffer_pos += head;
    indent -= static_cast<unsigned int>(head);

    if (JSON_HEDLEY_LIKELY(indent == 0))
    {
        return;
    }

    flush();
    std::memset(write_buffer.data(), indent_char, capacity);

    while (indent >= capacity)
    {
        write_buffer_pos = capacity;
        flush();
        indent -= static_cast<unsigned int>(capacity);
    }

    // the buffer still holds indentation characters throughout, so the tail
    // only has to be claimed, not written again
    write_buffer_pos = indent;
}
```

Counting `memset` calls and bytes inside `put_indent` over a `dump()`:

| indent | before | after |
|---|---|---|
| 4 | 1 call / 4 B | 1 call / 4 B |
| 2000 | 2 calls / 2000 B | 2 calls / 2046 B |
| 100000 | 98 calls / 100000 B | 2 calls / 2046 B |

Constant instead of proportional to the width, and ordinary indents are untouched. The `indent == 0` early return is worth having on its own — every outermost closing brace asks for a zero-width indent, and without it the common path paid for a zero-length `memset`.

Tests extended to cover several whole buffer-fulls and an exact multiple of the buffer size.

*(Written by Claude Code.)*

gregmarr: > 4. Arbitrary range of a string_t → left on put_chars(). Together with to_chars()'s output those are the only two callers now, and both are genuinely a pointer into the middle of something plus a length; I didn't find a safer shape that didn't just move the arithmetic. Open to ideas.

This at least puts all the arithmetic inside the function.
```
template<typename StringOrArray>
void put_string(StringOrArray const &str, std::size_t start, std::size_t end)
{
  JSON_ASSERT(start < str.size());
  JSON_ASSERT(end< str.size());
  JSON_ASSERT(start < end);
  const std::size_t length = end - start;
  const char *s = str.data();
...
}
```

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -628,6 +688,126 @@ class serializer
...
+    @brief append a string literal to the write buffer
+
+    The length comes from the array bound rather than a hand-written count, so
+    it cannot drift out of sync with the literal. A literal always fits into the
+    buffer (checked at compile time), so unlike @ref put_chars this needs no
+    write-through path for oversized runs.
+    */
+    template<std::size_t N>
+    void put_literal(const char (&s)[N])
+    {
```
gregmarr, line 1182: I still like only computing the length once so you don't forget the `- 1`:
```
  const int length = N - 1;
```

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -189,137 +205,129 @@ class serializer
...
                         dump_integer(val.m_data.m_value.binary->back());
                     }
 
-                    o->write_characters("],\n", 3);
-                    o->write_characters(indent_string.c_str(), new_indent);
+                    put_literal("],\n");
+                    put_indent(new_indent);
 
-                    o->write_characters("\"subtype\": ", 11);
+                    put_literal("\"subtype\": ");
```
gregmarr, line 317: Not a change here, but the "subtype" about 30 lines later doesn't have a space after the colon.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -373,6 +381,19 @@ class serializer
         }
     }
 
+    /*!
+    @brief the indentation level to use for the children of the current value
+
+    A very large @a indent_step can wrap the unsigned accumulation on deep
+    nesting, which would silently truncate the indentation.
+    */
+    static unsigned int next_indent(const unsigned int current_indent, const unsigned int indent_step)
```
gregmarr: Should the indent be `size_t` then?

include/nlohmann/detail/output/serializer.hpp
nlohmann: Good catch — entirely my doing. I inserted `put_indent`/`put_literal`/`put_buffer` in front of `put_chars` and left its comment sitting on top of the first new function, so `put_indent` ended up with two doc blocks and `put_chars` with none. Reattached in 5567a0e.

*(Written by Claude Code.)*

include/nlohmann/detail/output/serializer.hpp
nlohmann: Yes — missed that one. `begin` is `number_buffer.data()`, so it is the same fixed-size-buffer case as the others; it is `put_buffer(number_buffer, ...)` as of 5567a0e.

That was the last one: `put_chars()` now has no callers outside `put_string()`/`put_buffer()`.

*(Written by Claude Code.)*

include/nlohmann/detail/output/serializer.hpp
nlohmann: Fair, and it reads better. Done in 5567a0e:

```cpp
static_assert(N >= 2, "put_literal expects a non-empty string literal");
// the array bound counts the terminating NUL, which is not written
constexpr std::size_t length = N - 1;
static_assert(length < write_buffer_size, "string literal must fit into the write buffer");
```

`constexpr` rather than `const` so the `static_assert` can still use it, and `std::size_t` to match what it is compared against.

*(Written by Claude Code.)*

include/nlohmann/detail/output/serializer.hpp
nlohmann: That does, and it is the better shape — added as `put_string()` in 5567a0e.

Two adjustments to the assertions. `end` is exclusive, so `end <= str.size()` rather than `<`: a run reaching the end of the string is the normal case in `dump_escaped`, and `<` would fire on it. And `start < end` rules out the empty range, which is harmless to write, so I made that `<=` too and dropped the separate `start` check as implied:

```cpp
template<typename StringType>
void put_string(const StringType& str, std::size_t start, std::size_t end)
{
    JSON_ASSERT(start <= end);
    JSON_ASSERT(end <= str.size());
    put_chars(str.data() + start, end - start);
}
```

With this and the `dump_float` fix from the other thread, `put_chars()` is down to two callers, both of them `put_string()`/`put_buffer()` — so nothing outside the buffer helpers passes a bare pointer and a count any more, which was your original point.

*(Written by Claude Code.)*

include/nlohmann/detail/output/serializer.hpp
nlohmann: Yes — changed in 5567a0e. It is a size, it is compared and combined with buffer sizes throughout, and `put_indent()` was casting to `std::size_t` and back on every call. `dump()`, `dump_internal()`, `next_indent()` and `put_indent()` all carry `std::size_t` now, and the casts are gone.

I kept `next_indent()`'s assertion: on a 64-bit `size_t` you would need an implausible nesting depth to wrap it, but it is still reachable where `size_t` is 32 bits, which is the same place the original `unsigned int` overflow was reachable.

*(Written by Claude Code.)*

include/nlohmann/detail/output/serializer.hpp
nlohmann: Checked this one rather than trusting the read — I think the two you are comparing are in different branches. Line 294 is the pretty-print path (`"subtype": ` with the space); the one about 30 lines later, at 322, is the compact path (`],"subtype":`), where nothing has spaces.

Actual output, and it is byte-identical to `develop`:

```
--- compact ---
{"b":{"bytes":[1,2,3],"subtype":42},"n":{"bytes":[1,2,3],"subtype":null}}
--- pretty ---
{
  "b": {
    "bytes": [1, 2, 3],
    "subtype": 42
  },
  ...
}
```

So both modes are internally consistent and neither changed here. No fix — but say the word if you were pointing at something else.

*(Written by Claude Code.)*

gregmarr: I couldn't tell from the restricted diff that one was pretty and one wasn't (one having newlines and one not should have given me a clue but didn't), so this difference makes sense.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -628,6 +690,173 @@ class serializer
...
+    }
+
+    /*!
+    @brief append the characters of @a str in [@a start, @a end)
+
+    Keeps the pointer arithmetic and the bounds checking inside the function
+    rather than at the call site, which is all @ref put_chars could offer.
+    */
+    template<typename StringType>
+    void put_string(const StringType& str, std::size_t start, std::size_t end)
```
gregmarr, line 1206: Looks like I didn't write out part of what I was thinking. I was thinking that this would *replace* `put_chars`.  That way there would not be a function that took a `const char *` and a `length`.  Then `put_buffer` is just `put_string(buffer, 0, length);`  That's why I had 
```
  const std::size_t length = end - start;
  const char *s = str.data();
...
```
as the rest of the function would be the current body of `put_chars()`.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -117,23 +118,39 @@ class serializer
...
+                       const std::size_t current_indent = 0,
+                       const std::size_t depth = 0)
     {
         switch (val.m_data.m_type)
         {
             case value_t::object:
             {
+                if (JSON_HEDLEY_UNLIKELY(depth >= dump_depth_limit()))
+                {
+                    dump_iteratively(val, pretty_print, ensure_ascii, indent_step, current_indent);
```
gregmarr, line 150: Rather than bailing out here to what is basically a copy of this function that can get out of sync, can you have the stack here and either recurse or push onto the stack based on the current depth?  I haven't analyzed in depth enough to know if that would be possible.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -1320,6 +1350,57 @@ class serializer
...
+            *out++ = static_cast<char>('0' + ((byte / 10) % 10));
+        }
+        else if (byte >= 10)
+        {
+            *out++ = static_cast<char>('0' + (byte / 10));
+        }
+
+        *out++ = static_cast<char>('0' + (byte % 10));
+
+        write_buffer_pos = static_cast<std::size_t>(out - write_buffer.data());
```
gregmarr: Would it be better to do write_buffer_pos++ along with each out++ and avoid the pointer math here?

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -780,13 +780,43 @@ class serializer
     characters by a sequence of "\u" followed by a four-digit hex
     representation. The escaped string is written to output stream @a o.
 
+    @param[in] s  the string to escape
+    @param[in] ensure_ascii  whether to escape non-ASCII characters with
+                             \uXXXX sequences
+
+    @complexity Linear in the length of string @a s.
+    */
+    /*!
```
gregmarr: Inserted the new function between the comment block and the old function again.

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -716,6 +1350,57 @@ class serializer
...
+    template<typename ByteType>
+    void dump_byte(const ByteType value, std::true_type /*is_plain_byte*/)
+    {
+        if (JSON_HEDLEY_UNLIKELY(write_buffer_pos + 3 > write_buffer.size()))
+        {
+            flush();
+        }
+
+        const auto byte = static_cast<unsigned>(value);
+        std::size_t pos = write_buffer_pos;
```
gregmarr, line 1382: Is this variable even needed?  Just use `write_buffer_pos` directly?

include/nlohmann/detail/output/serializer.hpp
nlohmann: Done in e614d61e6 — `put_chars()` is gone entirely.

Its body moved into `put_string()`, and `put_buffer()` is now just `put_string(buffer, 0, length)`: `std::array` already provides `data()` and `size()`, so it satisfies the same interface the assert needs. Nothing appends a run of characters without a bound any more.

<sub>(written by Claude Code)</sub>

include/nlohmann/detail/output/serializer.hpp
nlohmann: Fixed in e614d61e6. It was worse than detached this time — the comment block had been *duplicated*, so the same 15 lines appeared twice in a row before `dump_escaped()`.

The dispatcher now keeps the original block, and the templated worker got its own short comment explaining why `ensure_ascii` is a template parameter rather than a runtime flag.

<sub>(written by Claude Code)</sub>

include/nlohmann/detail/output/serializer.hpp
nlohmann: It is needed, and by a lot more than I expected — so I measured it rather than guess, and documented the reason in e614d61e6.

Writing through `write_buffer[]` is a `char` write, and a `char` lvalue may alias any object. So if `write_buffer_pos` is incremented in place, the compiler cannot keep it in a register across the digit stores: it has to reload and store the member around every one. The local breaks that and lets the whole byte be written from a register.

Dumping a 4 MiB binary value (median of 9, clang -O3):

| | |
|---|---|
| `std::size_t pos` local, stored back once | **7.4 ms** |
| `write_buffer_pos++` directly | **18.0 ms** |

2.4x. There is now a comment saying so, so it does not get simplified away later.

<sub>(written by Claude Code)</sub>

include/nlohmann/detail/output/serializer.hpp
nlohmann: Same answer as [the `pos` local thread](https://github.com/nlohmann/json/pull/5285#discussion_r3940010140) — this is the same function, before it was reworked to use an index instead of the `out` pointer.

Keeping the offset out of the member is what matters, and it is worth 2.4x here (7.4 ms vs 18.0 ms dumping a 4 MiB binary value): `char` writes may alias `write_buffer_pos`, so updating it in place forces a reload and store around every digit. Whether the local is a pointer or an index makes no difference; the current code uses an index, and e614d61e6 documents why it exists at all.

<sub>(written by Claude Code)</sub>

include/nlohmann/detail/output/serializer.hpp
nlohmann: Good instinct, and I tested the premise before answering: **the two paths agree today.**

I built two variants of the header, one with `dump_depth_limit()` raised so everything goes recursive, one with it set to 1 so everything goes iterative, and compared output over 4 005 randomised values × 5 dump modes (compact, indent 2, indent 4 + `ensure_ascii`, compact + `ensure_ascii`, tab indent), covering escapes, multi-byte UTF-8, binary values with their differing pretty/compact framing, empty containers, and deep chains at depths 5/50/129/200/500 nested through both arrays and objects. Byte-identical throughout.

So they have not drifted — but that is exactly your point: nothing *stops* them drifting.

On merging them, the honest answer is that it costs the thing this PR exists for. The recursive descent is the fast path; walking every value through the explicit stack measured **up to 20% slower** on object-heavy documents, which is why the split is there. A single function that holds the stack and either recurses or pushes based on depth still has to be able to *resume* mid-container on the stack path, so the container framing ends up written twice inside one function rather than across two — the duplication moves, it does not disappear.

Two options that do help, both cheaper than merging:

1. **Make the drift detectable.** Turn the parity check above into a unit test. That needs `dump_depth_limit()` to be overridable from tests, which is a small contained change, and then any future divergence between the paths fails CI. This is the one I would pick — it targets the actual risk rather than the structure.
2. **Share the framing primitives.** Factor "open brace / separator / indent / close brace" into small helpers both paths call, so the emitted bytes cannot drift even if the control flow stays separate.

I have not done either, since both are design calls on @nlohmann's PR rather than review fixes. Happy to implement whichever you two prefer.

<sub>(written by Claude Code)</sub>

gregmarr: Good analysis.  That answers my question well.

include/nlohmann/detail/input/string_scan.hpp
```diff
@@ -93,6 +93,58 @@ inline std::size_t find_string_special(const unsigned char* data, std::size_t n)
...
+        const std::uint64_t b = v ^ 0x5C5C5C5C5C5C5C5Cull; // '\\' (0x5C)
+        const std::uint64_t d = v ^ 0x7F7F7F7F7F7F7F7Full; // DEL  (0x7F)
+        const std::uint64_t stop = ((q - ones) & ~q & high)   // == '"'
+                                   | ((b - ones) & ~b & high)  // == '\\'
+                                   | ((d - ones) & ~d & high)  // == 0x7F
+                                   | ((v - 0x2020202020202020ull) & ~v & high) // < 0x20
+                                   | (v & high);               // >= 0x80
+        if (stop != 0)
+        {
+            for (std::size_t j = 0; j < 8; ++j)
```
gregmarr: Can you share the loop below by setting i equal to j and breaking out of the for loop?

```
if (stop != 0)
{
  i = j;
  break;
}
```

include/nlohmann/detail/output/serializer.hpp
```diff
@@ -107,219 +109,599 @@ class serializer
...
+                    const auto new_indent = next_indent(current_indent, indent_step);
 
                     // first n-1 elements
                     auto i = val.m_data.m_value.object->cbegin();
                     for (std::size_t cnt = 0; cnt < val.m_data.m_value.object->size() - 1; ++cnt, ++i)
                     {
-                        o->write_characters(indent_string.c_str(), new_indent);
-                        o->write_character('\"');
+                        put_indent(new_indent);
+                        put_char('\"');
```
gregmarr: You don't need to escape a double quote in a char literal.  (about 10 instances of this)

include/nlohmann/json.hpp
```diff
@@ -1347,7 +1347,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
 
         if (indent >= 0)
         {
-            s.dump(*this, true, ensure_ascii, static_cast<unsigned int>(indent));
+            s.dump(*this, true, ensure_ascii, static_cast<std::size_t>(indent));
```
gregmarr: Since these values of `serializer::dump` don't change over the life of the `serializer`, should they be given to the constructor instead of the `dump`, like is currently done for `indent_char`, and then they don't need to be passed to `dump`, `dump_internal` or `dump_iteratively`?

```
               const bool pretty_print,
              const bool ensure_ascii,
              const std::size_t indent_step,
 ```

File: include/nlohmann/detail/output/serializer.hpp [hunk 17]
```diff
@@ -70,7 +73,6 @@ class serializer
         , thousands_sep(loc->thousands_sep == nullptr ? '\0' : std::char_traits<char>::to_char_type(* (loc->thousands_sep)))
         , decimal_point(loc->decimal_point == nullptr ? '\0' : std::char_traits<char>::to_char_type(* (loc->decimal_point)))
         , indent_char(ichar)
-        , indent_string(512, indent_char)
```
gregmarr, line 73: not part of this review, but `serializer` holds a pointer to `std::localeconv()` even though it is only used to initialize `thousands_sep` and `decimal_point` in the member initialization list.  I assume it was used in more places in the past and just wasn't cleaned up, but even if that's not the case, I think it might be reasonable to initialize those two in the constructor body so that you don't need to hold onto the pointer.

include/nlohmann/detail/input/string_scan.hpp
gregmarr: Oh, right, `j` is just created here, even better.  :)

gregmarr (approved):

include/nlohmann/detail/output/serializer.hpp
nlohmann: Fixed in 84d2c238f: dropped the unneeded backslash-escapes on `'"'` char literals (10 spots in serializer.hpp).

Written by Claude Code.

include/nlohmann/detail/input/string_scan.hpp
nlohmann: Fixed in 84d2c238f: `find_ascii_copyable_run()` now just `break`s out of the SWAR loop on a hit and falls into the scalar tail loop below, instead of re-deriving the offset in a second, near-identical loop.

Written by Claude Code.

include/nlohmann/json.hpp
nlohmann: Done in 16a823e9e: `pretty_print`, `ensure_ascii` and `indent_step` are now constructor parameters/members, matching `indent_char`, and are no longer threaded through `dump()`, `dump_internal()`, `dump_iteratively()`, `dump_value()` or `dump_escaped()`.

Written by Claude Code.

include/nlohmann/detail/output/serializer.hpp
nlohmann: Fixed in 800b097bb: `loc` is now a local in the constructor body, used only to seed `thousands_sep`/`decimal_point` (which are no longer `const`, initialized in the body instead of the mem-initializer list). The serializer no longer holds onto the pointer.

Written by Claude Code.

gregmarr: Annoying that C++ doesn't allow you to assign const members inside the constructor body. :(  I forgot about that.  Been using too many other languages that allow you to do that.  Keeping them const would require holding onto the pointer or calling the function more than once since it's used for populating two members, neither of which are great.  The only other option I can think of involves putting them in a sub object and passing the pointer to its constructor.

nlohmann: Good point, and there was a third option: a small sub-object. Done in 772e44ff9 - locale_chars takes the lconv* in its own constructor and computes both chars there, so it's still only one call to std::localeconv(), nothing outlives the constructor, and thousands_sep/decimal_point stay const.

Written by Claude Code.

gregmarr (approved):

PR #5287, opened 2026-07-21
validate BSON document size against the bytes read

### LINKED ISSUE #5282: BSON document_size header is read but never validated (or used)
## Summary

The BSON parser reads the mandatory leading `document_size` field but never validates or otherwise uses it. This means structurally malformed BSON documents (including ones with a negative or nonsensical size header) are silently accepted rather than rejected.

## Affected code

`include/nlohmann/detail/input/binary_reader.hpp`:

```cpp
bool parse_bson_internal()
{
    std::int32_t document_size{};
    get_number<std::int32_t, true>(input_format_t::bson, document_size);
    // document_size is never checked (not > 0, not compared against
    // remaining input) and is never referenced again in this function.
    ...
}
```

The same pattern exists in `parse_bson_array()`.

Element-list parsing (`parse_bson_element_list`) terminates purely by encountering a `0x00` element-type byte in the stream — it does not use `document_size` as a bound at all. So today `document_size` is effectively dead code beyond consuming 4 bytes of input.

## Impact

Note: I initially assumed this could cause a loop-termination bug or signed-integer-overflow-driven memory issue, since that's the classic failure mode for an unchecked size header. That is **not** accurate for the current implementation — `document_size` isn't used in any arithmetic or comparison, so there's no active memory-safety path here.

The real impact is narrower: the parser doesn't enforce the BSON spec's minimum document size (5 bytes: 4-byte length + terminating `0x00`) or that the declared size is internally consistent with the actual content, so malformed/garbage BSON (e.g. a document size of `-1`, `0`, or a value wildly inconsistent with the real content) is processed as if valid instead of being rejected up front with a clear parse error. This is a correctness/spec-compliance gap and a defense-in-depth issue rather than an exploitable DoS/memory-corruption bug as currently written — but it should still be closed off, both for spec correctness and in case `document_size` is used for bounds-checking in a future refactor (where an unchecked negative value would become load-bearing).

## Proof of concept

```cpp
#include <nlohmann/json.hpp>

int main()
{
    // document_size = -1 (0xFFFFFFFF, little-endian), followed by an
    // immediate terminator. Spec requires a positive size >= 5.
    std::vector<std::uint8_t> data = {0xFF, 0xFF, 0xFF, 0xFF, 0x00};
    auto j = nlohmann::json::from_bson(data); // currently accepted
}
```

## Suggested fix

Validate `document_size` immediately after reading it in both `parse_bson_internal()` and `parse_bson_array()`:

```cpp
if (document_size < 5)
{
        parse_error::create(112, chars_read,
            exception_message(input_format_t::bson, "invalid document_size: " + std::to_string(document_size), "value"), nullptr));
}
```

Optionally, also verify `document_size` is consistent with the number of bytes actually available/consumed for the document.

## Environment

- nlohmann/json version: `develop` branch (current)

Angadi56:
The BSON reader takes the document length from the int32 at the start of every document and array, but never checks it. Parsing is driven entirely by the trailing `0x00`, so the declared length is read into a local and discarded. That lets a nested document claim a length that disagrees with where its terminator actually falls: an embedded document can declare far more bytes than it holds, terminate early, and the elements that its length said belonged to it get read as members of the enclosing document instead. On an embedded size of `FF FF FF 7F` in front of `{"a":null}` and then an `"h"` field, `from_bson` returns `{"d":{"a":null},"h":true}` where a length-aware parser rejects the input, so two BSON readers disagree on the shape of the same bytes. I found it while lining the BSON path up against the other decoders, which all validate their length fields. The fix records the offset before the size prefix and, once the element list and its terminator are read, checks that the bytes consumed match the declared size, in both `parse_bson_internal` and `parse_bson_array`. A negative or below-minimum size is caught by the same comparison, since at least the 5-byte minimum is always read.

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -164,11 +164,38 @@ class binary_reader
     //////////
 
     /*!
+    @brief Validate a BSON document's declared size against the bytes read.
+
+    A BSON document starts with an int32 that counts its own total length in
+    bytes, including that prefix and the trailing 0x00. The reader is driven
+    by the terminator rather than the declared length, so without this check a
+    nested document could declare a length that disagrees with where its
+    terminator actually falls and quietly hand the bytes in between to the
+    enclosing document. A well-formed document is at least 5 bytes (the prefix
+    plus the terminator); the equality also rejects those impossible sizes,
+    since at least 5 bytes are always consumed.
+
+    @param[in] document_start  value of chars_read before the size prefix
+    @param[in] document_size   the declared document size
+    @return whether the declared size matches the number of bytes read
+    */
+    bool check_bson_document_size(const std::size_t document_start, const std::int32_t document_size)
+    {
+        if (JSON_HEDLEY_UNLIKELY(document_size < 0 || static_cast<std::size_t>(document_size) != chars_read - document_start))
+        {
+            return sax->parse_error(chars_read, get_token_string(), parse_error::create(112, chars_read,
+                                    exception_message(input_format_t::bson, concat("document size ", std::to_string(document_size), " does not match the number of bytes read (", std::to_string(chars_read - document_start), ")"), "document"), nullptr));
+        }
+        return true;
+    }
+
+    /*!
     @brief Reads in a BSON-object and passes it to the SAX-parser.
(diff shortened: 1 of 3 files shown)


nlohmann (approved):
Looks good to me.

PR #5301, opened 2026-07-24
validate ndarray element types in write_bjdata_ndarray

Angadi56:
`to_bjdata` writes an object of the JData annotated form (`_ArrayType_`, `_ArraySize_`, `_ArrayData_`) back out as a BJData typed array. `write_bjdata_ndarray` picks a single `json_value` union member to read for every `_ArrayData_` element based on the `_ArrayType_` string, so a `uint64` array reads `number_unsigned`, a `double` array reads `number_float`, and so on, without ever checking that the element is actually stored as that kind of number. When the object comes from parsed input the three keys are ordinary strings a caller does not control, and `json::parse` builds this shape from untrusted text, so an element that is a string (or any non-number) has its storage reinterpreted: reading `number_unsigned` over a `std::string` yields the string's heap pointer, which `write_number` emits straight into the output. I ran into it feeding `{"_ArrayType_":"uint64","_ArraySize_":[1],"_ArrayData_":["A"]}` through `to_bjdata` and finding a live heap address in the last eight bytes. The same mismatch between an integer marker and a float element reinterprets the bit pattern and silently corrupts the encoded numbers. The fix checks each `_ArrayData_` element against the number kind the marker names, and each `_ArraySize_` entry for being an integer, before any union member is read; an object that does not match is written as a plain object, which is the fallback the function already uses for an unknown type or a size mismatch, so valid ndarrays are unaffected and still round-trip.

### DIFF
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -1647,6 +1647,12 @@ class binary_writer
         std::size_t len = (value.at(key).empty() ? 0 : 1);
         for (const auto& el : value.at(key))
         {
+            // the shape is read through the integer union member below; a
+            // non-integer entry would pun unrelated bytes as the dimension
+            if (!el.is_number_integer())
+            {
+                return true;
+            }
             len *= static_cast<std::size_t>(el.m_data.m_value.number_unsigned);
         }
 
@@ -1654,6 +1660,20 @@ class binary_writer
         if (value.at(key).size() != len)
         {
             return true;
+        }
+
+        // each element is read below through the one union member selected by
+        // dtype, so its stored type has to match (integer markers read the
+        // integer slot, 'd'/'D' the float slot). An element of any other type
+        // would reinterpret unrelated bytes, e.g. a string's heap pointer, as
+        // that number, so fall back to writing the value as a plain object.
+        const bool ndarray_is_float = (dtype == 'd' || dtype == 'D');
+        for (const auto& el : value.at(key))
+        {
+            if (ndarray_is_float ? !el.is_number_float() : !el.is_number_integer())
+            {
+                return true;
+            }
         }
 
         oa->write_character('[');
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -1647,13 +1647,15 @@ class binary_writer
         std::size_t len = (value.at(key).empty() ? 0 : 1);
         for (const auto& el : value.at(key))
         {
-            // the shape is read through the integer union member below; a
-            // non-integer entry would pun unrelated bytes as the dimension
-            if (!el.is_number_integer())
+            // a dimension is read as an unsigned value below, so anything that
+            // is not a non-negative integer is rejected: a non-integer entry
+            // would pun unrelated bytes as the dimension, and a negative one
+            // would wrap into a nonsensical length
+            if (!el.is_number_integer() || (!el.is_number_unsigned() && el.template get<std::int64_t>() < 0))
             {
                 return true;
             }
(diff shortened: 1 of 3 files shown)


include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1647,6 +1647,12 @@ class binary_writer
         std::size_t len = (value.at(key).empty() ? 0 : 1);
         for (const auto& el : value.at(key))
         {
+            // the shape is read through the integer union member below; a
+            // non-integer entry would pun unrelated bytes as the dimension
+            if (!el.is_number_integer())
```
nlohmann: `is_number_integer()` is true for both `number_integer` and `number_unsigned` values, but the read a few lines below is always through the `.number_unsigned` union member specifically. A shape entry that's stored as `number_integer` (e.g. a negative value) would pass this check and then have its `.number_unsigned` member read while `.number_integer` is the active alternative — the same class of union type-pun this PR is fixing, just for `_ArraySize_`. Since sizes can never legitimately be negative, this should require `el.is_number_unsigned()` instead of the looser `is_number_integer()`.

---

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1656,6 +1662,20 @@ class binary_writer
...
+        // integer slot, 'd'/'D' the float slot). An element of any other type
+        // would reinterpret unrelated bytes, e.g. a string's heap pointer, as
+        // that number, so fall back to writing the value as a plain object.
+        const bool ndarray_is_float = (dtype == 'd' || dtype == 'D');
+        for (const auto& el : value.at(key))
+        {
+            if (ndarray_is_float ? !el.is_number_float() : !el.is_number_integer())
+            {
+                return true;
+            }
```
nlohmann, line 1682: Same gap here on the non-float branch: `!el.is_number_integer()` accepts both `number_integer` and `number_unsigned` elements, but the write dispatch a few lines down is sign-specific — `i`/`I`/`l`/`L` read `.number_integer`, while `U`/`u`/`m`/`M`/`C`/`B` read `.number_unsigned`. An `int64`-typed ndarray whose data was parsed from JSON as a positive literal (which `json::parse` stores as `number_unsigned` by default) will pass this check and then read the inactive `.number_integer` member. That's benign for values that fit in the positive range (same bit pattern either way) but is still technically UB, and it silently breaks for `number_unsigned` values above `INT64_MAX`.

Suggest matching sign exactly, e.g.:
```cpp
const bool ndarray_is_unsigned = (dtype == 'U' || dtype == 'u' || dtype == 'm' || dtype == 'M' || dtype == 'C' || dtype == 'B');
...
if (ndarray_is_float ? !el.is_number_float()
    : ndarray_is_unsigned ? !el.is_number_unsigned() : !(el.is_number_integer() && !el.is_number_unsigned()))
```
Would also be good to add a test that round-trips through `json::parse` (rather than only the C++ initializer-list API), since that's the path where the unsigned-by-default parsing behavior actually bites.

---

nlohmann (changes requested):
Good fix for the main vulnerability — the type-category checks correctly stop string/bool/array/object elements from having their storage reinterpreted as numbers, closing the heap-pointer-leak case described in the PR.

There's a residual gap on the integer side, though: both new checks use `is_number_integer()`, which accepts *both* `number_integer` and `number_unsigned` values, while the actual write dispatch a few lines below is sign-specific. See the two inline comments for the exact locations and a suggested fix (requiring `is_number_unsigned()` for `_ArraySize_`, and an exact sign match for `_ArrayData_`). This should be addressed before merging, since a security-fix PR should close the type-punning hole precisely rather than narrowing it.

---

nlohmann (approved):
Looks good to me.

PR #5313, opened 2026-07-26
Cosmetic cleanups in the BSON writer (stale comment, inconsistent signature)

### LINKED ISSUE #5312: Cosmetic cleanups in the BSON writer (stale comment, inconsistent signature)
### Description

Minor, non-functional cleanups spotted while reviewing `include/nlohmann/detail/output/binary_writer.hpp`.

**1. Stale comment (`binary_writer.hpp:988`).** `write_bson_entry_header` is the generic header writer used for every element type, but its first line still carries a `// boolean` comment left over from when the code was specific to booleans:

```cpp
void write_bson_entry_header(const string_t& name,
                             const std::uint8_t element_type)
{
    oa->write_character(to_char_type(element_type)); // boolean
```

**2. Inconsistent signature (`binary_writer.hpp:1085`).** `write_bson_unsigned` takes the whole `BasicJsonType` and reaches into `j.m_data.m_value.number_unsigned` three times:

```cpp
void write_bson_unsigned(const string_t& name, const BasicJsonType& j)
```

Every sibling (`write_bson_boolean`, `write_bson_double`, `write_bson_integer`, `write_bson_string`, …) takes the plain value instead. This looks like a leftover from when the function threw `out_of_range.407` and needed `&j` as diagnostic context; since #4286 removed that throw, the parameter no longer earns its keep. Taking `std::uint64_t` would match the rest of the file and let the call site at `binary_writer.hpp:1247` read like its neighbours.

Neither affects behaviour.

### Library version

`develop` @ 06ac77f

banaboi:
Two non-functional cleanups in `include/nlohmann/detail/output/binary_writer.hpp`:

1. **Remove stale `// boolean` comment** in `write_bson_entry_header`. That function
   writes the header for every BSON element type, not just booleans. The comment is
   left over from when the code was boolean-specific.

2. **Change `write_bson_unsigned` to take `std::uint64_t` instead of `const BasicJsonType&`.**
   It reached into `j.m_data.m_value.number_unsigned` three times, while every sibling
   (`write_bson_boolean`, `write_bson_double`, `write_bson_integer`, `write_bson_string`)
   takes the plain value. The `BasicJsonType` parameter was only needed as diagnostic
   context for the `out_of_range.407` throw that #4286 removed, so it no longer earns
   its keep. The call site now reads like its neighbours.

While there, fixed the doc comment on the adjacent `calc_bson_unsigned_size`, which still
referred to "the unsigned integer in `@a j`" despite already taking a `value` parameter.

No behaviour change.

Closes #5312

### DIFF
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -985,7 +985,7 @@ class binary_writer
     void write_bson_entry_header(const string_t& name,
                                  const std::uint8_t element_type)
     {
-        oa->write_character(to_char_type(element_type)); // boolean
+        oa->write_character(to_char_type(element_type));
         oa->write_characters(
             reinterpret_cast<const CharType*>(name.c_str()),
             name.size() + 1u);
@@ -1070,7 +1070,7 @@ class binary_writer
     }
 
     /*!
-    @return The size of the BSON-encoded unsigned integer in @a j
+    @return The size of the BSON-encoded unsigned integer @a value
     */
     static constexpr std::size_t calc_bson_unsigned_size(const std::uint64_t value) noexcept
     {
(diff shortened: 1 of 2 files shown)


nlohmann (approved):
Looks good to me.

PR #5314, opened 2026-07-27
to_bson() silently emits corrupt documents when a length exceeds INT32_MAX

### LINKED ISSUE #5307: to_bson() silently emits corrupt documents when a length exceeds INT32_MAX
### Description

Every BSON length field is written through an unchecked `static_cast<std::int32_t>`:

- `write_bson_object` — `include/nlohmann/detail/output/binary_writer.hpp:1287`
- `write_bson_array` — `binary_writer.hpp:1145`
- `write_bson_string` — `binary_writer.hpp:1030`
- `write_bson_binary` — `binary_writer.hpp:1165`

When a string, binary value, array or document exceeds `INT32_MAX` bytes, the cast wraps around and `to_bson()` returns successfully with **negative** length prefixes. The library then cannot read its own output back.

BSON is the only format in the library with a fixed 32-bit length field, so no other writer has this exposure.

### Reproduction steps

Serialize an object holding a binary value of 2 GiB (needs ~6 GB RAM).

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <cstring>
#include <iostream>
using json = nlohmann::json;

int main()
{
    const std::size_t n = 2147483648ULL; // 2^31 > INT32_MAX
    json j;
    j["b"] = json::binary(std::vector<std::uint8_t>(n, 0x42));

    const auto v = json::to_bson(j); // no exception

    std::int32_t declared_doc = 0;
    std::int32_t declared_bin = 0;
    std::memcpy(&declared_doc, v.data(), 4);
    std::memcpy(&declared_bin, v.data() + 7, 4);

    std::cout << "output size            = " << v.size() << '\n';
    std::cout << "declared document size = " << declared_doc << '\n';
    std::cout << "declared binary size   = " << declared_bin << '\n';
}
```

**Expected:** an exception (`out_of_range`), since the value cannot be represented in BSON.

**Actual:** `to_bson()` succeeds and produces a document with negative length prefixes:

```
output size            = 2147483661
declared document size = -2147483635
declared binary size   = -2147483648
```

Feeding that byte vector back into `from_bson()` fails with:

```
[json.exception.parse_error.112] parse error at byte 12: syntax error while parsing BSON binary: byte array length cannot be negative, is -2147483648
```

So the library produces output it cannot itself read, with no diagnostic at serialization time — silent data corruption.

### Suggested fix

Throw `out_of_range` when a computed length does not fit into `std::int32_t`. The writer already has the idiom for this (`out_of_range.409` at `binary_writer.hpp:975`); a new id (e.g. `412`) would fit the existing pattern, and would need a corresponding entry in `docs/mkdocs/docs/home/exceptions.md` plus a note in the `to_bson` API docs.

g++ 13.3.0 (Ubuntu 24.04), `-std=c++11 -O2`, Linux x86-64

### Library version

`develop` @ 06ac77f

banaboi:
Fixes #5307.

## What

BSON length fields are 32-bit signed integers, but all four sites that wrote one used an unchecked `static_cast<std::int32_t>`: `write_bson_string`, `write_bson_array`, `write_bson_binary`, and `write_bson_object`. When a string, binary value, array, or document exceeded `INT32_MAX` bytes the cast wrapped, and `to_bson()` returned successfully with negative length prefixes. Output that `from_bson()` cannot read back, with no diagnostic at serialization time.

This adds a checked helper, `to_bson_length()`, which throws a new `out_of_range.412` when a length does not fit into `std::int32_t`, and routes all four sites through it.

## Test

Reproducing the bug directly needs a 2 GiB value (~6 GB RAM), which is not suitable for CI. Instead the test defines a file-local `huge_binary_t` deriving from `std::vector<std::uint8_t>` that hides `size()` to report 2^31 without allocating, and uses it as the `BinaryType` of a `basic_json` specialization. The size calls are statically dispatched, so the writer sees an oversized value and throws before allocating or writing anything.

Verified against the unpatched header first, where the test crashes rather than emitting corrupt output: `write_bson_binary` reaches `write_characters()` and reads past the end of the empty buffer. With the fix it throws the expected message.

## Docs

New `json.exception.out_of_range.412` section in `docs/mkdocs/docs/home/exceptions.md`, plus the corresponding entry under Exceptions in `docs/mkdocs/docs/api/basic_json/to_bson.md`.

---

### DIFF
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -986,7 +986,7 @@ class binary_writer
     */
     static std::int32_t to_bson_length(const std::size_t size)
     {
-        if (JSON_HEDLEY_UNLIKELY(size > static_cast<std::size_t>((std::numeric_limits<std::int32_t>::max)())))
+        if (JSON_HEDLEY_UNLIKELY(!value_in_range_of<std::int32_t>(size)))
         {
             JSON_THROW(out_of_range::create(412, concat("BSON length ", std::to_string(size), " exceeds maximum of ", std::to_string((std::numeric_limits<std::int32_t>::max)())), nullptr));
         }
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -980,6 +980,21 @@ class binary_writer
     }
 
     /*!
+    @brief Checks that @a size fits into the 32-bit length field used by BSON
+    @return The size as a signed 32-bit integer
+    @throw out_of_range.412 if @a size exceeds the range of std::int32_t
+    */
+    static std::int32_t to_bson_length(const std::size_t size)
+    {
+        if (JSON_HEDLEY_UNLIKELY(size > static_cast<std::size_t>((std::numeric_limits<std::int32_t>::max)())))
+        {
+            JSON_THROW(out_of_range::create(412, concat("BSON length ", std::to_string(size), " exceeds maximum of ", std::to_string((std::numeric_limits<std::int32_t>::max)())), nullptr));
+        }
+
+        return static_cast<std::int32_t>(size);
+    }
+
+    /*!
     @brief Writes the given @a element_type and @a name to the output adapter
     */
     void write_bson_entry_header(const string_t& name,
@@ -1027,7 +1042,7 @@ class binary_writer
     {
         write_bson_entry_header(name, 0x02);
 
-        write_number<std::int32_t>(static_cast<std::int32_t>(value.size() + 1ul), true);
+        write_number<std::int32_t>(to_bson_length(value.size() + 1ul), true);
         oa->write_characters(
             reinterpret_cast<const CharType*>(value.c_str()),
             value.size() + 1);
@@ -1142,7 +1157,7 @@ class binary_writer
                           const typename BasicJsonType::array_t& value)
     {
         write_bson_entry_header(name, 0x04); // array
-        write_number<std::int32_t>(static_cast<std::int32_t>(calc_bson_array_size(value)), true);
+        write_number<std::int32_t>(to_bson_length(calc_bson_array_size(value)), true);
 
         std::size_t array_index = 0ul;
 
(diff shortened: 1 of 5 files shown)


include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -979,6 +979,21 @@ class binary_writer
...
     }
 
+    /*!
+    @brief Checks that @a size fits into the 32-bit length field used by BSON
+    @return The size as a signed 32-bit integer
+    @throw out_of_range.412 if @a size exceeds the range of std::int32_t
+    */
+    static std::int32_t to_bson_length(const std::size_t size)
+    {
+        if (JSON_HEDLEY_UNLIKELY(size > static_cast<std::size_t>((std::numeric_limits<std::int32_t>::max)())))
```
nlohmann: The library already has a generic "does this value fit in `OfType`" helper: `detail::value_in_range_of<OfType>(val)` in `include/nlohmann/detail/meta/type_traits.hpp` (used e.g. in `binary_reader.hpp` for the CBOR/MessagePack ndarray size checks). It correctly covers all signed/unsigned combinations via `value_in_range_of_impl2`.

For the `int32_t`/`size_t` pair here, `value_in_range_of_impl2<OfType, T, true, false>::test` does the same comparison this hunk reimplements:
```cpp
static_cast<CommonType>(val) <= static_cast<CommonType>((std::numeric_limits<OfType>::max)());
```

`binary_writer.hpp` already transitively includes `type_traits.hpp` via `binary_reader.hpp`, so no new include is needed. Could this be written as:

```cpp
static std::int32_t to_bson_length(const std::size_t size)
{
    if (JSON_HEDLEY_UNLIKELY(!detail::value_in_range_of<std::int32_t>(size)))
    {
        JSON_THROW(out_of_range::create(412, concat("BSON length ", std::to_string(size), " exceeds maximum of ",
                                                      std::to_string((std::numeric_limits<std::int32_t>::max)())), nullptr));
    }

    return static_cast<std::int32_t>(size);
}
```
instead of hand-rolling the bounds check? Reusing the shared trait avoids a second, independently-maintained copy of the same overflow logic.

---

nlohmann (commented):
Reviewed the fix — solid approach overall (single checked helper routing all four BSON length write sites). Left one inline note about reusing an existing helper for the bounds check.

---

nlohmann:
There is also a failed CO step with a warning that should be fixed.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks a lot!

Issue #5322, opened 2026-07-27
UBJSON/BJData high-precision numbers ('H') bypass the out_of_range.406 overflow check and silently yield infinity


nlohmann:
### Description

`get_ubjson_high_precision_number()` parses the digits of an `H` value with the same `detail::lexer` the text JSON parser uses, but it does not apply the overflow check the text parser applies afterwards.

The text parser rejects a float literal that does not fit into `number_float_t`:

https://github.com/nlohmann/json/blob/8ec98e2c9ea9c0adc15a791feb3d4d4468621743/include/nlohmann/detail/input/parser.hpp#L266-L275

The UBJSON path forwards the lexer result straight to the SAX handler with no such guard:

https://github.com/nlohmann/json/blob/8ec98e2c9ea9c0adc15a791feb3d4d4468621743/include/nlohmann/detail/input/binary_reader.hpp#L2811-L2818

So the same number text produces an exception through `parse()` and an infinity through `from_ubjson()`. The resulting `basic_json` holds a non-finite `number_float_t`, which `dump()` renders as `null` — the value is silently wrong rather than rejected, and it cannot be round-tripped back to JSON.

`from_bjdata()` shares the function and behaves identically.

### Reproduction steps

Decode an `H` value whose digits overflow `double`, and compare against `json::parse` on the same text.

### Expected vs. actual results

Expected: `out_of_range.406`, matching `parse()`.

Actual:

| number text | `from_ubjson(H …)` | `json::parse(…)` |
|---|---|---|
| `1e400` | **`inf`** (dumps as `null`) | throws `out_of_range.406` |
| `-1e400` | **`-inf`** (dumps as `null`) | throws `out_of_range.406` |
| `1e-400` | `0.0` | `0.0` (consistent) |

Underflow is already consistent between the two paths; only the overflow direction diverges.

```
H "1e400"  -> OK dump=null raw=inf
   json::parse -> THROW: [json.exception.out_of_range.406] number overflow parsing '1e400'
H "-1e400" -> OK dump=null raw=-inf
   json::parse -> THROW: [json.exception.out_of_range.406] number overflow parsing '-1e400'
```

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <iostream>
using json = nlohmann::json;

int main()
{
    // UBJSON: 'H' (high-precision) 'i' (int8 length) 5, then "1e400"
[...]

PR #5323, opened 2026-07-27
Fix UBJSON high-precision floating-point overflow handling

hum4nBeing:
This PR fixes issue #5322.

The UBJSON high-precision number parser previously accepted overflowing floating-point values (like 1e400) because it lacked the std::isfinite validation present in the standard JSON parser. This resulted in internal infinity values that serialized to null.

Changes:

Added std::isfinite check in binary_reader.hpp for token_type::value_float.

Replicates the out_of_range.406 exception handling used by the standard parser.

Added a regression test specifically for 1e400 in unit-ubjson.cpp.

### DIFF
--- a/tests/src/unit-ubjson.cpp
+++ b/tests/src/unit-ubjson.cpp
@@ -819,6 +819,8 @@ TEST_CASE("UBJSON")
                     CHECK_THROWS_WITH_AS(_ = json::from_ubjson(vec2), "[json.exception.parse_error.115] parse error at byte 5: syntax error while parsing UBJSON high-precision number: invalid number text: 1A", json::parse_error);
                     std::vector<uint8_t> const vec3 = {'H', 'i', 2, '1', '.'};
                     CHECK_THROWS_WITH_AS(_ = json::from_ubjson(vec3), "[json.exception.parse_error.115] parse error at byte 5: syntax error while parsing UBJSON high-precision number: invalid number text: 1.", json::parse_error);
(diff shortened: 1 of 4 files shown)


amalgamation.patch
nlohmann, line 1: Please do not commit this file.

nlohmann (changes requested):

tests/src/unit-ubjson.cpp
```diff
@@ -819,6 +819,8 @@ TEST_CASE("UBJSON")
                     CHECK_THROWS_WITH_AS(_ = json::from_ubjson(vec2), "[json.exception.parse_error.115] parse error at byte 5: syntax error while parsing UBJSON high-precision number: invalid number text: 1A", json::parse_error);
                     std::vector<uint8_t> const vec3 = {'H', 'i', 2, '1', '.'};
                     CHECK_THROWS_WITH_AS(_ = json::from_ubjson(vec3), "[json.exception.parse_error.115] parse error at byte 5: syntax error while parsing UBJSON high-precision number: invalid number text: 1.", json::parse_error);
+                    std::vector<uint8_t> const vec_overflow = {'H', 'i', 5, '1', 'e', '4', '0', '0'};
```
nlohmann, line 822: Please also add a similar overflow regression to unit-bjdata.cpp for parity with the UBJSON test.

nlohmann (changes requested):

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #5325, opened 2026-07-27
Reject nested indefinite-length CBOR string chunks

### LINKED ISSUE #5317: CBOR: nested indefinite-length strings/byte strings are accepted (RFC 8949 3.2.3 requires definite-length chunks)
### Description

RFC 8949 §3.2.3 requires that the chunks of an indefinite-length string be *definite-length* strings of the same major type:

> Each chunk of an indefinite-length string is a definite-length string of the same major type.

`get_cbor_string()` and `get_cbor_binary()` read each chunk by recursing into themselves ([`binary_reader.hpp:1025-1037`](https://github.com/nlohmann/json/blob/8ec98e2c9ea9c0adc15a791feb3d4d4468621743/include/nlohmann/detail/input/binary_reader.hpp#L1025-L1037) and [`1125-1137`](https://github.com/nlohmann/json/blob/8ec98e2c9ea9c0adc15a791feb3d4d4468621743/include/nlohmann/detail/input/binary_reader.hpp#L1125-L1137)), and the recursive call re-enters the full `switch` including the `0x7F`/`0x5F` case. So a chunk may itself be an indefinite-length string, to arbitrary depth, and the library accepts it.

The cross-major-type rule is already enforced correctly — a byte-string chunk inside a text string is rejected, and vice versa. It is only the "definite-length" half of the rule that is missing.

Secondary effect: because the nesting is unbounded, `0x7F` repeated ~50 000 times segfaults. That aspect overlaps with #5104 / #5293, but a depth guard alone would only turn the crash into a clean `parse_error` — the two-level case below would still be accepted rather than rejected. The conformance fix is independent and cheaper: reject `0x7F`/`0x5F` when already inside an indefinite-length string, which also removes this particular recursion path entirely.

### Reproduction steps

Parse a two-level indefinite-length text string and byte string.

Expected: `parse_error.113`, same as any other invalid chunk head.

Actual:

| input | meaning | result |
|---|---|---|
| `7F 7F 61 61 FF FF` | indefinite tstr containing an indefinite tstr | **accepted**, yields `"a"` |
| `5F 5F 41 61 FF FF` | indefinite bstr containing an indefinite bstr | **accepted**, yields `h'61'` |
| `7F 41 61 FF` | bstr chunk inside a tstr | rejected (correct) |
| `5F 61 61 FF` | tstr chunk inside a bstr | rejected (correct) |

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <iostream>
using json = nlohmann::json;

int main()
{
    // 0x7F ( 0x7F ( "a" ) )  -- a chunk that is itself indefinite-length
    const std::vector<std::uint8_t> v{0x7F, 0x7F, 0x61, 0x61, 0xFF, 0xFF};
    std::cout << json::from_cbor(v) << "\n";   // prints "a"; should throw

    // unbounded nesting -> stack exhaustion (cf. #5104)
    std::vector<std::uint8_t> deep(50000, 0x7F);
    deep.insert(deep.end(), 50000, 0xFF);
    (void)json::from_cbor(deep);               // segfault
}
```

### Error messages

None — the malformed input is accepted. (The deep case terminates with SIGSEGV.)

### Library version

`develop` @ 8ec98e2

No API/ABI change, and the behavior change is confined to input that is malformed per the spec.

- **What changes:** byte sequences that RFC 8949 forbids and that other CBOR implementations reject would start throwing `parse_error.113` instead of being accepted. There is no valid encoder that produces them, so no legitimate document is affected.
- **Round-trip safety:** the library never *emits* indefinite-length strings, so nothing it writes can trip the new check.
- **Risk:** the only realistic breakage is a user whose non-conforming producer happens to emit nested chunks and who relies on this library's leniency. That seems remote, and tightening it matches how the reader already handles the cross-major-type half of the same rule.
- **Docs/tests:** the "Deserialization" section of `docs/mkdocs/docs/features/binary_formats/cbor.md` doesn't currently say anything about chunk structure; a sentence could be added. I did not find an existing unit test that covers nested chunks, so nothing in `unit-cbor.cpp` should need to be relaxed.

Fixing this also shrinks the surface of #5104: with nesting rejected, indefinite-length strings stop being a recursion vector at all, and the depth guard in #5293 is left to handle the genuinely recursive cases (nested arrays/maps and chained tags).

JosephDemarest:
Closes #5317.

Reject nested indefinite-length CBOR text and byte-string chunks as required by RFC 8949 Section 3.2.3. This update is rebased onto develop c41152e62092c16cf4fc4625becacc3eb9d9319f and builds on the iterative reader introduced by #5502, preserving its bounded stack use and direct chunk appends.

This is a behavior-visible change: nested indefinite-length chunks that older versions accepted leniently now raise parse_error.113 at the nested marker (or return discarded when exceptions are disabled through the from_cbor argument). Valid definite and indefinite strings remain supported.

The follow-up addresses the August 19 review:

- Keep successful byte-for-byte decoding coverage using the existing fixture's valid prefix and cbor_binary.out; add empty and multiple definite-chunk success cases.
- Make malformed-chunk diagnostics list only definite-length forms while preserving the top-level diagnostic.
- Share nested-indefinite rejection formatting between text and binary readers.
- Retain use of the expected-output fixture, eliminating the orphaned-fixture concern.
- Update the upstream deep-input tests to expect immediate rejection, retain the issue regression coverage in its current regression3 file, and regenerate the single header.

Validation with Clang 20 / MSVC STL on Windows:

- C++17 focused CBOR and regression3 checks: 3/3 CTest entries pass (includes the data setup entry).
- C++11 focused checks: 3/3 pass.
- C++17 generated single-header focused checks: 3/3 pass.
- Broader C++17 run: 79/80 pass. The diagnostic-position test fails the same two swap-position assertions on untouched develop c41152e6. The five exhaustive Unicode executables and tests labeled not_reproducible/git_required were excluded.
- Artistic Style 3.4.13, reproducible amalgamation, and git diff --check pass.
- LSP did not provide a clean gate: the MCP transport failed, and direct clangd reported style diagnostics in existing declarations and refactoring self-test failures. Compilation and test evidence are reported separately.

Local code coverage was not measured; upstream coverage/CI remains to be run after publication.

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -995,10 +995,11 @@ class binary_reader
     Additionally, CBOR's strings with indefinite lengths are supported.
 
     @param[out] result  created string
+    @param[in] allow_indefinite  whether an indefinite-length string is allowed
 
     @return whether string creation completed
     */
-    bool get_cbor_string(string_t& result)
+    bool get_cbor_string(string_t& result, const bool allow_indefinite = true)
     {
         if (JSON_HEDLEY_UNLIKELY(!unexpect_eof(input_format_t::cbor, "string")))
         {
@@ -1062,10 +1063,16 @@ class binary_reader
 
             case 0x7F: // UTF-8 string (indefinite length)
             {
+                if (JSON_HEDLEY_UNLIKELY(!allow_indefinite))
+                {
+                    auto last_token = get_token_string();
+                    return sax->parse_error(chars_read, last_token, parse_error::create(113, chars_read,
+                                            exception_message(input_format_t::cbor, concat("indefinite-length string is not allowed inside indefinite-length string; last byte: 0x", last_token), "string"), nullptr));
+                }
                 while (get() != 0xFF)
                 {
                     string_t chunk;
-                    if (!get_cbor_string(chunk))
+                    if (!get_cbor_string(chunk, false))
                     {
                         return false;
                     }
@@ -1091,10 +1098,11 @@ class binary_reader
     Additionally, CBOR's byte arrays with indefinite lengths are supported.
 
     @param[out] result  created byte array
+    @param[in] allow_indefinite  whether an indefinite-length byte array is allowed
 
     @return whether byte array creation completed
     */
-    bool get_cbor_binary(binary_t& result)
+    bool get_cbor_binary(binary_t& result, const bool allow_indefinite = true)
     {
         if (JSON_HEDLEY_UNLIKELY(!unexpect_eof(input_format_t::cbor, "binary")))
         {
@@ -1162,10 +1170,16 @@ class binary_reader
 
             case 0x5F: // Binary data (indefinite length)
             {
+                if (JSON_HEDLEY_UNLIKELY(!allow_indefinite))
+                {
+                    auto last_token = get_token_string();
+                    return sax->parse_error(chars_read, last_token, parse_error::create(113, chars_read,
+                                            exception_message(input_format_t::cbor, concat("indefinite-length binary array is not allowed inside indefinite-length binary array; last byte: 0x", last_token), "binary"), nullptr));
+                }
                 while (get() != 0xFF)
                 {
                     binary_t chunk;
-                    if (!get_cbor_binary(chunk))
+                    if (!get_cbor_binary(chunk, false))
                     {
                         return false;
                     }
--- a/tests/src/unit-cbor.cpp
+++ b/tests/src/unit-cbor.cpp
@@ -2493,10 +2493,8 @@ TEST_CASE("examples from RFC 8949 Appendix A")
     {
         const auto packed = utils::read_binary_file(TEST_DATA_DIRECTORY "/binary_data/cbor_binary.cbor");
         json j;
-        CHECK_NOTHROW(j = json::from_cbor(packed));
-
-        const auto expected = utils::read_binary_file(TEST_DATA_DIRECTORY "/binary_data/cbor_binary.out");
-        CHECK(j == json::binary(expected));
+        // The fixture ends with nested indefinite-length byte strings, which RFC 8949 Section 3.2.3 forbids.
+        CHECK_THROWS_WITH_AS(j = json::from_cbor(packed), "[json.exception.parse_error.113] parse error at byte 513: syntax error while parsing CBOR binary: indefinite-length binary array is not allowed inside indefinite-length binary array; last byte: 0x5F", json::parse_error&);
 
         // 0xd8
         CHECK(json::to_cbor(json::binary(std::vector<uint8_t> {}, 0x42)) == std::vector<uint8_t> {0xd8, 0x42, 0x40});
--- a/docs/mkdocs/docs/features/binary_formats/cbor.md
+++ b/docs/mkdocs/docs/features/binary_formats/cbor.md
@@ -131,6 +131,7 @@ The library maps CBOR types to JSON value types as follows:
 | Byte string            | binary          | 0x59       |
 | Byte string            | binary          | 0x5A       |
 | Byte string            | binary          | 0x5B       |
+| Byte string            | binary          | 0x5F       |
 | UTF-8 string           | string          | 0x60..0x77 |
 | UTF-8 string           | string          | 0x78       |
 | UTF-8 string           | string          | 0x79       |
@@ -155,6 +156,9 @@ The library maps CBOR types to JSON value types as follows:
 | Half-Precision Float   | number_float    | 0xF9       |
 | Single-Precision Float | number_float    | 0xFA       |
 | Double-Precision Float | number_float    | 0xFB       |
+
+Indefinite-length UTF-8 strings (0x7F) and byte strings (0x5F) are supported. Each chunk must be a definite-length
+string of the same major type, as required by [RFC 8949, Section 3.2.3](https://www.rfc-editor.org/rfc/rfc8949.html#section-3.2.3).
 
 !!! warning "Incomplete mapping"
 
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -11589,10 +11589,11 @@ class binary_reader
     Additionally, CBOR's strings with indefinite lengths are supported.
 
     @param[out] result  created string
+    @param[in] allow_indefinite  whether an indefinite-length string is allowed
 
     @return whether string creation completed
     */
-    bool get_cbor_string(string_t& result)
+    bool get_cbor_string(string_t& result, const bool allow_indefinite = true)
     {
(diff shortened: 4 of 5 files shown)


include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -1024,10 +1025,16 @@ class binary_reader
 
             case 0x7F: // UTF-8 string (indefinite length)
             {
+                if (JSON_HEDLEY_UNLIKELY(!allow_indefinite))
+                {
+                    auto last_token = get_token_string();
+                    return sax->parse_error(chars_read, last_token, parse_error::create(113, chars_read,
+                                            exception_message(input_format_t::cbor, concat("expected length specification (0x60-0x7B) or indefinite string type (0x7F); last byte: 0x", last_token), "string"), nullptr));
```
nlohmann: This can yield confusing error messages like

```
expected length specification (0x60-0x7B) or indefinite string type (0x7F); last byte: 0x7F
```

Maybe we should rather have something like

```cpp
concat("indefinite-length string is not allowed inside indefinite-length string; last byte: 0x", last_token)
```

or just drop the ` or indefinite string type (0x7F)`.

nlohmann (changes requested):

nlohmann:
Clang was updated and the latest develop branch fixes a CI issue. Please rebase to the latest develop branch.

JosephDemarest:
Thanks — I rebased onto the current `develop` branch and updated the diagnostics for both nested indefinite-length text and byte strings so they describe the invalid context directly. I also updated the exact-message regression tests and regenerated `single_include/nlohmann/json.hpp`.

The affected C++11/C++17 multi-header tests and C++17 single-header tests pass locally. A clean bounded C++17 run also passed 86/86 tests, excluding the five exhaustive Unicode executables.

tests/src/unit-cbor.cpp
```diff
@@ -2493,10 +2493,8 @@ TEST_CASE("examples from RFC 8949 Appendix A")
     {
         const auto packed = utils::read_binary_file(TEST_DATA_DIRECTORY "/binary_data/cbor_binary.cbor");
         json j;
-        CHECK_NOTHROW(j = json::from_cbor(packed));
-
-        const auto expected = utils::read_binary_file(TEST_DATA_DIRECTORY "/binary_data/cbor_binary.out");
-        CHECK(j == json::binary(expected));
+        // The fixture ends with nested indefinite-length byte strings, which RFC 8949 Section 3.2.3 forbids.
+        CHECK_THROWS_WITH_AS(j = json::from_cbor(packed), "[json.exception.parse_error.113] parse error at byte 513: syntax error while parsing CBOR binary: indefinite-length binary array is not allowed inside indefinite-length binary array; last byte: 0x5F", json::parse_error&);
```
nlohmann, line 2497: This replaces the only test that verified *correct byte-for-byte decoding* of a real, multi-chunk CBOR byte-array fixture (previously `CHECK_NOTHROW(...); CHECK(j == json::binary(expected));` against the 513-byte `cbor_binary.cbor`/`cbor_binary.out` pair) with a pure `throws` check. The remaining coverage for successful indefinite-length chunk concatenation is now just two trivial 2-chunk inline cases (`unit-cbor.cpp:1593-1599` and `:2489`). Worth adding either a truncated-valid-prefix decode check or an inline 3+-chunk success test to keep the concatenation logic itself covered.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -1062,10 +1063,16 @@ class binary_reader
 
             case 0x7F: // UTF-8 string (indefinite length)
             {
+                if (JSON_HEDLEY_UNLIKELY(!allow_indefinite))
```
nlohmann, line 1066: This rejection block is duplicated near-verbatim in `get_cbor_binary`'s `case 0x5F` (and mirrored again in `single_include/nlohmann/json.hpp`, so 4 copies total of the same logic). A small shared helper (e.g. taking the type-name string) would avoid the two messages drifting out of sync on a future edit.

nlohmann (commented):
_Review written by Claude Code on behalf of @nlohmann._

Core fix looks correct: it enforces RFC 8949 §3.2.3 by rejecting a second level of indefinite-length chunks, and I traced the byte-position/error-message math for both new regression tests against the actual behavior — it matches. A few smaller things worth a look before merging:

1. **Test coverage regression** (see inline comment on `tests/src/unit-cbor.cpp`): the only test that verified *correct decoding* of a multi-chunk CBOR byte-array fixture was replaced by a pure `throws` check. The remaining coverage of successful chunk concatenation is two trivial 2-chunk cases.
2. **Misleading error message**: the pre-existing `default:` case in both `get_cbor_string` (line 1088) and `get_cbor_binary` (line 1195) still says `0x7F`/`0x5F` would be an acceptable byte here, but that's no longer true when this branch is reached while parsing a chunk (`allow_indefinite=false`) — confirmed reproducible with `json::from_cbor({0x7F, 0x7F, 0x00})`.
3. **Duplicated error blocks** (see inline comment): the new rejection block is copy-pasted between `get_cbor_string`/`get_cbor_binary` (and mirrored again in `single_include/nlohmann/json.hpp`), which risks message drift on future edits. Could be a small shared helper.
4. **Orphaned fixture**: `tests/data/binary_data/cbor_binary.out` is no longer referenced anywhere after this diff (only `cbor_binary.cbor` is still read, to prove the throw).
5. **PR description**: could use an explicit note that this is a behavior-visible change — code that previously parsed (leniently, non-conformant) nested indefinite-length CBOR chunks will now throw `parse_error.113`.

None of these block the core fix, which is sound.

nlohmann (changes requested):
Check https://github.com/nlohmann/json/pull/5325#pullrequestreview-4975398688

JosephDemarest:
Rebased onto current develop and adapted this to the iterative CBOR reader from #5502, preserving the nonrecursive parsing path.

The review points are addressed: successful multi-chunk coverage is restored (including a byte-for-byte check against cbor_binary.out for the valid fixture prefix); malformed-chunk errors no longer suggest an indefinite marker is valid inside another indefinite string; nested rejection formatting uses a shared helper; and the description explicitly calls out the compatibility change to parse_error.113.

Focused C++11, C++17, and single-header tests pass. The broader C++17 run passes 79/80 tests; the diagnostic-position test fails the same two swap-position assertions on untouched develop c41152e6. The single header is regenerated and reproducible.

nlohmann (changes requested):
Please rebase once more - then the CI should be green.

JosephDemarest:
Thanks — I rebased onto current `develop` (`4daca40d7`). The PR patch is equivalent to the previous revision (`git range-diff` shows `=`). Focused C++11, C++17, and generated single-header CBOR/regression tests pass; the diagnostic-position test that failed on the old base now passes too. Regenerating and formatting the header produces no content change. I'll check the new CI results.

PR #5327, opened 2026-07-28
reject out-of-range array index in const json_pointer access

Angadi56:
The const `operator[](json_pointer)` overload resolves an array reference token with `ptr->operator[](array_index(token))`, which lands on the const `operator[](size_type)`, a bare `std::vector::operator[]` with no range check. `array_index()` accepts any decimal in `0 .. SIZE_MAX-1`, so on `{"a":[1,2,3]}` a pointer like `/a/5` computes `array->data() + idx * sizeof(basic_json)` and hands back a const reference built from whatever heap bytes follow the array. AddressSanitizer reports a heap-buffer-overflow read there. Since the type byte and the `json_value` union are then read from unrelated memory, a type byte that happens to land on `object`/`array`/`string`/`binary` makes the following 8 bytes get used as a pointer and dereferenced.

The check belongs in `get_unchecked()` rather than at the call site: a caller passes a string, so it cannot tell whether a token is in range without resolving the pointer itself. The same function already throws `out_of_range.402` for `-`, which is likewise an index past the end, and `get_checked()` a few lines below performs exactly the `idx >= size()` test that was missing here, so this mirrors it and reuses its 401 message. That makes `j["/a/5"_json_pointer]` report what `j.at("/a/5"_json_pointer)` already reports for the same pointer. In-range pointers resolve unchanged, and the non-const overload keeps filling the array with nulls as documented.

Verified with the full test suite (109/109) plus `make check-amalgamation` and `cpplint`. The new checks in `unit-json_pointer.cpp` fail on develop and pass here.

### DIFF
--- a/docs/mkdocs/docs/api/basic_json/operator[].md
+++ b/docs/mkdocs/docs/api/basic_json/operator[].md
@@ -79,6 +79,8 @@ Strong exception safety: if an exception occurs, the original value stays intact
       JSON pointer `ptr` begins with '0'.
     - Throws [`parse_error.109`](../../home/exceptions.md#jsonexceptionparse_error109) if an array index in the passed
       JSON pointer `ptr` is not a number.
+    - Throws [`out_of_range.401`](../../home/exceptions.md#jsonexceptionout_of_range401) if an array index in the
+      passed JSON pointer `ptr` is out of range for the const version.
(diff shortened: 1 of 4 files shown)


nlohmann (commented):
But operator[] is not supposed to do range checks - this is what at() is for. Or what am I missing?

PR #5328, opened 2026-07-28
Avoid allocating temporary basic_json for cbor and msgpack object keys

### LINKED ISSUE #5318: to_cbor()/to_msgpack() allocate a temporary basic_json for every object key
### Description

`write_cbor()` serializes object keys by calling itself on the key:

https://github.com/nlohmann/json/blob/8ec98e2c9ea9c0adc15a791feb3d4d4468621743/include/nlohmann/detail/output/binary_writer.hpp#L403-L407

`el.first` is a `string_t`, but the parameter is `const BasicJsonType&`, so each key is implicitly converted to a temporary `basic_json` — which heap-allocates a `string_t` (plus its buffer for keys beyond SSO) — purely to reach the `value_t::string` case a few lines above. `write_msgpack()` has the identical pattern.

The UBJSON and BSON writers in the same file already do it the direct way, writing the key's characters straight to the output adapter without constructing a `basic_json`.

### Measurement

Counting global `operator new` calls while serializing a 1000-entry object (`-O2`, g++ 13.3.0):

| case | allocations |
|---|---|
| `to_cbor`, array of 1000 ints (baseline) | 13 |
| `to_cbor`, object, 1000 short keys (SSO) | **1016** |
| `to_cbor`, object, 1000 64-char keys | **2016** |
| `to_msgpack`, object, 1000 short keys | **1016** |
| `to_ubjson`, object, 1000 short keys | 17 |
| `to_bson`, object, 1000 short keys | 14 |

So CBOR and MessagePack pay one allocation per key even when the key fits in the small-string buffer (the `string_t` object itself is heap-allocated by `basic_json`), and two per key otherwise. UBJSON and BSON pay none.

### Minimal code example

```cpp
static long allocs = 0;
void* operator new(std::size_t n) { ++allocs; void* p = std::malloc(n); if (!p) throw std::bad_alloc(); return p; }

#include <nlohmann/json.hpp>
using json = nlohmann::json;

int main()
{
    json j = json::object();
    for (int i = 0; i < 1000; ++i) { char k[8]; std::snprintf(k, sizeof k, "k%04d", i); j[k] = i; }

    allocs = 0; (void)json::to_cbor(j);   std::printf("cbor   %ld\n", allocs); // 1016
    allocs = 0; (void)json::to_ubjson(j); std::printf("ubjson %ld\n", allocs); // 17
}
```

### Suggested fix

Factor the string-writing body of the `value_t::string` case out of `write_cbor` into a helper taking `const string_t&` (say `write_cbor_string`), call it from both the `value_t::string` case and the key loop, and do the same for `write_msgpack`. This is the same shape as the BSON writer cleanup in #5313, which changed `write_bson_unsigned` to take the underlying value rather than the `basic_json`.

g++ 13.3.0, Ubuntu 24.04, `-std=c++14`, `-O2`

### Library version

`develop` @ 8ec98e2

No. This is a purely internal refactor of two `private` member functions of `detail::binary_writer`:

- **Byte-for-byte identical output.** The extracted helper would run the same length-prefix and character-writing code the `value_t::string` case runs today, so `to_cbor`/`to_msgpack` produce exactly the same bytes. Existing round-trip tests should pass unchanged.
- **No API/ABI surface.** `binary_writer` is in `detail` and the affected functions are not part of the documented API. No signature in `basic_json` changes.
- **One thing to watch:** object keys are `object_t::key_type`, which is not necessarily `string_t` for a custom `ObjectType`. The helper should be written so key types that merely convert to `string_t` still compile — either by templating it or by keeping `write_cbor(el.first)` as a fallback for non-`string_t` key types. Getting this wrong would break exotic `basic_json` specializations at compile time, so it deserves a test with a custom object type.

I'm happy to open a PR for this if it's wanted.

alexprabhat99:
This PR optimizes CBOR and MessagePack serialization by avoiding the construction of a temporary basic_json object for every object key during serialization.

Previously, object keys were passed through write_cbor()/write_msgpack(), which expected a BasicJsonType and therefore implicitly constructed a temporary basic_json for each key. This introduced unnecessary heap allocations during serialization.

This change factors out the string serialization logic into dedicated helper functions and serializes object keys directly. The implementation preserves compatibility with custom object key types while eliminating the extra basic_json allocations for the common case where object_t::key_type is string_t.

A regression test has been added to verify compatibility with custom object key types.

Fixes #5318

Compatibility note: This does not change the public API (detail::binary_writer is internal). For custom object_t::key_type implementations, CBOR and MessagePack object keys are now required to be implicitly convertible to string_t. Previously, a custom key type handled only through to_json could be serialized through the generic JSON serializer, potentially producing non-string map keys that the library's own CBOR/MessagePack readers would reject.


alexprabhat99:
Hi @nlohmann 
All checks are successful now. 
Thanks!

gregmarr (approved):
Looks good to me.

nlohmann:
Thanks for tackling this! Overall this looks good and does exactly what #5318 asked for. A few minor comments below.

### The claimed win is real

I compiled the allocation-counting program from #5318 against `single_include/nlohmann/json.hpp` from `develop` and from this branch (clang, `-std=c++14 -O2`), 1000-key object:

| case | develop | this PR |
|---|---|---|
| `to_cbor`, short (SSO) keys | 1015 | **15** |
| `to_cbor`, 64-char keys | 2015 | **15** |
| `to_msgpack`, short keys | 1015 | **15** |
| `to_msgpack`, 64-char keys | 2015 | **15** |
| `to_ubjson` / `to_bson` (unchanged) | 16 / 14 | 16 / 14 |

CBOR and MessagePack now sit at the same allocation level as UBJSON and BSON. The extracted helper bodies are byte-for-byte the originals, and the `value_t::string` cases correctly reach the non-template overload via `*j.m_data.m_value.string`.

### Comments

**1. Missing SPDX header on the new test file**

`tests/src/custom_object_key_type.hpp` starts straight at `#pragma once`. Every other file under `tests/src` carries the ASCII-art banner plus `SPDX-FileCopyrightText` / `SPDX-License-Identifier` from the `json_support` REUSE template. `ci_reuse_compliance` passes only because `.reuse/dep5` has a `Files: *` catch-all, so CI can't catch this. Running `make reuse` should fix it.

**2. `key::c_str()` in the test helper looks like dead code**

It was added in d10514d31 ("fixes for failing ci"), but nothing on the CBOR/MessagePack path calls it — the template overload routes through `string_t(value)`. I deleted it locally and the test still compiles and passes under `-std=c++11/14/17/20/23`. If the intent was UBJSON compatibility, the type would also need `size()`, which it doesn't have (`binary_writer.hpp` line 940 uses both), so `to_ubjson` wouldn't compile with this key type anyway.

**3. The unconstrained template overloads may not be needed, and are undocumented**

`write_cbor_string(const KeyType&)` and `write_msgpack_string(const KeyType&)`. The BSON writer already solves the same problem by just declaring `write_bson_element(const string_t& name, ...)` and letting implicit conversion handle custom key types. I removed both templates from this branch's header and the new tests still compile and pass — the test's `key` has an implicit `operator std::string()`, so `const string_t&` binds directly.

So the templates' only added capability (key types with an *explicit*-only conversion to `string_t`) isn't exercised by the new tests, and they carry a small maintenance hazard: any future `write_cbor_string(x)` on a non-`string_t` silently allocates a temporary. Either drop them for consistency with the BSON writer, or keep them with a one-line comment explaining why — and make the test key's conversion operator `explicit` so the overload is actually load-bearing.

**4. Worth noting the source-compatibility nuance in the PR description**

Previously keys went through `write_cbor()` / `write_msgpack()`, i.e. through the JSON serializer. A key type with a `to_json` but no string conversion used to compile (emitting a non-string CBOR/MessagePack key, which this library's own reader then rejects); it now fails to compile. A key type with *both* a `to_json` and a string conversion changes its encoding.

I think the new behavior is the correct one, but since it's the convention here to state API impact: no public API break (`detail::binary_writer` is internal), with this source-level change for exotic `object_t::key_type`s.

**5. Nit**

Double blank line after the `write_msgpack_string` template, in both `include/nlohmann/detail/output/binary_writer.hpp` and the amalgamated `single_include/nlohmann/json.hpp`. astyle doesn't collapse those, so CI stayed green.

### Optional

There's no regression test guarding the allocation count, so this could silently regress later. Given it needs a global `operator new` override, it's probably not worth adding to the doctest binaries — just noting it.

<sub>Reviewed with the help of Claude Code.</sub>

nlohmann (changes requested):
Check https://github.com/nlohmann/json/pull/5328#issuecomment-5346167177

alexprabhat99:
Hi @nlohmann 
I have addressed the review comments. The one failing test seems unrelated to the PR changes.

nlohmann:
Please check the [DCO requirements](https://github.com/nlohmann/json/pull/5328/checks?check_run_id=97435561176).

alexprabhat99:
Hi @nlohmann 
All checks have passed.

alexprabhat99:
Hi @nlohmann
Gentle ping! All checks have passed.

nlohmann (changes requested):
Thanks a lot, and sorry for the delay.

Two issues I would like you to address:

1. A `static_assert` on `std::is_convertible<object_t::key_type, string_t>` in the two key loops, so a rejected key type produces one readable line instead of a template wall.
2. One sentence in `docs/mkdocs/docs/api/basic_json/object_t.md` stating that `object_t::key_type` must be convertible to `string_t`. The page already implies it by describing the type as `ObjectType<StringType, ...>`, but the new tests make it a contract, so let's write it down.

**Minor:** the key in `tests/src/custom_object_key_type.hpp` carries a `c_str()` that no CBOR/MsgPack test uses. That's what UBJSON would need - it calls `el.first.size()` and `.c_str()` on the key directly - and in fact this key type still doesn't compile with `to_ubjson`. Pre-existing and out of scope here, but it means "custom key types are supported" isn't uniformly true yet. Either drop the unused `c_str()` or leave a comment saying why it's there; the UBJSON gap can be a follow-up issue.

nlohmann:
Please fix the [DCO](https://github.com/nlohmann/json/pull/5328/checks?check_run_id=102472478551) requirements.

single_include/nlohmann/json.hpp
```diff
@@ -17342,10 +17152,14 @@ class binary_writer
                 }
                 // LCOV_EXCL_STOP
 
+                static_assert(
```
gregmarr: Looks like you added this manually and only to the `single_include` file.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -402,7 +370,7 @@ class binary_writer
                 // step 2: write each element
                 for (const auto& el : *j.m_data.m_value.object)
                 {
-                    write_cbor(el.first);
+                    write_cbor_string(el.first);
```
nlohmann, line 380: The `static_assert(std::is_convertible<typename BasicJsonType::object_t::key_type, string_t>::value, ...)` that guards this call in `single_include/nlohmann/json.hpp` isn't present here in the modular header. Regenerating `single_include` from `include/` (as `check_amalgamation.yml` does on every PR) would drop those 4 lines, producing a nonzero diff and failing that CI job. Same issue applies to the MessagePack object loop further down in this file.

nlohmann (commented):
Automated review findings — posted via [Claude Code](https://claude.com/claude-code) on behalf of @nlohmann.

Two issues found with high confidence:

1. The `static_assert` added to `single_include/nlohmann/json.hpp` guarding `object_t::key_type` convertibility isn't mirrored in the modular `include/` header it's generated from. Since `single_include` is produced from `include/` via `make amalgamate`, this diff will fail the `check_amalgamation` CI job, and the modular-header build silently loses the static_assert.
2. The new `custom_object_key_type.hpp` test fixture is missing a `size()` method to match its `c_str()` shim, making it incompatible with UBJSON's key-writing path (verified by compiling `to_ubjson` against it) despite the comment claiming broader serializer compatibility.

See inline comments below for exact locations.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -724,7 +663,7 @@ class binary_writer
                 // step 2: write each element
                 for (const auto& el : *j.m_data.m_value.object)
                 {
-                    write_msgpack(el.first);
+                    write_msgpack_string(el.first);
```
nlohmann, line 679: Same missing `static_assert` issue as the CBOR object loop above — it's present in `single_include/nlohmann/json.hpp` for this MessagePack loop too, but not here.

tests/src/custom_object_key_type.hpp
```diff
@@ -0,0 +1,74 @@
...
+        : m_value(std::move(value))
+    {}
+
+    operator std::string() const
+    {
+        return m_value;
+    }
+
+    // Kept for compatibility with serializers that access object keys as C strings.
+    const char* c_str() const noexcept
```
nlohmann, line 39: The comment says this is "kept for compatibility with serializers that access object keys as C strings," but there's no matching `size()`. UBJSON's `write_ubjson` accesses object keys via `el.first.size()` and `el.first.c_str()` directly (not through the `string_t` conversion operator CBOR/MsgPack/BSON use), so `custom_object_key_test::json::to_ubjson(...)` fails to compile with "no member named 'size' in 'custom_object_key_test::key'" if this shared fixture is ever reused for a UBJSON test. Not exercised by this PR (CBOR/MsgPack only), so it's latent for now.

nlohmann:
Please rebase to the latest develop to resolve conflicts.

PR #5330, opened 2026-07-29
Document BSON interoperability and subtype-less binary round trips

### LINKED ISSUE #5311: BSON docs: missing interoperability warning for marker 0x11, and binary round-trip asymmetry not mentioned
### Description

Two gaps on the [BSON feature page](https://json.nlohmann.me/features/binary_formats/bson/) (`docs/mkdocs/docs/features/binary_formats/bson.md`).

**1. Marker `0x11` interoperability is only covered in one direction.**

The serialization table maps `number_unsigned` in `9223372036854775808..18446744073709551615` to marker `0x11`, with no caveat. In the BSON spec, `0x11` is **Timestamp** (a MongoDB-internal uint64 type), not a general unsigned integer — so documents this library writes will be read as timestamps by mongod and other drivers.

The existing note covers only the reading direction:

> BSON type 0x11 is used to represent uint64 numbers. This library treats these values purely as uint64 numbers and does not parse them into date-related formats.

It appears under "Deserialization" and says nothing about what happens when *other* implementations read what this library writes. Since #4286 (v3.12.0) made `to_bson()` emit `0x11` rather than throwing `out_of_range.407`, the serialization side needs the same warning.

**2. Binary round-trip is not identity.**

A binary value with no subtype is written with subtype `0x00` and read back *with* subtype `0`, so `from_bson(to_bson(j)) != j`:

```cpp
json j;
j["b"] = json::binary({1, 2, 3});          // has_subtype() == false
const auto k = json::from_bson(json::to_bson(j));
// k["b"].get_binary().has_subtype() == true
// j == k  -->  false
```

This is stated in `docs/mkdocs/docs/features/binary_values.md` ("If no subtype is given, the generic binary subtype 0x00 is used"), but not on the BSON page, and the consequence for round-tripping is not spelled out anywhere. It is also untested: `tests/src/unit-bson.cpp` only round-trips binary values that already carry an explicit subtype (`json::binary(s, 0)` and `json::binary(md5hash, 5)`), so nothing would catch a regression here.

### Suggested fix

- Add an interoperability warning next to the `0x11` row in the serialization table, or move/extend the existing note so it covers both directions.
- Note the subtype-less binary asymmetry on the BSON page and add a round-trip test for `json::binary({...})` without a subtype.

### Notes

This issue covers only the documentation gaps from a wider BSON review. The unbounded-recursion concern found in the same pass is already tracked in #5104 and addressed by #5293, so it is deliberately not repeated here.

### Library version

`develop` @ 06ac77f

YingqiDuan:
## Summary

- document that `number_unsigned` values above `INT64_MAX` are serialized as BSON type `0x11`, which the BSON specification defines as Timestamp and other implementations may therefore read as Timestamps
- clarify that deserializing BSON type `0x11` preserves its 64-bit value as `number_unsigned` but does not preserve the Timestamp type information
- document the subtype-less binary asymmetry: serialization uses the generic subtype `0x00`, so the deserialized value has an explicit subtype and `from_bson(to_bson(j)) != j`
- add a BSON unit test that verifies the encoded subtype and the round-trip result

This makes both behaviors and their consequences clear on the BSON feature page, and adds coverage for the previously untested subtype-less case.

Closes #5311.

## Validation

- full CMake build completed successfully; CTest: 91/91 passed
- `make build` in `docs/mkdocs` passed, including the documentation structure check
- reviewed the rendered BSON page locally
- `git diff --check upstream/develop...HEAD` was clean

## Checklist

- [ ] Code coverage — N/A (no library implementation code changed; a focused unit test was added).
- [ ] `make amalgamate` — N/A (no files under `include/nlohmann` changed).

### DIFF
--- a/docs/mkdocs/docs/features/binary_formats/bson.md
+++ b/docs/mkdocs/docs/features/binary_formats/bson.md
@@ -35,6 +35,19 @@ The library uses the following mapping from JSON values types to BSON types:
     The mapping is **incomplete**, since only JSON-objects (and things contained therein) can be serialized to BSON.
     Also, keys may not contain U+0000, since they are serialized a zero-terminated c-strings.
 
+!!! warning "BSON type 0x11 interoperability"
+
+    The BSON specification defines type `0x11` as a Timestamp. This library uses marker `0x11` when serializing
+    `number_unsigned` values in the range `9223372036854775808..18446744073709551615`. Other BSON implementations may
+    therefore interpret these values as Timestamps instead of unsigned integers.
+
+!!! info "Binary values without a subtype"
+
+    BSON requires every binary value to have a subtype. If a binary value has no subtype, this library serializes it
+    with the generic subtype `0x00`. After deserialization, `has_subtype()` returns `true` and `subtype()` returns `0`.
+    As a result, serializing and deserializing a JSON object containing such a value produces a different JSON object,
+    even though the binary data is unchanged.
+
 ??? example
 
     ```cpp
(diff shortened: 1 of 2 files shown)


nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #5331, opened 2026-07-29
Fix CBOR tag handlers not recognizing tags 0-5 and 21-23

### LINKED ISSUE #5315: CBOR: cbor_tag_handler_t::ignore does not ignore tags 0-5 and 21-23 (head bytes 0xC0-0xC5, 0xD5-0xD7)
### Description

`cbor_tag_handler_t::ignore` is documented as "ignore tags", but `from_cbor(…, cbor_tag_handler_t::ignore)` still throws `parse_error.112` for CBOR tags 0–5 and 21–23.

Every byte in `0xC0`–`0xDB` is major type 6 (tag). The `switch` in `parse_cbor_internal()` enumerates only `0xC6`–`0xD4` and `0xD8`–`0xDB` ([`binary_reader.hpp:763-781`](https://github.com/nlohmann/json/blob/8ec98e2c9ea9c0adc15a791feb3d4d4468621743/include/nlohmann/detail/input/binary_reader.hpp#L763-L781)), so head bytes `0xC0`–`0xC5` and `0xD5`–`0xD7` fall through to `default:` and are reported as "invalid byte" — in *every* handler mode, including `ignore` and `store`.

The gap is at both ends of the range and covers the tags most likely to appear in real data: 0 (RFC 3339 date/time string), 1 (epoch time), 2/3 (bignum), 4 (decimal fraction), 5 (bigfloat), and 21/22/23 (expected base64url/base64/base16 conversion).

There is also an interop consequence. An encoder using preferred serialization (RFC 8949 §4.1) encodes a tag number below 24 in the 1-byte head, e.g. tag 23 as `0xD7`. This library writes binary subtypes as `0xD8 0x17` instead (see `write_cbor`, `binary_writer.hpp:310-332`), so its own round-trip works — but a byte string tagged 23 by a conforming encoder cannot be read back at all, in any mode.

### Reproduction steps

Parse a tagged value with `cbor_tag_handler_t::ignore` for each tag head byte in `0xC0`–`0xDB`.

Expected: `ignore` skips the tag head and parses the enclosed data item, uniformly for all of `0xC0`–`0xDB`.

Actual, sweeping the whole range with the tagged item being a byte string (`Y` = parses, `n` = `parse_error.112`):

| head byte | tag | `error` | `ignore` | `store` |
|---|---|---|---|---|
| `0xC0`–`0xC5` | 0–5 | n | **n** | **n** |
| `0xC6`–`0xD4` | 6–20 | n | Y | Y |
| `0xD5`–`0xD7` | 21–23 | n | **n** | **n** |
| `0xD8`–`0xDB` | 1/2/4/8-byte argument | n | Y | Y |

A realistic document fails:

```
map with tag-0 datetime, ignore : THROW: [json.exception.parse_error.112]
    parse error at byte 4: syntax error while parsing CBOR value: invalid byte: 0xC0
```

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <iostream>
using json = nlohmann::json;

int main()
{
    // {"t": 0("2026-07-27T00:00:00Z")} -- a text string tagged with tag 0
    const std::vector<std::uint8_t> v{
        0xA1, 0x61, 0x74, 0xC0, 0x74, '2','0','2','6','-','0','7','-','2','7',
        'T','0','0',':','0','0',':','0','0','Z'};

    // works: tag 6 in the same position is skipped
    const std::vector<std::uint8_t> w{0xC6, 0x18, 0x2A};
    std::cout << json::from_cbor(w, true, true, json::cbor_tag_handler_t::ignore) << "\n";

    // throws: tag 0 is not skipped
    std::cout << json::from_cbor(v, true, true, json::cbor_tag_handler_t::ignore) << "\n";
}
```

### Error messages

```
[json.exception.parse_error.112] parse error at byte 4: syntax error while parsing CBOR value: invalid byte: 0xC0
```

### Library version

`develop` @ 8ec98e2

Not in the API/ABI sense — no signatures, types, or enumerators change, and the amalgamated header stays source- and binary-compatible.

Behaviorally it is a narrow, opt-in-scoped change:

- **Default behavior is untouched.** `cbor_tag_handler_t::error` is the default and must keep throwing on every tag, including these. Callers who never pass a tag handler see no difference at all.
- **`ignore`/`store` callers would see input that currently throws start to parse.** That is a real behavior change, but it moves in the direction those callers explicitly asked for. Code that relies on `ignore` throwing for tag 0 while silently accepting tag 6 seems very unlikely to be intentional.
- **This is not a request to interpret the tags.** No bignum/date decoding is implied — the tag head is skipped and the enclosed data item is parsed on its own, exactly as tags 6–20 are handled today. A tagged text string stays a string; `0xC2` over a byte string stays a byte string, not an integer.
- **Test and doc churn.** `unit-cbor.cpp`'s "all CBOR first bytes" test lists `0xc0`–`0xc5` and `0xd5`–`0xd7` in its `unsupported` set, and the "Incomplete mapping" warning in `docs/mkdocs/docs/features/binary_formats/cbor.md` lists them as parse errors. Both would need updating, which also means the current behavior is documented rather than accidental — so this may warrant a release note even though nothing breaks.

A conservative alternative, if changing `ignore` is considered too much: leave the reader as is and only document the asymmetry more explicitly. That keeps the interop gap, though.

Related: #1968 added tag skipping in 3.9.0 and closed on the `0xD9 0xD9 0xF7` self-describe case, which lands inside the supported `0xD8`–`0xDB` range — the two omitted ranges were likely just missed then.

sahilkamate03:
Fixes #5315.

## What changed
- Added the missing `case 0xC0`-`0xC5` and `case 0xD5`-`0xD7` labels to the tagged-item switch in binary_reader::parse_cbor_internal()` (`include/nlohmann/detail/input/binary_reader.hpp`), so the full `0xC0`-`0xDB` tag-header range is handled the same way.
- Extended the `"Tagged values"` test in `tests/src/unit-cbor.cpp` to cover `0xC0`-`0xD7` (previously only `0xC6`-`0xD4` was tested) under `error`, `ignore`, and `store` handlers.
- Updated `docs/mkdocs/docs/features/binary_formats/cbor.md` to state the tagged-item range as `0xC0`..`0xDB` instead of leaving it unspecified.
- Regenerated `single_include/nlohmann/json.hpp` via `make amalgamate`.

## Why
- CBOR tags 0-5 (`0xC0`-`0xC5`: RFC 3339 date/time, epoch time, bignum, decimal, bigfloat) and tags 21-23 (`0xD5`-`0xD7`: base64url, base64, base16 conversion hints) are valid major-type-6 tags per RFC 8949, but the switch statement only recognized `0xC6`-`0xD4` and `0xD8`-`0xDB`. 
- The gap made these bytes fall through to the `default` case and throw `parse_error.112` ("invalid byte"), even under `cbor_tag_handler_t::ignore` and `::store` — handlers that are documented to accept *any* tag. This broke interop with compliant ncoders that use preferred (shortest) serialization, since e.g. tag 23 is encoded as the single byte `0xD7`.

## Checklist

- [x] Code coverage remains at 100% — full suite passes 109/109.
- [x] Documentation updated.
- [x] Source amalgamated via `make amalgamate` (verified: diff matches the `include/nlohmann` source change exactly).

## Test plan

- [x] `cmake -S. -B build && cmake --build build -j10 && ctest --test-dir build -j10` → 100% tests passed, 0 failed out of 109
- [x] `test-cbor_cpp11` covers the new `0xC0`-`0xD7` tag range under `error`, `ignore`, and `store` handlers


tests/src/unit-cbor.cpp
```diff
@@ -2529,11 +2529,13 @@ TEST_CASE("Tagged values")
...
 
-    SECTION("0xC6..0xD4")
+    SECTION("0xC0..0xD7")
     {
         for (const auto b : std::vector<std::uint8_t>
     {
-        0xC6, 0xC7, 0xC8, 0xC9, 0xCA, 0xCB, 0xCC, 0xCD, 0xCE, 0xCF, 0xD0, 0xD1, 0xD2, 0xD3, 0xD4
+        0xC0, 0xC1, 0xC2, 0xC3, 0xC4, 0xC5,
+        0xC6, 0xC7, 0xC8, 0xC9, 0xCA, 0xCB, 0xCC, 0xCD, 0xCE, 0xCF, 0xD0, 0xD1, 0xD2, 0xD3, 0xD4,
+        0xD5, 0xD6, 0xD7
```
nlohmann, line 2577: The tests only wrap a string value for each new tag byte. Might be worth a case wrapping a binary payload under `cbor_tag_handler_t::store` to confirm 0xC0-0xC5/0xD5-0xD7 take the same "unwrap, don't treat as subtype" path as 0xC6-0xD4 already do.

docs/mkdocs/docs/features/binary_formats/cbor.md
nlohmann, line 1: The "Incomplete mapping" warning still lists `date/time (0xC0..0xC1)`, `bignum (0xC2..0xC3)`, `decimal fraction (0xC4)`, `bigfloat (0xC5)`, and `expected conversions (0xD5..0xD7)` as types that "are not supported and will yield parse errors." That's now only true for the default `error` handler — under `ignore`/`store` these parse fine, same as `0xC6..0xD4`/`0xD8..0xDB`, which were already correctly omitted from this list. Since the PR already updates the "Tagged items" warning just below to say the full `0xC0..0xDB` range is tag-handler-dependent, these five bullets should be deleted from the "Incomplete mapping" box to avoid the two warnings contradicting each other.

nlohmann (changes requested):

nlohmann:
Apologies for the AI-assisted review here — I don't have time for a deeper pass on this one right now, so this is a quick comparison rather than a full re-review.

We got a second PR for the same issue, #5367, which has the identical code fix (same `case 0xC0`-`0xC5` / `0xD5`-`0xD7` additions in `binary_reader.hpp`). It also independently fixes the docs issue flagged above — removing the stale "Incomplete mapping" bullets for tags 0-5/21-23 and adding a cross-reference note — which you could port over here.

The other ask from the review above (a `store`-mode test with a binary payload, not just a string, to confirm the new tag bytes take the same "unwrap, don't treat as subtype" path as the existing ones) is still outstanding in both PRs.

Given this one already has a clean CI run and is otherwise the more complete PR, it's the better one to land — just needs the docs fix and the extra test case.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #5335, opened 2026-07-30
Update CBOR references and half-float assertion bounds

### LINKED ISSUE #5319: CBOR: docs and comments cite the obsoleted RFC 7049, and the half-precision decoder has two tautological assertions
### Description

Two small cleanups in the same corner of the CBOR support.

**1. RFC 7049 was obsoleted by RFC 8949 (December 2020).** All six references in the tree still point at 7049, and none at 8949:

| location | text |
|---|---|
| `docs/mkdocs/docs/features/binary_formats/cbor.md:10` | "RFC 7049 - the CBOR specification" |
| `docs/mkdocs/docs/features/binary_formats/cbor.md:15` | "according to the CBOR specification (RFC 7049)" |
| `docs/mkdocs/docs/api/basic_json/binary_t.md:16` | "CBOR's RFC 7049 describes this type as:" |
| `include/nlohmann/detail/input/binary_reader.hpp:899` | "Code from RFC 7049, Appendix D, Figure 3" |
| `include/nlohmann/detail/input/binary_reader.hpp:2487` | same comment in the MessagePack half-float path |
| `tests/src/unit-cbor.cpp:2312` | `TEST_CASE("examples from RFC 7049 Appendix A")` |

The two `tools.ietf.org` links should also move to `rfc-editor.org`, and the Appendix D/Figure 3 citations need renumbering — the half-precision decoder sample is Appendix D in 7049 but the corresponding material sits in RFC 8949 §D/Appendix D with different figure numbering, so the comment should be checked rather than mechanically renamed. #2779 asked which RFC the library targets and was answered in the affirmative for 8949's data model; the docs never caught up.

**2. Tautological bounds in the half-precision decoder.**

https://github.com/nlohmann/json/blob/8ec98e2c9ea9c0adc15a791feb3d4d4468621743/include/nlohmann/detail/input/binary_reader.hpp#L910-L913

`exp` is `(half >> 10u) & 0x1Fu`, so its range is [0, 31]; `mant` is `half & 0x3FFu`, so its range is [0, 1023]. Both assertions therefore hold for every possible input and can never fire. The intended bounds are presumably `exp <= 31` and `mant <= 1023`. `0 <= exp` is also always true for a value derived from a masked unsigned and may draw `-Wtype-limits` on some configurations.

The decode arithmetic itself is correct — I checked subnormals (`ldexp(mant, -24)`), normals (`ldexp(mant + 1024, exp - 25)`), ±0, ±infinity and NaN, and they all decode to the right values. This is only about the assertions being misleading.

Expected: references name the current specification; assertions state the actual invariants.

Actual: as above.

N/A (documentation and assertions)

### Library version

`develop` @ 8ec98e2

No, on both counts.

- The RFC references are prose and comments only. Nothing in the mapping tables changes: the CBOR data model this library implements is the same in 7049 and 8949, and the encodings it reads and writes are unaffected. The only visible change is a test case *name*, which is not part of any API.
- Correcting the assertion bounds cannot change behavior: the old predicates were true for all inputs and the new ones are true for all inputs too, so no build — with or without `NDEBUG` — behaves differently. If anything it silences a latent warning.

Worth noting that a few nearby doc statements are worth re-checking against 8949 while touching this, in particular the "preferred serialization" wording (§4.1), which the writer does not follow for tag heads below 24 — see #5315.

Patrick10199:
## Summary

- Correct the CBOR half-precision exponent and mantissa assertion bounds.
- Update CBOR documentation, comments, and test naming from RFC 7049 to RFC 8949.
- Regenerate the amalgamated header.

## Rationale

The decoder masks the exponent with `0x1F` and the mantissa with `0x3FF`, so their maximum values are 31 and 1023. The previous bounds of 32 and 1024 could never be reached.

RFC 8949 obsoletes RFC 7049. The existing Appendix D, Figure 3 citation remains valid in RFC 8949. This change corrects debug invariants and stale references without changing CBOR encoding or decoding behavior.

## Testing

- `cmake --build build --parallel`
- `test-cbor_cpp11` passed.
- Full local CTest run: 106/109 passed. The three remaining MinGW floating-point parsing failures reproduced unchanged on the unmodified base commit `2e2368709`.
- Formatted with Artistic Style 3.4.13 and regenerated `single_include/nlohmann/json.hpp`.

Closes #5319

### DIFF
--- a/docs/mkdocs/docs/api/basic_json/binary_t.md
+++ b/docs/mkdocs/docs/api/basic_json/binary_t.md
@@ -13,9 +13,8 @@ is compatible with both of the binary data formats that use binary subtyping, (t
 incompatible with each other, and it is up to the user to translate between them). The subtype is added to `BinaryType`
 via the helper type [byte_container_with_subtype](../byte_container_with_subtype/index.md).
 
-[CBOR's RFC 7049](https://tools.ietf.org/html/rfc7049) describes this type as:
-> Major type 2: a byte string. The string's length in bytes is represented following the rules for positive integers
-> (major type 0).
+[CBOR's RFC 8949](https://www.rfc-editor.org/rfc/rfc8949.html#section-3.1) describes this type as:
+> Major type 2: A byte string. The number of bytes in the string is equal to the argument.
 
 [MessagePack's documentation on the bin type
 family](https://github.com/msgpack/msgpack/blob/master/spec.md#bin-format-family) describes this type as:
--- a/docs/mkdocs/docs/features/binary_formats/cbor.md
+++ b/docs/mkdocs/docs/features/binary_formats/cbor.md
@@ -7,12 +7,12 @@ extremely small code sizes, fairly small message size, and extensibility without
 
     - [CBOR Website](http://cbor.io) - the main source on CBOR
     - [CBOR Playground](http://cbor.me) - an interactive webpage to translate between JSON and CBOR
-    - [RFC 7049](https://tools.ietf.org/html/rfc7049) - the CBOR specification
+    - [RFC 8949](https://www.rfc-editor.org/rfc/rfc8949.html) - the CBOR specification
 
 ## Serialization
 
 The library uses the following mapping from JSON values types to CBOR types according to the CBOR specification
-([RFC 7049](https://www.rfc-editor.org/rfc/rfc7049.html)):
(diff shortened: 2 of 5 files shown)


nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

patrickarms:
Happy to help out! I'll keep poking around and see if there's anything else i can do.

/commented using the wrong account

PR #5339, opened 2026-07-30
fix: check CBOR tagged subtype reads

### LINKED ISSUE #5338: CBOR: get_number() return value discarded when reading tagged-binary subtype (0xD8-0xDB)
### Description

Following up on #5309 / #5332 (BSON reader): the same pattern — discarding the `bool` return value of `get_number()` — also exists in the CBOR tagged-binary-subtype handling in `parse_cbor_internal()`, for both the `ignore` and `store` tag-handler branches:

| location (`include/nlohmann/detail/input/binary_reader.hpp`, `develop` branch) | issue |
|---|---|
| line 800 (`case 0xD8`, `cbor_tag_handler_t::ignore`) | `get_number(input_format_t::cbor, subtype_to_ignore);` — return value ignored |
| line 806 (`case 0xD9`, `cbor_tag_handler_t::ignore`) | same, `std::uint16_t` |
| line 812 (`case 0xDA`, `cbor_tag_handler_t::ignore`) | same, `std::uint32_t` |
| line 818 (`case 0xDB`, `cbor_tag_handler_t::ignore`) | same, `std::uint64_t` |
| line 836 (`case 0xD8`, `cbor_tag_handler_t::store`) | `get_number(input_format_t::cbor, subtype);` — return value ignored, then used in `b.set_subtype(...)` |
| line 843 (`case 0xD9`, `cbor_tag_handler_t::store`) | same, `std::uint16_t` |
| line 850 (`case 0xDA`, `cbor_tag_handler_t::store`) | same, `std::uint32_t` |
| line 857 (`case 0xDB`, `cbor_tag_handler_t::store`) | same, `std::uint64_t` |

As with the BSON case, I have not found a way to turn this into an accepted malformed document — a truncated read here still produces `parse_error.110` from a later check (e.g. `get_cbor_binary`'s own EOF handling, or the subsequent `parse_cbor_internal(true, tag_handler)` recursive call hitting EOF). So this is a robustness/maintainability issue rather than a confirmed live bug, same as #5309 was before it was fixed.

But the same risk applies: in `cbor_tag_handler_t::store`, `set_subtype()` is called with a possibly-uninitialized `subtype` value before any EOF check happens, relying entirely on downstream validation to catch the truncation. If that downstream check ever changes (as happened with the BSON binary case, which was only safe because of the document-size check added in #5287), this could start passing malformed/truncated data as valid CBOR to the SAX consumer.

### Suggested fix

Same as #5332's approach for BSON: check the return value of each `get_number()` call and short-circuit (return `false` / propagate the failure) before using the subtype, instead of discarding it.

### Context

Noticed while reviewing #5332, which fixed the equivalent bug in the BSON reader.

tomatotomata:
Closes #5338\n\nThe CBOR tagged-binary subtype branches now propagate failed get_number() reads before using the subtype, for both the ignore and store handlers. The regression suite covers truncated 1-, 2-, 4-, and 8-byte subtype tags in both modes. The amalgamated single header was regenerated as required by the project.\n\nVerification:\n- git diff --check passes\n- Static inspection confirms all eight tagged subtype reads now check their return values\n- C++ tests could not be run in this Windows environment because no compiler or CMake is installed

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -807,25 +807,37 @@ class binary_reader
                             case 0xD8:
                             {
                                 std::uint8_t subtype_to_ignore{};
-                                get_number(input_format_t::cbor, subtype_to_ignore);
+                                if (!get_number(input_format_t::cbor, subtype_to_ignore))
+                                {
+                                    return false;
+                                }
                                 break;
                             }
                             case 0xD9:
                             {
                                 std::uint16_t subtype_to_ignore{};
-                                get_number(input_format_t::cbor, subtype_to_ignore);
+                                if (!get_number(input_format_t::cbor, subtype_to_ignore))
+                                {
+                                    return false;
+                                }
                                 break;
                             }
                             case 0xDA:
                             {
                                 std::uint32_t subtype_to_ignore{};
-                                get_number(input_format_t::cbor, subtype_to_ignore);
+                                if (!get_number(input_format_t::cbor, subtype_to_ignore))
+                                {
+                                    return false;
+                                }
                                 break;
                             }
                             case 0xDB:
                             {
                                 std::uint64_t subtype_to_ignore{};
-                                get_number(input_format_t::cbor, subtype_to_ignore);
+                                if (!get_number(input_format_t::cbor, subtype_to_ignore))
+                                {
+                                    return false;
+                                }
                                 break;
                             }
                             default:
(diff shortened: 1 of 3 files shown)


nlohmann (changes requested):
The logics looks good, but the amalgamation is wrong. Please re-amalgamate by running `make amalgamate`.

tomatotomata:
I applied the repository's amalgamation patch from the failed check and pushed 2f31abf45. That refreshes single_include/nlohmann/json.hpp and the affected regression-test formatting, with git diff --check clean. The PR is ready for another look; let me know what you think.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #5344, opened 2026-08-01
fix: restore the character that terminates a number

### LINKED ISSUE #5340: operator>> does not restore the character that terminates a number
### Summary

`operator>>` is documented to leave the stream "positioned right after" the value it parsed, so that it can be called repeatedly on a stream of concatenated JSON values. That holds for strings, arrays, objects and literals, but **not for numbers**: a number is only terminated by the character following it, and that character is consumed from the stream and never put back. The next extraction therefore starts one byte too late.

This came up while reviewing #5326, which documents the `parse()` vs `operator>>` strictness difference and repeats the positioning guarantee from [`operator>>`'s notes](https://json.nlohmann.me/api/operator_gtgt/#notes). Before we document the guarantee more prominently, we should decide whether to make it true or to qualify it.

### Reproduction

```cpp
#include <nlohmann/json.hpp>
#include <iostream>
#include <sstream>

int main()
{
    std::istringstream input("1true");
    nlohmann::json j1;
    input >> j1;                       // j1 == 1

    std::string rest;
    char c;
    while (input.get(c)) { rest += c; }
    std::cout << j1 << " | rest = \"" << rest << "\"\n";
}
```

Output:

```
1 | rest = "rue"
```

Expected (per the documented guarantee): `rest = "true"`, and a second `input >> j2` yielding `true`. Instead the second extraction throws:

```
[json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - invalid literal; last read: 'r'
```

More cases (current behaviour):

| input | first value | remaining stream |
| --- | --- | --- |
| `1true` | `1` | `rue` ❌ |
| `1 true` | `1` | `true` ✔ (the eaten byte was whitespace) |
| `truefalse` | `true` | `false` ✔ |
| `[1][2]` | `[1]` | `[2]` ✔ |
| `null null` | `null` | ` null` ✔ |

So the guarantee only breaks when a number is immediately followed by a non-whitespace character — `1true`, `1[2]`, `1{}`, `1"a"`. Concatenated values that are whitespace-separated, or whose first value is not a number, work as documented.

### Cause

`lexer::scan_number()` reads one character past the number and ungets it:

https://github.com/nlohmann/json/blob/dd24e2dffd734c24d71cd79d09dee32d72dde7ca/include/nlohmann/detail/input/lexer.hpp#L1277-L1280

But `lexer::unget()` is deliberately simulated — it only rewinds the lexer's own bookkeeping, not the input:

> We implement unget by setting variable next_unget to true. The input is not changed - we just simulate ungetting by modifying chars_read_total, chars_read_current_line, and token_string.

and `input_stream_adapter::get_character()` consumes via `sbumpc()` with no corresponding `sungetc()`:

https://github.com/nlohmann/json/blob/dd24e2dffd734c24d71cd79d09dee32d72dde7ca/include/nlohmann/detail/input/input_adapters.hpp#L127-L136

The simulated unget is correct and sufficient for `parse()`/`accept()`, which require the input to end after the value anyway. It is only observable when the caller keeps using the stream afterwards — i.e. `operator>>`, and `sax_parse` with `strict == false`.

### Options

1. **Fix the adapter.** Give `input_stream_adapter` a real unget (`sungetc()`, or `sputbackc()` for the character just read) and have `lexer::unget()` propagate it to the adapter when the adapter supports it. This makes the documented guarantee hold. It is a behaviour change: code that today relies on the extra byte being swallowed would see it again. `sungetc()` can also fail if the streambuf has no putback room, which needs a decision (currently `sbumpc()` is used precisely to avoid depending on `std::istream` state).
2. **Qualify the documentation.** Keep the implementation as is and state in [`operator>>`'s notes](https://json.nlohmann.me/api/operator_gtgt/#notes) and in [features/parsing](https://json.nlohmann.me/features/parsing/) that concatenated values must be separated by whitespace when the preceding value is a number. Cheapest, no behaviour change, but leaves a sharp edge.
3. **Both**: qualify the docs now (so #5326 can land accurately) and treat the adapter fix as a candidate for the next major release.

I lean towards 3, but since option 1 changes observable behaviour of `operator>>`, marking this for discussion.

### Which version of the library did you use?

develop (dd24e2dff)

---

*This issue was written by Claude Code.*

nlohmann:
Fixes #5340. **Draft:** this changes observable behavior of `operator>>` and `sax_parse(strict = false)`, so it is a candidate for the next major release rather than a patch release. The version numbers in the docs assume 4.0.0 and need adjusting if that changes.

## The bug

A number is the only JSON value whose end can only be detected by looking at the character that follows it. `lexer::scan_number()` reads that character and calls `lexer::unget()`, but `unget()` is deliberately *simulated* — it rewinds only the lexer's own bookkeeping (`chars_read_total`, `chars_read_current_line`, `token_string`), not the input. `input_stream_adapter::get_character()` consumed via `sbumpc()`, so the terminating character stayed consumed:

```cpp
std::istringstream input("1true");
json j1;
input >> j1;   // j1 == 1, but the stream now starts at "rue"
```

This is invisible to `parse()`/`accept()`, which require the input to end after the value anyway. It is only observable where the caller keeps using the input: `operator>>` and `sax_parse(strict = false)`.

## The fix

Following [@gregmarr's suggestion](https://github.com/nlohmann/json/pull/5344#issuecomment-5160531385), the adapter now **peeks instead of consuming**, rather than consuming and trying to put the character back afterwards.

- `input_stream_adapter::get_character()` returns `sb->sgetc()` and remembers that a character is pending; the *next* call steps over it with `sbumpc()` before peeking again. `get_elements()` and the destructor commit the pending character the same way, so every other user of the adapter — and the stream the caller gets back — sees the position it saw before.
- `input_stream_adapter::release_lookahead()` drops the pending character instead of committing it, leaving it in the stream. The adapter advertises this with `supports_lookahead`, detected with the same `is_detected` tag-dispatch idiom already used for `supports_seek`; other adapters compile the call away to a no-op.
- `lexer::release_lookahead()` forwards a pending *simulated* unget to the adapter and clears `next_unget`, so the character is read from the input again rather than replayed from `current`. A pending unget of EOF needs no special case: reaching EOF leaves no lookahead to release.
- `parser` calls it on the three non-strict paths (both `parse()` branches and `sax_parse`), folded into the existing `if (strict && …)` checks.

The invariant that makes this safe: `get()` clears `next_unget` whenever it consumes it, so a pending unget always refers to the most recent adapter read, and there is never more than one.

### Why not `sungetc()`

The first version of this PR consumed with `sbumpc()` and put the character back with `sungetc()`. That is best-effort: `sungetc()` fails when the streambuf has no putback position, and the character stays consumed. Peeking cannot fail — the character is never consumed in the first place, so no putback position is required. The `no_putback_streambuf` test (a `streambuf` with no get area whose `pbackfail()` always fails) covers exactly that case and now passes.

## Verification

| input | before | after |
| --- | --- | --- |
| `1true` | `rue` ❌ | `true` ✅ |
| `1 true` | `true` ❌ | ` true` ✅ |
| `1[2]` | `2]` ❌ | `[2]` ✅ |
| `1{}` | `}` ❌ | `{}` ✅ |
| `1"a"` | `a"` ❌ | `"a"` ✅ |
| `-0.5e3x` | `` ❌ | `x` ✅ |

Note `1 true` was wrong before too — the issue's table lists it as fine because the swallowed byte happened to be whitespace, but the position was still off by one.

- New tests in `tests/src/unit-deserialization.cpp` (`stream position after extraction (#5340)`): number terminators for every following value type, self-delimiting values, a number at end of input, repeated extraction of `1true[2]3"x"{"a":4}5`, `sax_parse(strict = false)`, strict parsing still rejecting trailing data, and the `streambuf` that cannot put back. **8 assertions in this section fail against `develop`** (and the repeated-extraction subcase throws).
- `unit-deserialization`, `unit-class_lexer`, `unit-class_parser`, `unit-class_parser_diagnostic_positions`, `unit-user_defined_input`, `unit-regression2`, and `unit-disabled_exceptions` pass against both `include/` and `single_include/` — 20,411 assertions each, 0 failures.
- Parse error messages and reported positions were diffed across 13 malformed inputs (string and stream paths) and are **byte-identical** to `develop`.

## Public API

**No breaking changes to the public API surface**: no signature, type, or name changes; nothing added to or removed from the public interface. `lexer::release_lookahead()`, `input_stream_adapter::release_lookahead()`, and `input_stream_adapter::supports_lookahead` are new members in `detail::`.

**There is a behavior change**, which is why this is a draft targeting the next major release:

- `operator>>` and `sax_parse(strict = false)` on a `std::istream` now leave the stream one byte earlier when the parsed value was a number. Code that relied on the terminating byte being swallowed will observe it again.
- Anyone who worked around the bug by inserting whitespace separators is unaffected — whitespace is still skipped on the next extraction.
- `parse()`, `accept()`, and all non-stream inputs (strings, iterators, containers, `FILE*`) are unchanged.

## Follow-up

`docs/mkdocs/docs/api/operator_gtgt.md` and `sax_parse.md` say "version 4.0.0"; these need updating if the change lands elsewhere. The admonition added in #5343 is replaced here by the version note.

---

- [x] An [existing issue](https://github.com/nlohmann/json/issues/5340) is referenced.
- [x] All new code is covered by tests (new section in `unit-deserialization.cpp`; verified to fail without the fix).
- [x] The [documentation](https://json.nlohmann.me) is updated.
- [x] `make amalgamate` was run; `single_include` matches `include` (+132/−5, no unrelated reformatting).

---

*This pull request was written by Claude Code.*


gregmarr:
> When sungetc() fails
sungetc() can fail if the streambuf has no putback position. Then the character stays consumed — which is exactly today's behavior, so a best-effort unget is never worse than the status quo. restore_pending_unget() reports this rather than asserting. Covered by a test using a streambuf whose pbackfail always fails.

Have you looked at using `peek()` to get the next character and then advancing the stream if appropriate, rather than getting the next character and having to do `unget`?  I haven't looked to see if it's feasible, just curious if you already had done this.

gregmarr:
@nlohmann Since you're looking at a lot of breaking things under flags, would this be a candidate?  Also, I had a question above about using peek() rather than get() until you are sure you want to use the character.

nlohmann:
Thanks and sorry, I never answered the `peek()` question in the thread. c021a09b0 switched to exactly that: `input_stream_adapter::get_character()` now peeks with `sgetc()` and only commits the character with `sbumpc()` on the next read, and `release_lookahead()` leaves it in the stream. `sungetc()` is gone, so there's no longer a putback position that can fail; the `no_putback_streambuf` test covers that.

I am currently unsure if this change should be implemented - even with a flag. Few users will notice. The limitation is already documented by the admonition from #5343.

What do you think? Is adding the some 130 lines of code + another flag is worth it?

gregmarr:
I see now that the original message was edited to take that suggestion.  I was on vacation at that time, so I didn't notice the changed behavior.

So in that case, is there still a failure case if this is implemented? 

> operator>> and sax_parse(strict = false) on a std::istream now leave the stream one byte earlier when the parsed value was a number. Code that relied on the terminating byte being swallowed will observe it again.

Is this still accurate?  If so, what is the case that would work before that would fail now?

nlohmann:
Yes, it's still accurate. The peek changes *how* the character is kept, not *where* the stream ends up. Two cases that behave differently (checked against develop and this branch):

- `while (is >> j)` over `1,2,3`: today it reads 1, 2, 3 because the comma after each number is swallowed. With this PR the second extraction throws at the `,`.
- `is >> j; std::getline(is, line);` over `42\nfoo`: today `line == "foo"`; with this PR `line == ""`.

In both cases the new behavior is what you already get for every non-number value (`"a","b"` fails at the comma today, and `{}\nfoo` gives an empty `getline`) and what `is >> int` does. So the old behavior only "worked" by accident, and only for numbers. It's still something a user could observe, though, which is why I'm hesitant to ship it outside a major release.

gregmarr:
Okay, I think `1,2,3` looks enough like "this should work", even though it shouldn't, that it needs a major bump or a flag.

nlohmann:
So what do you think? Ship it behind a macro? Or close it? (Postponing to 4.0.0 is a bad idea, because we should have macros for everything so clients can try the breaking changes before)

gregmarr:
If you want it to be testable before 4.0, then a macro.

PR #5361, opened 2026-08-05
Fix start_pos() for strings containing escape sequences

nlohmann:
## Problem

`start_pos()` reports the wrong offset for any string value that contains an escape sequence.

The position was derived by subtracting the *parsed* value's length from the end position:

```cpp
v.start_position = v.end_position - v.m_data.m_value.string->size() - 2;
```

Escape sequences make the source token longer than the value it parses to, so the reported start lands inside the string — one byte off per escape sequence:

```cpp
json j = json::parse(R"({"a":"\n\n\n\n\n\n"})");
const json& v = j.at("a");
// start_pos() == 11, so the reported range covers   n\n\n\n"
// expected     ==  5, covering                     "\n\n\n\n\n\n"
```

This contradicts the [documented behavior](https://json.nlohmann.me/api/basic_json/start_pos/) of `start_pos()` ("string → position of the opening `\"`"), and it also corrupts the `(bytes N-M)` part of `JSON_DIAGNOSTICS` exception messages for such values.

Strings containing multi-byte UTF-8 but no escapes were unaffected (`"é"`, `"🌞"` report correctly), which is likely why this went unnoticed — the existing tests cover escaped strings for their *values* but never checked their positions.

## Fix

Record the offset of the token in the lexer when it starts scanning it, and use that instead of reconstructing it from the parsed value. This is exact by construction rather than inferred.

`true`, `false`, `null` and numbers already reported correct positions (their token length is not affected by escaping) and are left unchanged.

The new lexer member and accessor are compiled only when `JSON_DIAGNOSTIC_POSITIONS` is enabled. That macro is already part of the ABI tag (`_dp`), so builds with the default `JSON_DIAGNOSTIC_POSITIONS=0` are byte-for-byte unaffected.

## Public API

**No breaking changes.** No public declaration changes. `basic_json::start_pos()` keeps its signature and simply returns the documented value in a case where it previously did not. The only new symbol is `detail::lexer::get_token_start_position()`, which is internal.

Behavior does change for code reading `start_pos()` on escaped strings — but only from an incorrect value to the documented one.

## Testing

Added a section to `unit-diagnostic-positions.cpp` covering escaped strings, quotes, backslashes, multi-byte UTF-8, and a root-level escaped string. **These assertions fail on `develop` (5 failures) and pass with this change.**

Verified with a property-based check that for every value in a parsed document, `src.substr(start_pos(), end_pos() - start_pos())` re-parses to that value. Over 300k randomly generated documents this reported 13 failures before the change — all of them escaped strings — and 0 after.

Existing suites show no regressions: `unit-class_parser_diagnostic_positions` (9234 assertions), `unit-diagnostic-positions`, `unit-diagnostic-positions-only` all pass. Confirmed correct across all input adapters (string, `istream`, iterator pair) and both the plain and callback DOM parsers. Compiles clean at C++11/14/17/20 with `-Wall -Wextra -Werror`. Amalgamation regenerated; `make check-amalgamation` passes.

---

— written by Claude Code on behalf of @nlohmann

### DIFF
--- a/include/nlohmann/detail/input/lexer.hpp
+++ b/include/nlohmann/detail/input/lexer.hpp
@@ -1357,6 +1357,11 @@ scan_number_done:
         token_buffer.clear();
         decimal_point_position = std::string::npos;
 
+#if JSON_DIAGNOSTIC_POSITIONS
+        // the first character of the token has already been read, hence the -1
+        token_start_position = position.chars_read_total - 1;
+#endif
+
         note_token_start(std::integral_constant<bool, lazy_token_string> {});
     }
 
@@ -1518,6 +1523,15 @@ scan_number_done:
     {
         return position;
     }
+
+#if JSON_DIAGNOSTIC_POSITIONS
+    /// return the offset of the first character of the last read token; unlike
+    /// the token's parsed value, this accounts for escape sequences
+    constexpr std::size_t get_token_start_position() const noexcept
+    {
+        return token_start_position;
+    }
+#endif
 
     /// seekable adapter: rebuild the last read token from the input on demand
     const std::vector<char_type>& collect_token_chars(std::vector<char_type>& out, std::true_type /*lazy*/) const
@@ -1718,6 +1732,12 @@ scan_number_done:
     /// start offset of the current token within the input, used to reconstruct
     /// the last read token on error for seekable adapters (see collect_token_chars)
     std::size_t token_string_start = 0;
+
+#if JSON_DIAGNOSTIC_POSITIONS
+    /// start offset of the current token within the input, used to report
+    /// diagnostic positions (see reset())
+    std::size_t token_start_position = 0;
+#endif
 
     /// buffer for variable-length tokens (numbers, strings)
     string_t token_buffer {};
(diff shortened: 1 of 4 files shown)


include/nlohmann/detail/input/lexer.hpp
```diff
@@ -1357,6 +1357,11 @@ class lexer : public lexer_base<BasicJsonType>
         token_buffer.clear();
         decimal_point_position = std::string::npos;
 
+#if JSON_DIAGNOSTIC_POSITIONS
+        // the first character of the token has already been read, hence the -1
+        token_start_position = position.chars_read_total - 1;
```
gregmarr, line 1362: How is this different than `token_string_start` recorded by `note_token_start()`?

nlohmann: note_token_start() records an index into the input adapter’s consumed stream, which only seekable adapters can supply — for istream and friends there is no such counter, and diagnostic positions have to work there too. token_start_position is the lexer’s own chars_read_total, the same counter end_position comes from.

PR #5367, opened 2026-08-07
Ignore CBOR tags 0-5 and 21-23 (head bytes 0xC0-0xC5, 0xD5-0xD7)

### LINKED ISSUE #5315: CBOR: cbor_tag_handler_t::ignore does not ignore tags 0-5 and 21-23 (head bytes 0xC0-0xC5, 0xD5-0xD7)

MsfPablo:
Fixes #5315.

Every byte in `0xC0`–`0xDB` is major type 6 (tag), but the `switch` in `parse_cbor_internal()` enumerated only `0xC6`–`0xD4` and `0xD8`–`0xDB`. Head bytes `0xC0`–`0xC5` and `0xD5`–`0xD7` fell through to `default:` and were reported as `parse_error.112` in *every* handler mode, including `ignore` and `store`.

This adds the missing head bytes to the tagged-item case list, so they are handled exactly like tags 6–20 already are:

- `cbor_tag_handler_t::error` (the default) still throws — unchanged.
- `ignore` skips the tag head and parses the enclosed data item.
- `store` parses the enclosed item; as with tags 6–20, no subtype is recorded (only the `0xD8`–`0xDB` forms carry one today).

No tag is interpreted: a text string tagged 0 stays a string, `0xC2` over a byte string stays a byte string.

### Changes

- `include/nlohmann/detail/input/binary_reader.hpp` (and the amalgamated header): add `case 0xC0`–`0xC5` and `case 0xD5`–`0xD7` to the tagged-item block.
- `tests/src/unit-cbor.cpp`: the "Tagged values" section that swept `0xC6..0xD4` now sweeps `0xC0..0xD7`, asserting throw in `error` mode and the original value in `ignore`/`store` mode for each head byte.
- `docs/.../cbor.md`: the "Incomplete mapping" list no longer claims tags 0–5 and 21–23 are unsupported; the tagged-items note now states the range and that tags are skipped, not interpreted.

The "all CBOR first bytes" test needed no change — it parses in the default `error` mode, where these bytes still throw `parse_error.112`.

### Verification

`test-cbor_cpp11` passes (the only failures in my local run are the test cases requiring the downloaded test-data fixtures, which I did not fetch). The reproduction from the issue now works:

```
{"t":"2026-07-27T00:00:00Z"}
```

and a sweep of `0xC0`–`0xD7` confirms every head byte throws in `error` mode and yields the enclosed value in `ignore` and `store` mode.

### DIFF
--- a/docs/mkdocs/docs/features/binary_formats/cbor.md
+++ b/docs/mkdocs/docs/features/binary_formats/cbor.md
@@ -160,13 +160,10 @@ The library maps CBOR types to JSON value types as follows:
 
     The mapping is **incomplete** in the sense that not all CBOR types can be converted to a JSON value. The following CBOR types are not supported and will yield parse errors:
 
-     - date/time (0xC0..0xC1)
-     - bignum (0xC2..0xC3)
-     - decimal fraction (0xC4)
-     - bigfloat (0xC5)
-     - expected conversions (0xD5..0xD7)
      - simple values (0xE0..0xF3, 0xF8)
      - undefined (0xF7)
+
+    Tagged items (0xC0..0xDB) are not interpreted either; see the note on tagged items below.
 
 !!! warning "Negative integer overflow"
 
@@ -181,7 +178,7 @@ The library maps CBOR types to JSON value types as follows:
 
 !!! warning "Tagged items"
 
-    Tagged items will throw a parse error by default. They can be ignored by passing `cbor_tag_handler_t::ignore` to function `from_cbor`. They can be stored by passing `cbor_tag_handler_t::store` to function `from_cbor`.
+    Tagged items (0xC0..0xDB) will throw a parse error by default. They can be ignored by passing `cbor_tag_handler_t::ignore` to function `from_cbor`, in which case the tag is skipped and the enclosed data item is parsed on its own. They can be stored by passing `cbor_tag_handler_t::store` to function `from_cbor`. Note that no tag is ever interpreted: for instance, a text string tagged with tag 0 (date/time) stays a string.
 
 ??? example
 
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -773,7 +773,13 @@ class binary_reader
             case 0xBF: // map (indefinite length)
                 return get_cbor_object(detail::unknown_size(), tag_handler);
 
-            case 0xC6: // tagged item
+            case 0xC0: // tagged item
+            case 0xC1:
+            case 0xC2:
+            case 0xC3:
+            case 0xC4:
+            case 0xC5:
+            case 0xC6:
             case 0xC7:
             case 0xC8:
             case 0xC9:
@@ -788,6 +794,9 @@ class binary_reader
             case 0xD2:
             case 0xD3:
             case 0xD4:
+            case 0xD5:
+            case 0xD6:
+            case 0xD7:
             case 0xD8: // tagged item (1 byte follows)
             case 0xD9: // tagged item (2 bytes follow)
             case 0xDA: // tagged item (4 bytes follow)
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -11395,7 +11395,13 @@ class binary_reader
             case 0xBF: // map (indefinite length)
                 return get_cbor_object(detail::unknown_size(), tag_handler);
 
-            case 0xC6: // tagged item
+            case 0xC0: // tagged item
+            case 0xC1:
+            case 0xC2:
+            case 0xC3:
+            case 0xC4:
+            case 0xC5:
+            case 0xC6:
             case 0xC7:
             case 0xC8:
             case 0xC9:
@@ -11410,6 +11416,9 @@ class binary_reader
             case 0xD2:
             case 0xD3:
             case 0xD4:
+            case 0xD5:
+            case 0xD6:
+            case 0xD7:
             case 0xD8: // tagged item (1 byte follows)
             case 0xD9: // tagged item (2 bytes follow)
             case 0xDA: // tagged item (4 bytes follow)
--- a/tests/src/unit-cbor.cpp
+++ b/tests/src/unit-cbor.cpp
@@ -2565,11 +2565,13 @@ TEST_CASE("Tagged values")
     const json j = "s";
     auto v = json::to_cbor(j);
 
-    SECTION("0xC6..0xD4")
+    SECTION("0xC0..0xD7")
     {
         for (const auto b : std::vector<std::uint8_t>
     {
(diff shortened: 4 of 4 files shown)


nlohmann:
Apologies for the AI-assisted review here — I don't have time for a deeper pass on this one right now, so this is a quick comparison rather than a full re-review.

This duplicates #5331, which has the identical code fix and was opened first. Your docs update here is actually better — it removes the stale "Incomplete mapping" bullets that #5331 left contradicting the paragraph right below it — but this PR is currently missing a `Signed-off-by` trailer, so the DCO check is failing and will block merge as-is.

Neither PR yet adds the test coverage requested on #5331 (a `store`-mode test with a binary payload, not just a string, to confirm 0xC0-0xC5/0xD5-0xD7 take the same "unwrap, don't treat as subtype" path as the existing tag range).

Given #5331 is further along (green CI, already reviewed), we'll likely take that one as the base and fold in the docs fix from here. Thanks for the contribution either way — closing this in favor of #5331 once that's updated, unless you'd rather rebase your docs fix onto it yourself.

nlohmann (changes requested):
Please fix the DCO requirements and sign off your commits.

MsfPablo:
Signed off, so DCO should be green now.

I also added the `store`-mode test you mentioned was missing on both PRs: a binary payload (`json::binary({0xCA, 0xFE})`) prefixed with each of 0xC0-0xD7, checking the result round-trips *and* that `has_subtype()` is false — i.e. the head byte carries the tag value itself, there's no subtype byte to read, so store unwraps rather than treating it as a subtype. It asserts the 0xD8 contrast case too (next byte does become the subtype), so the distinction is pinned rather than implied.

Happy either way on which PR is the base. If you'd rather take #5331, say the word and I'll strip this down to just the docs change and the test so it applies cleanly on top of it — no need to spend review time reconciling two full diffs.

(One note in case it saves someone confusion: `unit-cbor.cpp`'s "examples from RFC 8949 Appendix A" case fails for me on a clean `develop` checkout too, so that one isn't from this branch.)

MsfPablo:
Closing — contributing focus has shifted elsewhere. Feel free to pick this up if useful.

PR #5368, opened 2026-08-07
Install CMake package files from the Meson build

### LINKED ISSUE #3885: meson build does not install nlohmann_json*.cmake files
### Description

Issue identified that meson build does not install nlohmann_json*.cmake files into /usr/share/cmake/nlohmann_json, so packages expecting these .cmake files are unable to build. 

Ref: https://github.com/LibreELEC/LibreELEC.tv/pull/7250

we have reverted back to cmake builds for the time being. ** Our package build system prefers meson over cmake

### Reproduction steps

Complete a meson build

Use a cmake build

### Minimal code example

```Shell
Na
```

### Error messages

```Shell
Na
```

LibreELEC :master / gcc 12.2

### Library version

3.11.2

### Validation

MsfPablo:
Fixes #3885.

The Meson build installed only the headers and a pkg-config file, so consumers using `find_package(nlohmann_json)` could not build against a Meson-installed copy.

This generates and installs `nlohmann_jsonConfig.cmake`, `nlohmann_jsonConfigVersion.cmake` and `nlohmann_jsonTargets.cmake` into `<datadir>/cmake/nlohmann_json`, the same location the CMake build uses.

- `cmake/config.cmake.in` and `cmake/nlohmann_jsonConfigVersion.cmake.in` are reused as-is from the CMake build.
- The targets file is generated by CMake's `install(EXPORT ...)`, so there is nothing to reuse; `cmake/nlohmann_jsonTargets.cmake.in` is added for the Meson build. It computes `_IMPORT_PREFIX` relative to its own location so the installed package stays relocatable.
- Only runs when not built as a subproject, alongside the existing header/pkg-config install rules.

Verified locally: `meson install --destdir` places the three files under `share/cmake/nlohmann_json`, and a separate CMake project doing `find_package(nlohmann_json 3.12.0 REQUIRED)` + `target_link_libraries(t PRIVATE nlohmann_json::nlohmann_json)` configures, builds and runs against it.

### DIFF
--- a/cmake/nlohmann_jsonTargets.cmake.in
+++ b/cmake/nlohmann_jsonTargets.cmake.in
@@ -1,0 +1,18 @@
+# Hand-written equivalent of the target export file that the CMake build
+# generates via install(EXPORT ...). Used by the Meson build, which cannot
+# reuse CMake's exporter.
+
+# Compute the installation prefix relative to this file.
+get_filename_component(_IMPORT_PREFIX "${CMAKE_CURRENT_LIST_FILE}" PATH)
+@NLOHMANN_JSON_IMPORT_PREFIX_STEPS@
+if(_IMPORT_PREFIX STREQUAL "/")
+    set(_IMPORT_PREFIX "")
+endif()
+
+add_library(@NLOHMANN_JSON_EXPORTED_TARGET_NAME@ INTERFACE IMPORTED)
+set_target_properties(@NLOHMANN_JSON_EXPORTED_TARGET_NAME@ PROPERTIES
+    INTERFACE_COMPILE_FEATURES "cxx_std_11"
+    INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/@NLOHMANN_JSON_INCLUDE_INSTALL_DIR@"
+)
+
+set(_IMPORT_PREFIX)
--- a/meson.build
+++ b/meson.build
@@ -21,4 +21,46 @@ pkgc.generate(name: 'nlohmann_json',
     version: meson.project_version(),
     description: 'JSON for Modern C++'
 )
+
+# CMake package files, so that find_package(nlohmann_json) works after a Meson
+# install just like it does after a CMake install.
+cmake_config_dir = get_option('datadir') / 'cmake' / meson.project_name()
+
+# Number of directory levels between the config file and the install prefix,
+# used to make the generated target file relocatable.
+import_prefix_steps = []
+foreach component : cmake_config_dir.split('/')
+    import_prefix_steps += 'get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)'
+endforeach
+
+version_parts = meson.project_version().split('.')
+
+cmake_conf = configuration_data()
+cmake_conf.set('PROJECT_NAME', meson.project_name())
+cmake_conf.set('PROJECT_VERSION', meson.project_version())
+cmake_conf.set('PROJECT_VERSION_MAJOR', version_parts[0])
+cmake_conf.set('NLOHMANN_JSON_TARGET_NAME', 'nlohmann_json')
+cmake_conf.set('NLOHMANN_JSON_EXPORTED_TARGET_NAME', 'nlohmann_json::nlohmann_json')
+cmake_conf.set('NLOHMANN_JSON_TARGETS_EXPORT_NAME', meson.project_name() + 'Targets')
+cmake_conf.set('NLOHMANN_JSON_INCLUDE_INSTALL_DIR', get_option('includedir'))
+cmake_conf.set('NLOHMANN_JSON_IMPORT_PREFIX_STEPS', '\n'.join(import_prefix_steps))
+
+configure_file(
+    input: 'cmake/config.cmake.in',
+    output: meson.project_name() + 'Config.cmake',
+    configuration: cmake_conf,
+    install_dir: cmake_config_dir,
+)
+configure_file(
+    input: 'cmake/nlohmann_jsonConfigVersion.cmake.in',
+    output: meson.project_name() + 'ConfigVersion.cmake',
+    configuration: cmake_conf,
+    install_dir: cmake_config_dir,
+)
+configure_file(
+    input: 'cmake/nlohmann_jsonTargets.cmake.in',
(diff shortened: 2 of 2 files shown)


nlohmann (commented):
> **A note up front, and apologies for it:** this review was written by Claude Code, not by me personally. I don't have the time to give this PR the full review it deserves right now, and I'd rather send you concrete findings than leave the PR sitting untouched. Please treat everything below as input to check rather than as a maintainer verdict — push back on anything that looks wrong, and I'll take a proper look when I can.

Thanks for tackling this — #3885 has been open a long time, and the approach here is the right one. Reusing `cmake/config.cmake.in` and `cmake/nlohmann_jsonConfigVersion.cmake.in` unchanged is exactly what I'd want, and the hand-written targets file mirrors what CMake's exporter emits closely enough (same `get_filename_component` walk, same `set(_IMPORT_PREFIX)` cleanup at the end).

I verified your happy path locally and it works as described: a default `meson setup` + `meson install --destdir` puts the three files under `share/cmake/nlohmann_json`, and a separate project doing `find_package(nlohmann_json 3.12.0 REQUIRED)` + `target_link_libraries(t PRIVATE nlohmann_json::nlohmann_json)` configures, builds and runs against it.

Two things I'd like addressed before this goes in.

## 1. This breaks `meson setup` on the released `include.zip`

`Makefile:208` packs `meson.build` into `include.zip`, but only ships the `include/` tree, the two amalgamated headers, the Bazel files and `LICENSE.MIT` — no `cmake/*.in`. That archive is a working standalone Meson source tree today; with this change it fails at configure time:

```
meson.build:48:0: ERROR: File cmake/config.cmake.in does not exist.
```

(I reproduced the zip line exactly and ran `meson setup` on the result.)

Subproject/wrap consumers are fine, since everything here sits inside `if not meson.is_subproject()`. But please extend the zip line so the templates ship alongside `meson.build`:

```make
zip -9 --recurse-paths -X include.zip $(SRCS) $(AMALGAMATED_FILE) $(AMALGAMATED_FWD_FILE) BUILD.bazel MODULE.bazel meson.build cmake/config.cmake.in cmake/nlohmann_jsonConfigVersion.cmake.in cmake/nlohmann_jsonTargets.cmake.in LICENSE.MIT
```

Worth noting the mirror-image gap: `json.tar.xz` already rsyncs `cmake/*.in` but doesn't ship `meson.build`, so after this change neither release artifact would work as a standalone Meson tree. Fixing `include.zip` is enough; I just want it to be a deliberate choice.

## 2. Absolute `datadir` / `includedir` silently produce a broken package

`meson.build:32` derives the number of prefix-walk steps from `cmake_config_dir.split('/')`. An absolute path contributes a leading empty component, so the walk strips one level too many. Separately, `meson.build:45` interpolates `includedir` into `"${_IMPORT_PREFIX}/@NLOHMANN_JSON_INCLUDE_INSTALL_DIR@"` without checking whether it is already absolute.

Meson normalises absolute dir options that fall *under* the prefix, so the usual distro invocation (`--prefix=/usr -Ddatadir=/usr/share`) is fine — I checked. Paths outside the prefix are not normalised, and both fail:

| Configuration | Generated `INTERFACE_INCLUDE_DIRECTORIES` | Headers actually installed to |
|---|---|---|
| `--prefix=/opt/json -Ddatadir=/usr/share` | resolves one level *above* the prefix | `<prefix>/opt/json/include` |
| `--prefix=/usr/local -Dincludedir=/opt/hdr` | `"${_IMPORT_PREFIX}//opt/hdr"` | `/opt/hdr` |

Both installs succeed without a warning and produce a config package pointing at a directory that does not exist. In the first case I confirmed via `get_target_property` that `find_package` resolves the include directory one level above the real prefix.

Relative-path walking can't relocate a config dir that lives on a different branch of the filesystem than the prefix, so I'd rather fail loudly than emit something subtly wrong:

```meson
if cmake_config_dir.startswith('/') or get_option('includedir').startswith('/')
    error('Meson-generated CMake package files require datadir and includedir to be relative to the prefix')
endif
```

Emitting an absolute include directory in that case (giving up relocatability only for that configuration) would be fine too — your call.

## Smaller points

- **Please add a CI job.** There is currently no Meson coverage anywhere in `.github/workflows`, and `INTERFACE_COMPILE_FEATURES "cxx_std_11"` in the new template duplicates `CMakeLists.txt:132` while `INTERFACE_INCLUDE_DIRECTORIES` duplicates `CMakeLists.txt:150`. Nothing will catch it when one side changes. `tests/cmake_import/project/` is already a minimal `find_package` consumer and can be pointed straight at the Meson install: `meson setup` → `meson install --destdir` → `cmake -Dnlohmann_json_DIR=<destdir>/.../share/cmake/nlohmann_json tests/cmake_import/project` → build. This also guards item 2 above if you add a case with a non-default prefix.

- **Declare `meson_version`.** The `/` path-join operator and `str.join` both need Meson ≥ 0.49, and the previous `meson.build` used nothing newer than the 0.4x baseline. Since `project()` declares no `meson_version:`, older Meson fails with an opaque parse error. Please add `meson_version: '>= 0.49.0'`.

- `meson.build:43`: `NLOHMANN_JSON_EXPORTED_TARGET_NAME` hard-codes `'nlohmann_json::nlohmann_json'` while `PROJECT_NAME` and `NLOHMANN_JSON_TARGET_NAME` are set independently two lines above. Deriving it (`meson.project_name() + '::nlohmann_json'`) keeps the three from drifting apart.

- The template header calls itself an "equivalent" of the CMake export, but it deliberately omits two things: the `cmake_policy(PUSH)`/`POP` wrapper and double-inclusion guard (harmless — `config.cmake.in` already guards with `if(NOT TARGET ...)`), and the MSVC natvis entry that `CMakeLists.txt:157-162` adds to `INTERFACE_SOURCES` (correct, since the Meson build doesn't install the natvis at all). Could you note both in the comment so they read as intentional?

- Could you add the usual note to the PR description on whether the public API changes? For the record the answer is no — this is build-system only, no public C++ API is touched.

Nothing here looks structural — the design seems sound, and items 1 and 2 are the ones I'd want resolved before this is considered ready. Thanks again for the contribution, and sorry again for the automated review.

MsfPablo:
Closing — contributing focus has shifted elsewhere. Feel free to pick this up if useful.

PR #4560, opened 2024-12-18
Clean up and document project files

nlohmann:
- update GitHub community documents
  - update code of conduct
  - update contribution guidelines
  - update pull request template
  - update security policy
- document non-code files in `FILES.md`
- add community section to documentation
  - governance (new)
  - code of conduct (link)
  - contribution guidelines (link)
  - security policy (link)
  - quality assurance (new)
  - architecture (new, draft)
- add DCO requirement
- update GitHub actions
  - migrate stalebot to GitHub action
  - remove deprecated `.lgtm.yml` file
  - add CI steps for C++26
  - add Xcode 16.2 (macOS) and Clang 3.4 (Ubuntu) action
  - simplify Ubuntu workflow to use `ubuntu-latest` more often and reduce dependencies to more exotic containers
- overwork documentation
  - remove duplicate documentation files
  - fix index file for mkdocs after merging #4517 
  - update customers
  - update docset table of contents
  - overwork documentation of all serialization macros (cross references, decision table, removed redundancies)
  - fix duplicate keys in navigation (fixes #4564)
  - fix broken links
  - add documentation for `start_pos` and `end_pos`
  - add glossary
  - serve assets (fonts, CSS) locally
- overwork README
  - move package manager section to documentation
  - shortened serialization macro section (fixes #4260)
  - compilers use in the CI to documentation
  - add quick reference to README
- overwork package manager documentation
  - add Swift Package Manager and xmake
  - add example for Conda
  - add overview with logos
  - annotate Buckaroo, CocoaPods, and wsjcpp as outdated
- cleanup
  - remove `.pre-commit-config.yml` file
  - pull update for `tests/thirdparty/imapdl/filterbr.py`
  - move compiler flags to separate CMake files

PR #5376, opened 2026-08-11
Don't overwrite BUILD.bazel while making a release

jwnimmer-tri:
Instead, add automation to 'develop' to ensure that it's always kept up to date.

Add missing header file 'logic.hpp'.

Re-apply e509007d "Remove alwayslink=True Bazel flag", which was accidentally reverted previously by the broken update script.

### DIFF
--- a/.github/workflows/check_bazel.yml
+++ b/.github/workflows/check_bazel.yml
@@ -1,0 +1,31 @@
+name: "Check BUILD.bazel updated"
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@b09bb98e06d4d774595224525879c09bc6e98c40 # v2.20.1
+        with:
+          egress-policy: audit
+
+      - name: Checkout pull request
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
+
+      - name: Reset scripts to use the develop branch
+        run: |
+          git fetch origin develop
+          git checkout origin/develop -- cmake/scripts
+
+      - name: Check whether BUILD.bazel has been updated
+        run: |
+          cmake -DCHECK=ON -P cmake/scripts/gen_bazel_build_file.cmake
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -17,6 +17,7 @@ license(
 
 cc_library(
     name = "json",
+    # BEGIN GENERATED HDRS LIST
     hdrs = [
         "include/nlohmann/adl_serializer.hpp",
         "include/nlohmann/byte_container_with_subtype.hpp",
@@ -49,6 +50,7 @@ cc_library(
         "include/nlohmann/detail/meta/detected.hpp",
         "include/nlohmann/detail/meta/identity_tag.hpp",
         "include/nlohmann/detail/meta/is_sax.hpp",
+        "include/nlohmann/detail/meta/logic.hpp",
         "include/nlohmann/detail/meta/std_fs.hpp",
         "include/nlohmann/detail/meta/type_traits.hpp",
         "include/nlohmann/detail/meta/void_t.hpp",
@@ -65,9 +67,9 @@ cc_library(
         "include/nlohmann/thirdparty/hedley/hedley.hpp",
         "include/nlohmann/thirdparty/hedley/hedley_undef.hpp",
     ],
+    # END GENERATED HDRS LIST
     includes = ["include"],
     visibility = ["//visibility:public"],
-    alwayslink = True,
 )
 
 cc_library(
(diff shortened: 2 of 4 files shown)


jwnimmer-tri:
@Vertexwahn could you take a look?

.github/workflows/check_bazel.yml
```diff
@@ -0,0 +1,39 @@
...
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    env:
+      MAIN_DIR: ${{ github.workspace }}/main
```
Vertexwahn: not clear to me where MAIN_DIR and TOOL_DIR is used...

Vertexwahn (approved):

Vertexwahn:
@jwnimmer-tri LGTM - only the env variables `MAIN_DIR` and `TOOL_DIR` look unused to me...

jwnimmer-tri:
Thanks!  Fixed now.

nlohmann (changes requested):
## Review

**`check` job is currently failing because of a path bug in the new workflow.**

In [`.github/workflows/check_bazel.yml`](https://github.com/nlohmann/json/pull/5376/files#diff-*), the "Checkout scripts" step checks out the whole `develop` repo into `path: cmake/scripts`, so the script actually ends up at `cmake/scripts/cmake/scripts/gen_bazel_build_file.cmake`. The run step instead does:

```yaml
cd ${{ github.workspace }}
cmake -DCHECK=ON -P cmake/scripts/gen_bazel_build_file.cmake
```

which looks one level too shallow. This isn't hypothetical — it's exactly why this PR's own `check` job fails right now:
```
CMake Error: Not a file: cmake/scripts/gen_bazel_build_file.cmake
```
([run log](https://github.com/nlohmann/json/actions/runs/31600625577))

This matters more than a typical CI nit here: the Makefile change in this same PR removes `include.zip`'s implicit dependency on regenerating `BUILD.bazel` (the whole point of the PR), so `check_bazel.yml` is meant to be the *only* remaining safeguard against a stale `BUILD.bazel` shipping in a release. Right now it can't run at all.

**Even after the path is fixed, the check would validate the wrong tree.** `gen_bazel_build_file.cmake` derives `PROJECT_ROOT` from `CMAKE_CURRENT_LIST_DIR` (its own on-disk location), not from the `main` checkout. If the script that actually runs is the `develop`-checked-out copy under `cmake/scripts`, `PROJECT_ROOT` resolves inside *that* checkout, so the job would always compare `develop`'s own `BUILD.bazel` against `develop`'s own headers (always "up to date") rather than checking the PR's `main` checkout. That's a silent no-op rather than the crash seen today — worth fixing at the same time, e.g. by mirroring the pattern in `check_amalgamation.yml` (`cd $MAIN_DIR` and invoke the develop-checked-out tool via its full path).

**Minor:**
- `check_bazel.yml` pins `step-security/harden-runner` to `v2.20.0` (`bf7454d06d71f1098171f2acdf0cd4708d7b5920`), while every other workflow in the repo is already on `v2.20.1` (`b09bb98e06d4d774595224525879c09bc6e98c40`, bumped in #5375). Worth matching to avoid an immediate Dependabot follow-up.
- The new check-bazel gate does a strict text comparison against `file(GLOB_RECURSE ...)` output, whose enumeration order isn't guaranteed identical across platforms/filesystems. A contributor running `make update_bazel` locally on macOS could get headers in a different order than the Ubuntu-based CI job recomputes, and fail the hard `FATAL_ERROR` check even though the header set is correct.

Everything else looks good — the marker-based substitution in the cmake script is correct, `logic.hpp` is a real existing header that was missing from the generated list, and the `alwayslink` revert matches the stated intent.

jwnimmer-tri:
Thanks for the review.  I couldn't figure out an easy way to test it locally.

> A contributor running make update_bazel locally on macOS could get headers in a different order ...

I don't think this is true.  All of the header filenames are lowercase.

gregmarr:
> Thanks for the review. I couldn't figure out an easy way to test it locally.
> 
> > A contributor running make update_bazel locally on macOS could get headers in a different order ...
> 
> I don't think this is true. All of the header filenames are lowercase.

Stable `file(GLOB` order requires CMake 3.6, this project has 3.5 as the minimum.

jwnimmer-tri:
Ah, I was thinking only of the CI check, which has a modern CMake.  I guess I'll add an explicit sort call.

jwnimmer-tri:
> ... we encourage you to update or comment if it is still relevant.

Still relevant.  Rebased on `develop`.

nlohmann (changes requested):
Thank you for this PR, and for keeping it rebased. I owe you an apology: #5554, which I merged today, overlaps with it almost completely. It came out of an AI-driven cleanup of our backlog, and that run missed that your PR already covered the same ground. You were here first, and I should have caught it.

For the record, #5554:
- makes `cmake/scripts/gen_bazel_build_file.cmake` reproduce the whole `BUILD.bazel`, so regenerating it during a release no longer loses anything;
- sorts the header list, which settles the `file(GLOB)` ordering question from the review;
- adds the missing headers, including `logic.hpp`;
- checks in CI (as part of the amalgamation check) that `BUILD.bazel` matches the generator.

One part of your PR is still missing on `develop`: removing `alwayslink = True`. #4396 removed it in 2024, the old script put it back in #4560, and #5554 kept it because it reproduced the file as it was. Would you rebase this PR and reduce it to that change, in both the generator and `BUILD.bazel`? I'd be happy to merge that. If you'd rather not spend more time on it, just say so, and I'll make the change with credit to you.

jwnimmer-tri:
All set.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks! Will merge once the CI is green.

PR #5378, opened 2026-08-12
Guard <version> inclusion with JSON_HAS_CPP_20

chliny:
## Summary

Guard the inclusion of `<version>` with `JSON_HAS_CPP_20` and regenerate the amalgamated single-header file.

## Problem

`detail/macro_scope.hpp` currently probes for `<version>` using `__has_include` and includes it whenever the header is physically available.

However, `<version>` was introduced in C++20. A standard library installation may provide the header even when the translation unit is compiled in C++11, C++14, or C++17 mode. In that case, `__has_include(<version>)` succeeds despite the active language mode not supporting the header, potentially causing consumers to fail while including nlohmann/json.

The root cause is that header availability is checked without first checking the active C++ language standard.

## Changes

- Check `JSON_HAS_CPP_20` before probing for and including `<version>`.
- Regenerate `single_include/nlohmann/json.hpp` from the multiple-header sources using the repository's amalgamation tooling.

## Testing

The complete test suite was run on Arch Linux x86_64 with GCC 16.1.1 and libstdc++ 20260728:

```text
cmake -S . -B build
cmake --build build -j 10
ctest --test-dir build -j 10 --output-on-failure

100% tests passed, 0 tests failed out of 109
```

The configured test matrix covered C++11, C++14, C++17, C++20, and C++23.

Additional targeted compilation checks were performed with GCC 16.1.1 and Clang 22.1.8 for both the multiple-header and single-header variants:

- C++11 builds compile successfully and no longer include `<version>`.
- C++20 builds compile successfully and continue to include `<version>`.

## Regression test note

No dedicated unit test was added because this behavior cannot be meaningfully observed with a regular runtime assertion. A reliable regression test would need a separate compile-only fixture that places a deliberately incompatible fake `version` header before the standard library include paths, then builds both the multiple-header and single-header variants in a pre-C++20 language mode.

That requires additional build-system test infrastructure disproportionate to this preprocessor guard change. The targeted dependency and compilation checks described above were used instead.

- [x] No existing issue was identified for this fix.
- [ ] The code coverage remained at 100%. No dedicated regression test was added for the reason described above.
- [x] Documentation changes are not applicable because this does not change the public API.
- [x] The source code was amalgamated by running `make amalgamate`.

### DIFF
--- a/include/nlohmann/detail/macro_scope.hpp
+++ b/include/nlohmann/detail/macro_scope.hpp
@@ -58,9 +58,11 @@
     #define JSON_HAS_CPP_11
 #endif
 
-#ifdef __has_include
-    #if __has_include(<version>)
-        #include <version>
+#ifdef JSON_HAS_CPP_20
+    #ifdef __has_include
+        #if __has_include(<version>)
+            #include <version>
+        #endif
     #endif
 #endif
 
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -2433,9 +2433,11 @@ JSON_HEDLEY_DIAGNOSTIC_POP
     #define JSON_HAS_CPP_11
 #endif
 
-#ifdef __has_include
-    #if __has_include(<version>)
-        #include <version>
+#ifdef JSON_HAS_CPP_20
+    #ifdef __has_include
+        #if __has_include(<version>)
+            #include <version>
+        #endif
     #endif
 #endif
 


gregmarr:
What failure are you getting with the current code?
The entire purpose of `__has_include` and the feature test macros is to avoid having to test against full support for a specific language version.

nlohmann:
Do you have an example for this? Are there compilers out there that pass `__has_include(<version>)` but then fail to provide/compile `<version>`. Would be happy to see a

chliny:
Thanks for asking. The failure is caused by an include-path name collision rather than a compiler shipping a broken `<version>` header.

We have a very large monorepo. My subproject depends on `nlohmann/json`, while another subproject depends on an old version of JsonCpp. The combined Make build adds dependency directories to the include search path, including a directory containing this file:

```text
jsoncpp-src-0.5.0/version
```

The affected target is compiled in a pre-C++20 mode. Without the `JSON_HAS_CPP_20` guard, `__has_include(<version>)` searches all configured include directories and returns `true` because it finds the unrelated JsonCpp `version` file. The subsequent:

```cpp
#include <version>
```

then includes that metadata file instead of the C++ standard library header, causing the compilation to fail because it is not a C++ header.

In other words, `__has_include(<version>)` confirms that a file with that name can be resolved, but it cannot confirm that the file is the standard library's C++20 `<version>` header.

The same problem could occur in other large builds whenever an include directory from another dependency contains a file named `version`. Guarding the probe with `JSON_HAS_CPP_20` prevents this collision for C++11/14/17 targets.

gregmarr:
Ouch, that's a tough one.  I'd be worried about this "must have C++20 defined" breaking some other already-working case.  If you control the include path, can you put your own empty `version` file somewhere in the include path before that directory, so it hides that file which isn't a C++ header, or can you exclude/rename that file?

chliny:
Thanks, I understand the compatibility concern. There are three reasons why I still think the `JSON_HAS_CPP_20` guard is the preferable direction.

First, from the language standard's perspective, `<version>` is a C++20 standard library header. Although an implementation may expose it as an extension in an earlier language mode, probing for and including it only when `JSON_HAS_CPP_20` is defined is consistent with the standard.

Second, the PR has passed 157 CI checks across a broad range of compilers, standard library implementations, language modes, platforms, and single-/multi-header configurations. The only failing check, `clang (18.1.8)`, failed while Chocolatey was downloading MinGW due to an HTTP 504 error; it never reached compilation and is unrelated to this change. While CI cannot cover every possible extension configuration, these results provide fairly strong confidence that the guard does not affect the project's supported configurations.

Third, the suggested workarounds are difficult to maintain in our environment:

- Excluding the JsonCpp directory from the aggregated include path is difficult in a large monorepo. It would also address only this particular JsonCpp dependency, while the same collision could occur with any other dependency directory containing a file named `version`.
- Adding an empty `version` shim would work for our current pre-C++20 target, but it is a fragile and non-obvious workaround. It would be easy to forget, and if the target is later upgraded to C++20, the shim would still shadow the real standard library `<version>` header and could silently suppress the expected feature-test macros.

For these reasons, I believe guarding a C++20 standard header with `JSON_HAS_CPP_20` is cleaner and less error-prone than adding project-specific include-path workarounds.

If there is a known supported pre-C++20 configuration that relies on including `<version>` to expose library features, I would be happy to investigate it and look for a narrower solution.

gregmarr:
Adding a requirement of `JSON_HAS_CPP_20` around the `__has_include` below will change `JSON_HAS_EXPERIMENTAL_FILESYSTEM` from `1` to `0` for someone that is using C++17 mode and a standard library that has `<version>` with `__cpp_lib_experimental_filesystem` defined.  This doesn't break the tests, because the tests are made to be keyed off of `JSON_HAS_EXPERIMENTAL_FILESYSTEM`, but will break or at even silently reduce the capability of those users.

```
#ifdef __has_include
    #if __has_include(<version>)
        #include <version>
    #endif
#endif

#if !defined(JSON_HAS_FILESYSTEM) && !defined(JSON_HAS_EXPERIMENTAL_FILESYSTEM)
    #ifdef JSON_HAS_CPP_17
        #if defined(__cpp_lib_filesystem)
            #define JSON_HAS_FILESYSTEM 1
        #elif defined(__cpp_lib_experimental_filesystem)
            #define JSON_HAS_EXPERIMENTAL_FILESYSTEM 1
        #elif !defined(__has_include)
            #define JSON_HAS_EXPERIMENTAL_FILESYSTEM 1
        #elif __has_include(<filesystem>)
            #define JSON_HAS_FILESYSTEM 1
        #elif __has_include(<experimental/filesystem>)
            #define JSON_HAS_EXPERIMENTAL_FILESYSTEM 1
        #endif
```

I'm just a user, not a maintainer, but to me, that threat of breakage more than outweighs the desire to support someone that has a file named `version` in their include path.  In the end, @nlohmann will need to make that call.

gregmarr:
>Third, the suggested workarounds are difficult to maintain in our environment:
>
> Excluding the JsonCpp directory from the aggregated include path is difficult in a large monorepo. It would also address only this particular JsonCpp dependency, while the same collision could occur with any other dependency directory containing a file named version.

True, but do you expect that to be widespread?  I have seen no other reports of this, ever.

> Adding an empty version shim would work for our current pre-C++20 target, but it is a fragile and non-obvious workaround. It would be easy to forget, and if the target is later upgraded to C++20, the shim would still shadow the real standard library <version> header and could silently suppress the expected feature-test macros.

It would only do so if you put that include path before the standard library include path, instead of between the standard library include path and the JsonCpp include path.  If the JsonCpp include path is already before the standard library include path, then you are already broken for C++20.

> and could silently suppress the expected feature-test macros.

Which is exactly what will happen with your proposed solution with a compiler in C++17 mode with a standard library with a `<version>` header that enables some of the feature test macros.

gregmarr:
> The combined Make build adds dependency directories to the include search path, including a directory containing this file:

> jsoncpp-src-0.5.0/version

It looks like JsonCpp 0.5.0 is from 2010, prior to even the GitHub repo existing.  I found a copy of it at https://github.com/AaronHwang/jsoncpp-src-0.5.0

This `version` file is in the root of the repo.  I'm wondering why you're including the root of the repo in the include path.  I would expect `jsoncpp-src-0.5.0/include` or `jsoncpp-src-0.5.0/include/json`, but not just `jsoncpp-src-0.5.0`.

nlohmann:
Thanks for the detailed investigation - this looks like a build-configuration issue (including the JsonCpp repo root rather than its `include/` directory) rather than a defect in nlohmann/json, and guarding `<version>` behind `JSON_HAS_CPP_20` risks silently degrading filesystem feature detection on other C++17 configurations, so I'd rather not take this change.

PR #5380, opened 2026-08-13
Throw other_error.502 when UBJSON use_type is set without use_size

### LINKED ISSUE #5321: to_ubjson()/to_bjdata() with use_type=true and use_size=false aborts in debug builds and emits invalid output in release builds
### Description

`to_ubjson.md` documents `use_type` as "must be combined with `use_size = true`", but the combination is never enforced — it is only asserted:

https://github.com/nlohmann/json/blob/8ec98e2c9ea9c0adc15a791feb3d4d4468621743/include/nlohmann/detail/output/binary_writer.hpp#L810-L816

There are three such `JSON_ASSERT(use_count)` in `write_ubjson` — in the `array`, `binary`, and `object` cases. The consequences split by build type:

- **Debug builds:** `JSON_ASSERT` is `assert`, so `to_ubjson(j, false, true)` terminates the process with `Assertion 'use_count' failed`. An `abort()` is the wrong response to a user-supplied argument combination — asserts should guard internal invariants, not public API parameters, and there is no way for a caller to recover.
- **`NDEBUG` builds:** the assert vanishes and the writer emits a `$` type marker with no `#` count, which is invalid UBJSON. `to_ubjson()` returns successfully and the library's own parser then rejects the result.

`to_bjdata()` shares the code path and behaves identically.

Note that the two parameters are plain `bool`s with defaults (`to_ubjson(j, use_size = false, use_type = false)`), so `to_ubjson(j, false, true)` is an easy call to write by accident — and its failure mode is either a crash or silent corruption depending on how the translation unit was compiled.

### Reproduction steps

Call `to_ubjson` with `use_size = false, use_type = true` on a non-empty array or object, once with asserts enabled and once with `-DNDEBUG`.

Expected: a thrown exception naming the invalid argument combination.

Actual, with `-DNDEBUG`:

| input | bytes written | read back by `from_ubjson` |
|---|---|---|
| `[1,2,3]` | `5b 24 69 01 02 03 5d` | `parse_error.112` |
| `{"a":1,"b":2}` | `7b 24 69 69 01 61 01 69 01 62 02 7d` | `parse_error.112` |
| `[]` | 2 bytes | `[]` (fine) |
| `{}` | 2 bytes | `{}` (fine) |
| `1` | 2 bytes | `1` (fine) |

```
[json.exception.parse_error.112] parse error at byte 4: syntax error while parsing UBJSON size:
    expected '#' after type information; last byte: 0x01
```

Empty containers and scalars are unaffected because the guarded branch is only reached for non-empty containers — so the failure is input-dependent as well as build-dependent.

Without `-DNDEBUG` the same call aborts:

```
Assertion `use_count' failed.
```

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <iostream>
using json = nlohmann::json;

int main()
{
    const json j = json::array({1, 2, 3});
    const auto v = json::to_ubjson(j, /*use_size=*/false, /*use_type=*/true);

    std::cout << v.size() << " bytes\n";        // NDEBUG: 7; otherwise: abort
    std::cout << json::from_ubjson(v) << "\n";  // throws parse_error.112
}
```

### Error messages

```
[json.exception.parse_error.112] parse error at byte 4: syntax error while parsing UBJSON size: expected '#' after type information; last byte: 0x01
```

g++ 13.3.0, Ubuntu 24.04, `-std=c++11`, with and without `-DNDEBUG`

### Library version

`develop` @ 8ec98e2

No API/ABI change under any of the options below — no signature, default argument, type, or enumerator changes.

**Option A — throw (preferred).** Reject the combination up front with a `type_error`/`out_of_range`, or add a dedicated id. Behaviour change per build type:

- Debug builds: `abort()` becomes a catchable exception. Strictly an improvement; nothing can depend on an abort.
- `NDEBUG` builds: a call that currently returns unparseable bytes starts throwing. This is the only real behaviour change, and it is hard to see it breaking anyone, since the output cannot be read back by this library or any conforming UBJSON parser. Worth a release note precisely because it is build-dependent today — a user who tested only in release mode may not know they are hitting it.
- `to_ubjson()` documents a strong exception-safety guarantee; throwing before writing anything preserves it.

**Option B — silently upgrade `use_type = true` to imply `use_size = true`.** Produces valid output and never throws, but ignores an explicit argument, which I would argue is worse: the caller asked for no size and silently gets sizes.

**Option C — leave it, and only document harder.** Keeps a public API whose failure mode is `abort()` in half of all builds. Not recommended.

Whichever is chosen, the three `JSON_ASSERT(use_count)` calls should go: an assertion is the wrong mechanism for validating caller-supplied arguments regardless of what replaces it.

One detail worth deciding explicitly: whether the check fires eagerly on entry to `to_ubjson` (so `to_ubjson(json::array(), false, true)` also throws) or lazily where it does today (so empty containers and scalars keep working). Eager is more predictable and easier to document; lazy is a smaller behaviour delta. I'd suggest eager, with the release note calling it out.

Krishnanand-G:
`to_ubjson(..., use_size=false, use_type=true)` and the same flags on `to_bjdata()` hit `JSON_ASSERT(use_count)` in debug and wrote a `$` without `#` in release. `from_ubjson` then rejected the bytes.

`write_ubjson` throws `other_error.502` before it writes anything. Tests cover both formats. Docs list the exception. Amalgamated header regenerated.

Fixes #5321

### DIFF
--- a/docs/mkdocs/docs/api/basic_json/to_bjdata.md
+++ b/docs/mkdocs/docs/api/basic_json/to_bjdata.md
@@ -52,6 +52,11 @@ optional, `#!cpp bjdata_version_t::draft2` by default.
 
 Strong guarantee: if an exception is thrown, there are no changes in the JSON value.
 
+## Exceptions
+
+- Throws [`other_error.502`](../../home/exceptions.md#jsonexceptionother_error502) if `use_type` is true and `use_size`
+  is false.
+
 ## Complexity
 
 Linear in the size of the JSON value `j`.
--- a/docs/mkdocs/docs/api/basic_json/to_ubjson.md
+++ b/docs/mkdocs/docs/api/basic_json/to_ubjson.md
@@ -45,6 +45,11 @@ The exact mapping and its limitations are described on a [dedicated page](../../
 
 Strong guarantee: if an exception is thrown, there are no changes in the JSON value.
 
+## Exceptions
+
+- Throws [`other_error.502`](../../home/exceptions.md#jsonexceptionother_error502) if `use_type` is true and `use_size`
+  is false.
+
 ## Complexity
 
 Linear in the size of the JSON value `j`.
--- a/docs/mkdocs/docs/examples/parser_callback_t.cpp
+++ b/docs/mkdocs/docs/examples/parser_callback_t.cpp
@@ -9,13 +9,13 @@ int main()
     auto text = R"({"IDs": [116, 943], "Width": 800})";
 
     // discard the array when the parser reads its opening bracket
-    json j_array_start = json::parse(text, [](int /*depth*/, json::parse_event_t event, json & /*parsed*/)
+    json j_array_start = json::parse(text, [](int /*depth*/, json::parse_event_t event, json& /*parsed*/)
     {
         return event != json::parse_event_t::array_start;
     });
 
     // discard the same array when the parser reads its closing bracket
-    json j_array_end = json::parse(text, [](int /*depth*/, json::parse_event_t event, json & /*parsed*/)
+    json j_array_end = json::parse(text, [](int /*depth*/, json::parse_event_t event, json& /*parsed*/)
     {
         return event != json::parse_event_t::array_end;
     });
@@ -33,7 +33,7 @@ int main()
     });
 
     // discard the top-level object
-    json j_root = json::parse(text, [](int /*depth*/, json::parse_event_t event, json & /*parsed*/)
+    json j_root = json::parse(text, [](int /*depth*/, json::parse_event_t event, json& /*parsed*/)
     {
         return event != json::parse_event_t::object_end;
     });
--- a/docs/mkdocs/docs/home/exceptions.md
+++ b/docs/mkdocs/docs/home/exceptions.md
@@ -965,3 +965,19 @@ A JSON Patch operation 'test' failed. The unsuccessful operation is also printed
     ```
     [json.exception.other_error.501] unsuccessful: {"op":"test","path":"/baz","value":"bar"}
     ```
+
+### json.exception.other_error.502
+
+[`to_ubjson`](../api/basic_json/to_ubjson.md) and [`to_bjdata`](../api/basic_json/to_bjdata.md) were called with
+`use_type = true` but `use_size = false`. UBJSON requires a size marker (`#`) after a type marker (`$`).
+
+!!! failure "Example message"
+
+    ```
+    [json.exception.other_error.502] use_type requires use_size = true
+    ```
+
+!!! note
+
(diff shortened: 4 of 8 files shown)


nlohmann (changes requested):
Please fix the amalgamation and the DCO requirements:

- Run `make amalgamate`.
- Sign-off your commits.

Krishnanand-G:
Ran `make amalgamate` and signed the commit. Ready when you are.

Krishnanand-G:
@nlohmann I ran `make amalgamate` and signed off on `dea2800`. DCO and `ci_test_amalgamation` are green on that commit. The earlier review still points at `b76ded7`. Could you take another look?

nlohmann (changes requested):
`include/nlohmann/detail/output/binary_writer.hpp` (mirrored in `single_include/nlohmann/json.hpp`):

- L751: 🔴 bug: the `use_type && !use_count` check moved to the top of `write_ubjson` fires for *every* value, not just the array/object/binary cases where `use_type` and `use_count` actually interact. `to_ubjson(42, false, true)`, `to_ubjson(json::array(), false, true)`, and `to_ubjson(json::binary({}), false, true)` all throw `other_error.502` on this branch, but succeed cleanly on `develop` (verified both in release mode). The original three `JSON_ASSERT(use_count)` calls were each guarded (`use_type && !j...->empty()` / `bjdata_draft3 || !...empty()`) — this check widens the actual UBJSON invariant (can't write a `$` type marker into a container without a `#` size marker) into a blanket precondition on the whole function. Please restore the guard, or scope the check to just the array/object/binary branches.
- L751: 🔵 nit: the check re-runs identically on every recursive call for nested elements (array/object items call `write_ubjson` again), since `use_count`/`use_type` never change across recursion. Cheap, but pure waste — validate once at the `to_ubjson`/`to_bjdata` entry point instead.

Tests only cover non-empty top-level array/object (`{{"a",1},{"b",2}}`, `{1,2,3}`) for the new case — no coverage for scalars or empty containers, which is exactly where the regression above slipped through.

---
Posted by Claude Code on behalf of @nlohmann.

Krishnanand-G:
@nlohmann Fixed. Moved the check into the array, binary, and object branches where type prefixes are emitted instead of checking at the top of write_ubjson. Added tests covering scalars and empty containers to confirm they do not throw. Amalgamated and signed off.

Krishnanand-G:
@nlohmann Amalgamation was failing on fdb3316. Fixed in cc37fd4 by re-running make amalgamate, and all checks are green now.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #5382, opened 2026-08-13
Throw type_error.318 when serializing discarded values to binary formats

### LINKED ISSUE #4714: Binary formats invalid encoding for <discarded> values in arrays and objects
### Description

Binary formats (BSON, CBOR, MessagePack, UBJSON/BJData) are creating broken outputs when `<discarded>` values are included in arrays/objects.

### Reproduction steps

https://godbolt.org/z/WEPP5WrYe

```
  JSON array encoded: [<discarded>,<discarded>,<discarded>,1]
  JSON array  parsed: [json.exception.parse_error.101] parse error at line 1, column 2: syntax error while parsing value - invalid literal; last read: '[<'
UBJSON array encoded: [i]
UBJSON array  parsed: [1]
MSGPCK array encoded: 0x 94 01
MSGPCK array  parsed: [json.exception.parse_error.110] parse error at byte 3: syntax error while parsing MessagePack value: unexpected end of input
  CBOR array encoded: 0x 84 01
  CBOR array  parsed: [json.exception.parse_error.110] parse error at byte 3: syntax error while parsing CBOR value: unexpected end of input

  JSON object encoded: {"bar":<discarded>,"baz":2,"foo":1}
  JSON object  parsed: [json.exception.parse_error.101] parse error at line 1, column 8: syntax error while parsing value - invalid literal; last read: '"bar":<'
UBJSON object encoded: {ibaribaziifooi}
UBJSON object  parsed: [json.exception.parse_error.113] parse error at byte 9: syntax error while parsing UBJSON string: expected length type specification (U, i, I, l, L); last byte: 0x62
MSGPCK object encoded: 0x 83 a3 62 61 72 a3 62 61 7a 02 a3 66 6f 6f 01
MSGPCK object  parsed: [json.exception.parse_error.113] parse error at byte 10: syntax error while parsing MessagePack string: expected length specification (0xA0-0xBF, 0xD9-0xDB); last byte: 0x02
  CBOR object encoded: 0x a3 63 62 61 72 63 62 61 7a 02 63 66 6f 6f 01
  CBOR object  parsed: [json.exception.parse_error.113] parse error at byte 10: syntax error while parsing CBOR string: expected length specification (0x60-0x7B) or indefinite string type (0x7F); last byte: 0x02
```

N/A

### Library version

3.11.3

### Validation

ameliabarnabyhub:
Fixes #4714

Discarded JSON values silently produced broken CBOR, MessagePack, and UBJSON output when nested in arrays or objects (element count/header written but payload skipped). Serialization now throws `type_error.318` instead, matching the approach discussed in the issue thread.

Adds regression tests for top-level, array, and object cases. Documents the new exception id.

### DIFF
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -410,7 +410,9 @@ class binary_writer
 
             case value_t::discarded:
             default:
-                break;
+            {
+                throw_on_discarded(j);
+            }
         }
     }
 
@@ -578,6 +580,10 @@ class binary_writer
                     oa->write_character(to_char_type(0xDB));
                     write_number(static_cast<std::uint32_t>(N));
                 }
+                else
+                {
+                    assert_msgpack_size(N, &j);
+                }
 
                 // step 2: write the string
                 oa->write_characters(
@@ -606,6 +612,10 @@ class binary_writer
                     // array 32
                     oa->write_character(to_char_type(0xDD));
                     write_number(static_cast<std::uint32_t>(N));
+                }
+                else
+                {
+                    assert_msgpack_size(N, &j);
                 }
 
                 // step 2: write each element
@@ -684,6 +694,10 @@ class binary_writer
                     oa->write_character(to_char_type(output_type));
                     write_number(static_cast<std::uint32_t>(N));
                 }
+                else
+                {
+                    assert_msgpack_size(N, &j);
+                }
 
                 // step 1.5: if this is an ext type, write the subtype
                 if (use_ext)
@@ -720,6 +734,10 @@ class binary_writer
                     oa->write_character(to_char_type(0xDF));
                     write_number(static_cast<std::uint32_t>(N));
                 }
+                else
+                {
+                    assert_msgpack_size(N, &j);
+                }
 
                 // step 2: write each element
                 for (const auto& el : *j.m_data.m_value.object)
@@ -732,7 +750,9 @@ class binary_writer
 
             case value_t::discarded:
             default:
-                break;
+            {
+                throw_on_discarded(j);
+            }
         }
     }
 
@@ -954,11 +974,18 @@ class binary_writer
 
             case value_t::discarded:
             default:
-                break;
+            {
+                throw_on_discarded(j);
+            }
         }
     }
 
   private:
+    static void throw_on_discarded(const BasicJsonType& j)
+    {
+        JSON_THROW(type_error::create(318, concat("cannot serialize ", j.type_name(), " values to binary format"), &j));
+    }
+
     //////////
     // BSON //
     //////////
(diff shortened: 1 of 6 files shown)


include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -578,6 +580,10 @@ class binary_writer
                     oa->write_character(to_char_type(0xDB));
                     write_number(static_cast<std::uint32_t>(N));
                 }
+                else
```
gregmarr, line 583: Looks to be left over from #5381

single_include/nlohmann/json.hpp
```diff
@@ -3722,71 +3722,71 @@ NLOHMANN_JSON_NAMESPACE_END
 // SPDX-License-Identifier: MIT
 
 #ifndef INCLUDE_NLOHMANN_JSON_FWD_HPP_
-    #define INCLUDE_NLOHMANN_JSON_FWD_HPP_
+#define INCLUDE_NLOHMANN_JSON_FWD_HPP_
```
gregmarr, line 3725: Looks like you have issues with your amalgamate setup.

nlohmann (changes requested):
Thanks for tackling #4714 — the exception-based fix is the right approach. A few issues to resolve before merge, found by Claude Code:

**Build-breaking**

1. `calc_bson_element_size` doesn't compile under `-Werror` (`include/nlohmann/detail/output/binary_writer.hpp:1264`). The `discarded`/`default` case calls `throw_on_discarded(j)` with no `return` afterward, and `throw_on_discarded` isn't marked `[[noreturn]]`, so GCC can't prove the non-void function always exits via exception. Confirmed in this PR's own CI: `error: control reaches end of non-void function [-Werror=return-type]` (`ci_test_gcc`) — this is why essentially every other job in the run is currently red too.
   - Fix: mark `throw_on_discarded` `[[noreturn]]`/`JSON_HEDLEY_NO_RETURN`, or add an unreachable `return` after the call.

2. `single_include/nlohmann/json.hpp` and `tests/src/unit-msgpack.cpp` don't match the pinned amalgamate/astyle output — the diff includes large unrelated reformatting instead of a minimal diff mirroring the `include/` changes. The `check` job's own amalgamate+astyle step overwrites these files and then fails (exit 1). Please re-run `make amalgamate` with the pinned toolchain and re-commit.

**Other issues**

3. `out_of_range.412` is reused for a second, unrelated condition. It already means "BSON length exceeds int32 max" (documented); this PR repurposes it for "MessagePack size exceeds uint32 max" without updating the docs, which still only describes the BSON case. Please give this its own id or document both meanings.

4. No `to_bson()` regression test for discarded values — `write_bson_element`/`calc_bson_element_size` now also throw `type_error.318` there, but `unit-bson.cpp` didn't get the same "discarded" test sections CBOR/MessagePack/UBJSON received.

5. The new MessagePack oversized-length branches (`else { assert_msgpack_size(N, &j); }`) never write a header before falling through to the payload write — this relies entirely on the throw actually unwinding, otherwise it silently writes an unprefixed payload, the same corruption class this PR fixes for discarded values.

---
*Written by Claude Code on behalf of @niels.*

PR #5383, opened 2026-08-13
CBOR store: fall back to ignore for non-binary tagged items

### LINKED ISSUE #5316: CBOR: cbor_tag_handler_t::store rejects documents that ignore accepts, and drops tag numbers 6-20
### Description

`cbor_tag_handler_t::store` is not a superset of `cbor_tag_handler_t::ignore`. Two separate gaps in the `store` branch of `parse_cbor_internal()`:

**1. A tagged item that is not a byte string is a hard parse error.** After reading the tag argument for `0xD8`–`0xDB`, the code unconditionally does `get(); return get_cbor_binary(b) && sax->binary(b);` ([`binary_reader.hpp:864-865`](https://github.com/nlohmann/json/blob/8ec98e2c9ea9c0adc15a791feb3d4d4468621743/include/nlohmann/detail/input/binary_reader.hpp#L864-L865)). If the enclosed data item is anything other than a byte string, parsing fails with `parse_error.113` — even though `ignore` handles the same document fine.

The most visible casualty is the CBOR self-describe tag 55799 (RFC 8949 §3.4.6), encoded `0xD9 0xD9 0xF7`, which is widely used as a magic-number prefix and is exactly the case that motivated #1968:

```
magic 55799 + map, ignore : OK -> {"a":1}
magic 55799 + map, store  : THROW: [json.exception.parse_error.113] parse error at byte 4:
    syntax error while parsing CBOR binary: expected length specification (0x40-0x5B)
    or indefinite binary array type (0x5F); last byte: 0xA1
```

**2. Tag numbers 6–20 are silently discarded.** For head bytes `0xC6`–`0xD4` the `store` switch hits `default:` and just recurses ([`binary_reader.hpp:861-862`](https://github.com/nlohmann/json/blob/8ec98e2c9ea9c0adc15a791feb3d4d4468621743/include/nlohmann/detail/input/binary_reader.hpp#L861-L862)), so `store` behaves identically to `ignore` there and the tag number is lost with no indication:

```cpp
auto j = json::from_cbor(std::vector<std::uint8_t>{0xCA, 0x41, 0x00}, // tag 10 over h'00'
                         true, true, json::cbor_tag_handler_t::store);
j.get_binary().has_subtype();  // false
```

`cbor_tag_handler_t.md` does scope `store` to "bytes 0xd8..0xdb", so #2 is documented. Combined with #5315 it means `store` covers only 4 of the 24 low tag head encodings.

### Reproduction steps

Parse tag `0xD8 0x18` over each CBOR content type under both handlers.

Expected: `store` records the subtype where it can and otherwise degrades to `ignore`'s behavior; it should never reject a document `ignore` accepts.

Actual (tag `0xD8 0x18` over various content):

| tagged item | `ignore` | `store` |
|---|---|---|
| byte string | OK | OK |
| unsigned int | OK | **THROW** |
| text string | OK | **THROW** |
| array | OK | **THROW** |
| map | OK | **THROW** |
| bool | OK | **THROW** |

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <iostream>
using json = nlohmann::json;

int main()
{
    // 55799({"a": 1}) -- CBOR self-describe magic followed by a map
    const std::vector<std::uint8_t> v{0xD9, 0xD9, 0xF7, 0xA1, 0x61, 0x61, 0x01};

    std::cout << json::from_cbor(v, true, true, json::cbor_tag_handler_t::ignore) << "\n"; // {"a":1}
    std::cout << json::from_cbor(v, true, true, json::cbor_tag_handler_t::store)  << "\n"; // throws
}
```

### Error messages

```
[json.exception.parse_error.113] parse error at byte 4: syntax error while parsing CBOR binary: expected length specification (0x40-0x5B) or indefinite binary array type (0x5F); last byte: 0xA1
```

### Library version

`develop` @ 8ec98e2

The two halves differ a lot in risk, and they can be fixed independently.

**Part 1 (fall back to `ignore` when the tagged item is not a byte string) — low risk.** No API/ABI change. Input that currently throws `parse_error.113` would start parsing, which only affects callers who explicitly opted into `store`. Depending on `store` throwing for a tagged map is hard to justify as intentional; the current behavior looks like an oversight rather than a contract. A caller who genuinely wants strictness still has `cbor_tag_handler_t::error`.

**Part 2 (actually store subtypes for tags 6–20) — higher risk, and I would not bundle it.** This changes the *value* produced for input that already parses today: a byte string tagged 10 currently yields a `binary` with `has_subtype() == false`, and would start yielding `has_subtype() == true, subtype() == 10`. That silently changes comparison results (`operator==` on binary values compares subtypes), `dump()` output, and any round-trip through `to_cbor`, which would then emit `0xD8 0x0A` where it previously emitted nothing. It is also arguably the wrong shape for the general problem, since a tag over a non-binary item still has nowhere to go. Worth splitting into its own issue, or leaving as documented behavior.

So: part 1 is a bug fix and safe to backport; part 2 is a behavior change that deserves its own discussion and at minimum a release note.

Both are somewhat entangled with #5315 — if the missing head bytes `0xC0`–`0xC5`/`0xD5`–`0xD7` are added there, the `store` branch will need to decide what to do with them too, and "fall back to `ignore`" is the natural answer for both.

ameliabarnabyhub:
Fixes #5316 (part 1)

`cbor_tag_handler_t::store` assumed every tagged item is a byte string. Tagged maps, arrays, and other non-binary values therefore failed with `parse_error.113` even though `ignore` accepted the same input (notably CBOR self-describe tag 55799).

When the tagged item is not a binary head byte, fall back to `ignore` parsing instead of forcing binary decoding. Binary tags still store the subtype as before.

Does not change behavior for tag numbers 6–20 over binary strings (part 2 of #5316).

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -899,7 +899,11 @@ class binary_reader
                                 return parse_cbor_internal(true, tag_handler);
                         }
                         get();
-                        return get_cbor_binary(b) && sax->binary(b);
+                        if ((current >= 0x40 && current <= 0x5B) || current == 0x5F)
+                        {
+                            return get_cbor_binary(b) && sax->binary(b);
+                        }
+                        return parse_cbor_internal(false, cbor_tag_handler_t::ignore);
                     }
 
                     default:                 // LCOV_EXCL_LINE
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -3722,71 +3722,71 @@ NLOHMANN_JSON_NAMESPACE_END
 // SPDX-License-Identifier: MIT
 
 #ifndef INCLUDE_NLOHMANN_JSON_FWD_HPP_
-    #define INCLUDE_NLOHMANN_JSON_FWD_HPP_
+#define INCLUDE_NLOHMANN_JSON_FWD_HPP_
 
-    #include <cstdint> // int64_t, uint64_t
-    #include <map> // map
-    #include <memory> // allocator
-    #include <string> // string
-    #include <vector> // vector
+#include <cstdint> // int64_t, uint64_t
+#include <map> // map
+#include <memory> // allocator
+#include <string> // string
+#include <vector> // vector
 
-    // #include <nlohmann/detail/abi_macros.hpp>
+// #include <nlohmann/detail/abi_macros.hpp>
 
 
-    /*!
-    @brief namespace for Niels Lohmann
-    @see https://github.com/nlohmann
-    @since version 1.0.0
-    */
-    NLOHMANN_JSON_NAMESPACE_BEGIN
+/*!
+@brief namespace for Niels Lohmann
+@see https://github.com/nlohmann
+@since version 1.0.0
+*/
+NLOHMANN_JSON_NAMESPACE_BEGIN
 
-    /*!
-    @brief default JSONSerializer template argument
+/*!
+@brief default JSONSerializer template argument
 
-    This serializer ignores the template arguments and uses ADL
-    ([argument-dependent lookup](https://en.cppreference.com/w/cpp/language/adl))
-    for serialization.
-    */
-    template<typename T = void, typename SFINAE = void>
-    struct adl_serializer;
+This serializer ignores the template arguments and uses ADL
+([argument-dependent lookup](https://en.cppreference.com/w/cpp/language/adl))
+for serialization.
+*/
+template<typename T = void, typename SFINAE = void>
+struct adl_serializer;
 
-    /// a class to store JSON values
-    /// @sa https://json.nlohmann.me/api/basic_json/
-    template<template<typename U, typename V, typename... Args> class ObjectType =
-    std::map,
-    template<typename U, typename... Args> class ArrayType = std::vector,
-    class StringType = std::string, class BooleanType = bool,
-    class NumberIntegerType = std::int64_t,
-    class NumberUnsignedType = std::uint64_t,
-    class NumberFloatType = double,
-    template<typename U> class AllocatorType = std::allocator,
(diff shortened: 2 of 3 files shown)


nlohmann (changes requested):
Found two correctness issues while reviewing this PR:

**1. `single_include/nlohmann/json.hpp` has diverged from `include/`** (most severe — likely root cause of the CI failures)

The amalgamation commit adds new logic (`throw_on_discarded()`, `assert_msgpack_size()`, discarded values now throwing `type_error.318` instead of producing empty output) that never landed in `include/nlohmann/detail/output/binary_writer.hpp`. That header still has the old `break;`/`JSON_ASSERT(false)` paths — confirmed against develop and against this PR's diff, which only touches `binary_reader.hpp`, `single_include/json.hpp`, and `tests/src/unit-cbor.cpp`, never `binary_writer.hpp`.

Since the test suite compiles against `include/`, the updated discarded-value tests in `tests/src/unit-cbor.cpp` (expecting `CHECK_THROWS_WITH_AS(...type_error.318...)`) can't pass. This lines up with essentially every CI check failing on this PR, including `ci_test_amalgamation`.

**2. The CBOR fix itself has a nested-tag regression**

In `include/nlohmann/detail/input/binary_reader.hpp` (~line 904), the fallback for a non-binary tagged item under `store` mode hardcodes:
```cpp
return parse_cbor_internal(false, cbor_tag_handler_t::ignore);
```
instead of propagating the enclosing `tag_handler` (which is `store` here). So `tag(100) -> [tag(80) -> h'01']` parsed under `store` mode loses the inner tag's subtype, because everything inside the non-binary wrapper now gets parsed with `ignore` instead of `store`. Passing `tag_handler` through instead of the literal `ignore` should fix this without changing the intended behavior for the top-level non-binary tag.

---
*Posted by Claude Code on behalf of @nlohmann.*

PR #5389, opened 2026-08-20
Fix stack overflow when copying a deeply nested value (#5387)

### LINKED ISSUE #5387: Stack overflow in copy constructor and dump() on deeply nested json (destructor was fixed in #1436)
### Description

`basic_json`'s copy constructor and `dump()` both recurse once per nesting level, no depth limit. Nest a `json` value deep enough and you blow the stack and crash the process (SIGSEGV) — no exception involved, nothing to catch.

Same bug class as #832/#1835, fixed for the destructor in #1436 ("destroy by flattening", made iterative). That fix never made it to the copy constructor or the serializer.

#1693 already reported this for `dump()` back in 2019. nlohmann even confirmed it: "the dump function is recursive, so a limited stack size can be a problem there." Got closed the next day as a stale "known limitation" anyway, and the actual ask in that thread — a depth limit for parse/dump, reject instead of crash — never happened. Copy constructor doesn't even have its own issue.

This already crashes llama.cpp's HTTP server in the wild (ggml-org/llama.cpp#27434), which copies and serializes client-controlled JSON (tool-call definitions) with no depth check. That PR patches it locally in llama.cpp, but the actual bug is here.

Relates to https://github.com/nlohmann/json/issues/1788.

### Reproduction steps

1. Build a JSON document nested far deeper than any legitimate document would be (e.g. 100,000 nested arrays) — `json::parse` handles this fine, since the parser is already iterative.
2. Copy the parsed value, or call `.dump()` on it.
3. Process crashes with SIGSEGV; no exception is thrown, so this cannot be caught or handled by the calling application.

**Expected:** either graceful rejection of documents nested beyond a configurable limit, or stability regardless of depth — matching how the destructor now behaves after #1436.

**Actual:** the process crashes with a segmentation fault.

### Minimal code example

```Shell
#include <nlohmann/json.hpp>
#include <string>

int main() {
    const int depth = 100000;
    std::string text(depth, '[');
    text += "0";
    text.append(depth, ']');

    nlohmann::json j = nlohmann::json::parse(text); // fine, parser is iterative
    nlohmann::json copy = j;                         // crashes: SIGSEGV
    // std::string s = j.dump();                     // also crashes: SIGSEGV
}
```

### Error messages

```Shell
Process terminates with SIGSEGV (exit code 139), no exception, no diagnostic. Verified independently of any application code, with a fresh build against this repo's single_include/nlohmann/json.hpp.
```

Apple clang version 21.0.0 (clang-2100.1.1.101) Target: arm64-apple-darwin25.6.0 Thread model: posix

### Library version

3.12.0

### Validation

nlohmann:
## What & why

`basic_json`'s copy constructor copies an object or array by handing the container to its own copy constructor, which copy-constructs every element and so reaches the copy constructor again — once per nesting level. A value nested deeply enough exhausts the call stack and terminates the process with `SIGSEGV`: no exception, nothing the caller can catch. Parsing such a value works (the parser is iterative), and so does destroying one (#1436 made destruction iterative) — copying never got the same treatment.

This is reachable from untrusted input wherever a parsed value is copied, which is what makes it worth fixing rather than documenting: it is currently crashing llama.cpp's HTTP server (ggml-org/llama.cpp#27434), which copies client-controlled JSON.

Fixes #5387 **for the copy constructor**. `dump()` is still recursive — that half is being handled in #5285, which already owns and restructures `serializer::dump()`.

## The approach: bound the descent, don't remove it

The obvious fix — copy every level through an explicit worklist — makes *every* copy pay for a problem only pathologically deep values have. I measured that version: **3% to 9% slower** depending on the shape of the value, because it has to inspect the elements before it can pick a strategy.

So instead of taking the call stack away from the copy, this bounds how far it descends:

- The first `copy_depth_limit()` (128) levels are copied **exactly as they were** — the containers copy their own elements, by far the fastest way to fill them.
- Only once the copy has descended 128 levels is the value below it finished without the call stack, through an explicit worklist (`copy_iteratively`).

Copying can therefore no longer exhaust the stack, however deeply a value is nested, while a value nested less deeply than the bound — all but a vanishing minority — runs the very same code as before and pays only for one counter.

`operator=` takes its argument by value, so copy assignment is fixed along with it.

### The counter, and `JSON_NO_THREAD_LOCAL`

The level counter lives in `thread_local` storage, as one shared between threads would be raced. **This is the library's first use of `thread_local`**, so it can be switched off: with `JSON_NO_THREAD_LOCAL` defined, objects and arrays go through the worklist right away. That yields the same values and is just as safe against deep nesting, but is measurably slower (see below). Documented as a macro page, with nav and index entries.

### Values copied while a copy is going on

The deferred values are completed before the copy they belong to returns. An earlier version of this deferred them to the *outermost* copy's worklist, which is slightly faster but means a value copied while another copy is in progress — by a custom base class, say — could be left as a null value when its constructor returns, and if it were a temporary, the outer copy would later write through a dangling pointer. The version here is not affected by the copy it is nested in; there is a test for it.

## Measured

Medians of 9 interleaved runs, clang `-O3`, versus `develop`:

| value | this PR | deferring every level | `JSON_NO_THREAD_LOCAL` |
|---|---|---|---|
| array of strings | −1.3% | +0.0% | +2.8% |
| flat object, 20k leaf values | +0.0% | +8.0% | +81.3% |
| flat array, 80k numbers | +0.1% | +3.3% | +28.2% |
| nested arrays | +0.3% | +2.3% | +22.9% |
| nested objects | +0.6% | +6.9% | +42.6% |
| twitter-like document | +1.2% | +9.5% | +24.4% |

Copying a three-key object costs about ten nanoseconds more — the counter.

## Public API impact

**No breaking changes.** No public signature, type, or exception changes; `copy_structured` and its helpers are private members. The one addition is the opt-out macro `JSON_NO_THREAD_LOCAL`, which is not defined by default, so existing builds are unaffected — except that they now use `thread_local` storage, which is the one thing to weigh in review.

No new requirements are placed on `ArrayType`/`ObjectType`: the worklist path uses only the fill constructor and range constructor the library already relies on (`json.hpp:4288`, `1153`, `1160`).

## Verification

- **Differential vs. `develop`, byte-identical:** 8000 randomized values (`json` and `ordered_json`) across three dump modes and both equality operators; plus a second differential walking *every node* of a parsed corpus comparing type, `start_pos()`/`end_pos()`, diagnostic path and value. Both run under the default build, `JSON_DIAGNOSTICS`, `JSON_DIAGNOSTIC_POSITIONS`, both together, **and** with `JSON_NO_THREAD_LOCAL` — identical in every combination.
- **The boundary where the two strategies meet** is covered for **every depth from 1 to 300**, for arrays and objects, plus alternating shapes and values that are deep in one branch only — so the bound can be moved without silently breaking anything.
- The new tests fail (crash) against `develop` and pass here.
- Unit suites green in C++11/17/20, under `JSON_DIAGNOSTICS` and under `JSON_NO_THREAD_LOCAL`; ASan + UBSan clean; warning-clean under the CI clang flag set (`-Weverything -Werror` minus the project's exclusions).
- `make amalgamate` run; re-amalgamation is stable.

Note that the deep-nesting tests deliberately compare `copy.dump()`/walk the values by hand rather than using `operator==`: comparison recurses through the containers (`json.hpp:3669`) and would overflow the stack itself.

- **CI covers the iterative path with the whole suite.** The descent bound means the path this PR adds is otherwise only reached by the handful of tests that nest deeper than 128 levels. `JSON_NO_THREAD_LOCAL` switches the descent off entirely, so every value goes down it - which is how the `ordered_map` key-comparator bug in the follow-up was caught in the first place. The new `ci_test_no_thread_local` target runs the full suite that way (109/109 tests pass); the macro had no build coverage at all before.
- **The metadata a copy has to carry over is now tested.** Copying a nested value has to reproduce what the element-wise copy constructor did: the parents `JSON_DIAGNOSTICS` reports paths from, and the positions `JSON_DIAGNOSTIC_POSITIONS` reports. Neither was tested anywhere before this PR - and the positions *were* broken in an earlier revision of it. Both are now checked on either side of the bound, for objects and arrays. Removing either fix makes the new tests fail: positions at exactly `depth=129` (with 1/127/128 still passing, pinning it to the iterative path), and the diagnostic pointer collapsing from `/a/a/.../a` to empty.

## Why this PR also splits a test file

`tests/src/unit-regression2.cpp` is at the size limit the MinGW linker copes with, and the copy helpers push it over:
its object grows 6.3% (4,654,128 -> 4,944,920 bytes at `-O0`), and the `clang` job on Windows then fails with
`relocation truncated to fit: IMAGE_REL_AMD64_REL32 against '.rdata'`. `develop` links at the smaller size, so this is
caused by the change - but the file was already at the edge, and the job already carries a comment about the same test
hitting the same class of failure against `.debug_line`.

I first tried shrinking the object with `-O1` in that job. It links - and then **39 of 102 tests segfault before doctest
prints its first line**, on both clang 11.0.1 and clang 18.1.8. That approach is reverted; a comment in the workflow now
records it so nobody tries it again.

The fix is to make the object smaller instead: the test cases after `TEST_CASE("regression tests 2")` move to a new
`unit-regression3.cpp`, which brings the object to 4,687,888 bytes - 0.7% above the size that links today rather than
6.3%. Both files still build for C++11/17/20, and the same 9 test cases and 135 assertions run as before, just in two
binaries. `CONTRIBUTING.md` now points new regression tests at `unit-regression3.cpp`.

## What still recurses after this PR

This PR fixes copying, and nothing else. These still recurse once per nesting level on user-controlled data, so #5387 should not be closed as fully fixed when this merges:

| what | where | status |
|---|---|---|
| `operator==`, `operator<`, `operator<=>` | `json.hpp:3669` | fixed in #5390, which also fixes an exponential blow-up in the pre-C++20 `operator<` |
| `dump()` | `serializer.hpp` | fixed in #5285 |
| `to_cbor` / `to_msgpack` / `to_ubjson` / `to_bson` | `binary_writer.hpp` | open |
| `basic_json::diff` | `json.hpp:5089` | open |
| `basic_json::merge_patch` | `json.hpp:5231` | open |
| `json_pointer::flatten` | `json_pointer.hpp:861` | open |

## How other libraries handle this

Worth knowing where this sits, because the trade-off taken here is deliberately not the common one.

**Most libraries reject deeply nested input rather than support it** — a hard depth limit at parse time, which then protects everything downstream:

| library | limit |
|---|---|
| [Boost.JSON](https://www.boost.org/doc/libs/1_83_0/libs/json/doc/html/json/ref/boost__json__parse_options.html) | `parse_options::max_depth` = 32 |
| [serde_json](https://github.com/serde-rs/json/issues/162) | 128, then `RecursionLimitExceeded` |
| [Jackson](https://github.com/FasterXML/jackson-core/pull/943) | `maxNestingDepth` 1000 (500 in 3.0) |
| [simdjson](https://simdjson.org/api/0.9.0/md_doc_basics.html) | 1024, refuses to parse deeper |
| .NET `System.Text.Json` | 65 |

Jackson's limit exists because of [CVE-2025-52999](https://www.herodevs.com/blog-posts/cve-2025-52999-denial-of-service-via-stack-overflow-in-jackson-core), and [CVE-2026-29062](https://github.com/advisories/GHSA-6v53-7c9g-w56r) covers a parser that bypassed it; PostgreSQL [patched its OAuth JSON parser](https://www.postgresql.org/message-id/CAJ7c6TO_MwMnpuw3%2Bdub%2BGif7c2tkS90YjHbyR%2BtmzpSP_ooXw%40mail.gmail.com) the same way. Only a few libraries are iterative throughout instead: [yyjson](https://github.com/ibireme/yyjson) advertises "unlimited JSON nesting levels" as a headline feature, and [miniserde](https://docs.rs/miniserde) gives its value type a non-recursive `Drop`. A third approach is to grow the stack on demand, as [serde_stacker](https://github.com/dtolnay/serde-stacker) does.

**This library sits in the worst spot of the three.** Its parser is iterative, so it accepts input nested arbitrarily deeply — and then hands it to operations that cannot cope. [lovasoa/bad_json_parsers](https://github.com/lovasoa/bad_json_parsers), which measures exactly this, buckets it under *"unlimited → segfault"*. A library with a parse limit never builds such a value in the first place.

**The closest C++ peer has the same split and has fixed none of it.** RapidJSON's `kParseIterativeFlag` makes only *parsing* iterative, and is opt-in because the recursive parser is faster; its destructor is still recursive ([#2217](https://github.com/Tencent/rapidjson/issues/2217), open, asking for exactly what #1436 did here in 2019), `CopyFrom` still overflows on deep values, and `Value::Accept()` — its serializer — still recurses.

Two things that supports about the approach taken here: RapidJSON keeps recursion as the *default because it is faster*, which matches the measurement that writing every value iteratively costs up to 20%; and a bound of 128 sits in the same range as everyone else's hard limits — except that nothing is rejected at it, so this library stays strictly more permissive than all of them while no longer crashing.

---

- [x] If applicable, an [existing issue](https://github.com/nlohmann/json/issues) is referenced. _(#5387)_
- [x] If applicable, the [documentation](https://json.nlohmann.me) is updated. _(`JSON_NO_THREAD_LOCAL` macro page, nav entry, and supported-macros overview)_

---

Written by Claude Code.


include/nlohmann/json.hpp
```diff
@@ -821,6 +821,307 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+
+    /// @brief how many levels the copy going on in this thread has descended into
+    static std::size_t& copy_depth() noexcept
+    {
+        static thread_local std::size_t depth = 0; // NOLINT(misc-use-internal-linkage)
+        return depth;
+    }
+#endif
+
+#ifndef JSON_NO_THREAD_LOCAL
```
gregmarr: why close and reopen the ifndef?

include/nlohmann/json.hpp
```diff
@@ -821,6 +821,307 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+    class copy_depth_guard
+    {
+      public:
+        explicit copy_depth_guard(std::size_t& depth) noexcept
+            : m_depth(depth)
+        {
+            ++m_depth;
+        }
+
+        ~copy_depth_guard() noexcept
```
gregmarr: `noexcept` is the default for destructors.

include/nlohmann/json.hpp
```diff
@@ -821,6 +821,307 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+
+            // the container is complete and will not be modified again
+            dst_value->set_parents();
+
+            if (worklist.empty())
+            {
+                break;
+            }
+
+            src_value = worklist.back().first;
```
gregmarr: should this only call `.back()` once for performance reasons?

include/nlohmann/json.hpp
```diff
@@ -821,6 +821,307 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         return j;
     }
 
+#ifndef JSON_NO_THREAD_LOCAL
+    /// the number of levels the copy constructor descends into before it
+    /// finishes the value below without the call stack
+    static constexpr std::size_t copy_depth_limit()
```
gregmarr: Would it be better to use a smaller size variable here, such as `unsigned char` to keep down the amount of thread storage used?

include/nlohmann/json.hpp
```diff
@@ -821,6 +821,307 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+            // before this returns, so a copy made by a custom base class - or by
+            // anything else that runs while a copy is going on - is unaffected
+            // by the copy it is nested in.
+            copy_iteratively(src);
+            return;
+        }
+
+        const copy_depth_guard guard(depth);
+        copy_level(src);
+#endif
```
gregmarr: If you invert things a little, the no-thread-local and too-deep are the same code.

```
#ifndef JSON_NO_THREAD_LOCAL
        std::size_t& depth = copy_depth();

        if (JSON_HEDLEY_LIKELY(depth < copy_depth_limit()))
        {
          const copy_depth_guard guard(depth);
          copy_level(src);
          return;
       } 
#endif

       copy_iteratively(src);
```

You could also have `copy_depth_guard` do all the access and testing.
```
    class copy_depth_guard
    {
      public:
        explicit copy_depth_guard() noexcept
            : m_okay(copy_depth() < copy_depth_limit())
        {
            ++copy_depth();
        }

        ~copy_depth_guard()
        {
            --copy_depth();
        }

        bool okay() const { return m_okay; }

        copy_depth_guard(const copy_depth_guard&) = delete;
        copy_depth_guard& operator=(const copy_depth_guard&) = delete;
        copy_depth_guard(copy_depth_guard&&) = delete;
        copy_depth_guard& operator=(copy_depth_guard&&) = delete;

      private:
        bool m_okay;
    };
```

```
    void copy_structured(const basic_json& src)
    {
#ifndef JSON_NO_THREAD_LOCAL
      const copy_depth_guard guard;

      if (JSON_HEDLEY_LIKELY(guard.okay()))
      {
        copy_level(src);
        return;
      }
#endif

      copy_iteratively(src);
    } 
```

gregmarr: #5390 already does the proposed change for `copy_depth_guard` except that it does it in a separate class instead of reusing `copy_depth_guard`, but it also renames `copy_depth_guard` to make it more general and then doesn't use it more generally.

include/nlohmann/json.hpp
nlohmann: Will fix.

include/nlohmann/json.hpp
nlohmann: Right. Will remove.

include/nlohmann/json.hpp
nlohmann: Right. Will fix.

include/nlohmann/json.hpp
nlohmann: Will do.

include/nlohmann/json.hpp
nlohmann: Nice!

include/nlohmann/json.hpp
```diff
@@ -821,6 +821,336 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         return j;
     }
 
+    /// the number of levels an operation descends into before it finishes the
+    /// value below it without the call stack
+    static constexpr std::size_t nesting_depth_limit()
```
gregmarr: Isn't this only needed when `JSON_NO_THREAD_LOCAL` is not defined?

Also, I just realized that most of the other defines like this are `#if 1/0` not "ifdef/ifndef`.  Is that intentional?  I have a feeling that is for ones that are detected but the user can override.

include/nlohmann/json.hpp
```diff
@@ -821,6 +821,336 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+        static thread_local std::size_t depth = 0; // NOLINT(misc-use-internal-linkage)
+        return depth;
+    }
+#endif
+
+    /*!
+    @brief counts one level of a bounded descent for as long as it runs
+
+    The count is taken rather than looked up here, because the caller has looked
+    it up already to test it against the limit: reaching thread-local storage is
```
gregmarr: It is not required that the caller look it up first.

include/nlohmann/json.hpp
```diff
@@ -821,6 +821,336 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+    @sa https://github.com/nlohmann/json/issues/5387
+    */
+    void copy_structured(const basic_json& src)
+    {
+#ifndef JSON_NO_THREAD_LOCAL
+        std::size_t& depth = nesting_depth();
+
+        if (JSON_HEDLEY_LIKELY(depth < nesting_depth_limit()))
+        {
+            const nesting_depth_guard guard(depth);
```
gregmarr: You say that `nesting_depth_guard` takes the depth because the caller has already looked it up, but that isn't necessary as I showed in my last comment below.

```
#ifndef JSON_NO_THREAD_LOCAL
      const copy_depth_guard guard;

      if (JSON_HEDLEY_LIKELY(guard.okay()))
      {
        copy_level(src);
        return;
      }
#endif
```

include/nlohmann/json.hpp
```diff
@@ -821,6 +821,351 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+    not nested deeper than the limit - all but a vanishing minority - is copied
+    exactly as it was before, and this whole detour costs it one counter.
+
+    @sa https://github.com/nlohmann/json/issues/5387
+    */
+    void copy_structured(const basic_json& src)
+    {
+        const nesting_depth_guard guard;
+
+        if (JSON_HEDLEY_LIKELY(guard.okay()))
```
gregmarr, line 1155: Does it still make sense to say that this is likely?

nlohmann: I think it's still valid: 128 levels is far deeper than virtually any real JSON document nests, so `guard.okay()` stays true on the vast majority of calls. Sharing the counter with comparison doesn't change that, since the two never nest into each other.

gregmarr: Sorry I wasn't specific enough, I meant specifically because of the change to use this even when TLS isn't available, so `okay()` never returns true.  I guess in that case, because this is all template code, it hopefully reduces to `JSON_HEDLEY_LIKELY(false)` and the compiler should be smart enough to eliminate that.

nlohmann: You're right, and I checked: compiling `single_include/nlohmann/json.hpp` at `-O2` with `-DJSON_NO_THREAD_LOCAL` (clang), the copy constructor's generated code jumps straight to `copy_iteratively` unconditionally — no reference to `nesting_depth`/TLS survives, and the fast `copy_level` path isn't reachable code at all. Compiling without the macro, the thread-local `nesting_depth` access is present and the branch is real. So yes — `guard.okay()` folds to a compile-time `false` there and LLVM eliminates the whole branch (and the `LIKELY` hint with it) rather than leaving a mispredicted check at runtime.

gregmarr: Excellent, thanks for confirming.

include/nlohmann/json.hpp
```diff
@@ -822,6 +822,351 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+    {
+        // a custom base class is only required to be copy-constructible and
+        // move-assignable, so the copy has to go through a temporary
+        static_cast<json_base_class_t&>(dst) = json_base_class_t(static_cast<const json_base_class_t&>(src));
+
+#if JSON_DIAGNOSTIC_POSITIONS
+        dst.start_position = src.start_position;
+        dst.end_position = src.end_position;
+#else
+        static_cast<void>(src);
```
gregmarr, line 919: Is this needed for when `json_base_class_t` is empty?  If so, a comment would be helpful.

nlohmann: Confirmed — not needed, even for the empty-base-class case. `src` and `dst` are already referenced unconditionally on the line above (the base-class copy), so `-Wunused-parameter` doesn't fire either way; checked with `-Wall -Wextra -Wunused-parameter` under both `JSON_DIAGNOSTIC_POSITIONS` 0 and 1. Pushed a commit removing the two `static_cast<void>` lines.

gregmarr (approved):
Looks good other than the test file merge conflict.

PR #5390, opened 2026-08-20
Fix stack overflow and exponential runtime when comparing nested values

nlohmann:
## What & why

Comparing two values compares their containers, which compare their elements, which brings the comparison back once per nesting level. Two values nested deeply enough exhaust the call stack and terminate the process — the same bug as #5387, in the last operation of `basic_json` that still had it. (The copy constructor is #5389, `dump()` is #5285, the destructor was #1436.)

While fixing that, a second and worse problem turned up.

## An ordered comparison took exponentially long before C++20

`std::vector::operator<` is a lexicographical comparison: for each element it asks whether the element is less than its counterpart, and then whether the counterpart is less than *it*. Both questions recurse into everything below that element, so every level doubles the work.

Measured on `develop`, comparing two **equal** values nested *n* levels deep with `<`:

| depth | `develop` |
|---|---|
| 20 | 5.7 ms |
| 26 | 237 ms |
| **30** | **3778 ms** |
| 40 | ≈ 1 hour (extrapolated) |

Nothing about such a value is pathological — 30 levels of nesting is ordinary, and reaching this needs no crafted input, just `a < b` on two parsed documents. C++20 is unaffected: `std::lexicographical_compare_three_way` asks once, and `operator<` is derived from `<=>` there.

## The change

A value nested too deeply to descend into is compared on an explicit stack instead, in a single pass that yields *less*, *equal*, *greater* or *unordered* at once.

- **Equality** and the **three-way comparison** descend as they always did for the first 128 levels — nothing measurable changes for them — and finish iteratively below that.
- **An ordered comparison never descends**, which is what takes the exponent out of it.

Equality needs no ordering, so it no longer asks for any: the iterative path is templated on whether the values are being ordered, so a key or string type that can only be compared for equality still compiles.

## Reproducing the results exactly

Two subtleties had to be reproduced, both found by differential testing rather than by reading:

1. `std::lexicographical_compare` **steps over** a pair it cannot order — a NaN, say — and carries on with the next element, where `std::lexicographical_compare_three_way` **stops at it**. The iterative pass does whichever the caller needs.
2. An object compares its keys with `<` where its entries are ordered, but with `==` where they are only checked for equality — **not** with the object's own comparator, which for `nlohmann::ordered_map` is `std::equal_to` and would report every equal key as "less".

## Measured

Medians of 7 interleaved runs, clang `-O3`, C++11:

| workload | |
|---|---|
| two equal values nested 30 deep, `<` | **3778 ms → 0.002 ms** |
| ordering flat objects | **−33.6%** |
| equality, every shape | unchanged |
| ordering flat arrays of numbers | **+27.3%** |

The last row is the one shape that pays for the single pass, and it buys the row above it.

## Public API impact

**No breaking changes.** No public signature, type, or exception changes; the comparison helpers are private members. Results are unchanged for every pair of values — see below. Nothing is rejected that was accepted before.

## Verification

- **Differential vs. `develop`:** 68,121 comparisons — every pair drawn from a corpus covering NaN, discarded values, mixed number types, binary values, empty containers, both object types and every operator (`==`, `!=`, `<`, `<=`, `>`, `>=`) — **identical** in C++11, C++17 and C++20, with and without `thread_local` storage, and with `JSON_USE_LEGACY_DISCARDED_VALUE_COMPARISON`.
- `unit-comparison.cpp` (4097 assertions) green in all three standards and in legacy mode; full suite green in every configuration; ASan + UBSan clean; warning-clean under the CI clang flag set.
- New tests compare values nested 100,000 deep, for arrays and objects, with every operator.

`JSON_NO_THREAD_LOCAL` earned its keep here: it forces the iterative path for *every* value, and it is what caught the `ordered_map` key-comparator bug above — a divergence that would otherwise only appear below depth 128 and be practically untestable.

## Note for reviewers

This is stacked on #5389, which introduces the depth counter this reuses. Retarget to `develop` once that lands.

`binary_writer` (`to_cbor`/`to_msgpack`/`to_ubjson`/`to_bson`) is not in this PR because the four formats need four separate iterative writers, and BSON additionally computes each document's length up front through a second recursive walk. Worth its own change.

## What still recurses after this PR

| what | where | status |
|---|---|---|
| copy constructor (and everything built on it) | `json.hpp` | fixed in #5389 |
| `dump()` | `serializer.hpp` | fixed in #5285 |
| `operator==`, `operator<`, `operator<=>` | `json.hpp:3669` | **this PR** |
| `to_cbor` / `to_msgpack` / `to_ubjson` / `to_bson` | `binary_writer.hpp` | open |
| `basic_json::diff` | `json.hpp:5089` | open |
| `basic_json::merge_patch` | `json.hpp:5231` | open |
| `json_pointer::flatten` | `json_pointer.hpp:861` | open |

`diff`, `merge_patch` and `flatten` were not previously called out anywhere; they recurse once per nesting level on user-controlled data exactly as copying and comparison did. #5387 should not be closed as fully fixed until they are dealt with too.

---

- [x] If applicable, an [existing issue](https://github.com/nlohmann/json/issues) is referenced. _(#5387)_
- [x] If applicable, the [documentation](https://json.nlohmann.me) is updated. _(the `JSON_NO_THREAD_LOCAL` page and the macro overview now cover comparison as well as copying, including what the macro costs a comparison)_

---

Written by Claude Code.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -847,14 +847,32 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         static thread_local std::uint8_t depth = 0; // NOLINT(misc-use-internal-linkage)
         return depth;
     }
-
-    /// @brief how many levels the comparison going on in this thread has descended into
-    static std::size_t& compare_depth() noexcept
-    {
-        static thread_local std::size_t depth = 0; // NOLINT(misc-use-internal-linkage)
-        return depth;
-    }
 #endif
+
+    /*!
+    @brief whether a descent must stop here and finish without the call stack
+
+    @a may_descend says whether the operator descends at all; it is a constant
+    at every call site, and is passed rather than tested by the caller so that
+    the test does not become a constant condition there, which MSVC reports as
+    C4127.
+
+    The comparison operators use this rather than @ref nesting_depth_guard::okay,
+    because they are written as a macro and a macro cannot use the preprocessor
+    the way the guard's constructor does; @ref copy_structured, which can, asks
+    the guard instead and never calls this.
+    */
+    static bool nesting_depth_exhausted(bool may_descend = true) noexcept
+    {
+#ifdef JSON_NO_THREAD_LOCAL
+        // without a count of its own per thread, a descent cannot be bounded
+        // without racing another one, so none is made
+        static_cast<void>(may_descend);
+        return true;
+#else
+        return !may_descend || nesting_depth() >= nesting_depth_limit();
+#endif
+    }
 
     /*!
     @brief counts one level of a bounded descent for as long as it runs, and
@@ -1178,53 +1196,6 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     /// ordered at all, such as a discarded value or a NaN
     enum class compare_result { less, equal, greater, unordered };
 
-    /*!
-    @brief counts one level of a comparison for as long as it runs
-
-    Does nothing without thread_local storage, where no descent is made at all.
-    */
-    class compare_depth_guard
-    {
-      public:
-        compare_depth_guard() noexcept
-        {
-#ifndef JSON_NO_THREAD_LOCAL
-            ++compare_depth();
-#endif
-        }
-
-        ~compare_depth_guard() noexcept
-        {
-#ifndef JSON_NO_THREAD_LOCAL
-            --compare_depth();
-#endif
-        }
-
-        compare_depth_guard(const compare_depth_guard&) = delete;
-        compare_depth_guard& operator=(const compare_depth_guard&) = delete;
-        compare_depth_guard(compare_depth_guard&&) = delete;
-        compare_depth_guard& operator=(compare_depth_guard&&) = delete;
-    };
-
-    /*!
-    @brief whether a comparison must stop descending and finish iteratively
-
-    @a may_descend says whether the operator descends at all; it is constant at
-    every call site, and is passed rather than tested by the caller so that the
-    test does not become a constant condition there, which MSVC reports (C4127).
-    */
-    static bool compare_descent_exhausted(bool may_descend) noexcept
-    {
-#ifdef JSON_NO_THREAD_LOCAL
-        // without a counter of its own per thread, the descent cannot be
-        // bounded without racing another one, so none is made
-        static_cast<void>(may_descend);
-        return true;
-#else
-        return !may_descend || compare_depth() >= compare_depth_limit();
-#endif
-    }
-
 #if JSON_HAS_THREE_WAY_COMPARISON
     /// @brief the ordering that @a result stands for
     static std::partial_ordering to_partial_ordering(compare_result result) noexcept // *NOPAD*
@@ -1244,18 +1215,18 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     }
 #endif
 
-    /// the number of levels a comparison descends into before it compares what
-    /// is left without the call stack
-    static constexpr std::size_t compare_depth_limit()
-    {
-        return 128;
-    }
-
     /*!
     @brief compare two values that are not both an array or both an object
 
     Such a pair is compared by the operators themselves, which cannot descend
     into it and therefore cannot recurse.
+
+    That holds for a pair whose types differ as much as for a pair of leaves: an
+    array and an object are told apart by their types alone, because an operator
+    only ever descends into two values of the same type. So `==` reports them as
+    unequal without looking inside either, and an ordering falls back to the
+    order of the types - an object sorts before an array - exactly as it does
+    for a value that is not nested deeply enough to get here.
     */
     template<bool Ordered>
     static compare_result compare_leaves(const_reference lhs, const_reference rhs) noexcept
@@ -1329,7 +1300,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     /*!
     @brief compare @a lhs and @a rhs without descending into them
 
-    Reached once a comparison has descended @ref compare_depth_limit levels, so
+    Reached once a comparison has descended @ref nesting_depth_limit levels, so
     that comparing values cannot exhaust the call stack however deeply they are
     nested. The two values are walked in lockstep on an explicit stack and
     compared lexicographically, element by element in the order the containers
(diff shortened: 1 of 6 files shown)


include/nlohmann/json.hpp
```diff
@@ -835,33 +835,38 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         static thread_local std::size_t depth = 0; // NOLINT(misc-use-internal-linkage)
         return depth;
     }
+
+    /// @brief how many levels the comparison going on in this thread has descended into
```
gregmarr: Why do you need a separate variable for this, rather than renaming `copy_depth` to `nesting_depth`?

include/nlohmann/json.hpp
```diff
@@ -1116,12 +1121,299 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+    /// the result of comparing two values, including values that cannot be
+    /// ordered at all, such as a discarded value or a NaN
+    enum class compare_result { less, equal, greater, unordered };
+
+    /*!
+    @brief counts one level of a comparison for as long as it runs
+
+    Does nothing without thread_local storage, where no descent is made at all.
+    */
+    class compare_depth_guard
```
gregmarr: Why is this needed when you have `nesting_depth_guard`?

include/nlohmann/json.hpp
```diff
@@ -1116,12 +1121,299 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+                return std::partial_ordering::unordered;
+        }
+    }
+#endif
+
+    /// the number of levels a comparison descends into before it compares what
+    /// is left without the call stack
+    static constexpr std::size_t compare_depth_limit()
+    {
+        return 128;
```
gregmarr: Needed?  It's the same as the other depth limit.  Is this so that it can be different in the future?

include/nlohmann/json.hpp
```diff
@@ -1116,12 +1121,299 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+    /*!
+    @brief compare two values that are not both an array or both an object
+
+    Such a pair is compared by the operators themselves, which cannot descend
+    into it and therefore cannot recurse.
+    */
+    template<bool Ordered>
+    static compare_result compare_leaves(const_reference lhs, const_reference rhs) noexcept
+    {
+        if (lhs == rhs)
```
gregmarr, line 1233: What happens here if one is an array and one is an object?

gregmarr:
~Haven't looked at all of this in detail yet.~
Done.

include/nlohmann/json.hpp
nlohmann: compare_leaves is only reached when the pair is not both arrays or both objects, so lhs == rhs hits operator=='s type-mismatch path and returns without descending; order_leaves then falls back to comparing the value_t values.

include/nlohmann/json.hpp
nlohmann: Will fix.

include/nlohmann/json.hpp
nlohmann: Will fix.

include/nlohmann/json.hpp
nlohmann: Will fix.

include/nlohmann/json.hpp
gregmarr: Okay.

include/nlohmann/json.hpp
```diff
@@ -3952,11 +4242,25 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
-                \
+            }                                                                                            \
+            \
             case value_t::object:                                                                        \
+            {                                                                                            \
+                if (JSON_HEDLEY_UNLIKELY(nesting_depth_exhausted(may_descend)))                        \
+                {                                                                                        \
+                    return (deep_result);                                                                \
+                }                                                                                        \
+                const nesting_depth_guard guard;                                                         \
```
gregmarr, line 4260: Can you do it like this, or is this where you hit C4127?
```
            case value_t::array:
            {
                const nesting_depth_guard guard;
                if (may_descend && guard.okay())
                {
                    return (deep_result);
                }
                return (*lhs.m_data.m_value.array) op (*rhs.m_data.m_value.array);
            }
```

include/nlohmann/json.hpp
```diff
@@ -1151,6 +1194,253 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+        {
+            return compare_result::less;
+        }
+
+        if (rhs < lhs)
+        {
+            return compare_result::greater;
+        }
+
+        return compare_result::equal;
```
gregmarr, line 1264: Could this take advantage of three way compare when available?

include/nlohmann/json.hpp
nlohmann: Yes — that's exactly it. `nesting_depth_exhausted(may_descend)` exists because `may_descend` is a compile-time constant at each call site inside the `JSON_IMPLEMENT_OPERATOR` macro; testing it directly there (`if (may_descend && guard.okay())`) turns the condition into a constant one for some instantiations, which MSVC flags as C4127. Passing it into the function sidesteps that. `nesting_depth_guard` is still used the same way `copy_structured` uses it — this path just can't fold the test into the guard itself because it's a macro, not a function that can use `#ifdef` internally.

include/nlohmann/json.hpp
nlohmann: It could, but only for key types that actually support `<=>` — the library only requires `operator<` on `object_t::key_type` (that's the documented contract via `std::less<>`), so a custom key type isn't guaranteed to have it even when `JSON_HAS_THREE_WAY_COMPARISON` is set. Doing this properly would mean a SFINAE/concept-gated overload that falls back to the current double-`<` version otherwise. Given `compare_keys` is only on the ordering-operator path (not `==`/`!=`), I'd rather not add that complexity here unless it shows up in profiling.

include/nlohmann/json.hpp
gregmarr: Ah, yes, if `may_descend` is `false` then the whole thing becomes `if (false)`.

include/nlohmann/json.hpp
gregmarr: Sounds good.

gregmarr (approved):

PR #5443, opened 2026-08-28
Support custom object/array types and improve template parameter handling

nlohmann:
## Description

This PR documents the implicit requirements `basic_json` places on its eleven template parameters, fixes the places where those requirements were stricter than they needed to be, and drops the ones the library does not actually need.

### Key Changes

1. **Custom `ObjectType` support**
   - `actual_object_comparator` used `std::conditional`, which eagerly instantiates both branches, so any object type without a `key_compare` member failed to compile. It now uses `detected_or_t`. This was the single blocker for *every* hash map — none of them defines `key_compare`.
   - `erase_from_object()` handles object types whose `erase(iterator)` returns `void` (Abseil, `phmap`, `gtl`, `ankerl::unordered_dense`); the successor iterator is computed before erasing.
   - `erase(key)` is now optional: without it the library falls back to `find(key)` plus `erase(iterator)`.

2. **Custom `ArrayType` support**
   - New `has_capacity` trait and an `array_capacity()` pair of overloads. Array types without `capacity()` (for instance `std::deque`) are handled conservatively under `JSON_DIAGNOSTICS` instead of failing to compile.
   - `at(size_type)` is no longer required: `basic_json::at(size_type)` checks the index itself and uses `operator[]`. This also removes the hidden requirement that the array type throw `std::out_of_range`. The thrown exception, its message, and the behaviour under `JSON_NOEXCEPTION` are unchanged.

3. **Binary type fixes**
   - Explicit casts to `std::uint8_t` when hashing and when writing UBJSON, so binary types whose `value_type` is not an integer (`std::byte`, `char`) work in `dump()`, `std::hash`, and the binary writers.

4. **JSON pointer `unflatten()` fix**
   - `unflatten()` depended on the iteration order of the flattened object, so it produced wrong results with an unordered `ObjectType`. Array parents are now collected in a pre-pass, which keeps the previous semantics exactly and no longer depends on order.

5. **Reduced `string_t` requirements**
   - `c_str()` and `back()` are gone; every call site already knew the length, and the one that did not — the `JSON_DIAGNOSTICS` key path — uses `data()`, which the library requires to be null-terminated anyway.
   - `find(str, pos)`, `replace()`, and `substr()` are gone. RFC 6901 escaping rebuilt the string with one `replace()` per escaped character, which moves the tail every time — O(n²). Both `escape()` and `unescape()` now scan with `find_first_of()` and append whole runs. Escaping 64000 tildes drops from 717 ms to 20 ms; a string with nothing to escape gets faster too (8.4 ms → 5.8 ms).
   - `json_pointer::to_string()` accumulates with `detail::concat<string_t>` instead of letting `concat` default to `std::string` and converting afterwards, so streaming a `json_pointer` no longer requires `string_t` to be assignable from a `std::string` (this also unblocked several third-party string types).
   - The BSON writer writes the terminating null byte itself instead of taking it from the string's buffer.

6. **Documentation**
   - New `features/types/template_parameters.md` covering all eleven parameters: what is always required, what only particular API functions require, which concrete types are compatible, which are not and why. Every claim was checked by compiling and running against Boost 1.83, Abseil 20250127.0, Folly, EASTL 3.21, `ankerl::unordered_dense`, `phmap`, `gtl`, `robin_hood`, `tsl::ordered_map`, and Qt 6 — not by reading headers.
   - The page also lists the six requirement violations that are *not* caught at compile time, and records that an object type is instantiated while `basic_json` is still incomplete — which rules out `std::unordered_map` on libstdc++ 9.
   - `object_order.md` no longer recommends `tsl::ordered_map`, whose iterators expose the mapped value as `const`.
   - `JSON_CATCH_USER` no longer wraps a catch of `std::out_of_range`; its documentation describes what the library actually catches.

### Testing

- `unit-custom-object-type.cpp`: object types without `key_compare` and object types whose `erase(iterator)` returns `void`.
- `unit-custom-array-type.cpp`: `std::deque` as `ArrayType`, and a `std::vector` whose `at()` is hidden.
- `unit-custom-binary-type.cpp`: dump, hash, and the binary formats for binary types over `char` and over `std::byte`.
- `unit-alt-string.cpp`: `alt_string` loses the five dropped members, and gains coverage of the binary formats and of the RFC 6901 escaping paths.
- `unit-json_pointer.cpp`: `unflatten()` independent of key order.
- The escaping rewrite was fuzz-checked against the previous implementation over 400000 random strings.

### Checklist

- [x] Changes are described in detail
- [x] New code paths are covered by unit tests
- [x] Documentation updated with a template parameter guide
- [x] Source code amalgamated (`single_include/nlohmann/json.hpp` updated)

https://claude.ai/code/session_018hxZxz8svM54c6ATEvXp5E


include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1053,8 +1058,9 @@ class binary_writer
 
         write_number<std::int32_t>(to_bson_length(value.size() + 1ul), true);
         oa->write_characters(
-            reinterpret_cast<const CharType*>(value.c_str()),
-            value.size() + 1);
+            reinterpret_cast<const CharType*>(value.data()),
```
gregmarr, line 1061: Copy the comment from above about where the null terminator comes from?

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1145,7 +1151,8 @@ class binary_writer
 
         const std::size_t embedded_document_size = std::accumulate(std::begin(value), std::end(value), static_cast<std::size_t>(0), [&array_index](std::size_t result, const typename BasicJsonType::array_t::value_type & el)
         {
-            return result + calc_bson_element_size(std::to_string(array_index++), el);
+            const auto key = std::to_string(array_index++);
```
gregmarr, line 1159: Copy the comment from below about explicit conversion?

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+
+!!! warning "Requirements are not checked"
+
+    Apart from a `#!cpp static_assert` on the array iterator category, the requirements below are not diagnosed with
+    dedicated error messages. Violating most of them results in a compiler error somewhere inside the library. Six
+    violations are not caught at compile time at all:
+
+    - A [`StringType`](#stringtype) whose `data()` is not null-terminated compiles and silently misparses numbers,
+      because the lexer hands the buffer to `#!cpp std::strtoull`/`#!cpp std::strtoll`/`#!cpp std::strtod`.
+    - A [`BinaryType`](#binarytype) whose `value_type` is wider than one byte compiles and silently produces wrong
```
gregmarr: Could there be a static assert to test this?

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+      results, because the readers and writers reinterpret its storage as raw bytes.
+    - A stateful [`AllocatorType`](#allocatortype) compiles and silently ignores its state: allocation, deallocation,
+      and [`get_allocator()`](../../api/basic_json/get_allocator.md) each use a different default-constructed instance.
+    - A [`NumberUnsignedType`](#numberintegertype-and-numberunsignedtype) too narrow to hold the absolute value of
```
gregmarr: Could there be one of these somewhere?
```
static_assert(sizeof(NumberUnsignedType) == sizeof(NumberIntegerType));
```

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+| [`ObjectType`](#objecttype)                                       | `std::map`                        | [`nlohmann::ordered_map`](../../api/ordered_map.md), Abseil hash maps  |
+| [`ArrayType`](#arraytype)                                         | `std::vector`                     | `#!cpp std::deque`                                                     |
+| [`StringType`](#stringtype)                                       | `std::string`                     | `std::string`-like types over `char`                                   |
+| [`BooleanType`](#booleantype)                                     | `bool`                            | none worth using                                                       |
+| [`NumberIntegerType`](#numberintegertype-and-numberunsignedtype)  | `std::int64_t`                    | any signed integer type                                                |
+| [`NumberUnsignedType`](#numberintegertype-and-numberunsignedtype) | `std::uint64_t`                   | any unsigned integer type                                              |
```
gregmarr: `any unsigned integer type of the same size as NumberIntegerType`?

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+
+The library does not sort or de-duplicate keys itself; the behavior described in
+[`object_t`](../../api/basic_json/object_t.md) is entirely the behavior of the chosen container.
+
+### Compatible containers
+
+| Container                                                                                                     | Notes                                                                    |
+|---------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------|
+| `#!cpp std::map` (default)                                                                                    |                                                                          |
+| [`nlohmann::ordered_map`](../../api/ordered_map.md)                                                           | used by [`ordered_json`](../../api/ordered_json.md); keeps insertion order |
```
gregmarr: Since this table is entirely new, should it use the width of the widest row, which is only 5 spaces wider than the current width?

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+| `robin_hood::unordered_flat_map`                                                                              | adapter must drop the allocator                                          |
+| `folly::F14NodeMap`                                                                                           | through the adapter above; requires C++20, see the note above            |
+| `folly::sorted_vector_map`                                                                                    | alias must drop the allocator, whose value type it disagrees on          |
+
+### Containers that cannot be used
+
+| Container                                                                                                     | Reason                                                                   |
+|---------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------|
+| `absl::btree_map`, `phmap::btree_map`, `gtl::btree_map`, `robin_hood::unordered_node_map`, `folly::F14FastMap`, `eastl::hash_map` | require a complete mapped type                          |
```
gregmarr: multiple rows to keep the width down?

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+| `eastl::map`                                                                                                  | EASTL iterators do not work with `#!cpp std::iterator_traits`            |
+| `tsl::ordered_map`                                                                                            | its iterators expose the mapped value as `#!cpp const`                   |
+| `QMap`                                                                                                        | no `value_type` member type                                              |
+| `QHash`                                                                                                       | its `value_type` is the mapped type rather than a key/value pair, and its iterators dereference to the mapped value |
+| `#!cpp std::multimap`, `#!cpp std::unordered_multimap`                                                         | `emplace` does not return `#!cpp std::pair<iterator, bool>`              |
```
gregmarr: extra space in first column

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+    With [`JSON_DIAGNOSTICS`](../../api/macros/json_diagnostics.md) enabled, the library reads `array_t::capacity()`
+    to find out whether adding an element reallocated the array and moved its elements, which would invalidate the
+    parent pointers. An array type without a `capacity()` member function is handled conservatively: the parent
+    pointers of all elements are refreshed after every insertion, which makes adding *n* elements cost O(*n*²). Only
+    diagnostics builds pay this; without them `capacity()` is never called.
+
+### Compatible containers
+
+| Container                                                                               | Notes                                                    |
+|-------------------------------------------------------------------------------------------|----------------------------------------------------------|
```
gregmarr: two extra dashes in first column

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+| `boost::container::small_vector`, `folly::small_vector`                                 | through an alias that fixes the inline capacity          |
+| `boost::container::static_vector`                                                       | through the same kind of alias, for arrays that stay within the fixed capacity |
+| `folly::fbvector`                                                                       | requires C++20, see the note above                       |
+
+### Containers that cannot be used
+
+| Container                        | Reason                                                                                       |
+|----------------------------------|----------------------------------------------------------------------------------------------|
+| `#!cpp std::list`                | no `operator[]`, and no random-access iterators                                              |
+| `eastl::vector`, `QList`, `QVector` | no `max_size()`; they handle the incomplete value type fine                                |
```
gregmarr: align column widths

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+    parent pointers. An array type without a `capacity()` member function is handled conservatively: the parent
+    pointers of all elements are refreshed after every insertion, which makes adding *n* elements cost O(*n*²). Only
+    diagnostics builds pay this; without them `capacity()` is never called.
+
+### Compatible containers
+
+| Container                                                                               | Notes                                                    |
+|-------------------------------------------------------------------------------------------|----------------------------------------------------------|
+| `#!cpp std::vector` (default)                                                           |                                                          |
+| `#!cpp std::deque`                                                                      | references survive appends, but not insertions elsewhere; see the `capacity()` note above |
```
gregmarr: match all column widths to this one, only a few longer that the next shortest?

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+  `append(first, last)`, and `append(data, size)`.
+- The comparison operator `==` against another `StringType`, and `<` for use as a key of the chosen
+  [`ObjectType`](#objecttype) (with the default comparator, `#!cpp std::less<>` must be able to compare two
+  `StringType` values, and a `StringType` with the key types used for lookup). `!=` is never applied to a
+  `StringType`, and `==` against `#!cpp const char*` is resolved by the implicit `#!cpp const char*` constructor.
+
+### Required for the binary formats
+
+- `resize(n)`, used by the readers to make room for a block of bytes.
+- Non-const `operator[]`, into which the readers `#!cpp std::memcpy` those bytes.
```
gregmarr: Any reason to not use `data() + index` here to avoid this requirement?

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+  `replace(pos, count, const StringType&)` is required.
+- `empty()`.
+- `begin()` and `end()` -- used by
+  [`operator[](const json_pointer&)`](../../api/basic_json/operator%5B%5D.md) to decide whether a reference token
+  denotes an array index.
+
+### Required for other functionality
+
+| Functionality                                                                                                    | Additional requirement                                                              |
+|--------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------|
```
gregmarr: slightly mismatched column widths.

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+| [`to_bson`](../../api/basic_json/to_bson.md)                                                                    | `find(value_type)` and `npos`                                                         |
+| [`parse`](../../api/basic_json/parse.md) from a `string_t`                                                      | the input adapters must accept it; otherwise pass a character range                   |
+| `#!cpp operator<<(std::ostream&, const json_pointer&)`                                                           | streamability to `#!cpp std::ostream`                                                 |
+| exception messages                                                                                               | `data()` and `size()`, or `begin()` and `end()`                                       |
+
+### Compatible types
+
+| Type                                                            | Notes                                                                     |
+|-------------------------------------------------------------------|---------------------------------------------------------------------------|
```
gregmarr: two dashes wider than all first columns

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+| `#!cpp std::string` (default)                                   |                                                                           |
+| `#!cpp std::basic_string` with a custom **stateless** allocator |                                                                           |
+| `#!cpp std::pmr::string`                                        | see the warning below before relying on the memory resource               |
+| `boost::container::string`                                      | needs a user-supplied `#!cpp std::hash` specialization (Boost provides `boost::hash` instead) |
+| `folly::fbstring`                                               | requires C++20, see the note above                                        |
+| `eastl::string`                                                 | needs a user-supplied `#!cpp std::hash` and an ADL `int_to_string` (it is not assignable from a `#!cpp std::string`); [`parse`](../../api/basic_json/parse.md) does not accept it directly -- pass a character range or a `#!cpp std::string` |
+| a custom string class in a user-defined namespace               | if the requirements above are met                                          |
```
gregmarr: extra space in second column

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+
+### Types that cannot be used
+
+| Type                                                                | Reason                                                                |
+|-----------------------------------------------------------------------|-----------------------------------------------------------------------|
```
gregmarr: slightly mismatched column widths.

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+- It must support **incomplete types**: `AllocatorType<basic_json>` is instantiated inside the definition of
+  `basic_json` itself.
+- `#!cpp std::allocator_traits<AllocatorType<basic_json>>::pointer` becomes
+  [`basic_json::pointer`](../../api/basic_json/index.md#container-types), and iterators are constructed from raw
+  `#!cpp basic_json*` values. The `pointer` type must therefore be a plain pointer; fancy pointers are not supported.
+
+### Compatible types
+
+| Type                                                            | Support                                            |
+|-----------------------------------------------------------------|----------------------------------------------------|
```
gregmarr: add one character to first column for all but the last row

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+  The latter form is required for types that are not default-constructible; see
+  [Arbitrary Types Conversions](../arbitrary_types.md).
+- To support the [converting constructor](../../api/basic_json/basic_json.md) between different `basic_json`
+  specializations, `to_json` must be available for `boolean_t`, `number_integer_t`, `number_unsigned_t`,
+  `number_float_t`, `string_t`, `object_t`, `array_t`, and `binary_t` of the *source* specialization.
+
+### Compatible types
+
+| Type                                                              | Support                                                                 |
+|-------------------------------------------------------------------|-------------------------------------------------------------------------|
```
gregmarr: small mismatches in column widths

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+using binary_t = nlohmann::byte_container_with_subtype<BinaryType>;
+```
+
+### Always required
+
+- A non-`final` class type -- [`byte_container_with_subtype`](../../api/byte_container_with_subtype/index.md) derives
+  from it publicly.
+- A member type `value_type` that is **exactly one byte** wide (e.g., `#!cpp std::uint8_t`, `#!cpp char`, or
+  `#!cpp std::byte`). Readers and writers reinterpret the container's storage as raw bytes. A wider `value_type` is
+  **not diagnosed**: it compiles and silently produces wrong results.
```
gregmarr: could this be diagnosed?

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+- Contiguous storage: the binary readers `#!cpp std::memcpy` into `#!cpp &binary[n]`, the writers `reinterpret_cast`
```
gregmarr, line 638: Would it make sense for both to use `data() + n`?

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+
+`max_size()`, `at()`, `reserve()`, `erase()`, `pop_back()`, and `emplace_back()` are **not** used at all.
+
+See [`binary_t`](../../api/basic_json/binary_t.md) for how a non-default `BinaryType` changes the meaning of assigning
+such a container to a `basic_json` value.
+
+### Compatible containers
+
+| Container                                                                                                                  | Notes                                                              |
+|------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------|
```
gregmarr: two extra dashes in first column

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+| `#!cpp std::vector<std::uint8_t>` (default)                                                                                |                                                                    |
+| `#!cpp std::vector<char>`, `#!cpp std::vector<std::byte>`                                                                  | `dump()` writes the bytes as 0..255 whichever is used              |
+| `boost::container::vector<std::uint8_t>`, `boost::container::small_vector<std::uint8_t, N>`                                |                                                                    |
```
gregmarr: significantly wider than widest first column, default second column is only 7 characters narrower than widest.

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+| `absl::InlinedVector<std::uint8_t, N>`                                                                                     | usable here, unlike as an `ArrayType`, because the value type is complete |
+| `eastl::vector<std::uint8_t>`                                                                                              | usable here, unlike as an `ArrayType`, because `max_size()` is not needed |
+| `folly::fbvector<std::uint8_t>`                                                                                            | requires C++20, see the note above                                 |
+
+### Containers that cannot be used
+
+| Container                                            | Reason                                                                                  |
+|--------------------------------------------------------|-----------------------------------------------------------------------------------------|
```
gregmarr: two extra dashes in first column

docs/mkdocs/docs/features/types/template_parameters.md
```diff
@@ -0,0 +1,663 @@
...
+
+Note the namespace of `CustomBaseClass` becomes an associated namespace of `basic_json` for the purpose of
+argument-dependent lookup.
+
+See [`json_base_class_t`](../../api/basic_json/json_base_class_t.md) for an example.
+
+### Compatible types
+
+| Type                                                     | Support                                                      |
+|----------------------------------------------------------|--------------------------------------------------------------|
```
gregmarr: small mismatches in second column widths

docs/mkdocs/docs/features/types/template_parameters.md
nlohmann: The readers write through a single helper, `get_bytes()`, which serves both `StringType` and `BinaryType`. `std::string` has no non-`const` `data()` before C++17, and the library still supports C++11, so `operator[]` has to stay for the string side. The page now says why rather than leaving the requirement unexplained.

---

docs/mkdocs/docs/features/types/template_parameters.md
nlohmann: For `BinaryType` on its own it would work — `std::vector`'s non-`const` `data()` is already C++11. But the read path goes through `get_bytes()`, one helper shared with `StringType`, whose non-`const` `data()` is C++17 and later only. Splitting it in two just to change how the pointer is spelled did not seem worth it, so both stay on `operator[]`; the page now records the reason.

---

gregmarr (approved):

nlohmann:
`ci_test_documentation (ci_test_build_documentation)` is red on `37a3bcb`, and it is not this PR's failure.

`repology.org` is refusing connections from the Actions runners (`[Errno 111] Connection refused` on every badge URL). The `privacy` plugin logs the failed download as a warning but has already registered the asset as a file, so `mkdocs build` crashes later in `copy_static_files`:

```
FileNotFoundError: [Errno 2] No such file or directory:
  '.../.cache/plugin/privacy/assets/external/repology.org/badge/version-for-repo/homebrew/nlohmann-json.svg'
```

The badges live in `docs/mkdocs/docs/integration/package_managers.md`, which this PR does not touch. Controls: #5464, #5466 and #5467 fail the identical job in the same batch today, and #5464 changes no documentation at all. The job was re-run at 16:33Z, four hours after the first failure, and failed identically — so this is not a transient blip that a further re-run would clear.

There is no fix on `develop` to port, and adding one here would widen this PR, so I am standing down on it rather than pushing. Everything else on this head is green: Windows and macOS workflows pass in full, as does the rest of the Ubuntu workflow.

For whenever the docs build is hardened separately — any single unreachable external host currently takes this job down for every open PR — the targeted knob is `assets_exclude` on the `privacy` plugin in `docs/mkdocs/mkdocs.yml`:

```yaml
  - privacy:
      assets_exclude:
        - repology.org/*
```

Verified against mkdocs-material 9.7.7: patterns are matched with `fnmatch` against host + path, and a match returns from `_is_excluded()` before the asset is ever queued or registered as a file, so there is nothing left for `copy_static_files()` to trip over. The trade-off is that those badges would then be fetched by the visitor's browser instead of being inlined, which is a privacy decision rather than a purely mechanical one.

---

PR #5447, opened 2026-08-30
Make std::hash<basic_json> consistent with operator== for numbers

### LINKED ISSUE #5256: Int and uint compare equal but hashes do not

### LINKED ISSUE #5400: std::hash<basic_json> is inconsistent with operator== for equal cross-type numbers (breaks unordered containers)
### Description

`std::hash<basic_json>` can produce **different** hash values for two `json` values that compare **equal** under `operator==`. The C++ standard requires of a `Hash` used with `std::unordered_map`/`unordered_set` that `a == b` implies `hash(a) == hash(b)`; this invariant is violated for numbers of different internal types.

`operator==` converts between `number_integer`, `number_unsigned`, and `number_float` before comparing ([`include/nlohmann/json.hpp`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json.hpp), `JSON_IMPLEMENT_OPERATOR`), so `json(0) == json(0.0) == json(0u)` are all `true`. But `hash()` folds the `value_t` type tag into the result via `combine(type, h)` ([`include/nlohmann/detail/hash.hpp`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/hash.hpp)), giving each of the three a distinct hash.

Consequently a `std::unordered_set<json>` can hold `json(0)`, `json(0.0)`, and `json(0u)` as three separate elements even though they are all equal — silent duplicate keys / lookup misses.

The documentation ([`std_hash.md`](https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/api/basic_json/std_hash.md)) currently describes this as intended ("different hash values for `null`, `0`, `0U`, and `false`"). But `0`, `0U`, and `0.0` are *equal* under `operator==`, so distinguishing them by hash is precisely the contract violation. Only `false` vs `0` is legitimately distinguishable (those are not equal — different `value_t` and no cross-conversion). This looks like a case where the documented intent itself encodes the bug.

### Reproduction steps

Hash `json(0)`, `json(0.0)`, `json(0u)`; observe they are `==` but hash differently and coexist in an `unordered_set`.

- **Expected:** equal values hash equally; `unordered_set` of `{0, 0.0, 0u}` has size 1.
- **Actual:** three distinct hashes; `unordered_set` size is 3.

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <unordered_set>
#include <cstdio>
using json = nlohmann::json;

int main()
{
    json a = 0, b = 0.0, c = 0u;
    std::hash<json> H;
    std::printf("a==b:%d a==c:%d  hashes: %zu %zu %zu\n",
                (int)(a==b), (int)(a==c), H(a), H(b), H(c));
    std::unordered_set<json> s; s.insert(a); s.insert(b); s.insert(c);
    std::printf("unordered_set size (all == 0): %zu (expected 1)\n", s.size());
}
```

### Error messages

```
a==b:1 a==c:1  hashes: 2654436095 2654436221 2654436156
unordered_set size (all == 0): 3 (expected 1)
```

### Discussion / possible fix

Make `hash()` consistent with `operator==` for numbers, e.g. normalize numeric values to a canonical representation before hashing (so that any two numerically-equal values hash the same), or fold a single "number" category into the hash instead of the specific `value_t`. This is a semantics decision for the maintainer; the current behavior breaks the standard `Hash` requirement whenever numeric `json` values are used as keys in unordered containers.

### Library version

### Validation

afonsojanu:
Closes #5400.

`operator==` converts between `number_integer`, `number_unsigned`, and `number_float` before comparing, so `json(0)`, `json(0U)`, and `json(0.0)` all compare equal. `hash()` folded the specific `value_t` into the result for each of the three numeric cases, giving each a distinct hash and breaking the standard `Hash` requirement that `a == b` implies `hash(a) == hash(b)`. A `std::unordered_set` could hold all three as separate elements even though they compare equal, which is exactly the reproduction in the issue.

`hash()` now treats all three numeric variants the same way: it converts the value to `number_float_t` and combines it with a single shared type tag, so any two numbers `operator==` considers equal hash identically regardless of which internal type actually holds them.

I also updated the accompanying test to check this consistency directly (including via an actual `unordered_set`) instead of asserting that `0`, `0U`, and `0.0` hash differently, since that assumption was the bug. The function's own doc comment and the `std::hash` API docs described the old behavior as intended, so I corrected those too, and regenerated the doc example's output since `hash(0)` and `hash(0U)` are no longer different values.

Tested locally by building the amalgamated header and compiling the full test suite (minus the parts that need the separately-downloaded test data submodule, which isn't available in my environment) plus `unit-hash.cpp` in isolation against both the fixed and unfixed code, confirming the new assertions fail against the original `hash()` with the exact numbers from the issue and pass with the fix.


nlohmann (changes requested):
@afonsojanu Please fix the [DCO requirements](https://github.com/nlohmann/json/pull/5447/checks?check_run_id=99297208485). Otherwise good to go.

nlohmann:
This also fixes #5256, which is a duplicate report of the same `std::hash` contract violation (that issue's repro is `json(42) == json(42u)` but with different hashes — verified that this PR's approach makes them hash equal too).

There was a separate attempt at this in #5262, but it turned out to have a correctness bug of its own (hashing large integers in their native domain instead of through `number_float_t` breaks the `std::hash` contract precisely for the numbers where `operator==` relies on the float cast) — details in [this comment](https://github.com/nlohmann/json/pull/5262#issuecomment-5513247719). Recommending we go with this PR instead.

Fixes #5256

nlohmann (changes requested):
The [DCO](https://github.com/nlohmann/json/pull/5447/checks?check_run_id=99297208485) requirements are still unmet. Please check.

PR #5455, opened 2026-09-02
Fix to_bjdata() emitting unparsable output when _ArraySize_ is not an array

### LINKED ISSUE #5399: to_bjdata() emits output it cannot itself parse when _ArraySize_ is null or an object (round-trip guarantee violated)
### Description

When an object carries the three JData ndarray keys and `_ArraySize_` is **not an array** (e.g. `null` or an object), `to_bjdata()` produces a byte stream that `from_bjdata()` then rejects. This violates the documented guarantee ([`bjdata.md:58`](https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/features/binary_formats/bjdata.md)):

> Any BJData output created by `to_bjdata` can be successfully parsed by `from_bjdata`.

It also violates the fallback rule (same doc): a valid ndarray requires that "every entry of `_ArraySize_` is a non-negative integer" — an `_ArraySize_` that is not even a vector fails this and should be "serialized as a regular JSON object."

Root cause: `write_bjdata_ndarray` ([`include/nlohmann/detail/output/binary_writer.hpp`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/output/binary_writer.hpp)) never checks that `_ArraySize_` is an array:

- `std::size_t len = (value.at("_ArraySize_").empty() ? 0 : 1);` — for `null`, `.empty()` is `true`, so `len = 0`.
- The dimension-validation loop `for (const auto& el : value.at("_ArraySize_"))` iterates **zero** times for `null` (and, for an object, iterates its integer *values*, which pass the per-element non-negative-integer check).
- If `_ArrayData_`'s size then matches `len`, the writer emits `[ $ <type> #` followed by `write_ubjson(value.at("_ArraySize_"))` verbatim — which is `Z` for `null` or `{…}` for an object.

The result is a size specification that is not a valid BJData length, so `from_bjdata()` fails with `parse_error.113`.

(This is the same family of writer/reader-consistency gaps addressed for the overflow case in #5362, but the `_ArraySize_`-is-not-an-array case was not covered.)

### Reproduction steps

Serialize an object whose `_ArraySize_` is `null` (or an object) with a matching-size `_ArrayData_`, then feed the bytes back to `from_bjdata()`.

- **Expected:** fall back to a regular object encoding, and the output re-parses via `from_bjdata()`.
- **Actual:** `to_bjdata()` emits `[$U#Z` (for `null`); `from_bjdata()` throws `parse_error.113`.

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <cstdio>
using json = nlohmann::json;

int main()
{
    json j;
    j["_ArrayType_"] = "uint8";
    j["_ArraySize_"] = nullptr;              // or json::object({{"a", 1}, {"b", 2}})
    j["_ArrayData_"] = json::array();
    auto v = json::to_bjdata(j);             // no exception
    for (auto b : v) std::printf("%02x ", b);// 5b 24 55 23 5a  = "[$U#Z"
    std::printf("\n");
    json::from_bjdata(v);                    // throws parse_error.113
}
```

### Error messages

```
5b 24 55 23 5a
terminate called after throwing an instance of 'nlohmann::json_abi_v3_12_0::detail::parse_error'
  what():  [json.exception.parse_error.113] parse error at byte 5: syntax error while parsing BJData size: expected length type specification (U, i, u, I, m, l, M, L) after '#'; last byte: 0x5A
```

### Suggested fix

Require `value.at("_ArraySize_").is_array()` (and `return true` for the plain-object fallback otherwise) before entering the dimension loop. As a related observation in the same area, an object-valued `_ArrayData_` whose element count matches `len` is also serialized as raw ndarray data and silently loses its structure (e.g. `{"_ArrayData_":{"a":1,"b":2}, …}` round-trips to `[1,2]`) — an `is_array()` check on `_ArrayData_` would close that too.

### Library version

### Validation

qatcod:
Fixes #5399.

`write_bjdata_ndarray()` never checked that `_ArraySize_` is an array. Since the shape is written verbatim as the header length, a shape that is not an array cannot produce a valid one:

```cpp
json j_null{{"_ArrayType_", "uint8"}, {"_ArraySize_", nullptr}, {"_ArrayData_", json::array()}};
json::from_bjdata(json::to_bjdata(j_null));
// [json.exception.parse_error.113] parse error at byte 5: syntax error while parsing
// BJData size: expected length type specification (U, i, u, I, m, l, M, L) after '#';
// last byte: 0x5A
```

`0x5A` is `Z`, the null marker. An object shape fails the same way with `0x7B`, `{`. That breaks the guarantee in `bjdata.md` that any output of `to_bjdata` parses with `from_bjdata`.

Both cases got past the existing per-dimension validation for the reasons given in the issue: for `null`, `empty()` is `true` so the count starts at 0 and the loop body never runs, and for an object the loop iterates the values rather than dimensions, which can satisfy the non-negative integer check. Once `_ArrayData_.size()` matched that count, the writer committed to the ndarray path.

Requiring the shape to be an array makes both fall back to a plain object, which is what the fallback rule in the same doc already calls for, and they then round-trip exactly. A scalar shape is covered too, since it has the same problem.

A well-formed ndarray is unaffected: `{"_ArrayType_":"uint8","_ArraySize_":[2],"_ArrayData_":[1,2]}` still encodes to the same 10 byte optimized form.

Tests go in the section next to the overflow cases from #5362, since this is the same writer/reader consistency family. Reverting the writer change with the tests in place fails on the null case with the `parse_error.113` above. `unit-bjdata`, `unit-ubjson` and `unit-binary_formats` all pass, and `make check-amalgamation` is clean.

### DIFF
--- a/docs/mkdocs/docs/features/binary_formats/bjdata.md
+++ b/docs/mkdocs/docs/features/binary_formats/bjdata.md
@@ -125,6 +125,7 @@ The library uses the following mapping from JSON values types to BJData types ac
 
     - `"_ArrayType_"` is one of `uint8`, `int8`, `uint16`, `int16`, `uint32`, `int32`, `uint64`, `int64`, `single`,
       `double`, `char`, or `byte`,
+    - `"_ArraySize_"` is an array, since the dimensions are written as the ND-array header's length,
     - every entry of `"_ArraySize_"` is a non-negative integer, and their product is representable as a `std::size_t`,
     - `"_ArrayData_"` holds exactly that many elements, and
     - every element of `"_ArrayData_"` is a number of the kind named by `"_ArrayType_"` (a floating-point number for
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -1668,6 +1668,15 @@ class binary_writer
         CharType dtype = it->second;
 
         key = "_ArraySize_";
+        // the dimensions are written verbatim as the header length below, so a
+        // value that is not an array cannot produce a valid one: null emits 'Z'
+        // and an object emits '{', neither of which a reader accepts after '#'.
+        // Such an object is not a valid ndarray and falls back to a plain object.
+        if (!value.at(key).is_array())
+        {
+            return true;
+        }
+
         std::size_t len = (value.at(key).empty() ? 0 : 1);
         for (const auto& el : value.at(key))
         {
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -18611,6 +18611,15 @@ class binary_writer
         CharType dtype = it->second;
 
         key = "_ArraySize_";
+        // the dimensions are written verbatim as the header length below, so a
+        // value that is not an array cannot produce a valid one: null emits 'Z'
+        // and an object emits '{', neither of which a reader accepts after '#'.
+        // Such an object is not a valid ndarray and falls back to a plain object.
+        if (!value.at(key).is_array())
+        {
+            return true;
+        }
+
         std::size_t len = (value.at(key).empty() ? 0 : 1);
         for (const auto& el : value.at(key))
         {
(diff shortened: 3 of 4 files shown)


nlohmann (changes requested):
Please fix the DCO requirements.

Also: [bjdata.md:124](https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/features/binary_formats/bjdata.md#L124) lists the ndarray requirements as "all of the following" (4 bullets) but doesn't mention the new requirement that `_ArraySize_` itself must be an array - only implies it via "every entry of." Worth adding a 5th bullet so the doc stays the authoritative, exhaustive list the PR's own description assumes it already is.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

PR #5459, opened 2026-09-02
Compare integers with floats exactly instead of widening the integer

### LINKED ISSUE #5401: Mixed integer/float comparison near 2^63 is intransitive, violating strict-weak-ordering (std::sort/std::set on such values is UB)
### Description

When one operand of a comparison is an integer and the other a float, the integer is cast to `number_float_t` (double) before comparing (`JSON_IMPLEMENT_OPERATOR` in [`include/nlohmann/json.hpp`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json.hpp), used by `operator==`, `operator<`, and `operator<=>`). Near 2^63 this cast is lossy, which makes equality **intransitive** and the ordering **not a strict weak ordering**. Passing such values to `std::sort`, or using them as keys in `std::set`/`std::map`, is therefore undefined behavior.

Both `9223372036854775806` (2^63−2) and `9223372036854775807` (2^63−1) round to the same double `9223372036854775808.0` (2^63). With `x = int64(2^63−2)`, `y = double(2^63)`, `z = int64(2^63−1)`:

- `x == y` is `true` and `y == z` is `true`, but `x == z` is `false` (both integers, `2^63−2 ≠ 2^63−1`) → equality is intransitive.
- `equiv(x,y)` and `equiv(y,z)` are both `true` (where `equiv(p,q) = !(p<q) && !(q<p)`), yet `x < z` is `true`, so `equiv(x,z)` is `false` → transitivity of incomparability fails, which is exactly the property `std::sort`/`std::set` rely on.

The docs for the comparison operators only document NaN as unordered; the finite-number precision-loss inconsistency is not documented (they merely say numbers "are automatically converted before comparison"). This is distinct from the already-fixed unsigned-vs-signed integer case in #5210 — here the loss comes from the integer→`double` conversion.

### Reproduction steps

Compare the three values above; observe the intransitive equality and the SWO violation.

- **Expected:** a consistent equivalence relation and a valid strict weak ordering (e.g. exact integer/float comparison, as `operator<=>` implementations for mixed int/float types can do without precision loss).
- **Actual:** `x==y` and `y==z` but `x!=z`; `equiv(x,y)` and `equiv(y,z)` but `x<z`.

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <cstdint>
#include <cstdio>
using json = nlohmann::json;

int main()
{
    json x = std::int64_t(9223372036854775806LL); // 2^63-2
    json y = 9223372036854775808.0;                // double 2^63
    json z = std::int64_t(9223372036854775807LL);  // 2^63-1
    auto eq = [](const json& p, const json& q){ return !(p<q) && !(q<p); };
    std::printf("== x==y:%d y==z:%d x==z:%d | SWO eq(x,y):%d eq(y,z):%d eq(x,z):%d (x<z):%d\n",
                (int)(x==y),(int)(y==z),(int)(x==z),
                (int)eq(x,y),(int)eq(y,z),(int)eq(x,z),(int)(x<z));
}
```

### Error messages

```
== x==y:1 y==z:1 x==z:0 | SWO eq(x,y):1 eq(y,z):1 eq(x,z):0 (x<z):1
```

(Concrete transitivity-violating triple; sorting an array containing such a mix is UB.)

### Library version

### Validation

qatcod:
Fixes #5401.

The mixed number arms of `JSON_IMPLEMENT_OPERATOR` widen the integer to `number_float_t` before comparing, which is lossy past the mantissa. Reproducing the triple from the issue:

```
== x==y:1 y==z:1 x==z:0 | SWO eq(x,y):1 eq(y,z):1 eq(x,z):0 (x<z):1
```

Comparing exactly instead is enough to fix both the equality and the ordering. The integer's range is a power of two that the float represents exactly, so a float outside it is ordered by magnitude alone; inside it, truncating the float is exact, and then the integer parts and any fractional part decide.

The helper returns a pair whose comparison with the original operator reproduces that ordering, rather than an ordering value directly. That keeps each operator's return type exactly as it was, including `partial_ordering` for `<=>`, and it gives NaN somewhere to go: a NaN operand comes back in both members, so it stays false for the relational operators and unordered for the spaceship, which is what the float comparison did before.

### Checking

The ordering axioms, over a set concentrated on the lossy boundary and checked exhaustively rather than sampled:

| | transitivity of `<` | transitivity of equivalence |
|---|---|---|
| before | 0 | **30** |
| after | 0 | 0 |

(24 values, 13824 triples. Sampling triples at random is not much use here, the failing combination is too rare to land on.)

The comparison itself was then differentially tested against Python, which compares int and float exactly: 20221 signed and 20108 unsigned pairs, covering the boundaries at 2^53, 2^63 and 2^64, both infinities, and fractional values. No mismatches, first against the algorithm on its own and then through the library's operators.

Behaviour that should not move, and does not: `json(1) == json(1.0)`, `json(1u) == json(1.0)`, `json(-3) == json(-3.0)`, fractional ordering either way round, and NaN unordered against both integer kinds. The new section in `unit-comparison.cpp` sits next to the existing signed/unsigned one since it is the same family of problem; reverting the two headers with it in place fails 8 of its assertions.

Full suite passes, 91 of 91, and `make check-amalgamation` is clean.

### DIFF
--- a/include/nlohmann/detail/value_t.hpp
+++ b/include/nlohmann/detail/value_t.hpp
@@ -9,9 +9,12 @@
 #pragma once
 
 #include <array> // array
+#include <cmath> // isnan, ldexp, trunc
 #include <cstddef> // size_t
 #include <cstdint> // uint8_t
+#include <limits> // numeric_limits
 #include <string> // string
+#include <type_traits> // is_signed
 
 #include <nlohmann/detail/macro_scope.hpp>
 #if JSON_HAS_THREE_WAY_COMPARISON
@@ -113,6 +116,68 @@ inline bool operator<(const value_t lhs, const value_t rhs) noexcept
     return std::is_lt(lhs <=> rhs); // *NOPAD*
 }
 #endif
+
+
+/*!
+@brief compare an integer with a floating point number without precision loss
+
+Widening the integer to the floating point type loses precision beyond the
+float's mantissa, which makes equality intransitive: both 2^63-2 and 2^63-1
+round to 2^63, so each compares equal to that float while differing from each
+other. Ordering built on that is not a strict weak ordering, so sorting such
+values, or using them as keys in an ordered container, is undefined behavior.
+
+Returns a value to be compared against zero with the original operator, which
+reproduces the exact ordering. A NaN operand is returned as is, so comparing it
+against zero keeps NaN's semantics: false for the relational operators and
+unordered for `<=>`.
+*/
+template<typename IntegerType, typename FloatType>
+FloatType compare_integer_with_float(const IntegerType i, const FloatType f) noexcept
+{
+    const auto ordered = [](int c) noexcept
+    {
+        return static_cast<FloatType>(c);
+    };
+
+    if (std::isnan(f))
+    {
+        return f;
+    }
+
+    // values of IntegerType lie in [-bound, bound) when signed and in
+    // [0, bound) when unsigned; digits excludes the sign bit, so bound is a
+    // power of two that the float represents exactly
+    const FloatType bound = std::ldexp(static_cast<FloatType>(1), std::numeric_limits<IntegerType>::digits);
+    if (f >= bound)
+    {
+        return ordered(-1);
+    }
+    if (std::is_signed<IntegerType>::value ? (f < -bound) : (f < static_cast<FloatType>(0)))
+    {
+        return ordered(1);
+    }
+
+    // f is now within the integer's range, so truncating it is exact
+    const FloatType truncated = std::trunc(f);
+    const auto as_integer = static_cast<IntegerType>(truncated);
+    if (i != as_integer)
+    {
+        return ordered(i < as_integer ? -1 : 1);
+    }
+
+    // the integer parts agree, so any fractional part decides
+    const FloatType fraction = f - truncated;
+    if (fraction > static_cast<FloatType>(0))
+    {
+        return ordered(-1);
+    }
+    if (fraction < static_cast<FloatType>(0))
+    {
+        return ordered(1);
+    }
+    return ordered(0);
+}
 
 }  // namespace detail
 NLOHMANN_JSON_NAMESPACE_END
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -3703,19 +3703,19 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     }                                                                                                    \
     else if (lhs_type == value_t::number_integer && rhs_type == value_t::number_float)                   \
     {                                                                                                    \
-        return static_cast<number_float_t>(lhs.m_data.m_value.number_integer) op rhs.m_data.m_value.number_float;      \
+        return (detail::compare_integer_with_float(lhs.m_data.m_value.number_integer, rhs.m_data.m_value.number_float)) op (static_cast<number_float_t>(0)); \
     }                                                                                                    \
     else if (lhs_type == value_t::number_float && rhs_type == value_t::number_integer)                   \
     {                                                                                                    \
-        return lhs.m_data.m_value.number_float op static_cast<number_float_t>(rhs.m_data.m_value.number_integer);      \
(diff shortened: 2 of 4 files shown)


nlohmann (changes requested):
Please fix the [DCO](https://github.com/nlohmann/json/pull/5459/checks?check_run_id=100442377936) requirements.

nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

include/nlohmann/detail/value_t.hpp
```diff
@@ -114,5 +117,59 @@ inline bool operator<(const value_t lhs, const value_t rhs) noexcept
...
+    const FloatType truncated = std::trunc(f);
+    const auto as_integer = static_cast<IntegerType>(truncated);
+    if (i != as_integer)
+    {
+        return ordered(i < as_integer ? -1 : 1);
+    }
+
+    // the integer parts agree, so any fractional part decides
+    const FloatType fraction = f - truncated;
+    return ordered(fraction > static_cast<FloatType>(0) ? -1 : (fraction < static_cast<FloatType>(0) ? 1 : 0));
```
gregmarr: `clang_tidy` doesn't like this line.

value_t.hpp:171:65: error: conditional operator is used as sub-expression of parent conditional operator, refrain from using nested conditional operators [readability-avoid-nested-conditional-operator,-warnings-as-errors]
  171 |     return ordered(fraction > static_cast<FloatType>(0) ? -1 : (fraction < static_cast<FloatType>(0) ? 1 : 0));
      |                                                                 ^

include/nlohmann/json.hpp
```diff
@@ -3703,19 +3703,19 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
     }                                                                                                    \
     else if (lhs_type == value_t::number_integer && rhs_type == value_t::number_float)                   \
     {                                                                                                    \
-        return static_cast<number_float_t>(lhs.m_data.m_value.number_integer) op rhs.m_data.m_value.number_float;      \
+        return detail::compare_integer_with_float(lhs.m_data.m_value.number_integer, rhs.m_data.m_value.number_float) op static_cast<number_float_t>(0); \
```
gregmarr: /json.hpp:3706:119: error: macro argument should be enclosed in parentheses [bugprone-macro-parentheses,-warnings-as-errors]
 3706 |         return detail::compare_integer_with_float(lhs.m_data.m_value.number_integer, rhs.m_data.m_value.number_float) op static_cast<number_float_t>(0); \
      |                                                                                                                       ^
      |                                                                                                                       ( )

include/nlohmann/json.hpp
```diff
@@ -3703,19 +3703,19 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
     else if (lhs_type == value_t::number_float && rhs_type == value_t::number_integer)                   \
     {                                                                                                    \
-        return lhs.m_data.m_value.number_float op static_cast<number_float_t>(rhs.m_data.m_value.number_integer);      \
+        return static_cast<number_float_t>(0) op detail::compare_integer_with_float(rhs.m_data.m_value.number_integer, lhs.m_data.m_value.number_float); \
     }                                                                                                    \
     else if (lhs_type == value_t::number_unsigned && rhs_type == value_t::number_float)                  \
     {                                                                                                    \
-        return static_cast<number_float_t>(lhs.m_data.m_value.number_unsigned) op rhs.m_data.m_value.number_float;     \
+        return detail::compare_integer_with_float(lhs.m_data.m_value.number_unsigned, rhs.m_data.m_value.number_float) op static_cast<number_float_t>(0); \
```
gregmarr: /__w/json/json/include/nlohmann/json.hpp:3714:120: error: macro argument should be enclosed in parentheses [bugprone-macro-parentheses,-warnings-as-errors]
 3714 |         return detail::compare_integer_with_float(lhs.m_data.m_value.number_unsigned, rhs.m_data.m_value.number_float) op static_cast<number_float_t>(0); \

gregmarr:
@nlohmann There are workflows awaiting approval to run.

nlohmann:
Thanks!

PR #5461, opened 2026-09-04
Make diff() linear when an array shrinks

### LINKED ISSUE #5460: diff() is quadratic when an array shrinks, although it is documented as linear
### Description

`basic_json::diff()` is documented as *"Linear in the lengths of `source` and `target`"*
([diff.md](https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/api/basic_json/diff.md#complexity)).
It is linear when arrays grow or when elements are replaced, but when the source array is longer
than the target one it is quadratic in the number of removed elements.

The trailing `remove` operations have to be emitted highest index first. They are produced by
repeatedly inserting at the same position in the result array
([`json.hpp:5132`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json.hpp)):

```cpp
const auto end_index = static_cast<difference_type>(result.size());
while (i < source.size())
{
    // add operations in reverse order to avoid invalid indices
    result.insert(result.begin() + end_index, object(
    {
        {"op", "remove"},
        {"path", detail::concat<string_t>(path, '/', detail::to_string<string_t>(i))}
    }));
    ++i;
}
```

Every insert shifts all the operations already emitted at that position, so emitting *k* removals
costs O(k²) moves of `basic_json` values.

### Reproduction steps

Diff an array against a shorter one and vary the length.

- **Expected:** time grows linearly with the number of removed elements, as documented.
- **Actual:** time quadruples whenever the number of removed elements doubles.

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <chrono>
#include <cstdio>
using json = nlohmann::json;

int main()
{
    for (std::size_t n = 8000; n <= 64000; n *= 2)
    {
        json big = json::array();
        for (std::size_t i = 0; i < n; ++i) big.push_back(static_cast<int>(i));

        const auto t0 = std::chrono::steady_clock::now();
        const auto shrink = json::diff(big, json::array());          // n removals
        const auto t1 = std::chrono::steady_clock::now();
        const auto grow = json::diff(json::array(), big);            // n additions
        const auto t2 = std::chrono::steady_clock::now();

        std::printf("n=%6zu  shrink %8.1f ms   grow %6.1f ms\n", n,
                    std::chrono::duration<double, std::milli>(t1 - t0).count(),
                    std::chrono::duration<double, std::milli>(t2 - t1).count());
    }
}
```

### Measurement

Apple clang 17, `-O2`, macOS arm64:

| n | `diff(big, [])` (n removals) | `diff([], big)` (n additions) |
|---|---|---|
| 8,000 | 211 ms | 4.5 ms |
| 16,000 | 849 ms | 9.2 ms |
| 32,000 | 3,325 ms | 18.3 ms |
| 64,000 | **13,356 ms** | 36.9 ms |

Removals quadruple per doubling; additions and element-wise replacements are linear and unaffected.

A more ordinary shape, an array of 20,000 small objects with half of them deleted, takes 312 ms
where the same diff with the removals appended costs 6.8 ms.

### Suggested fix

Append the removals in descending index order instead of inserting each one at a fixed position.
The `add` branch that follows only runs when the target is *longer*, so the two are mutually
exclusive and appending lands at exactly the position the inserts targeted. The emitted patch is
unchanged.

I have this prepared with a regression test if it is wanted.

Apple clang 17.0.0, macOS 15 (arm64)

### Library version

`develop` @ `19386dd1`

### Validation

avionicharshit-byte:
Closes #5460.

`basic_json::diff()` is documented as linear in the lengths of `source` and `target`, but when the
source array is longer than the target it is quadratic in the number of removed elements.

The trailing `remove` operations must be emitted highest index first, and that ordering was
produced by inserting each one at the same position in the result array. Every insert shifts all
the operations already emitted at that position, so emitting *k* removals costs O(k²) moves of
`basic_json` values.

This appends them in descending index order instead. The `add` branch that follows only runs when
the target is longer, so the two loops are mutually exclusive and appending lands at exactly the
position the inserts targeted. **The emitted patch is unchanged.**

### Effect

Apple clang 17, `-O2`, macOS arm64, `diff(array_of_n, json::array())`:

| n | before | after |
|---|---|---|
| 8,000 | 211 ms | 5.4 ms |
| 16,000 | 849 ms | 7.7 ms |
| 32,000 | 3,325 ms | 12.5 ms |
| 64,000 | 13,356 ms | 25.4 ms |

An array of 20,000 small objects with half of them deleted goes from 312 ms to 6.8 ms. Growing and
element-wise diffs were already linear and are untouched.

### Verifying that the output does not change

- 50,000 random source/target pairs (nulls, booleans, numbers, strings, nested arrays and objects,
  keys containing `/` and `~`): the dumped patch is **byte-identical** before and after, and
  `source.patch(diff(source, target)) == target` holds in both.
- The [json-patch-tests](https://github.com/json-patch/json-patch-tests) suite still passes 108/108.
- The existing unit tests are unchanged.

### Tests

`tests/src/unit-json_patch.cpp` gains a test case pinning the order and content of the emitted
removals: an array shrunk to empty, a partial shrink following a replacement at a common index, a
nested array, and a 1,000 element shrink that must still round-trip.

- [x] The code coverage remained at 100%. A test case for every new line of code.
- [ ] If applicable, the documentation is updated. (No change: the documented complexity is what
      this restores.)
- [x] The source code is amalgamated.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -5128,19 +5128,17 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
                 // We now reached the end of at least one array
                 // in a second pass, traverse the remaining elements
 
-                // remove my remaining elements
-                const auto end_index = static_cast<difference_type>(result.size());
-                while (i < source.size())
+                // remove my remaining elements, highest index first; appending
+                // in that order avoids the quadratic reinsertion done before
+                for (std::size_t j = source.size(); j > i; --j)
                 {
-                    // add operations in reverse order to avoid invalid
-                    // indices
-                    result.insert(result.begin() + end_index, object(
+                    result.push_back(object(
                     {
                         {"op", "remove"},
-                        {"path", detail::concat<string_t>(path, '/', detail::to_string<string_t>(i))}
+                        {"path", detail::concat<string_t>(path, '/', detail::to_string<string_t>(j - 1))}
                     }));
-                    ++i;
                 }
+                i = source.size();
 
                 // add other remaining elements
                 while (i < target.size())
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -26482,19 +26482,17 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
                 // We now reached the end of at least one array
                 // in a second pass, traverse the remaining elements
 
-                // remove my remaining elements
-                const auto end_index = static_cast<difference_type>(result.size());
-                while (i < source.size())
+                // remove my remaining elements, highest index first; appending
+                // in that order avoids the quadratic reinsertion done before
+                for (std::size_t j = source.size(); j > i; --j)
                 {
-                    // add operations in reverse order to avoid invalid
-                    // indices
-                    result.insert(result.begin() + end_index, object(
+                    result.push_back(object(
                     {
                         {"op", "remove"},
-                        {"path", detail::concat<string_t>(path, '/', detail::to_string<string_t>(i))}
+                        {"path", detail::concat<string_t>(path, '/', detail::to_string<string_t>(j - 1))}
                     }));
-                    ++i;
                 }
+                i = source.size();
 
                 // add other remaining elements
                 while (i < target.size())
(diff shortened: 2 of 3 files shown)


nlohmann (approved):
Looks good to me.

nlohmann:
Thanks!

Issue #5422, opened 2026-08-26
Missing test coverage: lenient BSON input handling is documented but not pinned by tests


nlohmann:
## Description

#5333 documented three lenient behaviors of the BSON reader. None of them is asserted in `tests/src/unit-bson.cpp`, so the documented behavior is unpinned and could regress silently:

1. **"any non-zero byte is accepted as `true`"** — `binary_reader.hpp:350` does `sax->boolean(value != 0)`, but every `0x08` boolean in the test file uses payload `0x00` or `0x01` (`unit-bson.cpp:158, 182, 797, 815, 868, 990`). A payload like `0x02` → `true` is never tested.
2. **"array element keys are not checked against the required decimal sequence"** — `parse_bson_element_list(/*is_array*/true)` drops the key (`binary_reader.hpp:416`). No test feeds an array document with keys like `"5"`, `"x"`, or out-of-order indices.
3. **"the payload for binary subtype `0x02` is returned as-is, including its inner length prefix"** — only subtypes `0x00` and `0x05` appear in tests (`unit-bson.cpp:518, 550, 583, 931`). The "old binary" subtype `0x02` with its nested int32 length is never parsed.

Three small byte-vector test sections would pin all of this.

Related thin spot found while auditing error codes: `out_of_range.412` has a single throw site (`binary_writer.hpp:1000`) that is a shared size-check helper for BSON *binary*, *string*, and *document* lengths, but only the binary path is tested (`unit-bson.cpp:111`); the string/document overflow paths are untested.

PR #5462, opened 2026-09-04
test: cover lenient BSON input behavior

tvost2:
## Summary\n\nAdds regression coverage for the documented lenient BSON reader behavior from #5422:\n\n- non-zero boolean payloads are accepted as true\n- array element keys are not validated as decimal indices\n- old binary subtype 0x02 preserves its inner length prefix\n\n## Validation\n\n- 	est-bson_cpp11.exe: BSON test case, 79 assertions passed\n- External test data download was unavailable in this environment; file-backed BSON cases were excluded from the focused run.\n- git diff --check\n\nCommit: 1de43f418

### DIFF
--- a/tests/src/unit-bson.cpp
+++ b/tests/src/unit-bson.cpp
@@ -125,6 +125,45 @@ TEST_CASE("BSON")
         CHECK_THROWS_WITH_AS(_ = json::from_bson(v), "[json.exception.parse_error.112] parse error at byte 10: syntax error while parsing BSON string: string length must be at least 1, is -2147483648", json::parse_error&);
     }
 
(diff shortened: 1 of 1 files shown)


nlohmann:
Thanks for this — good coverage of the documented lenient BSON behaviors (non-zero-byte booleans, unchecked array element keys, and the subtype 0x02 raw-payload/inner-length-prefix case).

We landed equivalent coverage ourselves in #5478 (which also adds a couple of extra out_of_range.412 overflow tests for BSON strings and embedded documents), so we're moving forward with that PR and closing this one to avoid duplicate coverage. Appreciate you pinning this down.

— closed by Claude Code on behalf of @nlohmann

PR #5469, opened 2026-09-05
Reject MessagePack/BSON binary subtypes that don't fit their wire format

nlohmann:
## Summary

`byte_container_with_subtype::subtype()` is a `std::uint64_t`, but the MessagePack ext-type byte and the BSON binary subtype byte are each a single byte on the wire. Both writers (`include/nlohmann/detail/output/binary_writer.hpp`) cast the subtype down to a single byte without any range check, silently wrapping modulo 256 instead of erroring — a silent data-corruption bug:

```cpp
json::from_msgpack(json::to_msgpack(json::binary({1, 2}, 256))).get_binary().subtype();    // 0, not 256
json::from_msgpack(json::to_msgpack(json::binary({1, 2}, 70000))).get_binary().subtype();  // 112, not 70000
```

The valid range for both formats is actually the *full* 0–255 (a full unsigned byte), not just 0–127 — the MessagePack reader already round-trips 128–255 correctly via a signed/unsigned reinterpretation of the ext-type byte, and the BSON reader reads the subtype directly as `uint8_t`. Only subtypes ≥ 256 are affected. The CBOR writer already handles the full 64-bit range correctly (choosing 1/2/4/8-byte tag encodings), and the BSON writer already throws `out_of_range.407`-adjacent errors for other values it can't represent, so a check here is consistent with existing behavior.

## Fix

Add a range check (`subtype() > 255`) before each of the two truncating casts, throwing a new `out_of_range.413` (the next free out_of_range code after the existing `.412`) instead of silently truncating. A binary value with no subtype at all is unaffected.

## Test plan

- [x] Added `tests/src/unit-regression2.cpp` coverage: MessagePack/BSON subtype 0/200/255 still round-trip correctly (regression guard for the now-confirmed-correct 128–255 range); subtype 256/70000/300 now throw `out_of_range.413` with the exact expected message; CBOR (already correct) remains unaffected at subtype 70000; a binary value with no subtype at all is unaffected.
- [x] Documented the new exception code in `docs/mkdocs/docs/home/exceptions.md` (`json.exception.out_of_range.413`), matching the existing style.
- [x] Regenerated `single_include/nlohmann/json.hpp` via `make amalgamate`.

## Breaking change?

No breaking changes to the public API — this adds a new exception code (`out_of_range.413`) and only affects inputs that previously silently corrupted data (a subtype ≥ 256 being written and read back as a different, wrong value). No valid/previously-correct round-trip changes behavior; serializing a binary value whose subtype is ≥ 256 now throws instead of silently producing a wrong result.


docs/mkdocs/docs/api/basic_json/insert.md
gregmarr, line 91: Seems like this is in the wrong PR.  There is a lot of stuff here from other PRs.

nlohmann: This PR is stacked on #5468 (and further back, #5464-#5467) via `gh stack` — GitHub shows the cumulative unmerged diff for a PR whose base branch is itself an open PR, so `insert.md` (from #5468) and the other unrelated-looking files show up here until the earlier links in the chain merge. This PR's own commits only touch `binary_writer.hpp`, `exceptions.md`, and the msgpack/bson tests — nothing in `insert.md`. Once #5464-#5468 land, this diff will shrink to just those.

gregmarr: Odd, I've never run into that behavior before.

nlohmann: You were right and I take that back — I checked more carefully. It wasn't normal stacked-PR behavior: this branch had diverged from #5468 (12 commits ahead, 11 behind its own base per GitHub's compare), because #5468 had been force-pushed/updated several times after this branch was created on top of an older version of it. That divergence is exactly why GitHub's Files changed tab pulled in unrelated content like insert.md. The rest of the stack (#5464-#5468) was internally consistent — only this one link was stale. Rebased this branch's 2 real commits onto the current tip of #5468 and force-pushed; the diff is now just the 5 files this PR actually touches.

gregmarr (approved):

gregmarr (approved):

PR #5472, opened 2026-09-05
Reserve capacity in from_json() object conversion when the target container supports it

### LINKED ISSUE #5406: from_json into a std::unordered_map (and other reservable object types) never reserves, causing rehashing (~13% slower get<unordered_map>)
### Description

The object→C++ container conversion in `from_json` fills the target one element at a time with `emplace` and never reserves capacity, even when the target container supports `reserve()` and the source size is known:

[`include/nlohmann/detail/conversions/from_json.hpp:401-417`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/conversions/from_json.hpp):

```cpp
template<typename BasicJsonType, typename ConstructibleObjectType, ...>
inline void from_json(const BasicJsonType& j, ConstructibleObjectType& obj)
{
    if (JSON_HEDLEY_UNLIKELY(!j.is_object())) { JSON_THROW(...); }

    ConstructibleObjectType ret;
    const auto* inner_object = j.template get_ptr<const typename BasicJsonType::object_t*>();
    for (const auto& p : *inner_object)
    {
        ret.emplace(p.first, p.second.template get<typename ConstructibleObjectType::mapped_type>());
    }
    obj = std::move(ret);
}
```

For a `std::unordered_map` target this rehashes O(log n) times as it grows past successive load-factor thresholds. The **array** path already handles this correctly — `from_json_array_impl` (priority_tag&lt;1&gt;, `from_json.hpp:289`) does `ret.reserve(j.size())` via a `reserve`-detecting SFINAE overload. The object path has no equivalent.

### Measurement

`get<std::unordered_map<std::string,int>>()` on a 5,000-entry object, g++ 13.3.0 `-O2`, 200 iterations averaged:

| path | time | allocations |
|---|---|---|
| library `get<>` (no reserve) | 0.646 ms/call | 5009 |
| same fill **with** `reserve(j.size())` first | 0.572 ms/call | 5001 |

**~13% faster**; the 8 saved allocations are the bucket-array reallocations from rehashing (the per-node allocations, 5000, are inherent to `unordered_map`). The default `object_t` is `std::map`, which has no `reserve`, so only `unordered_map`-like targets benefit — but those are common conversion targets.

### Minimal code example

```cpp
#include <new>
#include <cstdlib>
#include <cstdio>
static long allocs = 0;
void* operator new(std::size_t n){ ++allocs; void* p=std::malloc(n); if(!p)throw std::bad_alloc(); return p; }

#include <nlohmann/json.hpp>
#include <unordered_map>
using json = nlohmann::json;

int main()
{
    json j = json::object();
    for (int i = 0; i < 5000; ++i) { char k[16]; std::snprintf(k, sizeof k, "k%06d", i); j[k] = i; }
    allocs = 0;
    auto m = j.get<std::unordered_map<std::string,int>>();
    std::printf("%ld allocs (reserve would save the rehash allocations)\n", allocs);
}
```

### Suggested fix

Add a `reserve`-detecting overload for the object path, mirroring the array path's `priority_tag` SFINAE: when `ConstructibleObjectType` has a `reserve(size_type)` member, call `ret.reserve(j.size())` before the emplace loop; otherwise fall back to the current loop unchanged. Behavior-preserving (same resulting map). Note the source size is always known here (`inner_object->size()`), so unlike input-iterator ranges there is no case where the size is unavailable.

### Library version

### Validation

nlohmann:
## Summary

`from_json()` for object-like C++ container targets (e.g. `std::map`, `std::unordered_map`, `std::multimap`, ...) filled the target container one element at a time via `emplace()`, without ever calling `reserve()` — even though the number of source elements is always known ahead of time (`inner_object->size()`), and even for container types that support `reserve(size_type)` (such as `std::unordered_map`). This causes avoidable rehashing while deserializing large objects into `std::unordered_map`.

The array conversion path (`from_json_array_impl` in `include/nlohmann/detail/conversions/from_json.hpp`) already solves the analogous problem for arrays using a `priority_tag`-based SFINAE dispatch: a higher-priority overload is selected when the target type has a usable `reserve()`, and a lower-priority fallback overload is used otherwise.

This PR mirrors that exact technique for the object conversion path:
- Added `from_json_object_impl(..., priority_tag<1>)`, enabled via expression SFINAE on `obj.reserve(size_type)`, which reserves `inner_object->size()` before the emplace loop.
- Added `from_json_object_impl(..., priority_tag<0>)` as the fallback, containing the original (unreserved) loop, used for object types without `reserve()` (e.g. `std::map`).
- `from_json()` for object-like containers now just dispatches to `from_json_object_impl(j, obj, priority_tag<1>{})`.

## Breaking change?

No breaking changes to the public API. This is a pure, behavior-preserving performance optimization: the resulting container contents are identical to before for every target type (verified with tests below), only unnecessary rehashing while filling `reserve()`-capable containers is avoided.

## Testing

Extended `tests/src/unit-conversions.cpp` with a new section ("reserve is called on containers that support it (#5406)") that builds a 100-entry JSON object and:
- Converts it to `std::unordered_map<std::string, int>` (uses the new reserve path) and checks size and every key/value pair.
- Converts it to `std::map<std::string, int>` (no `reserve()`, exercises the SFINAE fallback) and checks size and every key/value pair identically.

Ran locally (clang++, C++11 and C++17):
- `tests/src/unit-conversions.cpp`: all test cases pass (461 assertions, 0 failed).
- `tests/src/unit-udt.cpp`: all test cases pass (71 assertions, 0 failed).
- `tests/src/unit-allocator.cpp`: all test cases pass (17 assertions, 0 failed).

Also ran `make amalgamate` and committed the resulting `single_include/nlohmann/json.hpp` / `json_fwd.hpp` changes alongside the `include/` change.

Fixes #5406

### DIFF
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -398,6 +398,34 @@ inline void from_json(const BasicJsonType& j, CompatibleArrayType& bin)
     }
 }
 
+template<typename BasicJsonType, typename ConstructibleObjectType>
+auto from_json_object_impl(const BasicJsonType& j, ConstructibleObjectType& obj, priority_tag<1> /*unused*/)
+-> decltype(
+    obj.reserve(std::declval<typename ConstructibleObjectType::size_type>()),
+    void())
+{
+    ConstructibleObjectType ret;
+    const auto* inner_object = j.template get_ptr<const typename BasicJsonType::object_t*>();
+    ret.reserve(inner_object->size());
+    for (const auto& p : *inner_object)
+    {
+        ret.emplace(p.first, p.second.template get<typename ConstructibleObjectType::mapped_type>());
+    }
+    obj = std::move(ret);
+}
+
+template<typename BasicJsonType, typename ConstructibleObjectType>
+inline void from_json_object_impl(const BasicJsonType& j, ConstructibleObjectType& obj, priority_tag<0> /*unused*/)
+{
+    ConstructibleObjectType ret;
+    const auto* inner_object = j.template get_ptr<const typename BasicJsonType::object_t*>();
+    for (const auto& p : *inner_object)
+    {
+        ret.emplace(p.first, p.second.template get<typename ConstructibleObjectType::mapped_type>());
+    }
+    obj = std::move(ret);
+}
+
 template<typename BasicJsonType, typename ConstructibleObjectType,
          enable_if_t<is_constructible_object_type<BasicJsonType, ConstructibleObjectType>::value, int> = 0>
 inline void from_json(const BasicJsonType& j, ConstructibleObjectType& obj)
@@ -407,13 +435,7 @@ inline void from_json(const BasicJsonType& j, ConstructibleObjectType& obj)
         JSON_THROW(type_error::create(302, concat("type must be object, but is ", j.type_name()), &j));
     }
 
-    ConstructibleObjectType ret;
-    const auto* inner_object = j.template get_ptr<const typename BasicJsonType::object_t*>();
-    for (const auto& p : *inner_object)
-    {
-        ret.emplace(p.first, p.second.template get<typename ConstructibleObjectType::mapped_type>());
-    }
-    obj = std::move(ret);
+    from_json_object_impl(j, obj, priority_tag<1> {});
 }
 
 // overload for arithmetic types, not chosen for basic_json template arguments
--- a/include/nlohmann/detail/conversions/from_json.hpp
+++ b/include/nlohmann/detail/conversions/from_json.hpp
@@ -398,27 +398,23 @@ inline void from_json(const BasicJsonType& j, CompatibleArrayType& bin)
     }
 }
 
-template<typename BasicJsonType, typename ConstructibleObjectType>
-auto from_json_object_impl(const BasicJsonType& j, ConstructibleObjectType& obj, priority_tag<1> /*unused*/)
--> decltype(
-    obj.reserve(std::declval<typename ConstructibleObjectType::size_type>()),
-    void())
+template<typename ConstructibleObjectType>
+auto from_json_object_reserve(ConstructibleObjectType& obj, typename ConstructibleObjectType::size_type size, priority_tag<1> /*unused*/)
+-> decltype(obj.reserve(size), void())
 {
-    ConstructibleObjectType ret;
-    const auto* inner_object = j.template get_ptr<const typename BasicJsonType::object_t*>();
-    ret.reserve(inner_object->size());
-    for (const auto& p : *inner_object)
-    {
-        ret.emplace(p.first, p.second.template get<typename ConstructibleObjectType::mapped_type>());
-    }
-    obj = std::move(ret);
+    obj.reserve(size);
 }
 
+template<typename ConstructibleObjectType>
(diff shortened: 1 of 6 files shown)


nlohmann:
Verified this locally with a before/after benchmark comparing the amalgamated single-header build of `develop` (merge-base `09b6b6b5`) against this branch's tip, compiled with `clang++ -O2 -std=c++11` (Apple clang 21).

Benchmark: converting a 5,000-entry JSON object to `std::unordered_map<std::string, int>` via `get<>()`, repeated 80 times per run (5 runs, median reported), with a global `operator new` override counting heap allocations per conversion.

- Before (develop): 0.22 ms / 5012 allocs
- After (this branch): 0.17 ms / 5001 allocs
- **~22% faster** (allocation count drops by 11 — the rehash-driven bucket-array reallocations the missing `reserve()` used to cause)

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -398,15 +398,25 @@ inline void from_json(const BasicJsonType& j, CompatibleArrayType& bin)
...
+template<typename BasicJsonType, typename ConstructibleObjectType>
+auto from_json_object_impl(const BasicJsonType& j, ConstructibleObjectType& obj, priority_tag<1> /*unused*/)
+-> decltype(
+    obj.reserve(std::declval<typename ConstructibleObjectType::size_type>()),
+    void())
 {
-    if (JSON_HEDLEY_UNLIKELY(!j.is_object()))
+    ConstructibleObjectType ret;
+    const auto* inner_object = j.template get_ptr<const typename BasicJsonType::object_t*>();
+    ret.reserve(inner_object->size());
```
gregmarr: Would it make sense to have a `from_json_object_reserve` function that has this machinery that either does nothing or has the `reserve` and call that from `from_json` instead of duplicating the entire body between the two `from_json_object_impl`?

nlohmann: Good idea — done. Factored the reserve logic out into a `from_json_object_reserve(obj, size, priority_tag<1/0>)` overload pair (SFINAE-dispatched the same way as the rest of this file) that either calls `reserve()` or is a no-op; `from_json_object_impl()` now exists once and just calls it before the emplace loop. Pushed.

include/nlohmann/detail/conversions/from_json.hpp
```diff
@@ -398,24 +398,42 @@ inline void from_json(const BasicJsonType& j, CompatibleArrayType& bin)
...
-    }
+    obj.reserve(size);
+}
+
+template<typename ConstructibleObjectType>
+inline void from_json_object_reserve(ConstructibleObjectType& /*obj*/, std::size_t /*size*/, priority_tag<0> /*unused*/)
+{}
 
+template<typename BasicJsonType, typename ConstructibleObjectType>
+inline void from_json_object_impl(const BasicJsonType& j, ConstructibleObjectType& obj)
```
gregmarr: Why do you even need to factor out `from_json_object_impl` when it's just the last thing in `from_json`?

nlohmann: Fair point — it doesn't. Inlined it directly into `from_json`; `from_json_object_reserve` is the only helper left, since that one still needs the SFINAE overload pair. Pushed.

gregmarr (approved):

PR #5475, opened 2026-09-05
Undefine the four JSON_HEDLEY_* macros that leak after including json.hpp

### LINKED ISSUE #5408: Four JSON_HEDLEY_* macros leak into user code (PRAGMA, PREDICT_TRUE, PREDICT_FALSE, CLANG_HAS_DECLSPEC_ATTRIBUTE) — not undefined by hedley_undef.hpp
### Description

The library goes to lengths to undefine its internal macros after use — `detail/macro_unscope.hpp` includes [`thirdparty/hedley/hedley_undef.hpp`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/thirdparty/hedley/hedley_undef.hpp), which `#undef`s 148 `JSON_HEDLEY_*` names so they do not pollute the including translation unit. But four `JSON_HEDLEY_*` macros remain **defined** after `#include <nlohmann/json.hpp>`:

- `JSON_HEDLEY_PRAGMA`
- `JSON_HEDLEY_PREDICT_TRUE`
- `JSON_HEDLEY_PREDICT_FALSE`
- `JSON_HEDLEY_CLANG_HAS_DECLSPEC_ATTRIBUTE`

They are `#define`d in `thirdparty/hedley/hedley.hpp` but have no matching `#undef` in `hedley_undef.hpp`. This is internally inconsistent: for example `JSON_HEDLEY_LIKELY`/`JSON_HEDLEY_UNLIKELY` are correctly undefined, but their siblings `JSON_HEDLEY_PREDICT_TRUE`/`JSON_HEDLEY_PREDICT_FALSE` (the branch-prediction-with-probability variants) are not.

### Reproduction steps

```cpp
#include <nlohmann/json.hpp>
#include <cstdio>

int main()
{
    int leaked = 0;
#ifdef JSON_HEDLEY_PRAGMA
    std::puts("LEAK: JSON_HEDLEY_PRAGMA"); ++leaked;
#endif
#ifdef JSON_HEDLEY_PREDICT_TRUE
    std::puts("LEAK: JSON_HEDLEY_PREDICT_TRUE"); ++leaked;
#endif
#ifdef JSON_HEDLEY_PREDICT_FALSE
    std::puts("LEAK: JSON_HEDLEY_PREDICT_FALSE"); ++leaked;
#endif
#ifdef JSON_HEDLEY_CLANG_HAS_DECLSPEC_ATTRIBUTE
    std::puts("LEAK: JSON_HEDLEY_CLANG_HAS_DECLSPEC_ATTRIBUTE"); ++leaked;
#endif
#ifdef JSON_HEDLEY_LIKELY
    std::puts("LEAK: JSON_HEDLEY_LIKELY (should NOT appear)"); ++leaked;
#endif
    std::printf("leaked = %d (expected 0)\n", leaked);
}
```

- **Expected:** `leaked = 0` — every `JSON_HEDLEY_*` macro is undefined after the header, as intended by the scope/unscope design. `JSON_HEDLEY_LIKELY` correctly does not leak.
- **Actual (g++ 13.3.0, both the modular include tree and `single_include`):**
  ```
  LEAK: JSON_HEDLEY_PRAGMA
  LEAK: JSON_HEDLEY_PREDICT_TRUE
  LEAK: JSON_HEDLEY_PREDICT_FALSE
  LEAK: JSON_HEDLEY_CLANG_HAS_DECLSPEC_ATTRIBUTE
  leaked = 4
  ```

### Suggested fix

Add `#undef` lines for these four names to `thirdparty/hedley/hedley_undef.hpp`. Note that this file is **generated** (the Hedley vendoring renames the upstream `HEDLEY_*` macros to `JSON_HEDLEY_*`), and the gap likely originates in the upstream Hedley `HEDLEY_UNDEF` list, so the durable fix is in the generation/vendoring step (and possibly reported upstream) rather than a manual edit that the next regeneration would overwrite. A small amalgamation test asserting that none of the `JSON_HEDLEY_*` names remain defined after `#include <nlohmann/json.hpp>` would prevent regressions.

### Impact

Low severity — namespace pollution of prefixed macros, not a miscompile. But it contradicts the library's stated cleanup contract and can collide with user code that (re)defines any of these names. Compare the historical #417 (`json.hpp` forcibly defining `GCC_VERSION`).

### Library version

### Validation

nlohmann:
Fixes #5408.

## Problem

`include/nlohmann/detail/macro_unscope.hpp` includes `hedley_undef.hpp`, which is supposed to `#undef` every `JSON_HEDLEY_*` macro so none of them leak into the including translation unit. Four macros stayed defined after `#include <nlohmann/json.hpp>`:

- `JSON_HEDLEY_PRAGMA`
- `JSON_HEDLEY_PREDICT_TRUE`
- `JSON_HEDLEY_PREDICT_FALSE`
- `JSON_HEDLEY_CLANG_HAS_DECLSPEC_ATTRIBUTE`

## Root cause

`hedley_undef.hpp` is generated by the `update_hedley` Makefile target, which greps `hedley.hpp` for its own internal `#undef JSON_HEDLEY_X` redefinition guards (the `#if defined(JSON_HEDLEY_X) #undef JSON_HEDLEY_X #endif` idiom Hedley uses before most `#define`s). Two things go wrong for these four macros specifically:

- `JSON_HEDLEY_PRAGMA`, `JSON_HEDLEY_PREDICT_TRUE`, and `JSON_HEDLEY_PREDICT_FALSE` are defined in upstream Hedley without that redefinition-guard idiom at all, so they were never picked up by the grep.
- `JSON_HEDLEY_CLANG_HAS_DECLSPEC_ATTRIBUTE`'s guard has an upstream typo — it checks/undefs `JSON_HEDLEY_CLANG_HAS_DECLSPEC_DECLSPEC_ATTRIBUTE` (double `DECLSPEC`) instead of the macro it actually guards — so the grep captured the wrong (never-defined) name.

## Changes

- `include/nlohmann/thirdparty/hedley/hedley_undef.hpp`: corrected the `DECLSPEC_ATTRIBUTE` typo and added the three missing `#undef` lines, keeping the file's alphabetical ordering.
- `Makefile` (`update_hedley` target): changed the generation logic to extract macro names directly from every `#define JSON_HEDLEY_...` line in `hedley.hpp`, instead of from existing `#undef` guards. This way, a future `make update_hedley` run undefines every macro Hedley actually defines, including ones without a pre-existing redefinition guard, so this class of gap can't reappear on the next Hedley sync. I did **not** run `make update_hedley` in this PR, since that also pulls in an unrelated upstream Hedley sync (a separate, much larger change); `hedley_undef.hpp` was hand-patched instead, and `single_include` was regenerated with `make amalgamate`.
- `tests/src/unit-no-macro-leak.cpp` (new, picked up automatically by `tests/CMakeLists.txt`'s existing `unit-*.cpp` glob — no CMakeLists.txt edit needed): includes `<nlohmann/json.hpp>` and then does an `#ifdef`/`#error` check for every one of the 151 `JSON_HEDLEY_*` macro names (mechanically derived from `hedley_undef.hpp`), so a leak of any of them — not just these four — fails the build.

## Validation

- `tests/src/unit-no-macro-leak.cpp` compiles and passes against the fixed headers (both `include/` and `single_include/`).
- Confirmed the same test fails with exactly the 4 expected `#error`s when compiled against the pre-fix `hedley_undef.hpp`.
- Compiled `unit-meta.cpp` and `unit-udt_macro.cpp` (which exercise Hedley-derived macros) — both still pass.
- `make amalgamate` was run; the resulting `single_include/nlohmann/json.hpp` diff is limited to the same 4-line hedley_undef change.

## Breaking change?

No breaking changes. This only affects macros that are documented as internal/private (`JSON_HEDLEY_*`) and were never supposed to be visible to users in the first place.


gregmarr:
Is this not already done by #5415 ?

nlohmann:
> Is this not already done by #5415 ?

It is. But it seems that author created a bunch of PRs automatically and then was unresponsive since. I try to clean up the backlog a bit and get rid of these half-finished PRs.

gregmarr (approved):

PR #5476, opened 2026-09-05
Reserve capped array capacity in json_sax_dom_parser::start_array() for definite-length binary arrays

### LINKED ISSUE #5405: Binary readers don't reserve array capacity despite knowing the exact length, causing avoidable reallocations (~20% slower from_cbor/from_msgpack on large arrays)
### Description

When the binary readers decode a **definite-length** array (CBOR, MessagePack, and the optimized `[$type#count` form of UBJSON/BJData), the element count is known up front and is passed to `sax->start_array(len)`. But `json_sax_dom_parser::start_array(len)` uses `len` **only** for an overflow check — it never `reserve()`s the underlying `array_t` (a `std::vector` by default). Each element is then appended via `emplace_back` in `handle_value`, so the vector reallocates ~log₂(N) times and copies/moves its contents repeatedly while decoding an array whose final size was already known.

Relevant code:

- [`include/nlohmann/detail/input/json_sax.hpp:286-305`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/input/json_sax.hpp) — `start_array(len)` checks `len` against `max_size()` but does not reserve. `json_sax_dom_callback_parser::start_array` (`:637`) has the same gap. (`start_object` is not addressable this way: the default `object_t` is `std::map`, which has no `reserve`.)
- [`include/nlohmann/detail/input/json_sax.hpp:438-446`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/input/json_sax.hpp) — `handle_value` appends with `array->emplace_back(...)`.

The JSON **text** parser is unaffected: it passes `detail::unknown_size()` (it cannot know the count in advance), so no reserve happens and nothing changes for text parsing.

> ### ⚠️ Correction to an earlier version of this report
> An earlier version of this issue suggested an unconditional `reserve(len)` and claimed the `max_size()` check made it safe. **That is wrong** — see the DoS section below. `max_size()` for a `std::vector` is ~2⁶¹, far above any real allocation, so it does not bound a hostile `len`. The fix must **cap** the reservation. The numbers and suggested fix below have been corrected accordingly.

### Measurement (capped reserve)

Decoding an array of integers, g++ 13.3.0 `-O2`, counting `from_X` calls only (global `operator new` override). "CAPPED" reserves `min(len, 16384)` in `start_array` when `len != detail::unknown_size()`:

| case | baseline | capped reserve |
|---|---|---|
| `from_cbor`, array[1000]   | 0.111 ms, **17** allocs | 0.081 ms, **7** allocs  (**~27% faster**) |
| `from_cbor`, array[10000]  | 1.027 ms, **21** allocs | 0.942 ms, **7** allocs  (**~8% faster**) |
| `from_cbor`, array[100000] | 12.2 ms, **24** allocs  | 11.2 ms, **10** allocs (**~8% faster**) |
| `from_msgpack`, array[100000] | 12.1 ms, 24 allocs | 11.6 ms, 10 allocs |

Arrays **at or below the cap** (the overwhelming majority of real arrays) get the full benefit — a single reservation and the minimum allocation count. Arrays larger than the cap still improve (the initial cascade of tiny reallocations is removed) but pay a few geometric growths above the cap. An *unconditional* `reserve(len)` would give ~20% at 100k, but is unsafe (below). UBJSON's default array encoding is indefinite-length, so it legitimately does not change; the optimized `[$type#count` form and BJData ndarrays would.

### ⚠️ Why the reserve must be capped (memory-amplification DoS)

A definite-length header can claim an enormous element count with **no data following**. With an unconditional `reserve(len)`:

```cpp
#include <nlohmann/json.hpp>
#include <vector>
using json = nlohmann::json;
int main()
{
    // CBOR: 0x9A = array with uint32 length; claim 0xFFFFFFFF (~4.29e9) elements, no data
    std::vector<std::uint8_t> blob = {0x9A, 0xFF, 0xFF, 0xFF, 0xFF};
    try { auto r = json::from_cbor(blob); }
    catch (const std::exception& e) { std::printf("%s\n", e.what()); }
}
```

| version | result |
|---|---|
| **baseline** (develop) | `[json.exception.parse_error.110] ... unexpected end of input` — a few bytes of memory |
| unconditional `reserve(len)` | **`std::bad_alloc`** — attempts to reserve ~4.29e9 × `sizeof(basic_json)` ≈ 68 GB up front |
| **capped** `reserve(min(len, 16384))` | `[json.exception.parse_error.110]` — graceful, ≤ 256 KB reserved |

This is exactly the concern already documented for `get_bytes` (string/binary reading), which deliberately does **not** reserve on an untrusted length and instead reads in bounded 4096-byte chunks. The array path needs the same discipline.

### Minimal benchmark

```cpp
#include <new>
#include <cstdlib>
#include <cstdio>
static long allocs = 0; static bool count = false;
void* operator new(std::size_t n){ if(count)++allocs; void* p=std::malloc(n); if(!p)throw std::bad_alloc(); return p; }

#include <nlohmann/json.hpp>
#include <chrono>
using json = nlohmann::json;

int main()
{
    const int N = 1000;
    json arr = json::array();
    for (int i = 0; i < N; ++i) arr.push_back(i);
    auto blob = json::to_cbor(arr);
    long iters = 2000; volatile size_t sink = 0;
    count = true; allocs = 0;
    auto t0 = std::chrono::steady_clock::now();
    for (long it = 0; it < iters; ++it) { auto r = json::from_cbor(blob); sink += r.size(); }
    auto t1 = std::chrono::steady_clock::now();
    count = false;
    std::printf("%.4f ms/call, %ld allocs/call\n",
                std::chrono::duration<double,std::milli>(t1-t0).count()/iters, allocs/iters);
    (void)sink;
}
```

### Suggested fix

In `json_sax_dom_parser::start_array` (and `json_sax_dom_callback_parser::start_array`), after the existing `max_size()` overflow check, reserve a **capped** amount so a hostile length cannot force a large up-front allocation:

```cpp
if (len != detail::unknown_size())
{
    constexpr std::size_t reserve_cap = 16384; // bounds worst-case pre-allocation
    ref_stack.back()->m_data.m_value.array->reserve(len < reserve_cap ? len : reserve_cap);
}
```

Behavior-preserving (identical decoded value; the `max_size()` guard still runs first). The exact cap is a policy choice for the maintainers.

A more sophisticated alternative that recovers the *full* benefit safely for in-memory input: reserve `len` only when the input adapter can confirm at least `len` bytes remain (each array element encodes to ≥ 1 byte, so `len <= bytes_remaining` means the data really is present). That needs the input adapter to expose its remaining size, which the chunked `get_bytes` currently avoids relying on; the capped reserve above is the simple, self-contained fix.

### Library version

### Validation

nlohmann:
## Summary

CBOR, MessagePack, and the optimized `[$type#count` form of UBJSON/BJData all know the exact element count of a definite-length array up front and pass it to `sax->start_array(len)`. However, `json_sax_dom_parser::start_array()` (and the `json_sax_dom_callback_parser` variant) in `include/nlohmann/detail/input/json_sax.hpp` only used `len` for an overflow check against `max_size()` and never actually reserved capacity in the underlying `array_t` (`std::vector` by default). Each element is then appended one at a time via `emplace_back()`, causing a reallocation cascade for large arrays.

This PR adds a **capped** `reserve()` call after the existing `max_size()` check:

```cpp
if (len != detail::unknown_size())
{
    constexpr std::size_t reserve_cap = 16384;
    ref_stack.back()->m_data.m_value.array->reserve(len < reserve_cap ? len : reserve_cap);
}
```

**Why capped, not `reserve(len)` directly:** `max_size()` for a `std::vector` is astronomically large (~2^61), so it does not bound a hostile/corrupt `len` read from an untrusted input stream. A crafted CBOR array header (e.g. `0x9A` + a huge `uint32` count) can claim billions of elements with no data following it. An unconditional `reserve(len)` would then attempt a multi-gigabyte allocation for a few bytes of malicious input, turning a normal graceful `parse_error` into a memory-amplification DoS. Capping the reservation at 16384 elements keeps the fast-fail `parse_error` behavior for hostile/truncated input while still eliminating the reallocation cascade for realistic arrays (and improving, though not eliminating, the cascade above the cap).

`object_t` (`std::map` by default) has no `reserve()` and is intentionally untouched — only the array path is affected.

## Changes

- `include/nlohmann/detail/input/json_sax.hpp`: capped `reserve()` in `json_sax_dom_parser::start_array()` and `json_sax_dom_callback_parser::start_array()`.
- `single_include/nlohmann/json.hpp`: regenerated via `make amalgamate`.
- Tests added to `tests/src/unit-cbor.cpp`, `unit-msgpack.cpp`, `unit-ubjson.cpp`, `unit-bjdata.cpp`:
  - A DoS regression test: a definite-length array header with a huge claimed count and no element data still throws the expected `parse_error` (not `std::bad_alloc`, no excessive memory use).
  - Correctness tests decoding arrays of various sizes (small, exactly at the 16384 cap, above the cap) and confirming identical decoded values.
  - A test confirming `sax_parse` with a custom user SAX consumer is unaffected (the reserve is purely internal to the DOM builder).

## Breaking change?

No. This is a behavior-preserving performance optimization internal to `json_sax_dom_parser` / `json_sax_dom_callback_parser`. Decoded values, the public SAX interface, and error behavior for malformed/hostile input are all unchanged.

Fixes #5405

### DIFF
--- a/include/nlohmann/detail/input/json_sax.hpp
+++ b/include/nlohmann/detail/input/json_sax.hpp
@@ -8,6 +8,7 @@
 
 #pragma once
 
+#include <algorithm> // min
 #include <cstddef>
 #include <string> // string
 #include <type_traits> // enable_if_t
@@ -312,7 +313,7 @@ class json_sax_dom_parser
             // by max_size(), unlike e.g. std::vector) cannot trigger an oversized
             // allocation for a small or truncated input
             constexpr std::size_t reserve_cap = 16384;
-            ref_stack.back()->m_data.m_value.array->reserve(len < reserve_cap ? len : reserve_cap);
+            ref_stack.back()->m_data.m_value.array->reserve((std::min)(len, reserve_cap));
         }
 
         return true;
@@ -701,7 +702,7 @@ class json_sax_dom_callback_parser
                 // by max_size(), unlike e.g. std::vector) cannot trigger an oversized
                 // allocation for a small or truncated input
                 constexpr std::size_t reserve_cap = 16384;
-                ref_stack.back()->m_data.m_value.array->reserve(len < reserve_cap ? len : reserve_cap);
+                ref_stack.back()->m_data.m_value.array->reserve((std::min)(len, reserve_cap));
             }
         }
 
--- a/include/nlohmann/detail/input/json_sax.hpp
+++ b/include/nlohmann/detail/input/json_sax.hpp
@@ -305,6 +305,16 @@ class json_sax_dom_parser
             JSON_THROW(out_of_range::create(408, concat("excessive array size: ", std::to_string(len)), ref_stack.back()));
         }
 
+        if (len != detail::unknown_size())
+        {
+            // reserve upfront to avoid repeated reallocations while adding elements,
+            // but cap the reservation so a bogus/hostile length (which is not bounded
+            // by max_size(), unlike e.g. std::vector) cannot trigger an oversized
+            // allocation for a small or truncated input
+            constexpr std::size_t reserve_cap = 16384;
+            ref_stack.back()->m_data.m_value.array->reserve(len < reserve_cap ? len : reserve_cap);
+        }
+
         return true;
     }
 
@@ -682,6 +692,16 @@ class json_sax_dom_callback_parser
             if (JSON_HEDLEY_UNLIKELY(len != detail::unknown_size() && len > ref_stack.back()->max_size()))
             {
                 JSON_THROW(out_of_range::create(408, concat("excessive array size: ", std::to_string(len)), ref_stack.back()));
+            }
+
+            if (len != detail::unknown_size())
+            {
+                // reserve upfront to avoid repeated reallocations while adding elements,
+                // but cap the reservation so a bogus/hostile length (which is not bounded
+                // by max_size(), unlike e.g. std::vector) cannot trigger an oversized
+                // allocation for a small or truncated input
+                constexpr std::size_t reserve_cap = 16384;
+                ref_stack.back()->m_data.m_value.array->reserve(len < reserve_cap ? len : reserve_cap);
             }
         }
 
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -7892,6 +7892,7 @@ NLOHMANN_JSON_NAMESPACE_END
 
 
 
+#include <algorithm> // min
 #include <cstddef>
 #include <string> // string
 #include <type_traits> // enable_if_t
@@ -11025,7 +11026,7 @@ class json_sax_dom_parser
             // by max_size(), unlike e.g. std::vector) cannot trigger an oversized
             // allocation for a small or truncated input
             constexpr std::size_t reserve_cap = 16384;
-            ref_stack.back()->m_data.m_value.array->reserve(len < reserve_cap ? len : reserve_cap);
+            ref_stack.back()->m_data.m_value.array->reserve((std::min)(len, reserve_cap));
         }
 
         return true;
@@ -11414,7 +11415,7 @@ class json_sax_dom_callback_parser
                 // by max_size(), unlike e.g. std::vector) cannot trigger an oversized
                 // allocation for a small or truncated input
                 constexpr std::size_t reserve_cap = 16384;
-                ref_stack.back()->m_data.m_value.array->reserve(len < reserve_cap ? len : reserve_cap);
+                ref_stack.back()->m_data.m_value.array->reserve((std::min)(len, reserve_cap));
             }
         }
 
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -11018,6 +11018,16 @@ class json_sax_dom_parser
             JSON_THROW(out_of_range::create(408, concat("excessive array size: ", std::to_string(len)), ref_stack.back()));
         }
 
+        if (len != detail::unknown_size())
+        {
+            // reserve upfront to avoid repeated reallocations while adding elements,
+            // but cap the reservation so a bogus/hostile length (which is not bounded
+            // by max_size(), unlike e.g. std::vector) cannot trigger an oversized
+            // allocation for a small or truncated input
+            constexpr std::size_t reserve_cap = 16384;
+            ref_stack.back()->m_data.m_value.array->reserve(len < reserve_cap ? len : reserve_cap);
+        }
+
         return true;
     }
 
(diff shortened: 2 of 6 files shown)


nlohmann:
Verified this locally with a before/after benchmark comparing the amalgamated single-header build of `develop` (merge-base `09b6b6b5`) against this branch's tip, compiled with `clang++ -O2 -std=c++11` (Apple clang 21).

Benchmark: `json::from_cbor()` decoding a definite-length CBOR array of 10,000 integers, repeated 80 times per run (5 runs, median reported), with a global `operator new` override counting heap allocations per decode.

- Before (develop): 0.31 ms / 20 allocs
- After (this branch): 0.23 ms / 6 allocs
- **~26% faster, 70% fewer allocations** (20 → 6 — the capped `reserve()` collapses the vector's incremental reallocation cascade into a single allocation for arrays at or under the 16384 cap)

Also re-confirmed the safety property this PR is built around: a crafted CBOR header claiming ~4.3 billion elements with no following data (`0x9A FF FF FF FF`) still throws a clean `parse_error` rather than attempting an oversized allocation, both before and after this change.

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -301,6 +301,16 @@ class json_sax_dom_parser
...
         }
 
+        if (len != detail::unknown_size())
+        {
+            // reserve upfront to avoid repeated reallocations while adding elements,
+            // but cap the reservation so a bogus/hostile length (which is not bounded
+            // by max_size(), unlike e.g. std::vector) cannot trigger an oversized
+            // allocation for a small or truncated input
+            constexpr std::size_t reserve_cap = 16384;
+            ref_stack.back()->m_data.m_value.array->reserve(len < reserve_cap ? len : reserve_cap);
```
gregmarr: Would this be clearer as `(std::min)(len, reserve_cap)`?

include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -661,6 +671,16 @@ class json_sax_dom_callback_parser
...
             }
+
+            if (len != detail::unknown_size())
+            {
+                // reserve upfront to avoid repeated reallocations while adding elements,
+                // but cap the reservation so a bogus/hostile length (which is not bounded
+                // by max_size(), unlike e.g. std::vector) cannot trigger an oversized
+                // allocation for a small or truncated input
+                constexpr std::size_t reserve_cap = 16384;
+                ref_stack.back()->m_data.m_value.array->reserve(len < reserve_cap ? len : reserve_cap);
```
gregmarr: Would this be clearer as `(std::min)(len, reserve_cap)`?

include/nlohmann/detail/input/json_sax.hpp
nlohmann: Yes — switched both spots to `(std::min)(len, reserve_cap)`. Pushed.

include/nlohmann/detail/input/json_sax.hpp
nlohmann: Same fix applied here. Pushed.

gregmarr (approved):

Issue #5403, opened 2026-08-25
to_bjdata() silently truncates out-of-range _ArrayData_ elements (e.g. 256 as uint8 becomes 0)


nlohmann:
### Description

The BJData ndarray writer validates the *kind* of each `_ArrayData_` element (integer vs float, added in #5301) but not its *range*. An element that does not fit the declared `_ArrayType_` is written with a wrap-around `static_cast`, silently corrupting the data.

In `write_bjdata_ndarray` ([`include/nlohmann/detail/output/binary_writer.hpp`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/output/binary_writer.hpp)), each element is written as e.g. `write_number(static_cast<std::uint8_t>(el.get<std::uint64_t>()), true)`. For `_ArrayType_ = "uint8"` and an element `256`, the cast yields `0`, so the value round-trips to a different value with no error.

This is a softer case than #5398/#5399: the documented fallback conditions ([`bjdata.md`](https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/features/binary_formats/bjdata.md)) require each element to be "a number of the kind named by `_ArrayType_`", and `256` *is* an integer — the docs do not currently mention range. So this is reported as a lower-severity data-integrity issue / possible doc-vs-behavior gap for the maintainer to weigh: either range-check and fall back to a plain object encoding (consistent with the other annotation-validation failures), raise an error, or document that out-of-range elements are truncated.

### Reproduction steps

Serialize a `uint8` ndarray whose `_ArrayData_` contains a value > 255 and round-trip it.

### Expected vs. actual results

- **Expected:** fall back to plain-object encoding (or an `out_of_range` error); no silent value change.
- **Actual:** `256` is written as byte `0x00`; the array round-trips to `[1,0]`.

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <cstdio>
using json = nlohmann::json;

int main()
{
    json j;
    j["_ArrayType_"] = "uint8";
    j["_ArraySize_"] = json::array({2});
    j["_ArrayData_"] = json::array({1, 256});   // 256 does not fit uint8
    auto v = json::to_bjdata(j);
    std::printf("back=%s\n", json::from_bjdata(v).dump().c_str());  // back=[1,0]
}
```

### Error messages

```
back=[1,0]
```

### Compiler and operating system

g++ 13.3.0 (Ubuntu 24.04, x86-64)

### Library version

[...]

PR #5479, opened 2026-09-05
Fix to_bjdata() emitting the Draft-3-only 'B' marker in default Draft-2 mode

### LINKED ISSUE #5404: to_bjdata() emits the Draft-3-only 'B' (byte) marker for _ArrayType_:"byte" even in default Draft-2 mode, and the value does not round-trip
### Description

The BJData ndarray writer maps `_ArrayType_ = "byte"` to the type marker `B` unconditionally, with no gate on `bjdata_version`. But `B` (byte) is a Draft-3-only marker: the docs ([`bjdata.md`](https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/features/binary_formats/bjdata.md), "Binary values") state it "must be explicitly enabled using the `version` parameter of `to_bjdata`", and BJData Draft 2 has no `B` type.

In `write_bjdata_ndarray` ([`include/nlohmann/detail/output/binary_writer.hpp`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/output/binary_writer.hpp)) the type map contains `{"byte", 'B'}` and the marker is written regardless of the requested `version`. So with the **default** `version = draft2`, an annotated `byte` array produces `[$B#…`, a stream that is invalid under Draft 2 and that a strict third-party Draft-2 reader will reject.

Separately, unlike every other `_ArrayType_` (e.g. `uint8`, which round-trips back to the annotated object), the result parses back as a **binary** value, so `to_bjdata`/`from_bjdata` changes the value.

Low severity / spec-conformance and interop, and only reachable with JData-annotated input, but the behavior contradicts the documented version-gating of `B`.

### Reproduction steps

Serialize a `byte` ndarray with default (Draft-2) options and inspect the bytes / round-trip.

- **Expected (Draft-2 default):** either a fallback encoding that does not use the Draft-3 `B` marker, or an error; and round-trip equality like the other dtypes.
- **Actual:** bytes contain `B` (`0x42`); the value round-trips to a binary value instead of the original object.

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <cstdio>
using json = nlohmann::json;

int main()
{
    json j;
    j["_ArrayType_"] = "byte";
    j["_ArraySize_"] = json::array({2});
    j["_ArrayData_"] = json::array({1, 2});
    auto v = json::to_bjdata(j);              // default draft2
    for (auto b : v) std::printf("%02x ", b); // 5b 24 42 23 5b 69 02 5d 01 02  ("[$B#[i2]..")
    std::printf("\nback=%s\n", json::from_bjdata(v).dump().c_str());
}
```

### Error messages

```
5b 24 42 23 5b 69 02 5d 01 02
back={"bytes":[1,2],"subtype":null}
```

(The stream contains the Draft-3 `B` marker under the Draft-2 default, and the round-trip yields a binary value rather than the original annotated object.)

### Suggested fix

Only accept `"byte"`/emit `B` when `bjdata_version` selects Draft 3; otherwise fall back to a plain-object encoding (consistent with the other unrepresentable-annotation cases).

### Library version

### Validation

nlohmann:
## Stacked PR

This PR is stacked on top of #5473 (fix for #5403) and is based on its branch `issue-5403-bjdata-uint-range` rather than `develop`. It should be reviewed and merged after that PR.

## Summary

In the same BJData ndarray writer touched by #5473, `_ArrayType_ = "byte"` mapped unconditionally to the BJData type marker `'B'`, regardless of the `bjdata_version` requested for `to_bjdata()`. But `'B'` is defined only by [BJData Draft 3][BJDataBinArr]; the docs (`docs/mkdocs/docs/features/binary_formats/bjdata.md`, "Binary values") say the Draft 3 optimized binary array "must be explicitly enabled using the `version` parameter of `to_bjdata`".

With the default `version = draft2`, this produced a stream containing a marker that Draft 2 does not define and, unlike every other `_ArrayType_`, the value round-tripped back through `from_bjdata` as a **binary** value instead of the original annotated object — because a `'B'`-typed optimized array without the ndarray flag is otherwise the library's own encoding for binary data.

[BJDataBinArr]: https://github.com/NeuroJSON/bjdata/blob/master/Binary_JData_Specification.md#optimized-binary-array

```cpp
json j;
j["_ArrayType_"] = "byte";
j["_ArraySize_"] = json::array({2});
j["_ArrayData_"] = json::array({1, 2});
auto v = json::to_bjdata(j);  // default draft2 — used to emit the 'B' marker anyway
// from_bjdata(v) used to give a binary value, not the original object
```

## Fix

`write_bjdata_ndarray` now checks `bjdata_version` right after resolving `dtype` from `_ArrayType_`: if `dtype == 'B'` and `bjdata_version` is not `draft3`, it falls back to the same plain-object encoding already used by this function for other invalid-annotation cases (the same mechanism added/reused in #5473). Draft 3 explicitly selected keeps emitting the compact `'B'` ndarray encoding exactly as before.

## Tests

- Updated the existing round-trip check for `v_B` in the `"optimized ndarray (type and vector-size ndarray with JData annotations)"` section to pass `bjdata_version_t::draft3` explicitly, since `v_B` uses the Draft-3-only `'B'` marker.
- Split `"byte"` out of the generic type loop in `"ndarray parsed from text is written as a typed array"` into its own check that explicitly selects Draft 3.
- Added a new section, `"ndarray with _ArrayType_ "byte" is gated by the BJData draft version"`, covering:
  - default (Draft 2) falls back to a plain object and round-trips
  - explicit Draft 2 behaves the same as the default
  - explicit Draft 3 still uses the compact `'B'` ndarray encoding, with an exact expected byte sequence, and round-trips correctly

Ran the full `unit-bjdata` suite offline (compiled against `include/` with a stub `test_data.hpp`): 693941/693942 assertions pass; the one failure and the one skipped test case are pre-existing and unrelated (they require downloaded test data, unavailable in this offline setup) — identical to the baseline on `develop`. Also ran `unit-ubjson`, `unit-binary_formats`, and `unit-regression2` offline with no new failures.

## Breaking change?

No breaking changes to the public API. This changes only the wire output of `to_bjdata()` for annotated objects with `_ArrayType_ = "byte"` under the default (Draft 2) mode: such objects now serialize as plain objects (matching every other invalid-for-the-current-mode annotation already handled by this function) instead of emitting a Draft-3-only marker that Draft 2 readers do not define. Draft 3 output is unchanged.

Fixes #5404.

### DIFF
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -1691,6 +1691,16 @@ class binary_writer
         }
         CharType dtype = it->second;
 
+        // the 'B' (byte) marker is only defined by BJData Draft 3; emitting it
+        // under the default Draft 2 mode would produce a stream that Draft 2
+        // readers reject, so such an object falls back to a plain object
+        // encoding instead (see the "Binary values" section of the BJData
+        // documentation)
+        if (dtype == 'B' && bjdata_version != bjdata_version_t::draft3)
+        {
+            return true;
+        }
+
         key = "_ArraySize_";
         // the dimensions are written verbatim as the header length below, so a
         // value that is not an array cannot produce a valid one: null emits 'Z'
--- a/single_include/nlohmann/json.hpp
+++ b/single_include/nlohmann/json.hpp
@@ -20226,6 +20226,16 @@ class binary_writer
         }
         CharType dtype = it->second;
 
+        // the 'B' (byte) marker is only defined by BJData Draft 3; emitting it
+        // under the default Draft 2 mode would produce a stream that Draft 2
+        // readers reject, so such an object falls back to a plain object
+        // encoding instead (see the "Binary values" section of the BJData
+        // documentation)
+        if (dtype == 'B' && bjdata_version != bjdata_version_t::draft3)
+        {
+            return true;
+        }
+
         key = "_ArraySize_";
         // the dimensions are written verbatim as the header length below, so a
         // value that is not an array cannot produce a valid one: null emits 'Z'
--- a/tests/src/unit-bjdata.cpp
+++ b/tests/src/unit-bjdata.cpp
@@ -2586,7 +2586,12 @@ TEST_CASE("BJData")
                 CHECK(json::to_bjdata(json::from_bjdata(v_d), true, true) == v_d);
                 CHECK(json::to_bjdata(json::from_bjdata(v_D), true, true) == v_D);
                 CHECK(json::to_bjdata(json::from_bjdata(v_C), true, true) == v_C);
-                CHECK(json::to_bjdata(json::from_bjdata(v_B), true, true) == v_B);
+                // v_B uses the Draft-3-only 'B' marker, so it round-trips only when
+                // Draft 3 is explicitly selected (see GitHub issue #5404); the
+                // default Draft 2 falls back to a plain object instead, covered by
+                // the "ndarray with _ArrayType_ "byte" is gated by the BJData draft
+                // version" section below
+                CHECK(json::to_bjdata(json::from_bjdata(v_B), true, true, json::bjdata_version_t::draft3) == v_B);
             }
 
             SECTION("ndarray with data not matching _ArrayType_ is written as an object")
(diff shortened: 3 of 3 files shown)


include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1691,6 +1691,16 @@ class binary_writer
         }
         CharType dtype = it->second;
 
+        // the 'B' (byte) marker is only defined by BJData Draft 3; emitting it
+        // under the default Draft 2 mode would produce a stream that Draft 2
+        // readers reject, so such an object falls back to a plain object
+        // encoding instead (see the "Binary values" section of the BJData
+        // documentation)
+        if (dtype == 'B' && bjdata_version != bjdata_version_t::draft3)
```
gregmarr, line 1699: Is there a way to future-proof this to "at least draft3"?

nlohmann: Fair point — `bjdata_version_t` only has two values today, so `!=` and `<` happened to coincide. Pushed a change to `bjdata_version < bjdata_version_t::draft3`, which keeps working if a later draft keeps the `'B'` marker valid.

PR #5484, opened 2026-09-05
Skip integer conversion in accept()/SAX validation when the value is unused

### LINKED ISSUE #5411: accept()/SAX validation fully converts every number (strtoull/strtoll/strtof) although only finiteness is needed — up to 2.8x slower than necessary on numeric input
### Description

`lexer::scan_number()` **eagerly** converts every numeric token with `strtoull`/`strtoll`/`strtof` at the end of scanning ([`lexer.hpp:1282-1327`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/input/lexer.hpp)). The *syntactic* validity of the number is already decided by the scanner's state machine **before** that conversion; the conversion only produces the numeric *value*.

For `accept()` and any SAX validation consumer, that value is immediately discarded. The conversion's *only* remaining effect on the accept/reject decision is the parser's non-finite check ([`parser.hpp:266-275`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/input/parser.hpp)):

```cpp
const auto res = m_lexer.get_number_float();
if (JSON_HEDLEY_UNLIKELY(!std::isfinite(res)))
    return sax->parse_error(..., out_of_range::create(406, "number overflow ..."));
```

So the full conversion is performed on every number in the document purely to answer "does this float overflow to ±inf?" — and for integer-classified tokens, not even that:

- A `value_unsigned` / `value_integer` token is, by construction, one whose `strtoull`/`strtoll` succeeded without `ERANGE` — i.e. it fits 64 bits and is therefore always finite and **always accepted**. Its conversion result is *never* used by `accept()`.
- Only a `value_float` token can be rejected, and only when its magnitude exceeds the `double` range (≈1.8e308). That is a decimal-magnitude question that does not require a full `strtof`.

(Confirmed behavior: `accept("1e999") == false`, `accept("9999999999999999999999999999") == true` — the 28-digit integer overflows `uint64` but is finite as a `double`, so it is accepted as a float.)

### Measurement

g++ 13.3.0 `-O2`, `accept()` over a 100k-element array, comparing the current code against a build that short-circuits `scan_number`'s conversion (upper bound):

| input | baseline `accept` | conversion skipped | speedup |
|---|---|---|---|
| integers (100k) | 7.09 ms | 4.12 ms | **1.7×** |
| floats (100k) | 17.74 ms | 6.35 ms | **2.8×** |

On float-heavy input the `strtof` conversion is ~64% of `accept()`'s total time and is entirely discarded.

### Relationship to #5283

This is distinct from the open parser-performance PR #5283. That PR *speeds up* number conversion (Clinger fast path, contiguous fast path) but still performs the full conversion for the validation path — its `accept()` gains come from bulk string/UTF-8 scanning, not from numbers. Eliminating the conversion for validation is an orthogonal, additive win and is most naturally implemented on top of that PR's `number_parse.hpp` rework.

### Suggested direction

Make the value computation lazy / skippable on the validation path, preserving `accept`/reject exactly:

1. Thread a compile-time "values not required" property from the SAX consumer (the acceptor / `json_sax_acceptor`) so the parser can skip `get_number_*()` + `sax->number_*()` for tokens that cannot be rejected.
2. For `value_unsigned` / `value_integer` tokens: skip conversion entirely (always finite, always accepted). The scanner still needs to detect the overflow→float *reclassification*, which can be done by digit count (≤18 digits always fit; ≥21 always overflow `uint64`; only 19–20 need a precise compare) instead of a full `strtoull`.
3. For `value_float` tokens: replace `strtof` with a cheap decimal-magnitude finiteness check (reject when the effective base-10 exponent pushes the value outside the `double` range), keeping the exact `out_of_range.406` boundary.

Full `parse()` is unaffected — it still calls `get_number_*()` once per number and converts exactly as today.

### Behavior preservation

The optimization must keep `accept()`'s current results bit-for-bit, including: `1e999`/`1e400` rejected; huge-but-finite integers (e.g. 28 digits) accepted as finite doubles; the exact `406` overflow boundary. A differential test (`accept` old vs new over a large mixed + edge corpus, including magnitudes straddling `DBL_MAX`) should gate it.

### Library version

nlohmann:
## Summary

Fixes #5411.

`lexer::scan_number()` eagerly converts every numeric token with `strtoull()`/`strtoll()`/`strtof()` at the end of scanning, even though for `accept()` (and any SAX consumer that discards the numeric value, i.e. `json_sax_acceptor`) that converted value is never used. The only thing the conversion still affects for `accept()`/reject is the parser's non-finite check on `value_float` tokens.

This PR threads a `discard_number_values` flag from `basic_json::accept()` down through `parser`/`lexer`. When set, `lexer::scan_number()` can skip `strtoull()`/`strtoll()` entirely for `value_unsigned`/`value_integer` tokens **once the digit count alone guarantees the value fits into 64 bits** (a decimal number with up to 18 digits always fits into both `std::int64_t` and `std::uint64_t`, so `errno` could never have been set to `ERANGE`). Such tokens are, by construction, always finite and unconditionally accepted regardless of their actual value, so only the token classification is needed — not the converted value.

Numbers with 19+ digits (rare in practice) fall through to the exact, completely unmodified conversion code, so their handling — including reclassification to `value_float` when the value overflows 64 bits, and the existing `406` "number overflow" rejection when that reclassified value is not even finite as a `double` — is bit-for-bit identical to before this change.

### Scope and what was deliberately left out

The originating issue's "suggested direction" also proposed (2) a cheap digit-count-based reclassification check to fully replace `strtoull`/`strtoll` even near the 64-bit boundary, and (3) a cheap decimal-magnitude finiteness check to replace `strtof` for `value_float` tokens. Both are **out of scope for this PR**:

- `value_float` handling is completely unchanged — the existing `strtof`/`strtod`/`strtold` conversion is still always called, and the parser's `std::isfinite` check on it is untouched.
- For `value_unsigned`/`value_integer`, the fast path is only taken when the digit count makes 64-bit overflow provably impossible (≤18 digits); everything else (≥19 digits) uses the exact original code path, unchanged. I judged a full digit-count-based reclassification of the ≥19-digit boundary cases to be higher risk for a security-sensitive parser without more extensive validation than fits this PR, so I scoped this down to the always-safe subset.

`parse()` (and the public `sax_parse()` API for user-supplied SAX consumers) are completely unaffected — they never set `discard_number_values`, so they keep calling the full conversion exactly as before.

## Behavior preservation / testing

- Preserved invariants verified: `accept("1e999") == false`, `accept("9999999999999999999999999999") == true` (28-digit integer, overflows `uint64_t` but is finite as `double`).
- Added a differential regression test in `tests/src/unit-class_parser.cpp` (`SECTION("issue #5411 - skip conversion when accept() does not need the numeric value")`) comparing `json::accept()` against `json::parse()` across: small/large integers (both signs), the 18/19/20-digit boundary (both signs), the 64-bit boundaries (`INT64_MAX`/`INT64_MIN`/`UINT64_MAX`/`UINT64_MAX+1`), the 28-digit example from the issue, huge digit-only integers that overflow even a `double` (must reject), `1e999`/`1e400`-style exponent overflow (must reject), values straddling `DBL_MAX`, and a mix of valid/invalid numeric syntax.
- Ran a standalone differential harness comparing `accept()` output byte-for-byte between the pre-change and post-change headers (both the split `include/` headers and the amalgamated `single_include/nlohmann/json.hpp`) over ~2000 cases (hand-picked edge cases + randomized fuzz corpus of digit/exponent combinations): zero mismatches.
- Ran the full required test files locally (offline, without the `json_test_data` submodule available in this environment): `unit-class_lexer.cpp`, `unit-class_parser.cpp`, `unit-deserialization.cpp`, `unit-regression1.cpp`, `unit-regression2.cpp`, `unit-udt.cpp`, `unit-class_parser_diagnostic_positions.cpp` — all pass, with the exception of 7 pre-existing assertions in `unit-regression1.cpp` that fail identically on `develop` in this offline environment because they read fixture files from the (here, unavailable) `json_test_data` test corpus; unrelated to this change (verified by running the identical test against unmodified `develop` and getting the same 7 failures).
- `make amalgamate` was run and `single_include/nlohmann/json.hpp`/`json_fwd.hpp` are included in this PR.

## Breaking change?

No breaking changes. This only adds a new *defaulted* trailing parameter to the internal (`nlohmann::detail::parser`/`lexer`, and the `JSON_PRIVATE_UNLESS_TESTED` `basic_json::parser(...)` factory) constructors; the public `basic_json::accept()`/`parse()`/`sax_parse()` API signatures are unchanged.

### DIFF
--- a/include/nlohmann/detail/input/lexer.hpp
+++ b/include/nlohmann/detail/input/lexer.hpp
@@ -149,10 +149,11 @@ class lexer : public lexer_base<BasicJsonType>
   public:
     using token_type = typename lexer_base<BasicJsonType>::token_type;
 
-    explicit lexer(InputAdapterType&& adapter, bool ignore_comments_ = false) noexcept
+    explicit lexer(InputAdapterType&& adapter, bool ignore_comments_ = false, bool discard_number_values_ = false) noexcept
         : ia(std::move(adapter))
         , ignore_comments(ignore_comments_)
         , decimal_point_char(static_cast<char_int_type>(get_decimal_point()))
+        , discard_number_values(discard_number_values_)
     {}
 
     // deleted because of pointer members
@@ -1279,6 +1280,38 @@ scan_number_done:
         // we are done scanning a number)
         unget();
 
+        // If the caller does not need the converted value (only whether the
+        // input is syntactically valid; see json_sax_acceptor/accept()), an
+        // unsigned/integer token can be reported without calling
+        // strtoull()/strtoll() at all, *provided* we can already tell from
+        // the digit count alone that the conversion cannot overflow 64 bits.
+        // Such tokens are always finite and are accepted unconditionally by
+        // the parser regardless of their actual value (parser::sax_parse_internal()
+        // never checks finiteness for value_unsigned/value_integer), so the
+        // classification below is all that is needed.
+        //
+        // A decimal number with up to 18 digits is always representable in
+        // both std::uint64_t and std::int64_t (18 nines is ~1e18, well below
+        // both UINT64_MAX ~1.8e19 and INT64_MAX ~9.2e18), so strtoull()/strtoll()
+        // could not have set errno to ERANGE for it. Numbers with more digits
+        // (rare in practice) fall through to the exact code below, unchanged,
+        // so their handling -- including reclassification to value_float when
+        // the value overflows 64 bits, and rejection when it is not even
+        // finite as a double -- is bit-for-bit identical to before this
+        // optimization.
+        if (discard_number_values)
+        {
+            constexpr std::size_t safe_digit_count = 18;
+            if (number_type == token_type::value_unsigned && token_buffer.size() <= safe_digit_count)
+            {
+                return token_type::value_unsigned;
+            }
+            if (number_type == token_type::value_integer && token_buffer.size() - 1 <= safe_digit_count)
+            {
+                return token_type::value_integer;
+            }
+        }
+
         char* endptr = nullptr; // NOLINT(misc-const-correctness,cppcoreguidelines-pro-type-vararg,hicpp-vararg)
         errno = 0;
 
@@ -1754,6 +1787,13 @@ scan_number_done:
     const char_int_type decimal_point_char = '.';
     /// the position of the decimal point in the input
     std::size_t decimal_point_position = std::string::npos;
+
+    /// whether the caller (e.g. accept()/json_sax_acceptor) only needs the
+    /// token classification and never looks at the converted numeric value;
+    /// when set, scan_number() may skip strtoull()/strtoll() for
+    /// value_unsigned/value_integer tokens whose digit count guarantees they
+    /// fit into 64 bits (see scan_number())
+    const bool discard_number_values = false;
 };
 
 }  // namespace detail
--- a/include/nlohmann/detail/input/lexer.hpp
+++ b/include/nlohmann/detail/input/lexer.hpp
@@ -1299,6 +1299,26 @@ scan_number_done:
         // the value overflows 64 bits, and rejection when it is not even
         // finite as a double -- is bit-for-bit identical to before this
         // optimization.
+        //
+        // Note this reasons about std::uint64_t/std::int64_t, not about
+        // number_unsigned_t/number_integer_t (BasicJsonType's own, possibly
+        // narrower, template parameters -- e.g. std::uint32_t). That is fine
+        // *only* because discard_number_values is exclusively set by
+        // accept() (see json.hpp), and accept() always parses through the
+        // library's own json_sax_acceptor -- never a user-supplied SAX
+        // consumer -- whose number_unsigned()/number_integer()/number_float()
+        // callbacks unconditionally discard their argument and return true.
+        // So for every caller that can reach this branch, neither the token
+        // classification below nor the eventual (possibly narrowed, and on
+        // this fast path left stale/unset) value_unsigned/value_integer is
+        // ever consulted -- an unsigned/integer token is accepted outright,
+        // and even a >18-digit token that this fast path deliberately falls
+        // through for is, once reclassified to value_float, still finite
+        // (and thus accepted) for any digit count that fits in number_unsigned_t
+        // or number_integer_t regardless of that type's width. If this
+        // function is ever taught to run with discard_number_values true for
+        // a caller that *does* read the converted value, this reasoning (and
+        // the fast path below) would need to be revisited.
         if (discard_number_values)
         {
             constexpr std::size_t safe_digit_count = 18;
--- a/include/nlohmann/detail/input/parser.hpp
+++ b/include/nlohmann/detail/input/parser.hpp
@@ -72,9 +72,10 @@ class parser
                     parser_callback_t<BasicJsonType> cb = nullptr,
                     const bool allow_exceptions_ = true,
                     const bool ignore_comments = false,
-                    const bool ignore_trailing_commas_ = false)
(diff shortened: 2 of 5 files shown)


nlohmann:
Verified with a local benchmark (`json::accept()` over a 100k-element JSON array, `clang++ -O2 -DNDEBUG`, Apple Clang 21 / arm64 macOS, median of 80 timed iterations, comparing `develop` @ 09b6b6b5b vs this branch's tip):

- Integer array (100k plain integers, ~1.04 MB): 4.88 ms before → 3.41 ms after (about 30% faster)
- Float array (100k floats, ~1.14 MB): 5.71 ms before → 5.86 ms after (no measurable change, within noise — expected, since the fast path only applies to integer/unsigned tokens)

The integer-array speedup matches the PR's intent: skipping `strtoull()`/`strtoll()` for `accept()`-only parses whose digit count already guarantees no 64-bit overflow, while leaving float parsing and the exact/value-returning path untouched.

include/nlohmann/detail/input/lexer.hpp
```diff
@@ -1279,6 +1280,38 @@ class lexer : public lexer_base<BasicJsonType>
...
+        // both UINT64_MAX ~1.8e19 and INT64_MAX ~9.2e18), so strtoull()/strtoll()
+        // could not have set errno to ERANGE for it. Numbers with more digits
+        // (rare in practice) fall through to the exact code below, unchanged,
+        // so their handling -- including reclassification to value_float when
+        // the value overflows 64 bits, and rejection when it is not even
+        // finite as a double -- is bit-for-bit identical to before this
+        // optimization.
+        if (discard_number_values)
+        {
+            constexpr std::size_t safe_digit_count = 18;
```
gregmarr, line 1324: Can this ever be called with the number type being 32 bit?

nlohmann: Good catch, and yes — `number_unsigned_t`/`number_integer_t` are `BasicJsonType`'s own template parameters, so a custom `basic_json` could make them narrower than 64 bits (e.g. `std::uint32_t`).

That said, it doesn't matter here, because `discard_number_values` can only ever be `true` when it originates from `accept()` (see the three `accept()` overloads in json.hpp, which all call `parser(..., true).accept(true)`), and `parser::accept()` is hardcoded to always parse through the library's own `json_sax_acceptor` — never a user-supplied SAX consumer. That acceptor's `number_unsigned()`/`number_integer()`/`number_float()` callbacks unconditionally discard their argument and return `true` (its parameters are literally named `/*unused*/`). So for every caller that can actually reach this branch:

- An unsigned/integer classification is accepted outright regardless of the actual value or the width it would have narrowed to.
- Even a token this fast path *doesn't* short-circuit for (>18 digits) that gets reclassified to `value_float` by the unchanged code below is still finite for any digit count that could fit in a 32-bit (or narrower) `number_unsigned_t`/`number_integer_t`, so it'd be accepted either way too.

So the 64-bit reasoning in the comment is about `std::uint64_t`/`std::int64_t` specifically (what `strtoull`/`strtoll` themselves operate on), not about `BasicJsonType`'s configured integer width — and that's fine only because of the accept()-only, json_sax_acceptor-only coupling above, not because the digit-count math itself accounts for narrower types.

I pushed ef97a360c which spells this invariant out explicitly in the code comment, including a note that this reasoning would need revisiting if `discard_number_values` were ever wired up for a caller that *does* read the converted value. Thanks for pushing on this — it's exactly the kind of assumption that's easy to leave implicit.

gregmarr (approved):

Issue #5103, opened 2026-03-09
Build issue when compiling with gcc (any modules-capable version) in a C++ modules project


sebsjames:
### Description

This may be a gcc bug or an nlohmann::json bug - I suspect gcc though, because clang handles the code that I'm going to show. Nevertheless, I wanted to flag it here, as I know that modules support is complex and it might be useful to do so (I'll also report it to the gcc bug list).

To reproduce, you can build my minimal project at https://github.com/sebsjames/gcc15_modules_test with g++.

This compiles a couple of C++ modules, in one of which is `#included` nlohman::json.

The problem file is **test1.cppm**. This `#includes` **has_tools.h**, which `imports` a module called `mplot.tools`. 

```c++
module;

#include <has_tools.h> // not ok if tools is in this #included file and comes before nlohmann
#include <nlohmann/json.hpp>
// #include <has_tools.h> // ok if #included AFTER nlohmann json

export module test1;

// import mplot.tools; // ok if imported here

export namespace test
{
    class tt
    {
    public:
        tt() { this->init(); }
        ~tt() {}
        float get_value() const { return this->value; }
    protected:
        void init() { this->value = 2.0f; }
        float value = 0.0f;
    };
}
```

I have a workaround, which is to change the order of the included files, but I don't think that should be necessary.

### Reproduction steps

Prerequisites: cmake version 3.28.5 or higher, Ninja and g++ 15 or the 16 pre-release from their master branch. Then:
```bash
git clone --recursive https://github.com/sebsjames/gcc15_modules_test.git # pulls in nlohmann from develop branch
cd gcc15_modules_test
mkdir bgcc
cd bgcc
CXX=g++-15 cmake .. -GNinja
ninja
```

### Expected vs. actual results

The error output when g++ tries to compile that module is:
```
[11:46:26 gcc15_modules_test] ./build_g++.sh
In file included from /opt/gcc-15/include/c++/15.2.1/algorithm:63,
                 from json/include/nlohmann/json.hpp:21,
                 from test1.cppm:4:
/opt/gcc-15/include/c++/15.2.1/bits/stl_algo.h:5633:5: error: redefinition of ‘template<class _ForwardIterator, class _Compare> constexpr _ForwardIterator std::__min_element(_ForwardIterator, _ForwardIterator, _Compare)’
 5633 |     __min_element(_ForwardIterator __first, _ForwardIterator __last,
      |     ^~~~~~~~~~~~~
compilation terminated due to -Wfatal-errors.
[...]

PR #5485, opened 2026-09-05
Route hand-rolled diagnostic pragmas through Hedley

### LINKED ISSUE #5409: Route hand-rolled compiler diagnostic pragmas through Hedley (JSON_HEDLEY_DIAGNOSTIC_PUSH/POP, HAS_WARNING, PRAGMA)
### Description

The library bundles Hedley and uses many of its macros, but **none** of Hedley's diagnostic-control macros are used in application code. Every warning suppression is instead hand-rolled with a compiler `#if` guard plus a raw `#pragma`. There are 7 such blocks:

| location (push) | warning suppressed | guard |
|---|---|---|
| [`json.hpp:27`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json.hpp) / `:5474` | `-Wignored-attributes` | `__GNUC__ && !__clang__ && C++20` |
| [`json.hpp:3770`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json.hpp) | `-Wfloat-equal` | `__GNUC__` |
| [`json.hpp:3863`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json.hpp) | `-Wfloat-equal` | `__GNUC__` |
| [`detail/conversions/to_chars.hpp:1078`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/conversions/to_chars.hpp) | `-Wfloat-equal` | `__GNUC__` |
| [`detail/output/binary_writer.hpp:1844`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/output/binary_writer.hpp) | `-Wfloat-equal` | `__GNUC__` |
| [`detail/iterators/iteration_proxy.hpp:211`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/iterators/iteration_proxy.hpp) | `-Wmismatched-tags` | `__clang__` |
| [`detail/exceptions.hpp:36`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/exceptions.hpp) | `-Wweak-vtables` | `__clang__` |

### Proposal

Use the Hedley primitives the library already ships:

- `JSON_HEDLEY_DIAGNOSTIC_PUSH` / `JSON_HEDLEY_DIAGNOSTIC_POP` — portable push/pop (dispatches to `#pragma GCC/clang diagnostic` or `#pragma warning` per compiler).
- `JSON_HEDLEY_HAS_WARNING("-Wxxx")` — wraps `__has_warning` where available, with version-based fallbacks.
- `JSON_HEDLEY_PRAGMA(...)` — portable `_Pragma` wrapper for the specific `ignored` line.

Example (`to_chars.hpp`):

```cpp
// current
#ifdef __GNUC__
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wfloat-equal"
#endif
    if (value == 0) { ... }
#ifdef __GNUC__
#pragma GCC diagnostic pop
#endif

// proposed
JSON_HEDLEY_DIAGNOSTIC_PUSH
#if JSON_HEDLEY_HAS_WARNING("-Wfloat-equal")
    JSON_HEDLEY_PRAGMA(GCC diagnostic ignored "-Wfloat-equal")
#endif
    if (value == 0) { ... }
JSON_HEDLEY_DIAGNOSTIC_POP
```

### Benefits

1. **Consistency** — Hedley is a bundled dependency whose diagnostic layer is currently entirely unused; the codebase already relies on Hedley for `LIKELY`, `NON_NULL`, `WARN_UNUSED_RESULT`, `DEPRECATED_FOR`, etc.
2. **More precise gating** — `#ifdef __GNUC__` is also true for compilers that impersonate GCC (ICC classic, nvhpc, some others) but may not recognize `-Wfloat-equal`, which can itself provoke an "unknown warning group" diagnostic under `-Wpragmas`. `JSON_HEDLEY_HAS_WARNING` checks the warning actually exists before disabling it.

### Notes / caveats

- Hedley has **no** dedicated `DISABLE_*` macro for `-Wfloat-equal`, `-Wmismatched-tags`, `-Wweak-vtables`, or `-Wignored-attributes`, so the specific `ignored` line still needs a raw pragma — but wrapped in `HAS_WARNING`/`PRAGMA` rather than a bare `#ifdef`.
- This does **not** apply to `meta()` (`json.hpp:305-340`), which builds human-readable compiler *version strings* — Hedley's version macros are for conditionals, not string production — nor to the GCC spaceship-selection workaround in `detail/value_t.hpp:110` (#105200), which is a functional `#if`, not a diagnostic. Both are correctly hand-rolled and should stay.
- Behavior-preserving: same warnings suppressed in the same regions; output unchanged.

g++ 13.3.0 (Ubuntu 24.04, x86-64) — general/portability improvement

### Library version

nlohmann:
Fixes #5409.

Stacked on top of #5475 (`issue-5408-hedley-undef-leak`) since both PRs touch some of the same files. Only the last commit on this branch (`Route hand-rolled diagnostic pragmas through Hedley`) is new; the base commit is #5475's fix. This PR should be reviewed/merged after #5475 (or rebased onto `develop` once #5475 lands).

## Problem

Several places in the library hand-roll compiler diagnostic suppression with raw `#pragma`/`#ifdef __GNUC__`/`#ifdef __clang__` guards, instead of using the Hedley primitives already bundled and used elsewhere (`JSON_HEDLEY_LIKELY`, `WARN_UNUSED_RESULT`, `DEPRECATED_FOR`, etc.).

## Changes

Converted 6 of the 7 listed push/pop pairs to use `JSON_HEDLEY_DIAGNOSTIC_PUSH`/`JSON_HEDLEY_DIAGNOSTIC_POP` and `JSON_HEDLEY_PRAGMA(...)` instead of raw pragma text:

- `include/nlohmann/json.hpp` (~3770, ~3863): `-Wfloat-equal`
- `include/nlohmann/detail/conversions/to_chars.hpp` (~1078): `-Wfloat-equal`
- `include/nlohmann/detail/output/binary_writer.hpp` (~1844): `-Wfloat-equal`
- `include/nlohmann/detail/iterators/iteration_proxy.hpp` (~211): `-Wmismatched-tags`
- `include/nlohmann/detail/exceptions.hpp` (~36): `-Wweak-vtables`

`iteration_proxy.hpp` didn't previously `#include <nlohmann/detail/macro_scope.hpp>` itself — it only compiled because some earlier header in `json.hpp`'s include list happened to pull it in first. It now includes it directly, like the other detail headers that use Hedley macros, so it's self-contained.

Each pair now calls `JSON_HEDLEY_DIAGNOSTIC_PUSH`/`POP` unconditionally (a no-op on compilers that don't need it — that's exactly what the macro is for) and wraps the actual `... diagnostic ignored "..."` text in `JSON_HEDLEY_PRAGMA(...)`, going through Hedley's `_Pragma()`-based emission instead of a raw `#pragma` line — while **keeping** the original `#ifdef __GNUC__` / `#if defined(__clang__)` guard around the ignored-pragma itself.

## Deviation from the issue's suggested transformation

The issue's example replaces the `#ifdef __GNUC__` guard with `#if JSON_HEDLEY_HAS_WARNING("-Wfloat-equal")`. I did not adopt that part: `JSON_HEDLEY_HAS_WARNING` is implemented purely via Clang's `__has_warning` builtin —

```c
#if defined(__has_warning)
    #define JSON_HEDLEY_HAS_WARNING(warning) __has_warning(warning)
#else
    #define JSON_HEDLEY_HAS_WARNING(warning) (0)
#endif
```

— so it evaluates to `0` on real GCC (which doesn't define `__has_warning`). Adopting it verbatim would silently **stop suppressing `-Wfloat-equal` on GCC**, which is a real regression, not just a style change (the comment even says the guard exists specifically for GCC's behavior around `-Wfloat-equal` on `NaN`/`0.0` comparisons in `constexpr`-friendly code). I kept the existing `#ifdef __GNUC__` / `#if defined(__clang__)` guards around the ignored-pragma so the change stays strictly behavior-preserving, and only routed the push/pop/pragma-emission mechanism itself through Hedley, per the issue's actual stated goal ("This must be behavior-preserving").

## Two locations intentionally left unconverted

The `-Wignored-attributes` push at the very top of `json.hpp` and its matching pop after `#include <nlohmann/detail/macro_unscope.hpp>` (the workaround for #5103) were **not** converted:

- The push runs before `detail/macro_scope.hpp` (and therefore `hedley.hpp`) has been included anywhere in the translation unit — it's literally the first thing in the file. `JSON_HEDLEY_DIAGNOSTIC_PUSH` is not yet defined at that point.
- The pop runs *after* `macro_unscope.hpp`, which — via `hedley_undef.hpp` — has already `#undef`'d every `JSON_HEDLEY_*` macro (by design, so they don't leak to users, see #5408/#5475). `JSON_HEDLEY_DIAGNOSTIC_POP` is no longer defined by the time the pop is reached either.

Making this one pair route through Hedley would require either hoisting the ~2000-line vendored `hedley.hpp` to the very top of the amalgamated single header (a much bigger structural change to `single_include` than a pure mechanism swap) or special-casing this one pop ahead of the general macro cleanup at the end of the file. Both are riskier than the mechanical, behavior-preserving change the issue asks for, so I left this one pair as raw pragmas.

## Validation

- Compiled `include/nlohmann/json.hpp` and `single_include/nlohmann/json.hpp` with `-Wall -Wextra -Wfloat-equal -Wmismatched-tags -Wweak-vtables` (Clang, which also self-identifies as `__GNUC__`): no warnings, same as before the change.
- Compiled and ran `unit-to_chars.cpp`, `unit-conversions.cpp`, `unit-iterators1.cpp`, `unit-iterators2.cpp`, and `unit-class_parser.cpp`: all pass.
- Compiled `unit-msgpack.cpp`, `unit-bjdata.cpp`, and `unit-ubjson.cpp` (which exercise `binary_writer.hpp`'s `write_compact_float` extensively): all compile cleanly, and the overwhelming majority of assertions pass (the only failures are pre-existing, unrelated to this change — missing generated test-data files in this offline environment).
- `make amalgamate` was run; the `single_include` diff is limited to exactly the lines touched in `include/`, with no unrelated reordering.
- No real (non-Apple) GCC was available in this sandbox to test directly. The `_Pragma("GCC diagnostic ...")` text `JSON_HEDLEY_PRAGMA` emits is byte-identical to the previous raw `#pragma GCC diagnostic ...` text, and the surrounding `#ifdef __GNUC__` guard is unchanged, so GCC's behavior should be identical; CI covers the full GCC matrix.

## Breaking change?

No breaking changes. This is purely an internal mechanism change (compiler-diagnostic suppression plumbing); no public API, behavior, or emitted warnings change for any user of the library.

### DIFF
--- a/include/nlohmann/detail/conversions/to_chars.hpp
+++ b/include/nlohmann/detail/conversions/to_chars.hpp
@@ -1074,9 +1074,9 @@ char* to_chars(char* first, const char* last, FloatType value)
         *first++ = '-';
     }
 
+    JSON_HEDLEY_DIAGNOSTIC_PUSH
 #ifdef __GNUC__
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wfloat-equal"
+    JSON_HEDLEY_PRAGMA(GCC diagnostic ignored "-Wfloat-equal")
 #endif
     if (value == 0) // +-0
     {
@@ -1086,9 +1086,7 @@ char* to_chars(char* first, const char* last, FloatType value)
         *first++ = '0';
         return first;
     }
-#ifdef __GNUC__
-#pragma GCC diagnostic pop
-#endif
+    JSON_HEDLEY_DIAGNOSTIC_POP
 
     JSON_ASSERT(last - first >= std::numeric_limits<FloatType>::max_digits10);
 
--- a/include/nlohmann/detail/conversions/to_chars.hpp
+++ b/include/nlohmann/detail/conversions/to_chars.hpp
@@ -1074,8 +1074,8 @@ char* to_chars(char* first, const char* last, FloatType value)
         *first++ = '-';
     }
 
-    JSON_HEDLEY_DIAGNOSTIC_PUSH
 #ifdef __GNUC__
+    JSON_HEDLEY_DIAGNOSTIC_PUSH
     JSON_HEDLEY_PRAGMA(GCC diagnostic ignored "-Wfloat-equal")
 #endif
     if (value == 0) // +-0
@@ -1086,7 +1086,9 @@ char* to_chars(char* first, const char* last, FloatType value)
         *first++ = '0';
         return first;
     }
+#ifdef __GNUC__
     JSON_HEDLEY_DIAGNOSTIC_POP
+#endif
 
     JSON_ASSERT(last - first >= std::numeric_limits<FloatType>::max_digits10);
 
--- a/include/nlohmann/detail/exceptions.hpp
+++ b/include/nlohmann/detail/exceptions.hpp
@@ -32,9 +32,9 @@
 // functions to. As a result, we suppress this warning here to avoid client
 // code stumbling over this. See https://github.com/nlohmann/json/issues/4087
 // for a discussion.
+JSON_HEDLEY_DIAGNOSTIC_PUSH
 #if defined(__clang__)
-    #pragma clang diagnostic push
-    #pragma clang diagnostic ignored "-Wweak-vtables"
+    JSON_HEDLEY_PRAGMA(clang diagnostic ignored "-Wweak-vtables")
 #endif
 
 NLOHMANN_JSON_NAMESPACE_BEGIN
@@ -286,6 +286,4 @@ class other_error : public exception
 }  // namespace detail
 NLOHMANN_JSON_NAMESPACE_END
 
-#if defined(__clang__)
-    #pragma clang diagnostic pop
-#endif
+JSON_HEDLEY_DIAGNOSTIC_POP
--- a/include/nlohmann/detail/exceptions.hpp
+++ b/include/nlohmann/detail/exceptions.hpp
@@ -32,8 +32,8 @@
 // functions to. As a result, we suppress this warning here to avoid client
 // code stumbling over this. See https://github.com/nlohmann/json/issues/4087
 // for a discussion.
-JSON_HEDLEY_DIAGNOSTIC_PUSH
 #if defined(__clang__)
+    JSON_HEDLEY_DIAGNOSTIC_PUSH
     JSON_HEDLEY_PRAGMA(clang diagnostic ignored "-Wweak-vtables")
 #endif
 
@@ -286,4 +286,6 @@ class other_error : public exception
 }  // namespace detail
 NLOHMANN_JSON_NAMESPACE_END
 
-JSON_HEDLEY_DIAGNOSTIC_POP
+#if defined(__clang__)
+    JSON_HEDLEY_DIAGNOSTIC_POP
+#endif
--- a/include/nlohmann/detail/iterators/iteration_proxy.hpp
+++ b/include/nlohmann/detail/iterators/iteration_proxy.hpp
@@ -18,6 +18,7 @@
 #endif
 
 #include <nlohmann/detail/abi_macros.hpp>
+#include <nlohmann/detail/macro_scope.hpp>
 #include <nlohmann/detail/meta/type_traits.hpp>
 #include <nlohmann/detail/string_utils.hpp>
 #include <nlohmann/detail/value_t.hpp>
@@ -206,10 +207,10 @@ NLOHMANN_JSON_NAMESPACE_END
 namespace std
 {
 
+// Fix: https://github.com/nlohmann/json/issues/1401
+JSON_HEDLEY_DIAGNOSTIC_PUSH
 #if defined(__clang__)
-    // Fix: https://github.com/nlohmann/json/issues/1401
-    #pragma clang diagnostic push
-    #pragma clang diagnostic ignored "-Wmismatched-tags"
+    JSON_HEDLEY_PRAGMA(clang diagnostic ignored "-Wmismatched-tags")
 #endif
 template<typename IteratorType>
 class tuple_size<::nlohmann::detail::iteration_proxy_value<IteratorType>> // NOLINT(cert-dcl58-cpp)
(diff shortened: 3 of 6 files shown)


include/nlohmann/detail/conversions/to_chars.hpp
```diff
@@ -1074,9 +1074,9 @@ char* to_chars(char* first, const char* last, FloatType value)
         *first++ = '-';
     }
 
+    JSON_HEDLEY_DIAGNOSTIC_PUSH
```
gregmarr: What is the effect on compile time of having this here and all the other places on compilers other than `__GNUC__` that have diagnostic push and thus do something unnecessary here?

nlohmann: Good catch — this was an oversight, not intentional. `JSON_HEDLEY_DIAGNOSTIC_PUSH`/`POP` were unconditional while only the `JSON_HEDLEY_PRAGMA(...)` in between was guarded, so on any compiler where Hedley's push/pop expands to something real but the guard doesn't apply (MSVC, or the `__clang__`-guarded spots on non-Clang GCC), we'd push and immediately pop a diagnostic state with nothing suppressed in between — pure overhead with no benefit, unlike before this PR where the whole pair was inside the guard and non-GNU/Clang compilers saw nothing at all here. Moved the `#ifdef __GNUC__` / `#if defined(__clang__)` guard to also cover the push/pop in all 6 spots, restoring the original zero-cost behavior on other compilers while still emitting the pragma itself through Hedley. Pushed.

gregmarr (approved):

PR #5490, opened 2026-09-05
Speed up whitespace skipping in the lexer

### LINKED ISSUE #5412: Whitespace between tokens is skipped one byte at a time; bulk-skipping speeds up parse()/accept() on pretty-printed input (~25% of accept time)
### Description

Inter-token whitespace is consumed one character at a time ([`lexer.hpp:1615`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/detail/input/lexer.hpp)):

```cpp
void skip_whitespace()
{
    do
    {
        get();
    }
    while (current == ' ' || current == '\t' || current == '\n' || current == '\r');
}
```

`skip_whitespace()` runs before every token, and each `get()` also does per-byte position bookkeeping (`chars_read_total`, `chars_read_current_line`, line tracking on `\n`). For pretty-printed / indented input — a very common shape (config files, indented API payloads) — whitespace can be the majority of the bytes, so this loop is a meaningful fraction of parse/accept time.

### Measurement

g++ 13.3.0 `-O2`, `accept()` over the same data compact vs. `dump(4)`:

| input | size | `accept` | throughput |
|---|---|---|---|
| compact | 870 KB | 6.49 ms | 134 MB/s |
| pretty (`dump(4)`) | 2.63 MB (3.0×) | 8.70 ms | 302 MB/s |

The 1.76 MB of added whitespace is processed at ~800 MB/s and accounts for ~2.2 ms — roughly **25% of `accept()` time** on this pretty-printed document — all of it in the byte-at-a-time skip loop.

### Suggested direction

For contiguous input, skip whitespace in bulk: scan a run of `` ``/`\t`/`\n`/`\r` in the buffer (a small SWAR/`memspn`-style test) and update the position counters once per run (counting `\n`s in the run for line/column tracking) instead of once per byte. This reuses exactly the contiguous bulk-scan infrastructure that the open parser-performance PR #5283 introduces (`supports_bulk_scan` / `bulk_data()` / `bulk_skip()` on the iterator adapter) — that PR speeds up strings, numbers, and UTF-8 but does **not** touch whitespace skipping, so this is an additive extension best implemented on top of it. Streaming adapters keep the byte-at-a-time loop.

### Behavior preservation

Whitespace is semantically inert, so only throughput changes — same accept/reject, same values, same error byte offsets. Care is needed only to keep line/column bookkeeping identical when a run contains newlines (relevant to error positions and `JSON_DIAGNOSTIC_POSITIONS`); a differential test of error positions across compact vs. pretty inputs should gate it.

### Library version

nlohmann:
## Summary

Fixes #5412.

**Stacked on top of #5484 (fix for #5411)** — this PR's diff only makes sense on top of that branch; only the last commit here is new.

`lexer::skip_whitespace()` calls `get()` once per whitespace byte. `get()` itself checks a `next_unget` flag on every call to see whether it should replay a previously-ungotten character rather than reading a fresh one from the input adapter. That check only ever matters for the *first* character `skip_whitespace()` reads (a previous token, e.g. a number, may have ended with `unget()`, leaving one pending character to replay) — nothing inside `skip_whitespace()`'s loop itself calls `unget()`, so `next_unget` is provably `false` for every whitespace character after the first.

This PR reads the first character with the existing `get()` (unchanged), and every subsequent whitespace character with a new `get_ignoring_pending_unget()` that shares `get()`'s bookkeeping tail but skips the now-provably-dead `next_unget` branch. Behavior is unchanged for the first character read for every token as before.

### Scope and what was deliberately left out

The issue's suggested direction is a full contiguous-buffer bulk skip: scan a run of whitespace directly in the input adapter's buffer with a SWAR/`memspn`-style scan, and update the position counters once per run instead of once per byte. That depends on bulk-scan adapter infrastructure (`supports_bulk_scan`/`bulk_data()`/`bulk_skip()`) that the issue explicitly says is introduced by the **open, unmerged** parser-performance PR #5283. Per the scoping for this fix, I did not build that adapter infrastructure from scratch here, and did not depend on or replicate #5283. I judged writing new bulk-scan/adapter-peeking infrastructure targeted at a security-sensitive parser, without the review #5283 itself is still undergoing, to be out of scope for this PR.

Instead, this PR is limited to the safe, always-correct improvement described in the fallback scoping for this issue: removing redundant per-character bookkeeping from the existing byte-at-a-time loop. Full bulk-skipping is left as future work once #5283 (or equivalent contiguous-adapter support) lands upstream.

Given this narrower scope, the throughput improvement here is real but far more modest than the ~25% of `accept()` time cited in the issue for the full bulk-skip approach (which comes from amortizing per-byte counter updates across whole whitespace runs, not from removing one branch per byte).

## Behavior preservation / testing

- Line/column/byte-offset bookkeeping (`chars_read_total`, `chars_read_current_line`, `lines_read`) is completely untouched by this change — `get_ignoring_pending_unget()` performs the exact same steps `get()` does when `next_unget` is (provably) `false`.
- Added a differential regression test in `tests/src/unit-class_parser.cpp` (`SECTION("issue #5412 - whitespace skipping bookkeeping (compact vs. pretty-printed)")`) asserting the exact byte offset, line, column, and error message for parse errors in both compact and pretty-printed (`dump(4)`) malformed input, including input with an error after multiple embedded newlines/indentation.
- Verified with a standalone harness that error `byte`/`what()` output is byte-for-byte identical before and after this change across compact and pretty-printed inputs (including inputs truncated right before the final closing brace, and inputs with an invalid token after several indented lines).
- Ran the full required test files locally (offline, without the `json_test_data` submodule available in this environment): `unit-class_lexer.cpp`, `unit-class_parser.cpp`, `unit-deserialization.cpp`, `unit-regression1.cpp`, `unit-regression2.cpp`, `unit-udt.cpp`, `unit-class_parser_diagnostic_positions.cpp` (9234 assertions, all pass) — all pass, with the same 7 pre-existing `unit-regression1.cpp` failures present identically on unmodified `develop` in this offline environment (missing `json_test_data` fixture files; unrelated to this change).
- `make amalgamate` was run and `single_include/nlohmann/json.hpp` is included in this PR.

## Breaking change?

No breaking changes. This only adds a new private helper method to `nlohmann::detail::lexer` and changes the internal implementation of `skip_whitespace()`; no public API is touched.

### DIFF
--- a/include/nlohmann/detail/input/lexer.hpp
+++ b/include/nlohmann/detail/input/lexer.hpp
@@ -1705,9 +1705,20 @@ scan_number_done:
         // nothing below calls unget()
         get();
 
-        while (current == ' ' || current == '\t' || current == '\n' || current == '\r')
+        // this is written as an if-guarded do-while (rather than a plain
+        // while loop) because that shape is what lets both GCC and Clang
+        // keep the input adapter's read pointer in a register across
+        // iterations; the equivalent while-loop measurably defeated that
+        // optimization in testing, turning long whitespace runs (e.g. the
+        // indentation of pretty-printed JSON) from a register-only loop
+        // into one that reloads the pointer from memory every character
+        if (current == ' ' || current == '\t' || current == '\n' || current == '\r')
         {
-            get_ignoring_pending_unget();
+            do
+            {
+                get_ignoring_pending_unget();
+            }
+            while (current == ' ' || current == '\t' || current == '\n' || current == '\r');
         }
     }
 
--- a/include/nlohmann/detail/input/lexer.hpp
+++ b/include/nlohmann/detail/input/lexer.hpp
@@ -1459,6 +1459,14 @@ scan_number_done:
             current = ia.get_character();
         }
 
+        return track_after_read();
+    }
+
+    /// shared tail of get() / get_ignoring_pending_unget(): capture the
+    /// character for error messages (if needed) and update line/column
+    /// bookkeeping for the character now in `current`
+    char_int_type track_after_read()
+    {
         // seekable adapters reconstruct the token lazily on error (see
         // get_token_string), so the eager per-character copy is skipped
         capture_char(std::integral_constant<bool, lazy_token_string> {});
@@ -1470,6 +1478,30 @@ scan_number_done:
         }
 
         return current;
+    }
+
+    /*!
+    @brief like get(), but for call sites that can prove no unget() is pending
+
+    get() has to check the `next_unget` flag on every call, because a
+    previous token may have ended with unget() (e.g. scan_number() always
+    ungets the character that terminated the number, so the next call to
+    scan() can see it again). skip_whitespace() reads that first,
+    possibly-ungotten character via a plain get(), but every further
+    character it reads is guaranteed to be a fresh read: nothing between
+    those calls invokes unget(). This variant skips the (otherwise always
+    false) next_unget branch for those calls; it is not a general
+    replacement for get().
+    */
+    char_int_type get_ignoring_pending_unget()
+    {
+        JSON_ASSERT(!next_unget);
+
+        ++position.chars_read_total;
+        ++position.chars_read_current_line;
+        current = ia.get_character();
+
+        return track_after_read();
     }
 
     /// seekable adapter: nothing to capture, the token is rebuilt on error
@@ -1667,11 +1699,16 @@ scan_number_done:
 
     void skip_whitespace()
     {
-        do
+        // the first character may be a pending unget() left over from the
+        // previous token (see get_ignoring_pending_unget()); every
+        // subsequent character read by this loop is guaranteed fresh, since
+        // nothing below calls unget()
+        get();
+
+        while (current == ' ' || current == '\t' || current == '\n' || current == '\r')
         {
-            get();
+            get_ignoring_pending_unget();
         }
-        while (current == ' ' || current == '\t' || current == '\n' || current == '\r');
     }
 
     token_type scan()
--- a/include/nlohmann/detail/input/lexer.hpp
+++ b/include/nlohmann/detail/input/lexer.hpp
@@ -1704,6 +1704,12 @@ scan_number_done:
         return true;
     }
 
+    /// whether `current` is one of the four JSON whitespace characters
+    bool current_is_whitespace() const noexcept
+    {
+        return current == ' ' || current == '\t' || current == '\n' || current == '\r';
+    }
+
     void skip_whitespace()
     {
         // the first character may be a pending unget() left over from the
@@ -1712,6 +1718,11 @@ scan_number_done:
         // nothing below calls unget()
         get();
 
+        if (!current_is_whitespace())
+        {
+            return;
+        }
+
         // this is written as an if-guarded do-while (rather than a plain
         // while loop) because that shape is what lets both GCC and Clang
         // keep the input adapter's read pointer in a register across
@@ -1719,14 +1730,11 @@ scan_number_done:
         // optimization in testing, turning long whitespace runs (e.g. the
         // indentation of pretty-printed JSON) from a register-only loop
         // into one that reloads the pointer from memory every character
-        if (current == ' ' || current == '\t' || current == '\n' || current == '\r')
+        do
         {
-            do
-            {
-                get_ignoring_pending_unget();
-            }
-            while (current == ' ' || current == '\t' || current == '\n' || current == '\r');
+            get_ignoring_pending_unget();
         }
+        while (current_is_whitespace());
     }
 
     token_type scan()
--- a/include/nlohmann/detail/input/lexer.hpp
+++ b/include/nlohmann/detail/input/lexer.hpp
@@ -1446,8 +1446,7 @@ scan_number_done:
     */
     char_int_type get()
     {
-        ++position.chars_read_total;
-        ++position.chars_read_current_line;
+        advance_position();
 
         if (next_unget)
         {
@@ -1460,6 +1459,15 @@ scan_number_done:
         }
 
         return track_after_read();
+    }
+
+    /// shared head of get() / get_ignoring_pending_unget(): bump the
+    /// per-character position counters (line-count-on-'\n' bookkeeping is
+    /// handled afterwards, in track_after_read(), once `current` is known)
+    void advance_position() noexcept
+    {
+        ++position.chars_read_total;
+        ++position.chars_read_current_line;
     }
 
     /// shared tail of get() / get_ignoring_pending_unget(): capture the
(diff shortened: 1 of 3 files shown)


nlohmann:
Verified with a local benchmark (`json::accept()` over the same ~20k-object document rendered compact vs. `dump(4)`-pretty-printed, `clang++ -O2 -DNDEBUG`, Apple Clang 21 / arm64 macOS, median of 150 timed iterations, comparing this branch's base (`issue-5411-lexer-skip-conversion` tip) vs. this branch's tip):

- Compact (1.27 MB, no whitespace): 4.89 ms before → 4.74 ms after (about 3% faster)
- Pretty-printed (2.27 MB, ~1 MB of added whitespace): 6.13 ms before → 6.87 ms after (about 12% **slower**); overhead per added whitespace byte went from ~0.00000122 ms/byte to ~0.00000213 ms/byte (worse, not better)

I also tried a deeply-nested document (300 levels, indentation runs up to ~1200 spaces) to specifically stress long contiguous whitespace runs: pretty-printed `accept()` went from 0.74 ms before to 2.29 ms after (about 3x **slower**).

So on this machine/compiler I could not reproduce the claimed improvement for the whitespace-heavy case — it looks like a regression here, while compact (whitespace-free) input is marginally faster. Results were highly reproducible (<2% run-to-run variance) across `-O2`/`-O3` and with/without `-DNDEBUG`. This could be compiler/platform-specific (e.g. how Apple Clang 21 inlines/schedules the new `get_ignoring_pending_unget()` split), so it may be worth re-checking against your usual benchmark setup — but wanted to report the numbers as measured rather than assume they'd match the PR description.

include/nlohmann/detail/input/lexer.hpp
```diff
@@ -1452,6 +1460,30 @@ class lexer : public lexer_base<BasicJsonType>
...
+    character it reads is guaranteed to be a fresh read: nothing between
+    those calls invokes unget(). This variant skips the (otherwise always
+    false) next_unget branch for those calls; it is not a general
+    replacement for get().
+    */
+    char_int_type get_ignoring_pending_unget()
+    {
+        JSON_ASSERT(!next_unget);
+
+        ++position.chars_read_total;
```
gregmarr: Should these lines go in something like `advance_position` so that they're shared by both functions, just like the tail is?

nlohmann:
A follow-up benchmark (Apple Clang 21 arm64 and Homebrew GCC 12, both `-O2`/`-DNDEBUG`, macOS) found that this PR, as originally written, was a **regression**, not a speedup, on pretty-printed / deeply-nested input:

| case | Clang before (#5484 tip) | Clang PR (before this fix) | GCC before | GCC PR (before this fix) |
|---|---|---|---|---|
| 300-level deep nesting, `dump(4)` | 0.24 ms | 0.79 ms (**3.2x slower**) | 0.77 ms | 1.34 ms (**1.75x slower**) |
| 2000 objects, depth 3, `dump(4)` | 1.20 ms | 1.50 ms (**25% slower**) | 1.51 ms | 2.08 ms (**38% slower**) |
| isolated whitespace-skip cost (1M spaces) | 0.65 ns/char | 2.13 ns/char (**3.3x slower**) | 2.08 ns/char | 3.67 ns/char (**1.8x slower**) |

Compact JSON (no meaningful whitespace runs) was unaffected either way (~3.7-4.0 ms, noise-level differences), matching what was already reported.

**Root cause:** it wasn't the `get()`/`get_ignoring_pending_unget()` split itself — that still inlines completely into `scan()` on both compilers. The problem was the *shape* of the loop that resulted: rewriting `do { get(); } while (cond);` into `get(); while (cond) { get_ignoring_pending_unget(); }` defeated both Clang's and GCC's ability to keep the input adapter's read/end pointers in registers across loop iterations. Instead, both compilers regenerated code that reloaded those pointers from memory on every single character, turning a register-only loop into a memory-bound one — worst exactly on long whitespace runs (deep indentation), which matches what was observed.

**Fix (pushed as `c2ce707a8`):** keep the same two-function structure, but restore a do-while shape (guarded by an `if` for the "first character isn't whitespace" case):

```cpp
get();

if (current == ' ' || current == '\t' || current == '\n' || current == '\r')
{
    do
    {
        get_ignoring_pending_unget();
    }
    while (current == ' ' || current == '\t' || current == '\n' || current == '\r');
}
```

This is a pure control-flow reshaping with no behavior change. Verified after the fix (median of repeated runs, `-O2`/`-DNDEBUG`):

| case | Clang fixed | GCC fixed |
|---|---|---|
| 300-level deep nesting | 0.24 ms (matches pre-#5490 baseline) | 0.28 ms (**2.7x faster** than pre-#5490 baseline) |
| 2000 objects, depth 3 | 1.20-1.23 ms (matches baseline) | 0.99 ms (**1.5x faster** than baseline) |
| isolated whitespace-skip cost | 0.66 ns/char (matches baseline) | 0.78 ns/char (**2.7x faster** than baseline) |

So with the fix applied, this PR is at worst neutral (Clang) and meaningfully faster (GCC) on whitespace-heavy input, with no regression anywhere I tested, and the required test suites (`unit-class_lexer`, `unit-class_parser`, `unit-deserialization`, `unit-regression1`, `unit-regression2`, `unit-class_parser_diagnostic_positions`) all pass unchanged.

Recommend keeping this PR open with the fix applied rather than closing it.

include/nlohmann/detail/input/lexer.hpp
nlohmann: Good idea, done in 30922eac3 — factored the two increments into a small `advance_position()` helper shared by `get()` and `get_ignoring_pending_unget()`, right alongside the existing shared `track_after_read()` tail.

Given the loop-shape regression we just chased down in this same function, I re-ran the same before/after benchmark (deep-nesting and wide pretty-printed cases, both Clang and a real GCC build) after this change before pushing — no measurable difference either way on either compiler, and the full lexer/parser/diagnostic-positions test suite still passes. So this is a pure readability win with no perf cost.

include/nlohmann/detail/input/lexer.hpp
```diff
@@ -1647,11 +1686,27 @@ class lexer : public lexer_base<BasicJsonType>
...
+        // indentation of pretty-printed JSON) from a register-only loop
+        // into one that reloads the pointer from memory every character
+        if (current == ' ' || current == '\t' || current == '\n' || current == '\r')
         {
-            get();
+            do
+            {
+                get_ignoring_pending_unget();
+            }
+            while (current == ' ' || current == '\t' || current == '\n' || current == '\r');
```
gregmarr: What about using a `curremt_is_whitespace()` function to reduce more duplication?

```
get()

if (!current_is_whitespace()) return;

do { get_ignoring_pending_unget() } while (current_is_whitespace());
```

nlohmann: Done in 86dd10d5b — extracted `current_is_whitespace()` and used it in both the guard and the do-while condition, exactly as suggested.

Re-benchmarked before/after on both Clang and a real GCC build again (same deep-nesting/pretty-print cases as before) — no measurable difference, and the full lexer/parser/diagnostic-positions suite still passes.

gregmarr (approved):

Issue #5396, opened 2026-08-25
JSON Patch "remove" silently succeeds (no-op) when the target's parent is a primitive or null (RFC 6902 §4.2 violation)


nlohmann:
### Description

A JSON Patch `remove` operation whose target path points *through* a primitive (or `null`) parent succeeds without doing anything and without throwing, instead of reporting that the target location does not exist.

[RFC 6902 §4.2](https://datatracker.ietf.org/doc/html/rfc6902#section-4.2): *"The target location MUST exist for the operation to be successful."* The library's own [`patch.md`](https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/api/basic_json/patch.md) documents `out_of_range.403` when a JSON pointer inside the patch cannot be resolved, and does not document any deviation for `remove`.

Root cause: `operation_remove` inside `patch_inplace` ([`include/nlohmann/json.hpp:4913`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json.hpp)) handles `parent.is_object()` and `else if (parent.is_array())` but has **no final `else`**. When `result.at(parent_ptr)` resolves to a primitive or `null`, neither branch runs and the operation returns success.

Note that `replace` of the very same path correctly throws — only `remove` has the hole. Because a patch is documented to be atomic ("either all operations succeed … or an exception is thrown"), this also lets an invalid operation slip past in the middle of a multi-op patch.

### Reproduction steps

Apply a patch that removes a path whose parent is a number (`/a/b` where `/a` is `1`), or removes a member of a top-level `null`.

### Expected vs. actual results

- **Expected:** an exception (`out_of_range.403`, target location does not exist).
- **Actual:** `patch()` returns successfully with the document unchanged.

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <cstdio>
using json = nlohmann::json;

int main()
{
    json doc = {{"a", 1}};
    json r = doc.patch(json::parse(R"([{"op":"remove","path":"/a/b"}])"));
    std::printf("%s\n", r.dump().c_str());   // prints {"a":1} — no exception

    json r2 = json(nullptr).patch(json::parse(R"([{"op":"remove","path":"/a"}])"));
    std::printf("%s\n", r2.dump().c_str());  // prints null — no exception
}
```

### Error messages

```
{"a":1}
null
```

(No exception is thrown; both invalid removes are accepted as no-ops.)

### Suggested fix

[...]

PR #5497, opened 2026-09-05
Reject JSON Patch move when from is a proper prefix of path

### LINKED ISSUE #5397: JSON Patch "move" where "from" is a proper prefix of "path" succeeds with a wrong result (RFC 6902 §4.4 violation)
### Description

A JSON Patch `move` operation is allowed to move a location into one of its own children, which [RFC 6902 §4.4](https://datatracker.ietf.org/doc/html/rfc6902#section-4.4) explicitly forbids:

> The "from" location MUST NOT be a proper prefix of the "path" location; i.e., a location cannot be moved into one of its children.

`move` is implemented as remove-then-add in `patch_inplace` ([`include/nlohmann/json.hpp:5011`](https://github.com/nlohmann/json/blob/develop/include/nlohmann/json.hpp)) with no check that `from` is not a proper prefix of `path`. For object targets the subsequent add happens to throw `out_of_range.403` as a side effect, but for **array** targets the removal shifts indices so the path re-resolves to a different element and the operation "succeeds" with a nonsensical document.

This is a MUST violation that silently produces a wrong result rather than the required error. (The nearest prior report, #894, was a different — and fixed — bug about `copy`.)

### Reproduction steps

Apply a `move` whose `from` is a proper prefix of `path` on a nested array.

- **Expected:** an exception; the operation is invalid per RFC 6902 §4.4.
- **Actual:** `patch()` returns `[[[1,2],3]]` with no error.

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <cstdio>
using json = nlohmann::json;

int main()
{
    json doc = json::parse(R"([[1,2],[3]])");
    // from "/0" is a proper prefix of path "/0/0" -> must be rejected
    json r = doc.patch(json::parse(R"([{"op":"move","from":"/0","path":"/0/0"}])"));
    std::printf("%s\n", r.dump().c_str());   // prints [[[1,2],3]]
}
```

### Error messages

```
[[[1,2],3]]
```

(No exception; an invalid move is silently applied.)

### Suggested fix

Before performing the remove/add for `move`, reject the operation (e.g. `out_of_range.403` or a dedicated error) when `from.to_string()` is a proper prefix of `path.to_string()` at the reference-token level.

### Library version

### Validation

nlohmann:
Fixes #5397.

This is stacked on top of #5496 (fix for #5396), since both touch the same `patch_inplace` move/remove handling in `include/nlohmann/json.hpp`. It's based on the `issue-5396-patch-remove-primitive-parent` branch, so the diff shown here is only the second commit; the first commit is #5496.

## Problem

RFC 6902 (section 4.4) states: *"The 'from' location MUST NOT be a proper prefix of the 'path' location; i.e., a location cannot be moved into one of its children."* `move` is implemented as remove-then-add in `patch_inplace`, with no check for this. For object targets, the subsequent "add" happens to throw anyway, as a side effect of trying to resolve through the now-removed parent — but for **array** targets, removing the "from" element shifts subsequent indices, so "path" silently re-resolves to a different element and the operation "succeeds" with a silently corrupted document:

```cpp
json doc = json::parse(R"([[1,2],[3]])");
// from "/0" is a proper prefix of path "/0/0" -> must be rejected per RFC 6902 §4.4
doc.patch(json::parse(R"([{"op":"move","from":"/0","path":"/0/0"}])"));
// before this fix: returns [[[1,2],3]] with no error
```

## Fix

Before performing the remove/add for `move`, check whether `from` is a proper prefix of `path` **at the reference-token level**, not as raw pointer strings. `json_pointer` already splits a pointer into unescaped reference tokens (`~0` → `~`, `~1` → `/`) when parsing, and `basic_json` is declared a friend of `json_pointer` specifically to access this internal state, so the check compares `from_ptr.reference_tokens` against `ptr.reference_tokens` (a prefix-of-vector comparison), rather than doing a naive `std::string::rfind`/`substr` comparison on the undecoded pointer text — which a token containing an escaped `/` or `~` could defeat. When `from` is a proper prefix of `path`, this throws `out_of_range.414`.

## Tests

Added `tests/src/unit-json_patch.cpp` test case `"JSON patch - move where 'from' is a proper prefix of 'path' (regression #5397)"` covering:
- the array case from the issue
- an equivalent object case
- `from == path` (not a *proper* prefix) — confirmed **not** rejected by this check
- the key case proving token-level (not raw-string) comparison: `from="/ab"`, `path="/abc/x"` — `"ab"` is a raw string-prefix of `"abc"`, but `/ab` is **not** a pointer-token prefix of `/abc/x`, so this move is allowed
- `~0`/`~1` escaped tokens, confirming the comparison operates on unescaped tokens
- ordinary valid moves (unrelated paths, sibling paths, and the reverse relationship where "path" is a prefix of "from", which RFC 6902 does not forbid) still work unchanged

Ran the full pre-existing `unit-json_patch.cpp` suite (compiled offline against `include/`, with and without `JSON_DIAGNOSTICS`) on top of both this fix and the #5396 fix — all pre-existing assertions still pass; only the two test cases that require the downloaded `json-patch-tests` data fail in this offline setup (as expected, unrelated to this change).

## Documentation

Documented the new `out_of_range.414` exception in `docs/mkdocs/docs/home/exceptions.md`, and referenced it from `docs/mkdocs/docs/api/basic_json/patch.md` and `patch_inplace.md`.

## Breaking change?

No breaking change to the public API. This is a bug fix that makes the library correctly reject a JSON Patch document that itself violates RFC 6902 section 4.4: a `"move"` operation whose `"from"` is a proper prefix of its `"path"`. Previously this either happened to throw already (object targets) or silently produced a corrupted result (array targets, per the reproduction above) — never a well-defined success. It now consistently throws `out_of_range.414`. Any caller relying on the previous silently-corrupted array behavior for such an already-invalid patch would need to adjust, but no valid/spec-conformant JSON Patch document is affected.

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -4947,6 +4947,30 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
             }
         };
 
+        // RFC 6902 (section 4.4) forbids "from" from being a proper prefix
+        // of "path" for a "move" operation: a location cannot be moved into
+        // one of its own children. Compares reference tokens (already
+        // unescaped by json_pointer's parser) rather than the raw pointer
+        // strings, since a token may itself contain an escaped '/' or '~'
+        // that would defeat a naive string-prefix comparison. "from" equal
+        // to "path" is *not* a proper prefix and must return false.
+        const auto is_proper_prefix = [](const json_pointer & from, const json_pointer & to)
+        {
+            const auto from_size = from.reference_tokens.size();
+            if (from_size >= to.reference_tokens.size())
+            {
+                return false;
+            }
+            for (std::size_t i = 0; i < from_size; ++i)
+            {
+                if (!(from.reference_tokens[i] == to.reference_tokens[i]))
+                {
+                    return false;
+                }
+            }
+            return true;
+        };
+
         // type check: top level value must be an array
         if (JSON_HEDLEY_UNLIKELY(!json_patch.is_array()))
         {
@@ -5022,27 +5046,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
                     const auto from_path = get_value("move", "from", true).template get<string_t>();
                     json_pointer from_ptr(from_path);
 
-                    // RFC 6902 (section 4.4) forbids "from" from being a
-                    // proper prefix of "path": a location cannot be moved
-                    // into one of its own children. Compare the pointers'
-                    // reference tokens (already unescaped by json_pointer's
-                    // parser) rather than the raw pointer strings, since a
-                    // token may itself contain an escaped '/' or '~' that
-                    // would defeat a naive string-prefix comparison. This is
-                    // written as an explicit, manually-bounded loop (rather
-                    // than std::equal(first1, last1, first2), whose second
-                    // range has no explicit end iterator) so every access to
-                    // ptr.reference_tokens is visibly guarded by the same
-                    // index the loop condition already bounds against
-                    // from_size -- from_size < ptr.reference_tokens.size()
-                    // is checked once, up front, before the loop runs at all.
-                    const auto from_size = from_ptr.reference_tokens.size();
-                    bool from_is_proper_prefix_of_path = from_size < ptr.reference_tokens.size();
-                    for (std::size_t i = 0; from_is_proper_prefix_of_path && i < from_size; ++i)
-                    {
-                        from_is_proper_prefix_of_path = from_ptr.reference_tokens[i] == ptr.reference_tokens[i];
-                    }
-                    if (JSON_HEDLEY_UNLIKELY(from_is_proper_prefix_of_path))
+                    if (JSON_HEDLEY_UNLIKELY(is_proper_prefix(from_ptr, ptr)))
                     {
                         JSON_THROW(out_of_range::create(414, detail::concat("cannot move value: 'from' path '", from_path, "' is a proper prefix of 'path' '", path, "'"), &result));
                     }
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -5022,6 +5022,19 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
                     const auto from_path = get_value("move", "from", true).template get<string_t>();
                     json_pointer from_ptr(from_path);
 
+                    // RFC 6902 (section 4.4) forbids "from" from being a
+                    // proper prefix of "path": a location cannot be moved
+                    // into one of its own children. Compare the pointers'
+                    // reference tokens (already unescaped by json_pointer's
+                    // parser) rather than the raw pointer strings, since a
+                    // token may itself contain an escaped '/' or '~' that
+                    // would defeat a naive string-prefix comparison.
+                    if (JSON_HEDLEY_UNLIKELY(from_ptr.reference_tokens.size() < ptr.reference_tokens.size()
+                                             && std::equal(from_ptr.reference_tokens.begin(), from_ptr.reference_tokens.end(), ptr.reference_tokens.begin())))
+                    {
+                        JSON_THROW(out_of_range::create(414, detail::concat("cannot move value: 'from' path '", from_path, "' is a proper prefix of 'path' '", path, "'"), &result));
+                    }
+
                     // the "from" location must exist - use at()
                     basic_json const v = result.at(from_ptr);
 
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -5028,9 +5028,21 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
                     // reference tokens (already unescaped by json_pointer's
                     // parser) rather than the raw pointer strings, since a
                     // token may itself contain an escaped '/' or '~' that
-                    // would defeat a naive string-prefix comparison.
-                    if (JSON_HEDLEY_UNLIKELY(from_ptr.reference_tokens.size() < ptr.reference_tokens.size()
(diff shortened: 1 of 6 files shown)


include/nlohmann/json.hpp
```diff
@@ -5022,6 +5022,19 @@
...
 
+                    // RFC 6902 (section 4.4) forbids "from" from being a
+                    // proper prefix of "path": a location cannot be moved
+                    // into one of its own children. Compare the pointers'
+                    // reference tokens (already unescaped by json_pointer's
+                    // parser) rather than the raw pointer strings, since a
+                    // token may itself contain an escaped '/' or '~' that
+                    // would defeat a naive string-prefix comparison.
+                    if (JSON_HEDLEY_UNLIKELY(from_ptr.reference_tokens.size() < ptr.reference_tokens.size()
+                                             && std::equal(from_ptr.reference_tokens.begin(), from_ptr.reference_tokens.end(), ptr.reference_tokens.begin())))
```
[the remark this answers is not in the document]
gregmarr: That is absolutely false.  The first half of the if checks that `from_ptr.reference_tokens` is smaller than `ptr.reference_tokens`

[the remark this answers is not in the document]
nlohmann: You're right — the size check does guarantee `ptr.reference_tokens` has at least `from_ptr.reference_tokens.size()` elements before `std::equal` runs, so this was safe as written. Flawfinder can't see that invariant from the three-iterator `std::equal(first1, last1, first2)` call, since it has no visible end bound for the second range.

Rather than leave that as a judgment call for the next reader (human or tool), I replaced it in 95c1df59e with an explicit, manually-bounded loop — every access to `ptr.reference_tokens` is now guarded by the same index the loop condition bounds against `from_size`, so the safety is visible in the code itself rather than requiring the reader to trace back to the earlier size comparison. (I first tried the C++14 four-iterator `std::equal(first1, last1, first2, last2)` form, which would have been a more minimal change, but this codebase targets C++11 and that overload turned out not to be safely usable under `-std=c++11` with the standard library here — so the explicit loop it is.)

Thanks for flagging this.

include/nlohmann/json.hpp
```diff
@@ -5022,6 +5022,31 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+                    // range has no explicit end iterator) so every access to
+                    // ptr.reference_tokens is visibly guarded by the same
+                    // index the loop condition already bounds against
+                    // from_size -- from_size < ptr.reference_tokens.size()
+                    // is checked once, up front, before the loop runs at all.
+                    const auto from_size = from_ptr.reference_tokens.size();
+                    bool from_is_proper_prefix_of_path = from_size < ptr.reference_tokens.size();
+                    for (std::size_t i = 0; from_is_proper_prefix_of_path && i < from_size; ++i)
+                    {
+                        from_is_proper_prefix_of_path = from_ptr.reference_tokens[i] == ptr.reference_tokens[i];
```
gregmarr: `std::equal` will bail as soon as it finds a non-equal.  For long paths that differ at the very beginning, this can be significant, and you've been doing a lot of performance work lately.   I think making this a helper function would make it easier to bail early with false.

```
bool from_is_a_prefix_of_base(json_pointer &from, json_pointer &base)
{
  const auto from_size = from_ptr.reference_tokens.size();
  const auto base_size = ptr.reference_tokens.size();

  if (from_size > base_size) return false;

  std::size_t i = 0;
  do
  {
    if (from.reference_tokens[i] != base.reference_tokens[i]) return false;
  }
  while(i < from_size);
}
```

I suggest do/while simply based on the performance you saw in another PR.  Maybe that doesn't apply here because it's indexed and not pointers.

nlohmann: Good suggestion on extracting a helper — done in 47bd3e67c, following this function's existing convention of local lambdas (`operation_add`/`operation_remove` are defined the same way just above it) rather than a free function, so it has natural access to `json_pointer::reference_tokens`.

Two things I changed from your sketch, worth flagging:
- `if (from_size &gt; base_size) return false;` would let `from_size == base_size` fall through to the comparison loop and return true when all tokens match — i.e. it would treat `from == path` as a *proper* prefix. That's the one case this whole check must NOT reject (a no-op move, allowed by RFC 6902 and pinned by an existing test in this PR), so the guard needs to be `from_size &gt;= base_size`, not `&gt;`.
- Kept the early-exit `for` loop with `return false` inside it (rather than the do/while) — it already bails on the first mismatch, same as `std::equal` did, so there's no early-exit behavior to gain from a different loop shape here. Whether do/while-vs-for measurably matters at all is also a different question at this call site than in the lexer work: `reference_tokens` is index-accessed through a `std::vector`, not scanned via a raw pointer/adapter, and a JSON Pointer's token count is realistically single digits to maybe low hundreds in pathological cases — nowhere near the millions-of-characters-per-call scale where that loop-shape effect actually showed up. I don't think there's a measurable difference to chase here, so I kept the more conventional `for`.

Thanks for pushing on both the correctness and the readability.

gregmarr: I missed the testing of the comparison result in the for loop condition, it was indeed exiting early.

include/nlohmann/json.hpp
```diff
@@ -4947,6 +4947,30 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
...
+        const auto is_proper_prefix = [](const json_pointer & from, const json_pointer & to)
+        {
+            const auto from_size = from.reference_tokens.size();
+            if (from_size >= to.reference_tokens.size())
+            {
+                return false;
+            }
+            for (std::size_t i = 0; i < from_size; ++i)
+            {
+                if (!(from.reference_tokens[i] == to.reference_tokens[i]))
```
gregmarr, line 4966: Why `if (!(a == b))` rather than `if (a != b)`?

PR #5501, opened 2026-09-06
Return the parsed value by move from from_cbor() and friends

nlohmann:
Part of the series that makes the binary readers non-recursive (#5104); this one stands on its own.

## What

The binary entry points end with

```cpp
```

The conditional operator's second operand is an lvalue, so this is not a case where the return value can be elided or implicitly moved from. Every successful `from_cbor`, `from_msgpack`, `from_ubjson`, `from_bjdata` and `from_bson` call **deep-copies the value it just parsed**, and then destroys the original.

## Why it matters

`basic_json`'s copy constructor walks the whole value, so the cost is proportional to the document. Parsing a 2 MB CBOR document with 60,000 objects, median of 25 runs, clang 17 `-O3`:

| | before | after |
|---|---|---|
| `from_cbor` | 26.99 ms | **14.65 ms** |
| `from_msgpack` | 26.82 ms | **14.82 ms** |

There is a second reason. The copy constructor recurses once per nesting level, so it is also a stack-overflow path on the return side, on a value the reader has already accepted. That is currently masked because the readers themselves recurse and overflow first, but it has to be fixed for making them iterative to have any effect — see #5104 and the rest of this stack.

## API impact

**No breaking changes.** No signature, no return type and no observable behaviour changes; the parsed value is simply not copied on its way out. It is not used again after the return expression is evaluated.

## Checklist

### DIFF
--- a/include/nlohmann/json.hpp
+++ b/include/nlohmann/json.hpp
@@ -4501,8 +4501,11 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         basic_json result;
         auto ia = detail::input_adapter(std::forward<InputType>(i));
         detail::json_sax_dom_parser<basic_json, decltype(ia)> sdp(result, allow_exceptions);
-        const bool res = binary_reader<decltype(ia)>(std::move(ia), input_format_t::cbor).sax_parse(input_format_t::cbor, &sdp, strict, tag_handler); // cppcheck-suppress[accessMoved]
-        return res ? result : basic_json(value_t::discarded);
+        if (!binary_reader<decltype(ia)>(std::move(ia), input_format_t::cbor).sax_parse(input_format_t::cbor, &sdp, strict, tag_handler)) // cppcheck-suppress[accessMoved]
+        {
+            result = value_t::discarded;
+        }
+        return result;
     }
 
     /// @brief create a JSON value from an input in CBOR format (iterator pair, or iterator+sentinel pair for C++20 ranges support)
@@ -4518,8 +4521,11 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         basic_json result;
         auto ia = detail::input_adapter(std::move(first), std::move(last));
         detail::json_sax_dom_parser<basic_json, decltype(ia)> sdp(result, allow_exceptions);
-        const bool res = binary_reader<decltype(ia)>(std::move(ia), input_format_t::cbor).sax_parse(input_format_t::cbor, &sdp, strict, tag_handler); // cppcheck-suppress[accessMoved]
-        return res ? result : basic_json(value_t::discarded);
+        if (!binary_reader<decltype(ia)>(std::move(ia), input_format_t::cbor).sax_parse(input_format_t::cbor, &sdp, strict, tag_handler)) // cppcheck-suppress[accessMoved]
+        {
+            result = value_t::discarded;
+        }
+        return result;
     }
 
     template<typename T>
@@ -4544,8 +4550,11 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         auto ia = i.get();
         detail::json_sax_dom_parser<basic_json, decltype(ia)> sdp(result, allow_exceptions);
         // NOLINTNEXTLINE(hicpp-move-const-arg,performance-move-const-arg)
-        const bool res = binary_reader<decltype(ia)>(std::move(ia), input_format_t::cbor).sax_parse(input_format_t::cbor, &sdp, strict, tag_handler); // cppcheck-suppress[accessMoved]
-        return res ? result : basic_json(value_t::discarded);
+        if (!binary_reader<decltype(ia)>(std::move(ia), input_format_t::cbor).sax_parse(input_format_t::cbor, &sdp, strict, tag_handler)) // cppcheck-suppress[accessMoved]
+        {
+            result = value_t::discarded;
+        }
+        return result;
     }
 
     /// @brief create a JSON value from an input in MessagePack format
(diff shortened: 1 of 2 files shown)


include/nlohmann/json.hpp
```diff
@@ -4474,7 +4474,7 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         auto ia = detail::input_adapter(std::forward<InputType>(i));
         detail::json_sax_dom_parser<basic_json, decltype(ia)> sdp(result, allow_exceptions);
         const bool res = binary_reader<decltype(ia)>(std::move(ia), input_format_t::cbor).sax_parse(input_format_t::cbor, &sdp, strict, tag_handler); // cppcheck-suppress[accessMoved]
-        return res ? result : basic_json(value_t::discarded);
+        return res ? std::move(result) : basic_json(value_t::discarded);
```
gregmarr: Another way to do this would be to reset `result` to `discarded` if `sax_parse` returns false:
```
if (!binary_reader...sax_parse(...)) {
  result = basic_json::value_t::discarded;
}

return result;
```

nlohmann: Thanks, done!

include/nlohmann/json.hpp
```diff
@@ -4473,8 +4473,11 @@ class basic_json // NOLINT(cppcoreguidelines-special-member-functions,hicpp-spec
         basic_json result;
         auto ia = detail::input_adapter(std::forward<InputType>(i));
         detail::json_sax_dom_parser<basic_json, decltype(ia)> sdp(result, allow_exceptions);
-        const bool res = binary_reader<decltype(ia)>(std::move(ia), input_format_t::cbor).sax_parse(input_format_t::cbor, &sdp, strict, tag_handler); // cppcheck-suppress[accessMoved]
-        return res ? result : basic_json(value_t::discarded);
+        if (!binary_reader<decltype(ia)>(std::move(ia), input_format_t::cbor).sax_parse(input_format_t::cbor, &sdp, strict, tag_handler)) // cppcheck-suppress[accessMoved]
```
gregmarr, line 4504: Why does cppcheck complain about accessing a moved object here?  Is this just a left over comment that some previous change resolved?

nlohmann: Not leftover — I measured it both ways. Running the `ci_cppcheck` invocation over four variants of `json.hpp`:

| variant | `accessMoved` | `accessForwarded` |
| --- | --- | --- |
| develop, comments in place | 0 | 0 |
| develop, comments **stripped** | **19** | **1** |
| this branch, comments in place | 0 | 0 |
| this branch, comments **stripped** | **19** | **1** |

Identical before and after: 14 × `ia` at the `from_*` sites, plus `cb`/`first`/`last`/`i` at the four `parse` sites. Since `ci_cppcheck` runs with `--error-exitcode=1`, deleting the comments turns it red. Two extra runs with `--enable=information` reported no `unmatchedSuppression` on either variant, so none of the 17 is dead.

They came in with #4553, when `--inconclusive --check-level=exhaustive` were switched on, and have survived the cppcheck 1.4.6 → 1.5.1 bump since. This PR only reflows the statement and carries the comment along unchanged.

It *is* a false positive, though. `binary_reader` takes `InputAdapterType&&` and move-constructs its own `ia` member, and the caller's local is never read again — the only later mention is `decltype(ia)`, which is unevaluated. cppcheck points at column 64, which is the `ia` inside `std::move(ia)` itself. I couldn't reduce it to a minimal repro: neither `decltype` as a template argument nor a member sharing the local's name reproduces it under the same flags, so it seems specific to cppcheck's `--inconclusive` value-flow through this instantiation. That argues for leaving the comment until someone can characterise it.

One thing your question does expose: `ci_cppcheck` runs without `--enable=information`, so a suppression that *did* go stale would never be reported — nothing would catch it. Adding that (plus `--suppress=checkersReport`) would. I'll put it up as a separate PR.

<sub>Written by Claude Code.</sub>

gregmarr: Sounds good.

nlohmann: Following up on my own suggestion above — I tried the `--enable=information` change, and it does not hold up. Recording why, since the reasons are more interesting than the idea was.

**It reports a suppression that is not stale.** With the target's flags, the only thing it flags on unmodified `develop` is `json.hpp:726`:

```cpp
// cppcheck-suppress assertWithSideEffect
JSON_ASSERT(!check_parents || !is_structured() || std::all_of(begin(), end(), ...));
```

That comment is correct. It is unmatched only because `ci_cppcheck` passes `-UJSON_ASSERT`, so `JSON_ASSERT` expands to nothing and `assertWithSideEffect` can never fire in the configuration CI analyses. Enabling the severity would turn the target red for a suppression that is doing its job.

**"Unmatched" means "not reached", not "dead".** Dropping the `-U`/`-D` set takes the count from 2 to 9 — four `accessForwarded` suppressions and two `missingReturn` become "unmatched" purely because fewer configurations were analysed. So the signal tracks analysis coverage, not staleness, which is the opposite of what you want in a gate.

**And it silently does nothing under our current flags anyway.** `--relative-paths` suppresses the entire unmatched-suppression report. With everything else identical, planting a deliberately dead `// cppcheck-suppress nullPointer` in `json.hpp` gives:

| invocation | unmatched reported |
| --- | --- |
| target's flags, minus `--relative-paths` | 2 (the planted one, plus `:726`) |
| target's flags, as-is | 0 |

So even setting the false positives aside, the change would have been a no-op until `--relative-paths` was dropped, and dropping that changes every reported path in the output.

Conclusion: not worth gating CI on, so I am not opening that PR. Checking whether a suppression still matters stays a manual exercise — delete it, run the target, see what comes back. Which is what I did for the `accessMoved` comments here, and they are all still live.

<sub>Written by Claude Code.</sub>

gregmarr (approved):

Issue #537, opened 2017-03-28
recursive from_msgpack implementation will stack overflow


xgzeng:
`from_msgpack` function(etc) parses data recursively, which will cause stack overflow when process deep nested structure. 
It will be DOS vulnerabilities, if we use library to process data from outside.

Following is example code to generate malicious data. `msgpack_pack_array` is not shown here.
```cpp
for (int i = 0; i < 10000; ++i) {
    msgpack_pack_array(buffer, 1);
}
msgpack_pack_array(buffer, 0);
```

nlohmann:
The same issue was detected for CBOR in #517.

xgzeng:
Should we just add a defaut parameter to restrict nesting level?
I think it suits most use case.

nlohmann:
That would be an effective, yet ugly solution. I would rather have a non-recursive version that uses a stack internally so we can enforce a limit there.

nlohmann:
Closed in favor of #517.

Issue #517, opened 2017-03-17
Stack-overflow (OSS-Fuzz 869)


nlohmann:
```
Detailed report: https://oss-fuzz.com/testcase?key=5116967767769088

Project: json
Fuzzer: libFuzzer_json_parse_cbor_fuzzer
Job Type: libfuzzer_asan_json
Platform Id: linux

Crash Type: Stack-overflow
Crash Address: 0x7ffc0dea4b58
Crash State:
  nlohmann::basic_json<std::__1::map, std::__1::vector, std::__1::basic_string<cha
  
Sanitizer: address (ASAN)

Regressed: https://oss-fuzz.com/revisions?job=libfuzzer_asan_json&range=201701031958:201701032147

Reproducer Testcase: https://oss-fuzz.com/download/AMIfv95GDH2Mys_drMGzg9fUR7UPBBqUF_LhjJm-rgKw6x-oTwvsux765EavyU1j3OwyrVWiw_gW8F_Ia6rk-diY-dnPJ2rD592wN1mKQoik0vlg1QtZCaI2yCI5uBCQuaf1PBLkLbIsMQtTywZ5nmE_AZlrvhk2Mye_GJ4oM8wNGSM96vFsgM_IpJ8RXU_sE1htQ639Zyj55qzACG-LDlRAXFeA-Pxd9X2Os5u259dzhhjmH6eqLlXItedTywfMtygaPkYlc4R05g78-SwjpAyOv0BNR4dx2jkcD9xHXNjjAY-232t-qf9oWf2T66QqT_FoMbm9qtBf1FAI8H6Af4e9PUMVB2xjPOIe8uCQh50tqv58mnKY41ZAwOcstJChSYmBAAFsUSnV?testcase_id=5116967767769088


Issue filed automatically.

See https://github.com/google/oss-fuzz/blob/master/docs/reproducing.md for more information.

This bug is subject to a 90 day disclosure deadline. If 90 days elapse
without an upstream patch, then the bug report will automatically
become visible to the public.
```

[clusterfuzz-testcase-5116967767769088.zip](https://github.com/nlohmann/json/files/851775/clusterfuzz-testcase-5116967767769088.zip)

nlohmann:
This seems to be a classical stack overflow due to excessive nesting of arrays.

nlohmann:
ClusterFuzz has detected this issue as fixed in range 201705051619:201705061619.

PR #5505, opened 2026-09-06
Read MessagePack containers without recursing per nesting level

nlohmann:
Closes the MessagePack vector of #5104, and adds the machinery the rest of the stack reuses.

## What

`get_msgpack_array()` and `get_msgpack_object()` read their elements by calling back into `parse_msgpack_internal()`, which calls them again for a nested container. The native call stack grew with the nesting depth of the input, and each level costs **one byte** to encode: `0x91` is a one-element array, so a few hundred thousand of them crash the process before any of the input is rejected.

This was first reported for MessagePack in 2017 (#537), closed as a duplicate of #517, which was closed as "ClusterFuzz detected this issue as fixed" — a false negative, since no architectural fix ever landed.

## How

Keep the open containers on a heap stack, the way `parser::sax_parse_internal()` has always done for JSON text. A frame records how many elements are left and whether to close with `end_object()` or `end_array()`; `parse_msgpack_value()` reads a single value and, for a container, only opens it; and `parse_msgpack_internal()` loops, resuming the innermost container after each element. Whether the value that was begun is complete is answered by the stack being empty, so no separate bookkeeping is needed.

The switch that decodes a value is untouched apart from the six container cases. `enter_container()` is the only place a binary reader emits `start_object()`/`start_array()`, so a check that rejects a container can be added there once and is guaranteed to run before the start event — one of the review points on #5293.

## Verification

- `unit-msgpack` passes unchanged (1.2 M assertions), plus 12 new ones.
- Compared against the parent commit over empty, nested, counted (array 16/32, map 16/32) and truncated inputs: identical values, error codes, messages and byte offsets.
- 300,000 levels now report `parse_error.110` instead of crashing, and a well-formed 300,000-level value is read to completion through the SAX interface, where `develop` crashes.
- Timing is unchanged: 60,000 small objects and one array of a million integers are within run-to-run noise of `develop` either way.

## Note on the stack

Reading such a value into a `basic_json` needs #5501 as well, without which the recursive copy constructor overflows on the way out. That is the parent PR, and the test for the value path covers the two together.

## API impact

## Checklist

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -130,6 +130,7 @@ class binary_reader
                    const cbor_tag_handler_t tag_handler = cbor_tag_handler_t::error)
     {
         sax = sax_;
+        container_stack.clear();
         bool result = false;
 
         switch (format)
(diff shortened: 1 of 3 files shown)


include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -179,6 +180,57 @@ class binary_reader
...
+
+    The members are ordered widest first: frames are stored in a vector, and
+    declaring the `bool` first would pad the struct out for no reason.
+    */
+    struct container_frame
+    {
+        /// number of elements that have not been read yet
+        std::size_t remaining = 0;
+        /// whether to close this container with end_object() or end_array()
+        bool is_object = false;
```
gregmarr: ```    declaring the `bool` first would pad the struct out for no reason.```

The struct is the same size either way as struct alignment on x64 is 8.  You would need to do a packed struct to make it only 8, but then you wouldn't have the full `size_t`.  Putting the bool after the size_t is better for when there are other members.  In general, put the highest alignments first and work down.

The other option would be to have two arrays, one for remaining and one for is_object.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -179,6 +180,57 @@ class binary_reader
...
+    bool enter_container(const bool is_object, const std::size_t len)
+    {
+        if (JSON_HEDLEY_UNLIKELY(is_object ? !sax->start_object(len) : !sax->start_array(len)))
+        {
+            return false;
+        }
+
+        container_frame frame;
+        frame.remaining = len;
+        frame.is_object = is_object;
```
gregmarr:
```
        container_stack.emplace_back(len, is_object);
```

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -1578,7 +1640,7 @@ class binary_reader
             case 0x8D:
             case 0x8E:
             case 0x8F:
-                return get_msgpack_object(conditional_static_cast<std::size_t>(static_cast<unsigned int>(current) & 0x0Fu));
+                return enter_container(/*is_object*/true, conditional_static_cast<std::size_t>(static_cast<unsigned int>(current) & 0x0Fu));
```
gregmarr: Use `enter_object` and `enter_array` wrappers that handle the first parameter?

include/nlohmann/detail/input/binary_reader.hpp
gregmarr: This all makes sense in the later PRs.

include/nlohmann/detail/input/binary_reader.hpp

include/nlohmann/detail/input/binary_reader.hpp

gregmarr (approved):

PR #5506, opened 2026-09-06
Read CBOR containers and tags without recursing per nesting level

nlohmann:
Closes the CBOR container and tag vectors of #5104. CBOR has the most open OSS-Fuzz stack-overflow reports of any format here (issues 487903138 and 371248057).

## What

`get_cbor_array()` and `get_cbor_object()` read their elements by calling back into the value reader, which called them again for a nested container, and a tag was handled by reading the tagged value the same way. All three cost native stack, and all three cost a **single byte** to encode:

| bytes | meaning |
|---|---|
| `0x9F` | opens an indefinite-length array |
| `0x81` | a one-element array |
| `0xC2` | a tag |

Half a million of any of them crashes the process before the input runs out.

## How

The shape the MessagePack reader already uses (#5505): the open containers live on the heap stack, `parse_cbor_value()` reads a single value and only opens a container, and `parse_cbor_internal()` loops.

Two things are specific to CBOR:

- An **indefinite-length container** ends at a break marker rather than a count, and testing for that marker consumes a byte which is the first byte of the next element when it is not one. The frame's count is `npos` for those, and the driver tracks whether the next value starts at a fresh byte.
- A **tag is not a value of its own**. Instead of reading the tagged value by recursing, the value reader reports that a tag was read and the driver reads on, so a chain of tags costs no stack at all.

The switch that decodes a value is unchanged apart from the twelve container cases and the two tag sites.

## Verification

- `unit-cbor` passes unchanged (1.64 M assertions), plus 20 new ones.
- Compared against the parent commit over definite and indefinite arrays and maps, all four counted forms, empty containers, nesting of the forms inside each other, truncated inputs, and all three tag handlers: identical values, error codes, messages and byte offsets.
- 500,000 levels of each of the three vectors now report `parse_error.110` instead of crashing; a well-formed 200,000-level value is read to completion.

## Performance

Reported in full because it is not free. The driver does per element what a counted loop used to do per container, and CBOR pays for it more than MessagePack because the value reader also has to be told whether to fetch a byte. Over five alternating runs, parsing 60,000 small objects and one array of a million integers:

| | recursive | iterative |
|---|---|---|
| containers | 10.86 ms | 11.25 ms |
| scalars | 12.99 ms | 13.58 ms |

about **3–4 % slower**. Against `develop`, however, the same two inputs are about **44 % faster** (19.90 → 11.21 ms), because #5501 removes the copy the entry point used to make.

## API impact

## Checklist

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -1424,9 +1424,10 @@ class binary_reader
         {
             if (!container_stack.empty())
             {
-                // the reference is not held across parse_cbor_value() below,
-                // which can push onto the stack and reallocate it
-                container_frame& top = container_stack.back();
+                // a copy, not a reference: it must stay valid across the
+                // pop_back() below, which destroys the container_stack element
+                // it would otherwise alias
+                container_frame top = container_stack.back();
                 bool at_end = false;
 
                 if (top.remaining != npos)
@@ -1437,7 +1438,7 @@ class binary_reader
                     if (!at_end)
                     {
                         // claim the element about to be read
-                        --top.remaining;
+                        --container_stack.back().remaining;
                         if (top.is_object)
                         {
                             get();
@@ -1456,9 +1457,8 @@ class binary_reader
 
                 if (at_end)
                 {
-                    const bool is_object = top.is_object;
                     container_stack.pop_back();
-                    if (JSON_HEDLEY_UNLIKELY(is_object ? !sax->end_object() : !sax->end_array()))
+                    if (JSON_HEDLEY_UNLIKELY(top.is_object ? !sax->end_object() : !sax->end_array()))
                     {
                         return false;
                     }
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -575,9 +575,12 @@ class binary_reader
 
     @return whether a valid CBOR value was passed to the SAX parser
     */
-    bool parse_cbor_internal(const bool get_char,
-                             const cbor_tag_handler_t tag_handler)
+    bool parse_cbor_value(const bool get_char,
+                          const cbor_tag_handler_t tag_handler,
+                          bool& tag_pending)
     {
+        tag_pending = false;
+
         switch (get_char ? get() : current)
         {
             // EOF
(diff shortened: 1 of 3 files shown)


include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -757,37 +760,37 @@ class binary_reader
             case 0x95:
             case 0x96:
             case 0x97:
-                return get_cbor_array(
-                           conditional_static_cast<std::size_t>(static_cast<unsigned int>(current) & 0x1Fu), tag_handler);
+                return enter_container(/*is_object*/false,
```
gregmarr: As on #5505 have a helper for `enter_object` and `enter_array` that handles this first parameter so you don't need the comment?

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -1387,96 +1396,110 @@ class binary_reader
...
-            string_t key;
-            if (len != detail::unknown_size())
-            {
-                for (std::size_t i = 0; i < len; ++i)
+                if (at_end)
                 {
-                    get();
-                    if (JSON_HEDLEY_UNLIKELY(!get_cbor_string(key) || !sax->key(key)))
+                    const bool is_object = top.is_object;
+                    container_stack.pop_back();
```
gregmarr, line 1460: This invalidates `top`.  There should be a comment here about that, so someone doesn't change something later that allows `top` to be read again.  Alternatively, `top` could be a copy, and then `--top.remaining;` above becomes `--container_stack.back().remaining;`.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -1387,96 +1396,110 @@ class binary_reader
...
+        string_t key;
 
-        if (len != detail::unknown_size())
+        while (true)
         {
-            for (std::size_t i = 0; i < len; ++i)
+            if (!container_stack.empty())
             {
-                if (JSON_HEDLEY_UNLIKELY(!parse_cbor_internal(true, tag_handler)))
+                // the reference is not held across parse_cbor_value() below,
```
gregmarr: However, it is held across `container_stack.pop_back();`, comments below about mitigations or fixes.

nlohmann: Fixed in d05ac29f9: top is now a copy (container_frame top = container_stack.back();), so it stays valid regardless of what happens to container_stack afterward, including the pop_back() right below. The one mutation (--top.remaining) now goes through container_stack.back() directly.

Written by Claude Code.

include/nlohmann/detail/input/binary_reader.hpp
nlohmann: Fixed in d05ac29f9 - went with the copy approach you suggested, and updated the comment above to explain why (it is not about reallocation from a push anymore, it is about surviving the pop_back() here).

Written by Claude Code.

gregmarr (approved):

PR #5507, opened 2026-09-06
Read UBJSON and BJData containers without recursing per nesting level

nlohmann:
Closes the UBJSON and BJData vectors of #5104, including the oldest open OSS-Fuzz report for this project (issue 42541028, filed July 2018, 115 duplicates).

## What

`get_ubjson_array()` and `get_ubjson_object()` read their elements by calling back into the value reader, which called them again for a nested container. `[` alone opens a container, so half a million of them crashes the process before the input runs out.

The optimized forms reach the same path through a size or type annotation. And in **plain UBJSON** `[` and `{` are permitted as the *type* of an optimized container — BJData forbids them — so `[$[#i\x01` repeated nests just as deeply at six bytes a level. That variant was not in #5293's scope.

## How

Both readers now only open their container, and `parse_ubjson_internal()` loops: it closes the containers that have ended, claims the next element of the innermost one, reads its key when it is an object, and works out the marker of the value to read next. That last part is where the formats differ, and the loop follows what the four element loops used to do:

- a sized, typed container gives its elements no marker of their own
- a sized, untyped container reads one for each element
- a container that ends at a marker has the byte already, from the test against `]` or `}`; for an object it is the first byte of the key

The **ND-array wrapper** and the `'B'` binary shortcut stay as they are. Both read a complete value rather than opening a container, and their elements are always scalars — which is exactly why only plain UBJSON needed the type-marker case above.

A container of no-ops keeps its behaviour of holding no elements while still announcing its declared size to the SAX parser.

## Verification

- `unit-ubjson` and `unit-bjdata` pass unchanged — **1.39 million assertions** between them — plus 22 new ones.
- Compared against the parent commit over every container form (sized, unsized, typed, untyped, empty, no-op, ND-array, binary, and the forms nested inside one another): identical values, error codes, messages and byte offsets.
- 500,000 levels of each vector now report a parse error instead of crashing; a well-formed 100,000-level value is read to completion.

## Performance

About **3 %** on the two inputs described in #5506; reading the frame once per element rather than per branch halved what it cost initially (8 % → 3.5 % on the scalar case).

## API impact

## Checklist

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -2205,17 +2205,18 @@ class binary_reader
             }
 
             // advance to the next element, closing the containers that ended.
-            // The reference is not held across get_ubjson_value() above, which
-            // can push onto the stack and reallocate it.
+            // top is a copy, not a reference: it must stay valid across the
+            // pop_back() below, which destroys the container_stack element it
+            // would otherwise alias.
             for (;;)
             {
-                container_frame& top = container_stack.back();
+                container_frame top = container_stack.back();
 
                 if (top.remaining != npos)
                 {
                     if (top.remaining != 0)
                     {
-                        --top.remaining;
+                        --container_stack.back().remaining;
                         if (top.is_object)
                         {
                             key.clear();
@@ -2254,9 +2255,8 @@ class binary_reader
                     break;
                 }
 
-                const bool is_object = top.is_object;
                 container_stack.pop_back();
-                if (JSON_HEDLEY_UNLIKELY(is_object ? !sax->end_object() : !sax->end_array()))
+                if (JSON_HEDLEY_UNLIKELY(top.is_object ? !sax->end_object() : !sax->end_array()))
                 {
                     return false;
                 }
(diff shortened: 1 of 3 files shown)


include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -2169,7 +2175,103 @@ class binary_reader
...
+                    }
+                    else
+                    {
+                        prefix = current;
+                    }
+                    break;
+                }
+
+                const bool is_object = top.is_object;
+                container_stack.pop_back();
```
gregmarr, line 2258: This invalidates `top`, have a comment here, copy `top`, or just extract the needed `top` values.

nlohmann: Fixed in d46c80b98: same fix as the CBOR reader (#5506) - top is now a copy (container_frame top = container_stack.back();), so it stays valid regardless of what happens to container_stack afterward. The one mutation (--top.remaining) now goes through container_stack.back() directly.

Written by Claude Code.

gregmarr (approved):

PR #5508, opened 2026-09-06
Read BSON documents without recursing per nesting level

nlohmann:
Closes the last vector of #5104.

## What

An embedded document (record type `0x03`) or array (`0x04`) was read by calling back into the document reader, which read its element list, which called the element reader again for the next embedded one. The native call stack grew with the nesting depth of the input, and **about seven bytes buy a level**, so a document of a few hundred kilobytes crashes the process.

## How

The same shape as the other three formats: `open_bson_document()` reads the size prefix and opens the document, `parse_bson_element_internal()` calls it for both record types instead of recursing, and `parse_bson_internal()` loops over the element list of whichever document is innermost, closing it when its terminator is reached and resuming the one below.

`check_bson_document_size()` is unchanged, and so is **when** it runs: a document is still measured from the byte before its size prefix to the byte after its terminator, and still reported before the end event. The frame carries those two values, which is what a per-document check needs once the reads are interleaved rather than nested.

## Verification

- `unit-bson` passes unchanged, plus 8 new assertions.
- Round trips through `to_bson` of nested objects, arrays, arrays of objects and mixed nesting are identical to the parent commit, as are the errors for a truncated document, an unsupported record type, a negative size and a size that does not match, including their byte offsets.
- A 30,000-level document built by `to_bson` is now read to completion where it used to crash.

## Sequencing with #5185

#5185 changes `parse_bson_internal()`, the element list and the array reader — the exact functions this PR restructures. **It should land first**; this PR then keeps its checks and moves them onto the loop. I have deliberately not tried to pre-empt it.

## API impact

## Checklist

### DIFF
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -337,8 +337,10 @@ class binary_reader
 
             if (element_type == 0) // end of the innermost document
             {
-                const container_frame& top = container_stack.back();
-                const bool is_object = top.is_object;
+                // a copy, not a reference: it must stay valid across the
+                // pop_back() below, which destroys the container_stack
+                // element it would otherwise alias
+                const container_frame top = container_stack.back();
 
                 if (JSON_HEDLEY_UNLIKELY(!check_bson_document_size(top.start_position, top.declared_size)))
                 {
@@ -346,7 +348,7 @@ class binary_reader
                 }
 
                 container_stack.pop_back();
-                if (JSON_HEDLEY_UNLIKELY(is_object ? !sax->end_object() : !sax->end_array()))
+                if (JSON_HEDLEY_UNLIKELY(top.is_object ? !sax->end_object() : !sax->end_array()))
                 {
                     return false;
                 }
--- a/include/nlohmann/detail/input/binary_reader.hpp
+++ b/include/nlohmann/detail/input/binary_reader.hpp
@@ -206,9 +206,14 @@ class binary_reader
         /// number of elements that have not been read yet, or npos when the
         /// container is not sized and ends at a marker instead
         std::size_t remaining;
+        /// BSON: value of chars_read before this document's size prefix, which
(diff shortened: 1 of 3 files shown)


include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -271,31 +276,99 @@ class binary_reader
...
         }
 
-        if (JSON_HEDLEY_UNLIKELY(!sax->start_object(detail::unknown_size())))
+        if (JSON_HEDLEY_UNLIKELY(!enter_container(is_object, detail::unknown_size())))
         {
             return false;
         }
 
-        if (JSON_HEDLEY_UNLIKELY(!parse_bson_element_list(/*is_array*/false)))
+        container_stack.back().start_position = document_start;
```
gregmarr: only call `back` once?  The compiler may optimize this out anyway.

include/nlohmann/detail/input/binary_reader.hpp
```diff
@@ -283,31 +288,100 @@ class binary_reader
...
+            {
+                const container_frame& top = container_stack.back();
+                const bool is_object = top.is_object;
+
+                if (JSON_HEDLEY_UNLIKELY(!check_bson_document_size(top.start_position, top.declared_size)))
+                {
+                    return false;
+                }
+
+                container_stack.pop_back();
```
gregmarr, line 350: This invalidates `top`, have a comment here, copy `top`, or just extract the needed `top` values.

nlohmann: Fixed in f51ea1ce1: same fix as the CBOR (#5506) and UBJSON/BJData (#5507) readers - top is now a copy (const container_frame top = container_stack.back();), so it stays valid regardless of what happens to container_stack afterward. Nothing here mutates the live entry, so no field needs redirecting through container_stack.back().

Written by Claude Code.

gregmarr (approved):

PR #5511, opened 2026-09-07
Split unit-regression2.cpp so the MinGW linker can relocate it

nlohmann:
Prerequisite for the binary-reader series (#5104, PRs #5501–#5508); it does not change library code.

## What

Linking `test-regression2` with clang and MinGW fails once the translation unit grows past a certain size:

```
relocation truncated to fit: IMAGE_REL_AMD64_REL32 against `.rdata'
```

The code can no longer reach the read-only data it references within the range of a 32-bit relocation. `unit-regression2.cpp` is one of the largest files in the test suite and had been sitting just under that limit, so a change anywhere in the library is enough to tip it over — the iterative binary readers in this series do exactly that, on 9 of 9 clang versions.

Two things worth noting:

- This is the **second** time this file has hit a size limit: it exists because `unit-regression1.cpp` was split for size.
- `windows.yml` already carries a workaround for the *same limit* hitting the *same target* — `-DCMAKE_CXX_FLAGS_DEBUG="-g0"`, because linking it failed with `IMAGE_REL_AMD64_SECREL against '.debug_line'`. Dropping debug info fixed that relocation. This one is against `.rdata`, which no compiler flag avoids.

## How

Move the second half of the regression tests, and the helper types only they use, into `unit-regression3.cpp`. The sections are independent — every statement in `TEST_CASE("regression tests 2")` was already inside a `SECTION` — so they move unchanged.

Nothing is lost: **168 assertions before the split, 50 + 118 after.**

Both files end up comfortably smaller than the one that used to link (clang, `-O1`, C++20):

| | read-only data | text | object |
|---|---|---|---|
| before | 58,233 | 1,287,764 | 3,158,120 |
| `unit-regression2.cpp` | 48,161 | 1,012,988 | 2,522,296 |
| `unit-regression3.cpp` | 41,710 | 772,704 | 1,878,880 |

No CMake change is needed: `tests/CMakeLists.txt` globs `src/unit-*.cpp`, so the new file is built for every standard like its siblings.

`CONTRIBUTING.md` pointed contributors at `unit-regression2.cpp` for new bug tests; it now points at the smaller file and explains why the two exist, so the split doesn't quietly undo itself.

## API impact

**No breaking changes.** Test-only change; no library code is touched.

## Checklist


.github/CONTRIBUTING.md
```diff
@@ -108,7 +108,9 @@ The tests are located in [`tests/src/unit-*.cpp`](https://github.com/nlohmann/js
 are structured along the features of the library or the nature of the tests. Usually, it should be clear from the
 context which existing file needs to be extended, and only very few cases require creating new test files.
 
-When fixing a bug, edit `unit-regression2.cpp` and add a section referencing the fixed issue.
+When fixing a bug, edit `unit-regression3.cpp` and add a section referencing the fixed issue.
+`unit-regression2.cpp` holds the older tests; the two files exist because a single one grew large enough for the
+MinGW linker to fail relocating it, so please keep adding to the smaller file rather than growing one of them.
```
gregmarr: "growing the larger one"?

gregmarr (approved):

.github/CONTRIBUTING.md
nlohmann: Will fix - once the running CI jobs finish.

gregmarr (approved):

PR #5515, opened 2026-09-08
handled when size exceed uint32

### LINKED ISSUE #5320: to_msgpack() silently emits corrupt output (or reads out of bounds) above the documented 4294967295 size limit

dsp0redy:
Fixes : https://github.com/nlohmann/json/issues/5320

to_msgpack() now throws out_of_range.412 when a string, array, binary, or object reports a size above UINT32_MAX, matching the documented MessagePack limit and the BSON fix pattern from https://github.com/nlohmann/json/pull/5314.

Adds a regression test using a container type that reports an oversized size() without allocating multi-gigabyte data.
----------------------------------------------------------------------
 - [x] If applicable, the [documentation](https://json.nlohmann.me/) is updated.
 - [x] The source code is amalgamated by running make amalgamate.


include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1937,4 +1957,4 @@ class binary_writer
 };
 
 }  // namespace detail
-NLOHMANN_JSON_NAMESPACE_END
+NLOHMANN_JSON_NAMESPACE_END
```
gregmarr: You lost the newline at the end of the file.

tests/src/unit-msgpack.cpp
```diff
@@ -1967,4 +1967,169 @@ TEST_CASE("MessagePack with std::byte")
...
+    j.get_binary().push_back(0x01);
+    j.get_binary().push_back(0x02);
+
+    CHECK_THROWS_WITH_AS(
+        huge_json::to_msgpack(j),
+        "[json.exception.out_of_range.412] MessagePack size 4294967296 exceeds maximum of 4294967295",
+        json::out_of_range&);
+}
+
+#endif
```
gregmarr: No newline.

gregmarr:
There is no update to the amalgamated header.  There will be a CI notification for that eventually.

nlohmann (commented):
Automated review findings — posted via [Claude Code](https://claude.com/claude-code) on behalf of @nlohmann.

This PR's CI is currently failing (`gh pr checks 5515`) for reasons visible directly in the diff:

1. A local `using huge_json = ...` alias is redeclared with the same name in 3 of the 4 new `TEST_CASE`s, shadowing the global alias declared for the array test. This trips `-Werror -Wshadow=global` and is the root cause of the `ci_test_gcc`, `ci_test_standards_gcc(23)`, `ci_test_standards_clang(14)`, MSVC, MSVC-arm64, mingw, and clang-cl failures on the current run.
2. `single_include/nlohmann/json.hpp` was never regenerated for this change (the diff touches only 2 files), so `ci_test_amalgamation` also fails.
3. The 4 new regression tests ended up nested inside the pre-existing `#ifdef JSON_HAS_CPP_17` block — the old `#endif` that closed it was deleted and only re-added after the new tests — even though none of them need C++17. They're compiled out entirely under this project's C++11/C++14 test builds.

See inline comments below for exact locations.

tests/src/unit-msgpack.cpp
```diff
@@ -1967,4 +1967,169 @@ TEST_CASE("MessagePack with std::byte")
...
+        if (fake_size)
+        {
+            return (std::numeric_limits<std::uint32_t>::max)() + 1ULL;
+        }
+
+        return base::size();
+    }
+};
+
+using huge_json = nlohmann::basic_json <
```
nlohmann: This global `huge_json` alias is shadowed by local `using huge_json = ...` redeclarations at lines 2037, 2078, and 2111 (the object/string/binary test cases). Under this project's `-Werror -Wshadow=global` build this is a hard compile error — confirmed from the PR's actual `ci_test_gcc` failure log:
```
tests/src/unit-msgpack.cpp:2037:11: error: declaration of 'using huge_json = ...' shadows a global declaration [-Werror=shadow]
tests/src/unit-msgpack.cpp:1990:7: note: shadowed declaration is here
```
(and again at 2078 and 2111). Renaming the three local aliases (e.g. `huge_object_json`, `huge_string_json`, `huge_binary_json`) would fix it.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -578,6 +578,11 @@ class binary_writer
                     oa->write_character(to_char_type(0xDB));
                     write_number(static_cast<std::uint32_t>(N));
                 }
+                else
+                {
+                    JSON_THROW(out_of_range::create(412, concat("MessagePack size ", std::to_string(N), " exceeds maximum of ",
```
nlohmann, line 583: `single_include/nlohmann/json.hpp` isn't updated anywhere in this diff even though this adds new throwing branches to `write_msgpack()`. Since `single_include` is generated from `include/` via `make amalgamate`, and the PR checklist above claims that was run, this looks like an oversight — it's currently failing the `ci_test_amalgamation` CI job.

tests/src/unit-msgpack.cpp
nlohmann: This closes the `#ifdef JSON_HAS_CPP_17` block that opens at line 1893 (originally right before the pre-existing "MessagePack with std::byte" test). This diff deleted the `#endif` that used to close that block right after that test, and only re-added it here — so all 4 new regression tests above (and the `huge_array`/`huge_map`/`huge_string`/`huge_binary` helpers) are now nested inside the C++17 guard, even though none of them use any C++17-only feature. This project builds separate test binaries per C++ standard (11/14/17/20/23); under the C++11 and C++14 ones, `JSON_HAS_CPP_17` is undefined, so these regression tests are silently skipped there.

nlohmann (approved):
Looks good to me.

unaPoloGTIc:
This branch last synced with develop on Sep 11 and CI is failing across all platforms. Could you rebase onto latest develop? The msgpack fix itself looks solid — just needs a fresh CI run.

nlohmann:
@dsp0redy Can you please take a look at the PR to finalize it?

dsp0redy:
> @dsp0redy Can you please take a look at the PR to finalize it?

Sure I will check

Issue #5393, opened 2026-08-21
Stack overflow in diff(), merge_patch(), and flatten() due to unbounded recursion


nlohmann:
## Summary

Three value algorithms recurse once per nesting level with no depth limit, so each crashes the process on a deeply nested value: `json::diff()`, `basic_json::merge_patch()` and `json_pointer::flatten()` (reached through `basic_json::flatten()`).

The text parser is iterative, so `json::parse` accepts input of any depth. Any of these three called on such a value is a crash, which matters wherever the value came from somewhere untrusted — JSON Patch and JSON Merge Patch are typically applied to exactly that.

`unflatten()` is **not** affected: it iterates the object and lets `get_and_create` walk each pointer, and it handles a 500,000-level pointer without trouble. I checked, so that nobody has to.

## Reproduction

Against `develop` (734fd305a), `-O2`, macOS/arm64 with the default 8 MB stack. All three crash at depth 100,000:

```cpp
#include <nlohmann/json.hpp>
#include <string>

int main()
{
    const std::size_t depth = 100000;
    const auto deep = [&](const char* leaf)
    {
        return nlohmann::json::parse(std::string(depth, '[') + leaf + std::string(depth, ']'));
    };

    const nlohmann::json a = deep("0");
    const nlohmann::json b = deep("1");

    const auto patch = nlohmann::json::diff(a, b);   // SIGSEGV
    return static_cast<int>(patch.size());
}
```

| call | result |
|---|---|
| `json::diff(a, b)` | SIGSEGV |
| `target.merge_patch(patch)` | SIGSEGV |
| `j.flatten()` | SIGSEGV |
| `j.unflatten()` | fine — iterative |

`merge_patch` is measured with nested objects, since it only descends into objects.

## Root cause

- **`diff`** — `basic_json::diff()` (json.hpp:5089) calls itself per element: json.hpp:5120 (arrays), :5168 (objects)
- **`merge_patch`** — `basic_json::merge_patch()` (json.hpp:5231) calls itself per member: json.hpp:5247
- **`flatten`** — `json_pointer::flatten()` (json_pointer.hpp:861) calls itself per element: json_pointer.hpp:879 (arrays), :898 (objects)

## Suggested fix

The same shape used for the destructor in #1436 and for copying, serializing and comparing in #5389 / #5285 / #5390: descend a bounded number of levels, then finish the rest on an explicit stack.

[...]

PR #5517, opened 2026-09-09
Prevent stack overflow in json_pointer::flatten

dajiaohuang:
## Summary

Replace the recursive implementation of `json_pointer::flatten()` with an explicit work stack so deeply nested values no longer exhaust the call stack. Existing traversal order and empty-container handling are preserved.

This addresses the `flatten()` portion of #5393.

## Validation

- `test-regression2_cpp11` and `test-regression2_cpp20` pass, including a 100,000-level nesting regression.
- `test-json_pointer_cpp11` and `test-json_pointer_cpp20` pass.
- The generated-header `single_detail_json_pointer` smoke executable passes.
- `git diff --check` passes.

- [x] An existing issue is referenced.
- [x] The source code is amalgamated.

### DIFF
--- a/include/nlohmann/detail/json_pointer.hpp
+++ b/include/nlohmann/detail/json_pointer.hpp
@@ -910,7 +910,7 @@ class json_pointer
                         {
                             const auto index = i - 1;
                             stack.emplace_back(detail::concat<string_t>(current.reference_string, '/', std::to_string(index)),
-                                                &current.value->m_data.m_value.array->operator[](index));
+                                               &current.value->m_data.m_value.array->operator[](index));
                         }
                     }
                     break;
--- a/include/nlohmann/detail/json_pointer.hpp
+++ b/include/nlohmann/detail/json_pointer.hpp
@@ -876,58 +876,81 @@ class json_pointer
                         const BasicJsonType& value,
                         BasicJsonType& result)
     {
-        switch (value.type())
+        struct flatten_task
         {
-            case detail::value_t::array:
-            {
-                if (value.m_data.m_value.array->empty())
-                {
-                    // flatten empty array as null
-                    result[reference_string] = nullptr;
-                }
-                else
-                {
-                    // iterate array and use index as a reference string
-                    for (std::size_t i = 0; i < value.m_data.m_value.array->size(); ++i)
-                    {
-                        flatten(detail::concat<string_t>(reference_string, '/', std::to_string(i)),
-                                value.m_data.m_value.array->operator[](i), result);
-                    }
-                }
-                break;
-            }
+            string_t reference_string;
+            const BasicJsonType* value;
+        };
 
-            case detail::value_t::object:
-            {
-                if (value.m_data.m_value.object->empty())
-                {
-                    // flatten empty object as null
-                    result[reference_string] = nullptr;
-                }
-                else
-                {
-                    // iterate object and use keys as reference string
-                    for (const auto& element : *value.m_data.m_value.object)
-                    {
-                        flatten(detail::concat<string_t>(reference_string, '/', detail::escape(element.first)), element.second, result);
-                    }
-                }
-                break;
-            }
+        std::vector<flatten_task> stack;
+        stack.push_back({reference_string, &value});
 
-            case detail::value_t::null:
-            case detail::value_t::string:
-            case detail::value_t::boolean:
-            case detail::value_t::number_integer:
-            case detail::value_t::number_unsigned:
-            case detail::value_t::number_float:
-            case detail::value_t::binary:
-            case detail::value_t::discarded:
-            default:
+        while (!stack.empty())
+        {
+            auto current = std::move(stack.back());
+            stack.pop_back();
+
+            switch (current.value->type())
             {
-                // add a primitive value with its reference string
-                result[reference_string] = value;
-                break;
+                case detail::value_t::array:
+                {
+                    if (current.value->m_data.m_value.array->empty())
+                    {
+                        // flatten empty array as null
+                        result[current.reference_string] = nullptr;
+                    }
+                    else
+                    {
+                        // iterate array and use index as a reference string
+                        for (std::size_t i = current.value->m_data.m_value.array->size(); i > 0; --i)
+                        {
+                            const auto index = i - 1;
+                            stack.push_back({detail::concat<string_t>(current.reference_string, '/', std::to_string(index)),
+                                             &current.value->m_data.m_value.array->operator[](index)});
+                        }
+                    }
+                    break;
+                }
+
+                case detail::value_t::object:
+                {
+                    if (current.value->m_data.m_value.object->empty())
+                    {
+                        // flatten empty object as null
+                        result[current.reference_string] = nullptr;
+                    }
+                    else
+                    {
+                        // iterate object and use keys as reference string
+                        std::vector<flatten_task> children;
+                        children.reserve(current.value->m_data.m_value.object->size());
+                        for (const auto& element : *current.value->m_data.m_value.object)
+                        {
+                            children.push_back({detail::concat<string_t>(current.reference_string, '/', detail::escape(element.first)),
+                                                &element.second});
+                        }
+                        for (auto it = children.rbegin(); it != children.rend(); ++it)
+                        {
+                            stack.push_back(std::move(*it));
+                        }
+                    }
+                    break;
+                }
+
+                case detail::value_t::null:
+                case detail::value_t::string:
+                case detail::value_t::boolean:
+                case detail::value_t::number_integer:
+                case detail::value_t::number_unsigned:
+                case detail::value_t::number_float:
+                case detail::value_t::binary:
+                case detail::value_t::discarded:
+                default:
+                {
+                    // add a primitive value with its reference string
+                    result[current.reference_string] = *current.value;
+                    break;
+                }
             }
         }
     }
--- a/include/nlohmann/detail/json_pointer.hpp
+++ b/include/nlohmann/detail/json_pointer.hpp
@@ -880,10 +880,14 @@ class json_pointer
         {
             string_t reference_string;
             const BasicJsonType* value;
+
+            flatten_task(string_t reference_string_, const BasicJsonType* value_)
+                : reference_string(std::move(reference_string_)), value(value_)
+            {}
         };
 
         std::vector<flatten_task> stack;
-        stack.push_back({reference_string, &value});
+        stack.emplace_back(reference_string, &value);
 
         while (!stack.empty())
         {
@@ -905,8 +909,8 @@ class json_pointer
                         for (std::size_t i = current.value->m_data.m_value.array->size(); i > 0; --i)
                         {
                             const auto index = i - 1;
-                            stack.push_back({detail::concat<string_t>(current.reference_string, '/', std::to_string(index)),
-                                             &current.value->m_data.m_value.array->operator[](index)});
+                            stack.emplace_back(detail::concat<string_t>(current.reference_string, '/', std::to_string(index)),
+                                                &current.value->m_data.m_value.array->operator[](index));
                         }
                     }
                     break;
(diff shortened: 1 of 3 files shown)


include/nlohmann/detail/json_pointer.hpp
```diff
@@ -876,58 +876,81 @@ class json_pointer
...
-        switch (value.type())
+        struct flatten_task
         {
-            case detail::value_t::array:
+            string_t reference_string;
+            const BasicJsonType* value;
+        };
+
+        std::vector<flatten_task> stack;
+        stack.push_back({reference_string, &value});
```
gregmarr: Should be able to use `emplace_back(reference_string, &value);`

include/nlohmann/detail/json_pointer.hpp
```diff
@@ -876,58 +876,81 @@ class json_pointer
...
+                    }
+                    else
                     {
-                        flatten(detail::concat<string_t>(reference_string, '/', std::to_string(i)),
-                                value.m_data.m_value.array->operator[](i), result);
+                        // iterate array and use index as a reference string
+                        for (std::size_t i = current.value->m_data.m_value.array->size(); i > 0; --i)
+                        {
+                            const auto index = i - 1;
+                            stack.push_back({detail::concat<string_t>(current.reference_string, '/', std::to_string(index)),
```
gregmarr: `emplace_back`

include/nlohmann/detail/json_pointer.hpp
```diff
@@ -876,58 +876,81 @@ class json_pointer
...
+                    }
+                    else
                     {
-                        flatten(detail::concat<string_t>(reference_string, '/', detail::escape(element.first)), element.second, result);
+                        // iterate object and use keys as reference string
+                        std::vector<flatten_task> children;
+                        children.reserve(current.value->m_data.m_value.object->size());
+                        for (const auto& element : *current.value->m_data.m_value.object)
+                        {
+                            children.push_back({detail::concat<string_t>(current.reference_string, '/', detail::escape(element.first)),
```
gregmarr: `emplace_back`

include/nlohmann/detail/json_pointer.hpp
```diff
@@ -876,58 +876,81 @@ class json_pointer
...
+                    {
+                        // flatten empty array as null
+                        result[current.reference_string] = nullptr;
+                    }
+                    else
                     {
-                        flatten(detail::concat<string_t>(reference_string, '/', std::to_string(i)),
-                                value.m_data.m_value.array->operator[](i), result);
+                        // iterate array and use index as a reference string
+                        for (std::size_t i = current.value->m_data.m_value.array->size(); i > 0; --i)
```
gregmarr, line 909: Comment about why this is iterating backwards?

include/nlohmann/detail/json_pointer.hpp
dajiaohuang: Implemented in 46d2b4d: added the C++11-compatible task constructor so the initial frame uses emplace_back(reference_string, &value). The focused C++11/C++20 regression builds pass.

include/nlohmann/detail/json_pointer.hpp
dajiaohuang: Implemented in 46d2b4d: array child frames now use emplace_back with the same traversal and pointer semantics. The focused C++11/C++20 regression builds pass.

include/nlohmann/detail/json_pointer.hpp
dajiaohuang: Implemented in 46d2b4d: object child frames now use emplace_back; the generated single-header smoke build also passes.

include/nlohmann/detail/json_pointer.hpp
dajiaohuang: Documented in 46d2b4d: children are pushed in reverse because the work stack is LIFO, preserving the original object iteration order. The issue #5393 regression passes in C++11 and C++20.

dajiaohuang:
Applied the amalgamation-patch artifact from run 34485452558. Commit fe2f71eaddf88d771c9d40d929cd65edf8a69520 updates only include/nlohmann/detail/json_pointer.hpp and single_include/nlohmann/json.hpp to match the generated formatting, and is pushed to fix/5393-iterative-flatten. Validation passed: Release C++11 and C++20 issue #5393 regression selections (1 test, 3 assertions each), Release single_detail_json_pointer build and smoke run, and git diff --check.

dajiaohuang:
The GCC C++26 standards jobs exposed a -Werror=noexcept failure for the iterative flatten stack when instantiated with the alternative string type: latten_task's move construction was inferred as potentially throwing. Commit 7121dfd44 marks that constructor 
oexcept in both the source and amalgamated headers. git diff --check passes; the focused C++26 CI jobs will provide the hosted validation.

dajiaohuang:
Amended the latest fix commit with the repository-required DCO sign-off and force-updated the fork branch. New head: 0eabb210e. No source changes beyond the Signed-off-by trailer; the previous C++26 noexcept fix remains intact.

gregmarr:
There are build failures related to the new type having pointers but not implementing or deleting the copy/move assignment operators.

include/nlohmann/detail/json_pointer.hpp
nlohmann: Already applied in 46d2b4d73 (`stack.emplace_back(...)`), same as the other two sites — just wasn't marked resolved. Resolving.

dajiaohuang:
Declared the flatten_task copy/move special members in the source and generated headers in commit 1d819ec5f14b31257079e57119f0718418ddb794 to satisfy GCC -Weffc++ while preserving the existing non-owning pointer semantics. git diff --check passes; local C++ toolchain is unavailable on Windows, so hosted CI is the remaining compile validation.

include/nlohmann/detail/json_pointer.hpp
```diff
@@ -876,58 +876,91 @@ class json_pointer
...
+            const BasicJsonType* value;
+
+            flatten_task(string_t reference_string_, const BasicJsonType* value_) noexcept
+                : reference_string(std::move(reference_string_)), value(value_)
+            {}
+
+            flatten_task(const flatten_task&) = default;
+            flatten_task(flatten_task&&) noexcept = default;
+            flatten_task& operator=(const flatten_task&) = default;
+            flatten_task& operator=(flatten_task&&) noexcept = default;
```
nlohmann: Rule of zero fixes both CI failures: GCC 4.8 rejects the `noexcept` defaulted moves for `alt_string`, and clang-tidy wants a destructor if these are declared. The implicit members get the right exception specifications. The constructor moves a `string_t` that may throw, so it shouldn't be `noexcept` either:

```suggestion
            flatten_task(string_t reference_string_, const BasicJsonType* value_)
                : reference_string(std::move(reference_string_)), value(value_)
            {}
```

tests/src/unit-regression2.cpp
```diff
@@ -1571,4 +1571,14 @@ TEST_CASE("issue #5402 - update(merge_objects=true) overwrites a primitive with
     CHECK(mixed == json({{"keep", {{"a", 1}, {"b", 2}}}, {"replace", {{"x", 2}}}}));
 }
 
+TEST_CASE("issue #5393 - flatten deeply nested values without overflowing the stack")
```
nlohmann: Could this move to `tests/src/unit-large_json.cpp`? That's where the other deep-nesting tests live, and `unit-regression2.cpp` is at the size the MinGW linker still copes with (it had to be split in #5511).

nlohmann (changes requested):
Thanks, this does exactly what #5393 asks for `flatten()`, and the work stack keeps the original traversal order.

I verified it locally, merged into current `develop` (the branch is 44 commits behind):

- **Identical output.** I compared `flatten()` and `unflatten(flatten(x))` against `develop` on 10,010 random values: `json` and `ordered_json`, keys containing `/` and `~` (so escaping is exercised), empty arrays and objects, and chains 100-200 levels deep with siblings. All 20,020 outputs match exactly. On the branch *as is*, some `unflatten` results differ, but that comes from #5443 landing on `develop` since your base, not from this PR. Merged, it is identical.
- **Deep values.** It handles 1,000,000 levels of nested arrays and of nested objects on a **1 MiB** stack, where `develop` crashes at 100,000. Memory improves too: 5 MB instead of 466 MB at 20,000 levels, because the recursive version kept every level's reference string alive.

**Two CI failures block it, and one change fixes both:**

1. **GCC 4.8** (`ci_test_compilers_gcc_old (4.8)`) rejects `flatten_task& operator=(flatten_task&&) noexcept = default;` because it is "defaulted ... with an exception-specification that differs from the implicit declaration". The instantiation that trips it is `unit-alt-string.cpp`, whose `alt_string` has a move assignment that is not `noexcept`. For such string types the explicit `noexcept` is not merely rejected by old compilers, it is wrong.
2. **clang-tidy** (`cppcoreguidelines-special-member-functions`) complains that the struct declares the copy/move members but no destructor.

Dropping all four explicitly defaulted members fixes both (rule of zero). The implicit ones are generated with the right exception specifications, so moves stay `noexcept` wherever `string_t`'s are. I'd also drop the `noexcept` on the converting constructor: it moves a `string_t` that may throw, and `noexcept` would turn that into `std::terminate`. See the suggestion. I ran clang-tidy locally with the repository's `.clang-tidy` on the suggested version, and the special-member finding is gone.

Two smaller things:

- Please move the test from `unit-regression2.cpp` to `unit-large_json.cpp`, where the other deep-nesting tests live. `unit-regression2.cpp` is at the size the MinGW linker still copes with; it already had to be split once (#5511).
- Please rebase onto `develop` so CI runs against the current code.

Optional: a newer clang-tidy than CI's (23) additionally suggests `const auto current = std::move(stack.back());` (`misc-const-correctness`). It's harmless either way.

---
_This review was written by [Claude Code](https://claude.com/claude-code) on behalf of @nlohmann._

include/nlohmann/detail/json_pointer.hpp
gregmarr: These were added because of build failures without them.  https://github.com/nlohmann/json/pull/5517#issuecomment-5637965914

gregmarr:
```
error: 'flatten_task' has pointer data members [-Werror=effc++]
  879 |         struct flatten_task
      |                ^~~~~~~~~~~~
error:   but does not declare 'flatten_task(const flatten_task&)' [-Werror=effc++]
error:   or 'operator=(const ::flatten_task&)' [-Werror=effc++]
note: pointer member 'flatten_task::value' declared here
  882 |             const BasicJsonType* value;
      |                                  ^~~~~
```

nlohmann: You're right, thanks. My rule-of-zero suggestion brings back exactly that error; I reproduced it with GCC 16 and the flags from `cmake/gcc_flags.cmake`. GCC's `-Weffc++` check fires for a class that has pointer members, a user-declared constructor and a non-trivial destructor (the `string_t` member makes it non-trivial), unless it declares the copy constructor and copy assignment. So the special members have to stay, and the review body's "one change fixes both" is wrong.

What each CI failure actually wants:

- **`-Weffc++`**: declare the copy constructor and copy assignment.
- **GCC 4.8**: no *explicit* `noexcept` on the defaulted moves. For `alt_string`, whose moves may throw, it differs from the implicit specification.
- **clang-tidy** (`cppcoreguidelines-special-member-functions`): once those are declared, declare the destructor too.
- **`-Wnoexcept`** (the C++26 job): keep `noexcept` on the converting constructor, as you already have.

Declaring all five as defaulted, without explicit exception specifications, satisfies all of them. The implicit specifications still make the moves `noexcept` wherever `string_t`'s are:

```suggestion
            flatten_task(string_t reference_string_, const BasicJsonType* value_) noexcept
                : reference_string(std::move(reference_string_)), value(value_)
            {}

            flatten_task(const flatten_task&) = default;
            flatten_task(flatten_task&&) = default;
            flatten_task& operator=(const flatten_task&) = default;
            flatten_task& operator=(flatten_task&&) = default;
            ~flatten_task() = default;
```

Checked on this PR merged into `develop`:

- GCC 16 with the CI flag set compiles `unit-alt-string.cpp`, `unit-json_pointer.cpp` and `unit-regression2.cpp` cleanly in both C++11 and C++26.
- The repository's clang-tidy config reports no special-member finding.
- For comparison: the rule-of-zero version fails `-Weffc++` in all six compiles, and the current version fails clang-tidy.

I couldn't run GCC 4.8 locally. Its error was specifically about the explicit `noexcept`, which this removes, so CI will confirm.

The same trap caught a frame struct in my own #5548, fixed the same way.

---
_Written by [Claude Code](https://claude.com/claude-code) on behalf of @nlohmann._

Issue #5392, opened 2026-08-21
Stack overflow in binary format writers (CBOR/MessagePack/UBJSON/BSON) due to unbounded recursion


nlohmann:
## Summary

The binary format **writers** (`binary_writer.hpp`) recurse once per nesting level with no depth limit, so serializing a deeply nested value crashes the process. This is the counterpart of #5104, which covers the binary *readers*; both directions are affected, and neither is fixed.

The text parser is iterative, so `json::parse` accepts input of any depth. A value that the library happily builds can therefore crash it on the way back out.

## Affected functions

- `json::to_cbor()`
- `json::to_msgpack()`
- `json::to_ubjson()` (and BJData)
- `json::to_bson()`

## Reproduction

Against `develop` (734fd305a), on macOS/arm64 with the default 8 MB stack:

```cpp
#include <nlohmann/json.hpp>
#include <string>

int main()
{
    const std::size_t depth = 200000;
    const nlohmann::json j = nlohmann::json::parse(
        std::string(depth, '[') + "0" + std::string(depth, ']'));

    const auto v = nlohmann::json::to_cbor(j);   // SIGSEGV
    return static_cast<int>(v.size());
}
```

| function | `-O0` | `-O2` |
|---|---|---|
| `to_cbor` | crashes at depth 100,000 | crashes at depth 200,000 |
| `to_msgpack` | crashes at depth 100,000 | crashes at depth 200,000 |
| `to_ubjson` | crashes at depth 100,000 | crashes at depth 100,000 |
| `to_bson` | crashes at depth 100,000 | crashes at depth 100,000 |

The depth needed varies with build and stack size, as usual for stack exhaustion; CBOR and MessagePack survive depth 100,000 at `-O2` on this machine and fail at 200,000. BSON is measured with a nested object, since BSON requires an object at the root.

## Root cause

Each writer calls itself for every element:

- **CBOR**: `write_cbor()` (binary_writer.hpp:96) → binary_writer.hpp:303 (array), :405/:406 (object)
- **MessagePack**: `write_msgpack()` (:420) → :614 (array), :727/:728 (object)
- **UBJSON/BJData**: `write_ubjson()` (:747) → :842 (array), :944 (object)
- **BSON**: `write_bson_element()` (:1238) → `write_bson_object()` (:1302) / `write_bson_array()` (:1160) → `write_bson_element()` (:1306/:1166)

[...]

Issue #5308, opened 2026-07-26
to_bson() is O(size × nesting depth), not linear as documented


nlohmann:
### Description

BSON documents are length-prefixed, so `write_bson_object` (`include/nlohmann/detail/output/binary_writer.hpp:1285`) calls `calc_bson_object_size` to compute the prefix — which recursively walks the entire subtree. It then recurses into each child, and each child walks *its* subtree again. `write_bson_array` (`binary_writer.hpp:1141` / `calc_bson_array_size` at `:1118`) does the same.

The result is that every nesting level re-measures everything below it, giving O(size × depth) instead of O(size).

`docs/mkdocs/docs/api/basic_json/to_bson.md` currently states:

> ## Complexity
>
> Linear in the size of the JSON value `j`.

which is not correct as written.

### Reproduction steps

Serialize objects with a roughly constant payload but increasing nesting depth and time `to_bson`.

### Minimal code example

```cpp
#include <nlohmann/json.hpp>
#include <chrono>
#include <iostream>
using json = nlohmann::json;

int main()
{
    for (int depth : {200, 400, 800, 1600})
    {
        json j = json::object();
        json* p = &j;
        for (int i = 0; i < depth; ++i)
        {
            (*p)["k"] = json::object();
            p = &(*p)["k"];
        }
        for (int i = 0; i < 200; ++i)
        {
            (*p)[std::to_string(i)] = i;
        }

        const auto t0 = std::chrono::steady_clock::now();
        const auto v = json::to_bson(j);
        const auto t1 = std::chrono::steady_clock::now();

        std::cout << "depth " << depth << "  bytes " << v.size() << "  time "
                  << std::chrono::duration<double, std::milli>(t1 - t0).count() << " ms\n";
    }
}
```

### Expected vs. actual results

**Expected:** runtime roughly proportional to output size.

**Actual:** runtime grows quadratically in depth — 27× the time for 4.4× the output:

```
depth 200   bytes 3295   time 1.35 ms
depth 400   bytes 4895   time 3.50 ms
depth 800   bytes 8095   time 10.08 ms
depth 1600  bytes 14495  time 36.04 ms
```

36 ms to produce 14 KB of BSON is a poor cost for the format, and it makes `to_bson()` a cheap amplification target when serializing attacker-shaped documents.

### Suggested fix

Either:

1. memoize sizes in a single bottom-up pass, or
[...]

PR #5518, opened 2026-09-10
Stop binary writers overflowing the stack on deep values

ayush-singh-0601:
See #5392.

`to_cbor`, `to_msgpack`, and `to_ubjson` recurse once per nesting level. Parsing is iterative, so a value this library will happily build can still SIGSEGV when you serialize it.

This keeps the current recursive write for the first 128 levels (normal documents) and finishes anything deeper on a heap stack. Encoding is the same.

BSON is not in this PR. Its writer also walks the tree a second time to compute document lengths, so that belongs with the size-pass work.

- [x] If applicable, the documentation is updated.

### DIFF
--- a/include/nlohmann/detail/output/binary_writer.hpp
+++ b/include/nlohmann/detail/output/binary_writer.hpp
@@ -93,8 +93,14 @@ class binary_writer
     /*!
     @param[in] j  JSON value to serialize
     */
-    void write_cbor(const BasicJsonType& j)
+    void write_cbor(const BasicJsonType& j, const int depth = 0)
     {
+        if (JSON_HEDLEY_UNLIKELY(depth >= binary_write_depth_limit()) && (j.is_array() || j.is_object()))
+        {
+            write_cbor_iterative(j);
+            return;
+        }
+
         switch (j.type())
         {
             case value_t::null:
@@ -268,39 +274,10 @@ class binary_writer
 
             case value_t::array:
             {
-                // step 1: write control byte and the array size
-                const auto N = j.m_data.m_value.array->size();
-                if (N <= 0x17)
-                {
-                    write_number(static_cast<std::uint8_t>(0x80 + N));
-                }
-                else if (N <= (std::numeric_limits<std::uint8_t>::max)())
-                {
-                    oa->write_character(to_char_type(0x98));
-                    write_number(static_cast<std::uint8_t>(N));
-                }
-                else if (N <= (std::numeric_limits<std::uint16_t>::max)())
-                {
-                    oa->write_character(to_char_type(0x99));
-                    write_number(static_cast<std::uint16_t>(N));
-                }
-                else if (N <= (std::numeric_limits<std::uint32_t>::max)())
-                {
-                    oa->write_character(to_char_type(0x9A));
-                    write_number(static_cast<std::uint32_t>(N));
-                }
-                // LCOV_EXCL_START
-                else if (N <= (std::numeric_limits<std::uint64_t>::max)())
-                {
-                    oa->write_character(to_char_type(0x9B));
-                    write_number(static_cast<std::uint64_t>(N));
-                }
-                // LCOV_EXCL_STOP
-
-                // step 2: write each element
+                write_cbor_array_prefix(j.m_data.m_value.array->size());
                 for (const auto& el : *j.m_data.m_value.array)
                 {
-                    write_cbor(el);
+                    write_cbor(el, depth + 1);
                 }
                 break;
             }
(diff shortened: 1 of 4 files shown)


nlohmann (changes requested):
Please update from the develop branch.

nlohmann:
Please rebase to the latest develop branch as I had to fix some tests.

ayush-singh-0601:
Rebased onto the latest develop.

The output-sink rewrite landed while this was open, so the branch no longer applied. The gcc job was also failing `-Weffc++` because the heap frames never initialized their iterators.

That is fixed. The depth limit sits on the new writer. Anything under 128 levels still uses the same recursive write, and anything deeper finishes on a heap stack. The bytes match what the recursive writer produced. I checked CBOR, MessagePack, UBJSON, and BJData against current develop, including values that cross the limit, and a 100000-deep array serializes instead of blowing the stack. BSON is still left out of this change.

Ready for another look.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -143,10 +143,26 @@ class binary_writer
     }
 
     /*!
-    @param[in] j  JSON value to serialize
+    @param[in] j      JSON value to serialize
+    @param[in] depth  nesting level of @a j, counted from the value passed to @ref write_cbor
```
gregmarr: Does this `@ref` need clarification, since this function IS `write_cbor`?  Is this referring to `basic_json::write_cbor`?  Same comment for other functions where `depth` was added.

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1037,6 +906,495 @@ class binary_writer
     }
 
   private:
+    /// the number of levels the recursive writers descend into before handing
+    /// over to the iterative writers below
+    static constexpr std::size_t binary_write_depth_limit()
```
gregmarr: Should this reuse the existing recursion limit rather than adding a new one?

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1037,6 +906,495 @@ class binary_writer
...
+        // detail/iterators/internal_iterator.hpp
+        struct frame
+        {
+            explicit frame(const BasicJsonType* value_) noexcept
+                : value(value_)
+                , started(false)
+            {}
+
+            const BasicJsonType* value;
+            bool started;
```
gregmarr: Should this be `bool started = false;` instead of manually setting it in the constructor?

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1037,6 +906,495 @@ class binary_writer
...
+    {
+        struct frame
+        {
+            explicit frame(const BasicJsonType* value_) noexcept
+                : value(value_)
+                , started(false)
+            {}
+
+            const BasicJsonType* value;
+            bool started;
```
gregmarr: as above

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1037,6 +906,495 @@ class binary_writer
...
 
   private:
+    /// the number of levels the recursive writers descend into before handing
+    /// over to the iterative writers below
+    static constexpr std::size_t binary_write_depth_limit()
+    {
+        return 128;
+    }
+
+    void write_cbor_array_prefix(const std::size_t N)
```
nlohmann: `std::size_t` makes the `uint64_t` comparison below (line 939) tautological on 64-bit targets, which fails every clang job. The container's own `size_type` is dependent on `BasicJsonType`, so clang doesn't diagnose it (that is why the replaced `const auto N = ...` code was fine):

```suggestion
    void write_cbor_array_prefix(const typename BasicJsonType::array_t::size_type N)
```

include/nlohmann/detail/output/binary_writer.hpp
```diff
@@ -1037,6 +906,495 @@ class binary_writer
...
+        // LCOV_EXCL_START
+        else if (N <= (std::numeric_limits<std::uint64_t>::max)())
+        {
+            oa.write_character(to_char_type(0x9B));
+            write_number(static_cast<std::uint64_t>(N));
+        }
+        // LCOV_EXCL_STOP
+    }
+
+    void write_cbor_object_prefix(const std::size_t N)
```
nlohmann: Same here, for line 970:

```suggestion
    void write_cbor_object_prefix(const typename BasicJsonType::object_t::size_type N)
```

nlohmann (changes requested):
Thanks for picking this up and for keeping the encoding identical. It follows the same bounded-descent pattern `dump()` uses, which makes it easy to review. My request from 2026-09-12 to update from `develop` is done: the branch is on top of #5286.

I verified it locally before reviewing:

- **Byte-identical output.** I compared against `develop` on 1,420 random values, including many built to nest 126, 127, 128, 129, 130, 200 and 300 levels deep so they cross the switch-over, with siblings on both sides of it. That covers `json` and `ordered_json`, plus JData ndarray objects. Every writer and mode matched: CBOR and MessagePack (vector-returning and adapter overloads), UBJSON and BJData draft 2/3 in all four `use_size`/`use_type` combinations. That's 28,400 outputs in C++11 and C++20.
- **Deep values.** All four writers now handle 1,000,000 levels of nested arrays and of nested objects on a **1 MiB** stack, where `develop` crashes at 100,000. Reading the output back gives the same depth and leaf.
- `unit-large_json.cpp` passes in C++11 and C++17.
- **Speed.** On an ordinary 3,000-object document it is consistently 1-4% slower (median of 41 runs, repeated three times), from the extra depth argument. That's fine with me; I'm only noting it.

**One thing blocks it:** every clang job fails with `-Wtautological-type-limit-compare` at `binary_writer.hpp:939` and `:970`. The new `write_cbor_array_prefix`/`write_cbor_object_prefix` take a plain `std::size_t`, so `N <= numeric_limits<uint64_t>::max()` is always true on a 64-bit target. The code these replace declared `const auto N = j.m_data.m_value.array->size();`, a type that depends on `BasicJsonType`, and clang doesn't diagnose tautological comparisons on dependent types. Taking the containers' own `size_type` restores that; see the two suggestions. I checked it locally: with them, the `-Weverything` set from `cmake/clang_flags.cmake` compiles `to_cbor` cleanly again.

BSON is not part of this PR, and that's fine: I'll cover it in a separate PR, since its writer also needs the recursive document-size pass reworked (#5308).

---
_This review was written by [Claude Code](https://claude.com/claude-code) on behalf of @nlohmann._

gregmarr (commented):
I wrote these comments yesterday but didn't submit the review.  Hopefully they're not out of date.

Issue #5418, opened 2026-08-26
Test suite: reduce runtime — 52 % is Unicode byte sweeps, plus redundant re-parsing in binary roundtrip tests


nlohmann:
## Description

Measured with GCC 13, default (unoptimized) build, `ctest --no-skip` (what the Linux gcc/clang/coverage/sanitizer CI jobs run): **562 s CPU total**, extremely concentrated:

| test | time | share |
|---|---|---|
| test-unicode4_cpp11 | 178 s | 32 % |
| test-msgpack (cpp11+cpp17) | 118 s | 21 % |
| test-cbor_cpp11 | 54 s | 10 % |
| test-unicode3_cpp11 | 54 s | 10 % |
| test-unicode5_cpp11 | 42 s | 7 % |
| test-binary_formats_cpp11 | 30 s | 5 % |
| test-ubjson + test-bjdata | 30 s | 5 % |

Note `doctest::skip()` only helps `JSON_FastTests=ON` runs (Windows/macOS/Cirrus); the Linux CI jobs pass `--no-skip` and run everything, so only real iteration cuts help there.

## Where the time goes, and proposed reductions

**1. Unicode "wrong Nth byte" sections sweep all combinations of the *other* bytes.** unicode4 runs 5,517,507 iterations (its own progress counter, `unit-unicode4.cpp:126`); e.g. "wrong 2nd byte" is 3·192·64·64 = 2.36 M iterations. Each iteration calls both `check_utf8string` (1 × `json::parse`) and `check_utf8dump` (**8–9 × `dump()`** across error-handler/ensure_ascii combinations) — roughly 50 M `dump()` calls suite-wide. The property tested ("byte N out of range ⇒ error") is independent of the other trailing bytes. Pinning unrelated bytes to one representative value removes ~73 % of the 8.9 M unicode iterations; calling `check_utf8dump` only on well-formed sweeps (or a sample of ill-formed ones) cuts ~8× more. Must be coordinated with the byte3/byte4 typo fix (#5416), which would otherwise *add* 3.1 M iterations.

[...]

PR #5519, opened 2026-09-10
Cut test suite runtime in binary roundtrips and integer sweeps

ayush-singh-0601:
See #5418.

Linux CI runs `--no-skip`, so the expensive binary roundtrip cases actually run there. Each corpus file was parsed four times (once per input adapter). Parse once and reuse it.

The 16-bit integer sweeps still prove width selection: stride 7 hits every low-byte residue, and the endpoints stay in the loop.

`unit-large_json` used depth 5,000,000 to check the non-recursive destructor from #1419. 500,000 is enough. `jeopardy.json` is ~500 MB of serialization output, so it is now its own skipped test; canada/twitter/citm/sample keep running.

Unicode byte sweeps are left to #5426.

- [x] If applicable, the documentation is updated.

### DIFF
--- a/tests/src/test_utils.hpp
+++ b/tests/src/test_utils.hpp
@@ -9,6 +9,7 @@
 #pragma once
 
 #include <cstdint> // uint8_t
+#include <cstddef> // size_t
 #include <fstream> // ifstream, istreambuf_iterator, ios
 #include <vector> // vector
 
@@ -23,6 +24,23 @@ namespace utils
 // ordinary function call does.
 template<typename T>
 inline void ignore_return_value(T&& /*unused*/) noexcept {}
+
+// Walk [first, last] inclusive with a stride, always visiting last.
+// stride 7 is coprime to 256, so every low-byte residue is still hit.
+template<typename T>
+T next_integer_sample(T i, T last, T stride)
+{
+    if (i >= last)
+    {
+        return static_cast<T>(last + 1);
+    }
+    if (stride > 0 && i > static_cast<T>(last - stride))
+    {
+        return last;
+    }
+    const T n = static_cast<T>(i + stride);
+    return n < last ? n : last;
+}
 
 inline std::vector<std::uint8_t> read_binary_file(const std::string& filename)
 {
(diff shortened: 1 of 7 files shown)


nlohmann:
Please rebase to the latest develop branch as I had to fix some tests.

ayush-singh-0601:
Rebased onto the latest develop.

The gcc job was failing as well, so I fixed that in the same push. `static_cast<int32_t>` on the sampled loop bounds is a useless cast there (`int32_t` is just `int`), and `ci_test_gcc` builds with `-Werror`. Those bounds are plain ints now. The ranges, the stride of 7, and the endpoints are unchanged.

nlohmann (changes requested):
Thanks! Most of this PR is fine to merge. The stride-7 integer sweeps still hit every boundary and every byte value, the roundtrip hoisting only removes duplicated parsing, and 500k depth is still plenty for `unit-large_json`.

I'd like one change: please revert `tests/src/unit-binary_formats.cpp` to develop's version.

The split doesn't save time, because of how skipped tests run in CI (`cmake/test.cmake`):

| Job type | Command | Runs `skip()` tests? |
|---|---|---|
| Default (gcc/clang/sanitizer/coverage) | `test-foo --no-skip` | yes |
| `JSON_FastTests=ON` | `test-foo` | no |
| Valgrind | `valgrind test-foo` | no |

On develop, "Binary Formats" is skipped, so it runs only in the default jobs. With this PR:

- **Default `--no-skip` jobs** still run both test cases, `jeopardy.json` included, so there is no saving.
- **FastTests and valgrind jobs** never ran any of this before. Now they run canada/twitter/citm/sample. Under valgrind, `test-binary_formats_valgrind` went from about 2 s to about 223 s (compared with develop CI run 35829411620). That makes the valgrind job slower overall (ctest time 390 s → 527 s), even though the other changes should have made it faster.

So the cheaper files don't "keep running": they never ran in those jobs. It isn't a coverage gain either. The only checks that run in more places are the encoded-size checks for four files, and the default and coverage jobs already run those.

If you'd like something from that file in the fast jobs, un-skipping just `sample.json` would be acceptable, since it's small. Otherwise, a plain revert is best.

Nit: the comment on `next_integer_sample` in `tests/src/test_utils.hpp` says "Walk [first, last]", but the function has no `first` parameter.

---
*This review was written by Claude Code on behalf of @nlohmann.*

PR #5522, opened 2026-09-13
Only reserve array capacity if the array type supports it

nlohmann:
## Problem

#5476 added a capped `reserve()` call to `start_array()` in both `json_sax_dom_parser` and `json_sax_dom_callback_parser`, so that definite-length arrays from the binary formats do not reallocate repeatedly.

`ArrayType` is a `basic_json` template parameter, though, and is not required to have a `reserve()` member function. Since `parser::parse()` instantiates both SAX parsers, the call is a hard compile error for any array type without `reserve()`:

```
include/nlohmann/detail/input/json_sax.hpp:705:57: error: no member named 'reserve' in
  'std::deque<nlohmann::basic_json<std::map, std::deque>>'
  705 |     ref_stack.back()->m_data.m_value.array->reserve((std::min)(len, reserve_cap));
```

That breaks `parse()`, `sax_parse()`, and all binary readers for such a type. `develop` itself is green because no test currently instantiates `basic_json` with a `reserve()`-less array type — #5443 adds one, which is why its AppVeyor x64 jobs are red ([build 54705259](https://ci.appveyor.com/project/nlohmann/json/builds/54705259)).

## Fix

Move the capped reservation into a `reserve_array()` helper and pick a no-op overload via `priority_tag` when `ArrayType` has no `reserve()`. This mirrors the existing `from_json_object_reserve()` in `from_json.hpp`, and also removes the duplicated comment and cap constant from the two call sites.

Behaviour for `std::vector`-based array types is unchanged, including the 16384-element cap that keeps a bogus or hostile length from triggering an oversized allocation.

## Tests

`tests/src/unit-regression3.cpp` gains a regression test covering:

- `basic_json<std::map, std::deque>` — text parsing, CBOR/MsgPack round-trips, and the callback parser
- `std::vector`-based `json` still reserves (`capacity() >= 100` after a 100-element CBOR array)
- the reservation stays capped (a CBOR array header announcing 2^32-1 elements, truncated, is rejected)

## API changes

**No breaking changes.** `reserve_array()` is a new function in `nlohmann::detail`, which is not part of the public API. No public type, signature, or behaviour changes: array types that have `reserve()` reserve exactly as before, and array types without it now compile instead of failing.

## Note

While writing the test I found that `std::list` is *not* usable as `ArrayType` today, for reasons unrelated to this change — `serializer::dump_internal()` does `cend() - 1` and `exception::diagnostics()` does `array->operator[](i)`, both of which need random access. [`array_t.md`](https://github.com/nlohmann/json/blob/develop/docs/mkdocs/docs/api/basic_json/array_t.md) names `std::list` as an example, so either the docs or the library needs a follow-up. The test therefore covers `std::deque` only.


include/nlohmann/detail/input/json_sax.hpp
```diff
@@ -151,6 +152,29 @@ constexpr std::size_t unknown_size()
...
+for a small or truncated input.
+
+The overload below is selected for array types without reserve() (e.g.,
+std::deque), which are then left untouched.
+*/
+template<typename ArrayType>
+auto reserve_array(ArrayType& arr, std::size_t len, priority_tag<1> /*unused*/)
+-> decltype(arr.reserve(len), void())
+{
+    constexpr std::size_t reserve_cap = 16384;
```
gregmarr, line 170: Oh, good, and now there's a single place for this magic number.

gregmarr (approved):

include/nlohmann/detail/input/json_sax.hpp
nlohmann: Indeed!

PR #5532, opened 2026-09-15
Treat a NUL byte in the input as an ordinary byte, not end-of-input

### LINKED ISSUE #5530: A trailing NUL byte after a complete JSON value silently truncates parsing instead of raising a trailing-data error
### Description

`json::parse()` raises `parse_error.101` ("unexpected additional data") when non-whitespace content follows a syntactically complete value — for every trailing byte *except* `'\0'`. A trailing NUL byte (and everything after it) is silently dropped instead, and the call succeeds.

This reproduces through ordinary iterator/container-range parsing (`std::string`, `std::vector<uint8_t>`, …), not only through the null-terminated `const char*` convenience overloads that the special-casing was presumably added for.

### Reproduction

```cpp
#include <nlohmann/json.hpp>
#include <iostream>
using json = nlohmann::json;

int main() {
    std::cout << json::parse(std::string("123") + "x").dump() << "\n";
    // throws parse_error.101: unexpected additional data — correct

    std::cout << json::parse(std::string("123") + std::string(1, '\0')).dump() << "\n";
    // succeeds, prints 123 — the trailing NUL is silently swallowed

    std::cout << json::parse(std::string("123") + std::string(1, '\0') + "x").dump() << "\n";
    // still succeeds, prints 123 — everything from the NUL onward is dropped

    std::cout << json::parse(std::string("\"abc\"") + std::string(1, '\0')).dump() << "\n";
    // succeeds, prints "abc" — same behavior for strings
}
```

Traced to [`lexer.hpp#L2156-L2160`](https://github.com/nlohmann/json/blob/3bfe2b6da7393af5cfd68c44f58a1058e60499e2/include/nlohmann/detail/input/lexer.hpp#L2156-L2160):

```cpp
// end of input (the null byte is needed when parsing from
// string literals)
case '\0':
case char_traits<char_type>::eof():
    return token_type::end_of_input;
```

and the trailing-content check in [`parser.hpp#L170`](https://github.com/nlohmann/json/blob/3bfe2b6da7393af5cfd68c44f58a1058e60499e2/include/nlohmann/detail/input/parser.hpp#L170), `if (result && strict && (get_token() != token_type::end_of_input))`, which treats the lexer's report of `end_of_input` at the NUL byte as if the input had genuinely ended there.

### Why this seems worth a look

The comment states the NUL-as-EOF shortcut exists "for parsing from string literals" — i.e., to make `json::parse("123")` work when handed a `const char*` where a naive caller might otherwise include the implicit terminating NUL in the range. That rationale doesn't obviously extend to `parse(iterator, iterator)` calls over `std::string`/`std::vector` ranges, where the byte range is explicit and a NUL has no special meaning as an end-of-string marker — yet the same short-circuit applies there too, silently truncating otherwise-well-formed input that happens to contain an embedded or trailing NUL.

I could not find this behavior mentioned in the public docs (parsing, exceptions, or FAQ pages) — only in the source comment above.

### Suggested fix / alternatives

Scope the NUL-as-EOF shortcut to the specific null-terminated `const char*`/string-literal entry point(s) that need it, so byte-range/iterator-based parsing treats `'\0'` like any other regular (non-whitespace) byte for the trailing-content check — or, if the current scope is intentional, document it explicitly (e.g. alongside [RFC 8259 §8.1](https://json.nlohmann.me/home/faq/#parse-errors-reading-non-ascii-characters)-related notes in the FAQ) so it's a known tradeoff rather than a surprise.

### Context

Found while investigating a third-party compliance report ([#5526](https://github.com/nlohmann/json/issues/5526)) — most of that report's findings were either testing against superseded JSON RFCs or already-documented, intentional behavior (duplicate keys), but this one (the `n_multidigit_number_then_00` JSONTestSuite case) survived verification against the current `develop` header (commit `3bfe2b6da7393af5cfd68c44f58a1058e60499e2`).

nlohmann:
## Summary

A `'\0'` (NUL) byte in the input was silently treated as end-of-input by the lexer, instead of the normal `parse_error.101` any other unexpected byte gets:

```cpp
json::parse(std::string("123") + "x");                          // throws parse_error.101 — correct
json::parse(std::string("123") + std::string(1, '\0'));          // succeeded, parsed as 123 — wrong (before this PR)
json::parse(std::string("123") + std::string(1, '\0') + "x");    // succeeded, parsed as 123 — wrong (before this PR)
```

Fixes #5530.

## Root cause

`include/nlohmann/detail/input/lexer.hpp`'s token dispatch had:
```cpp
// end of input (the null byte is needed when parsing from
// string literals)
case '\0':
case char_traits<char_type>::eof():
    return token_type::end_of_input;
```

The comment's rationale doesn't hold up: `input_adapter(const char*)` computes its length with `strlen()`, which already excludes the terminating NUL from the range the lexer ever sees — it always hits real EOF (`char_traits<char_type>::eof()`), never `'\0'`. `case '\0'` only ever fires on a genuine NUL byte present in actual input data — exactly the bug. (Git history backs this up: the comment traces to the old re2c-generated lexer, where `'\0'` was an internal buffer-refill padding sentinel unrelated to string literals; the rationale carried over inaccurately when the scanner was hand-rewritten in 2017.)

Two more sites in the same file had the identical issue, inside `ignore_comments` handling — a NUL embedded in a `//` or `/* */` comment stopped the comment scan early instead of only a real newline/`*/`/EOF doing so.

All three `case '\0':` labels are removed; a NUL now falls through to the same `default:`/"invalid literal" (or, after a complete value, the same "expected end of input") handling any other unexpected byte already gets.

## A deviation the above reasoning missed, found and fixed during implementation

Verifying `input_adapter(const char*)` alone isn't the whole story: a string literal argument like `json::parse("123")` doesn't decay to that overload — it binds directly to the sibling `template<typename T, size_t N> input_adapter(T(&array)[N])` overload (array-reference binding is preferred over pointer decay), which passed the array's full extent, implicit trailing `'\0'` included. Removing the lexer's special case alone would have broken `json::parse("<literal>")` for every literal in this codebase (verified empirically: it regresses to `invalid literal; last read: '123<U+0000>'`).

Fixed by splitting that overload on `is_same<remove_cv_t<T>, char>`: a `char` array now strips a single trailing NUL if present (mirroring the `const char*` overload, but only ever removing the one implicit terminator — not strlen-style truncation at the *first* NUL, which would silently reintroduce the same truncation bug for a literal with an embedded NUL followed by real content). Any other element type (`uint8_t` buffers used for CBOR/MessagePack/etc.) is untouched, full extent passed through unchanged, since a trailing zero byte there may be real data.

I verified this handles the edge cases correctly: a literal with an embedded (non-trailing) NUL now correctly reports the embedded byte as trailing data rather than silently dropping it, and a `char` array with no NUL terminator at all is passed through unchanged (nothing incorrectly stripped).

## Base branch

This PR targets `claude/issue-5340-restore-unget` (#5344) rather than `develop`: that PR reworks this same area (`lexer.hpp`/`parser.hpp`'s handling of what happens at/after end of input, via its new `release_lookahead()` mechanism), so this fix is written against its already-restructured code rather than risking a rebase later. `parser.hpp` itself needed no changes — the fix is fully contained to how the lexer classifies bytes.

## Test plan

- [x] `tests/src/unit-class_parser.cpp`: new section covering a trailing NUL after a value, a NUL where a value is expected, NUL inside both comment styles (confirming the scan no longer stops early), a regression guard that `json::parse("123")` (and other literals) are unaffected, and a regression guard that the pre-existing "NUL inside a quoted string" behavior is unchanged.
- [x] Fixed several pre-existing tests that had unknowingly depended on the old behavior: a couple of `std::array<uint8_t, 5>` fixtures relying on implicit zero-padding to spell `"true"` (now sized to 4), and the `#2546 std::byte` regression test's `std::span` over a C array (now excludes the array's trailing NUL explicitly, since a `std::span` over an array captures its full extent).
- [x] Updated one existing expected-message string (the `/*`-unterminated-comment test) — it no longer shows a spurious `<U+0000>` in "last read", since the array-literal fix now excludes the trailing NUL before the lexer ever sees it.
- [x] Full existing suites pass unchanged, including PR #5344's own new "stream position after extraction" tests.
- [x] `make amalgamate` run; `single_include/nlohmann/json.hpp` confirmed in sync with `include/`.

## Docs

Added a "NUL bytes in the input" FAQ entry (`docs/mkdocs/docs/home/faq.md`) explaining the corrected behavior and why the plain C-string convenience overload is unaffected.

## Breaking change?

No breaking changes to the public API. Input that happens to contain a literal NUL byte — trailing, embedded, or via a zero-padded buffer — that previously parsed silently (dropping everything from the NUL onward) now correctly raises `parse_error.101`, consistent with how any other unexpected byte is already handled. `json::parse()` with a `const char*`/string-literal argument (with no NUL as data) is unaffected.

https://claude.ai/code/session_01N4RQ1Ahan5YAGbnAQGjZTY

---
_Generated by [Claude Code](https://claude.ai/code/session_01N4RQ1Ahan5YAGbnAQGjZTY)_


include/nlohmann/detail/input/input_adapters.hpp
```diff
@@ -697,7 +697,29 @@ contiguous_bytes_input_adapter input_adapter(CharT b)
...
 
-template<typename T, std::size_t N>
+// char arrays are usually string literals (e.g. json::parse("[1,2,3]")),
+// which the compiler pads with a trailing '\0' that is not part of the
+// text to parse; mirror the const char* overload above (which computes
+// its length with strlen()) and exclude a single trailing NUL terminator,
+// if present, so parsing a literal behaves the same whether the argument
+// decays to a pointer or binds directly to this array overload.
+template < typename T, std::size_t N,
+           typename std::enable_if<std::is_same<typename std::remove_cv<T>::type, char>::value, int>::type = 0 >
```
gregmarr, line 707: also need to check for `unsigned char` and `signed char`?

nlohmann:
Closing without merging — on reflection, changing this behavior is an observable, backwards-incompatible change for any caller who already (even unknowingly) depends on the current NUL-as-EOF handling, and that's not something to ship silently in a patch-level fix. Thanks for the review, @gregmarr — your question about `unsigned char`/`signed char` was a good catch of a real gap in the array-overload logic here, but with the underlying behavior change itself off the table, it's moot rather than something to resolve.

Replaced by #5534, which documents the current behavior clearly instead of changing it.

---

gregmarr:
Is it worth adding this with a compile time option to enable, or leaving it open and targeting 4.0.0?

nlohmann:
> Is it worth adding this with a compile time option to enable, or leaving it open and targeting 4.0.0?

Added — `JSON_STRICT_NUL_HANDLING`, off by default, on #5534.

---

PR #5534, opened 2026-09-15
Document that a NUL byte in the input is treated as end of input

### LINKED ISSUE #5530: A trailing NUL byte after a complete JSON value silently truncates parsing instead of raising a trailing-data error

nlohmann:
## Summary

Fixes #5530 — as documentation plus an opt-in macro, not an unconditional behavior change (see below).

A NUL byte (`'\0'`) anywhere in the input — trailing, or embedded ahead of more otherwise well-formed JSON — is currently treated the same as genuine end of input, so `json::parse()` silently stops there instead of raising the `parse_error.101` any other unexpected byte triggers:

```cpp
json::parse(std::string("123") + '\0');          // == 123, no error
json::parse(std::string("123") + '\0' + "true"); // == 123, the "true" is silently ignored too
json::parse("123x");                              // throws parse_error.101 — any OTHER trailing byte is rejected
```

This falls out of the same convention used when no explicit input length is given at all — `json::parse(const char*)` already stops at the first NUL byte via `strlen()`, since a bare pointer has no length of its own — just applied uniformly rather than only when a length is genuinely unavailable.

## Why documentation instead of an unconditional fix

An earlier version of this work (#5532, closed) changed this behavior unconditionally. On reflection: for a foundational, extremely widely-deployed header-only library, silently changing behavior that real callers may already (even unknowingly) depend on is the wrong trade, even to fix a genuine surprise. This PR documents the current, verified behavior clearly (`docs/mkdocs/docs/home/faq.md`, "NUL bytes in the input"), and — per @gregmarr's suggestion below — adds an opt-in macro so callers who want the corrected behavior today don't have to wait for 4.0.0.

## `JSON_STRICT_NUL_HANDLING`

New macro, default `0` (off — today's behavior, byte-for-byte unchanged), following the one existing precedent for this "opt-in now, default in a future major version" shape in this codebase, `JSON_BRACE_INIT_COPY_SEMANTICS`:
- Gates the three lexer sites that currently treat `'\0'` as end-of-input/comment-terminator (`include/nlohmann/detail/input/lexer.hpp`).
- `include/nlohmann/detail/input/input_adapters.hpp`: when enabled, a `char` array (including a string-literal argument like `json::parse("123")`) has its one compiler-appended trailing `'\0'` trimmed if present, so literals keep working; every other byte — and every other array element type (`unsigned char`, `std::uint8_t`, as used for CBOR/MessagePack) — is read as real data in both states. This checks only the array's own last element (bounded by its compile-time-known size `N`); it deliberately does **not** decay to the `strlen()`-based pointer overload for this, since that would read past the bounds of a same-sized-but-not-NUL-terminated `char` array — confirmed as a real AddressSanitizer-flagged stack-buffer-overflow before settling on the shipped, bounds-safe check.
- New CMake option `JSON_StrictNulHandling` (`OFF` by default) and a `ci_test_strict_nul_handling` target, wired into `ubuntu.yml`'s `ci_cmake_options` matrix — unlike `JSON_BRACE_INIT_COPY_SEMANTICS`'s own CI target, which (traced while implementing this) is never invoked by any GitHub Actions workflow, so its opt-in state has never actually been built in CI.
- Documented via a new dedicated reference page (`docs/mkdocs/docs/api/macros/json_strict_nul_handling.md`, mirroring the precedent's structure) plus updates to `parse.md`, `accept.md`, and `operator_gtgt.md` — including an explicit note that `operator>>` is only affected by a NUL *while a value is being read*, not by one following a complete value (verified empirically: `operator>>` parses non-strictly and never reaches the trailing-content check either way).
- Test coverage across `parse()`, `accept()`, and `operator>>`, in both macro states, built and run locally before pushing (plus a handful of pre-existing test fixtures across three files that unknowingly relied on the old NUL-as-EOF behavior for an incidental trailing byte, fixed alongside).

Both configurations (default and `-DJSON_StrictNulHandling=ON`) were built and their full test suites run locally, with additional targeted AddressSanitizer verification of the char-array safety property described above.

## Test plan

- [x] Every example in the FAQ entry and the new macro reference page was verified by compiling and running it, not just reasoned about.
- [x] Both macro states built and tested locally (CMake + direct compiles); zero incremental regressions in the opt-in state beyond a pre-existing, environment-only test-data-download limitation that affects both states identically.
- [x] AddressSanitizer-verified: a non-NUL-terminated `char` array is never read out of bounds in either macro state; `unsigned char`/`std::uint8_t` arrays are never truncated in either state.
- [x] `make amalgamate` run; `single_include/nlohmann/json.hpp` confirmed in sync with `include/`.

## Breaking change?

No. The macro defaults to off, preserving current behavior exactly. Enabling it is an explicit, documented opt-in.

https://claude.ai/code/session_01N4RQ1Ahan5YAGbnAQGjZTY


gregmarr:
Is it worth considering a compile-time flag that people can use to enable the correct behavior until 4.0.0?

nlohmann:
> Is it worth considering a compile-time flag that people can use to enable the correct behavior until 4.0.0?

Good point - another moving part though. At the same time, this makes the step toward 4.0 a lot easier if we could eventually just explain that 4.0 is the behavior of a list of DEFINEs. I'll add something.

nlohmann:
> Is it worth considering a compile-time flag that people can use to enable the correct behavior until 4.0.0?

Added: `JSON_STRICT_NUL_HANDLING`, off by default (preserving today's behavior), with a dedicated CI target now wired into `ubuntu.yml` so the opt-in state is actually built — full writeup in the PR description above. One thing worth flagging since it came up directly while implementing this: `JSON_BRACE_INIT_COPY_SEMANTICS`'s own opt-in test in `unit-regression3.cpp` guards with `#if defined(JSON_BRACE_INIT_COPY_SEMANTICS) && (...==1)` *after* `#include <nlohmann/json.hpp>` — but the header `#undef`s that macro itself before returning control to the including file (`detail/macro_unscope.hpp`), so that guard is always false regardless of the `-D` flag passed at compile time. Combined with that macro's CI target also never being invoked by any workflow, its opt-in behavior appears to have never actually been exercised by CI. Left it untouched as out of scope here, but wanted to surface it since it's a real, separate finding. (Worked around the same pitfall for the new macro's own tests by capturing the command-line value into a test-local macro before the `#include`.)

---
_Generated by [Claude Code](https://claude.ai/code)

---

docs/mkdocs/docs/api/macros/json_strict_nul_handling.md
```diff
@@ -0,0 +1,115 @@
+# JSON_STRICT_NUL_HANDLING
+
+```cpp
+#define JSON_STRICT_NUL_HANDLING /* value */
+```
+
+When defined to `1`, a `'\0'` (NUL) byte anywhere in the input is rejected with `parse_error.101`, like any other
```
gregmarr, line 7: Isn't there a place involving binary data that still accepts them?

gregmarr (commented):
Maybe I'm misremembering something from yesterday's review.  If so, then this looks fine.

docs/mkdocs/docs/api/macros/json_strict_nul_handling.md
nlohmann: Good catch - the "anywhere in the input" wording was too broad. The binary formats never go through the JSON lexer, so 0x00 stays ordinary data there, and a bare const char* is still cut off at its first NUL by strlen(). The page now lists these exceptions.

gregmarr (approved):

